PR #2, opened 2017-11-13
Basic binary build with a library dependency.

Fuuzetsu:
This was only tested on flat library/binary. No dependencies for the
library itself.

```
$ nix-shell -p bazel -p haskell.packages.ghc821.ghc
$ bazel test //tests:Runhello
$ bazel run @examples//hello-world
```

Many parts are pretty hacky but we can start filling in stuff.
Definitely need to figure out more proper ways to do some things.

I have renamed hello.hs to Main.hs for now as my current rules are too
stupid to deal with it (GHC produces Main.o while rules expect hello.o
based on rule name).

### DIFF
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -1,38 +1,120 @@
-HASKELL_FILETYPE = FileType([".hs"])
-
-# TODO
-haskell_library = 1
-
-_haskell_common_attrs = {
-    "srcs": attr.label_list(allow_files = HASKELL_FILETYPE),
-    "deps": attr.label_list(),
-}
+load(":toolchain.bzl",
+     "HaskellPackageInfo",
+     "ar_args",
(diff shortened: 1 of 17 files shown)


haskell/haskell.bzl
```diff
@@ -1,38 +1,124 @@
-HASKELL_FILETYPE = FileType([".hs"])
+# TODO: using run_shell doesn't work, some wrong nix store path to
```
mboes: This might be fixed in latest Nixpkgs. Did you try that? I haven't encountered anything similar (I'm also on NixOS).

haskell/toolchain.bzl
```diff
@@ -0,0 +1,192 @@
+HaskellPackageInfo = provider()
```
mboes: Is this just a stub? I'm not sure what it does if this provider has no arguments.

PR #14, opened 2017-12-09
Build rules sufficient for building inline-java

Fuuzetsu:
This large PR multiple features. Sadly it's far too sequential to easily split up without wasting whole day as each feature depends on changes in last. This is also the reason why couple of tests are tacked on in the end: discarding all work on top, writing the tests, reapplying work and updating the tests to use new needed features is time better spent on actually getting things to work.

Most commits are self-explanatory w.r.t. what they add.

Code itself is quite messy. I want to spend some time tomorrow on fixing that and adding more tests once that's done.

There's a test missing for C-related things as I could not come up with one on the spot.

I'm attaching a `git diff` from `inline-java` repository with BUILD files that work with this version. My nixpkgs is `6f8601288b1292726188d2600cfeaa6c5be658ce` FWIW.

[inline-java.txt](https://github.com/tweag/rules_haskell/files/1544942/inline-java.txt)

### DIFF
--- a/examples/hello-lib/BUILD
+++ b/examples/hello-lib/BUILD
@@ -8,5 +8,6 @@ load(
 haskell_library(
   name = 'hello-lib',
   srcs = glob(['**/*.hs']),
-  deps = ["@examples//hello-sublib"]
+  deps = ["@examples//hello-sublib"],
+  prebuilt_dependencies = ["base"]
 )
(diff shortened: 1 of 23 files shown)


Fuuzetsu:
@mboes I have cleaned up the rules quite a bit, things are a fair bit more separate now. I updated all tests and checked they work + updated the examples. Also attaching BUILD files I'm using in `inline-java` as reference.

[inline-java.diff.txt](https://github.com/tweag/rules_haskell/files/1546091/inline-java.diff.txt)

mboes:
Thanks, will have a look.

PR #25, opened 2017-12-18
Introduce haskell_import. Replaces external_libraries field.

### LINKED ISSUE #15: external_libraries requires absolute paths on OS
These paths are used in shared object libraries. We could get the needed libs from bazel in case of something like inline-java but it's unusable: best we can do is a relative path to `external` and during builds, the shared library moves around: the relative path becomes useless.

I don't know how to sanely solve this. For now I have something like

```
  external_libraries = {
    "jvm": "/nix/store/f4vdrkqqh71wffq3zz10ymywg3x28acw-openjdk-8u152b16/lib/openjdk/jre/lib/amd64/server",
…
}
```

but it's not really a solution…

mboes:
Prebuilt dynamic libraries, i.e. what Cabal calls `extra-libraries`,
are dependencies like any other. It's just that it's not a Haskell
dependency. The idiomatic way to express this in Bazel is to mix
dependencies of different types all in the `deps` field. So we move to
the content of `external_libraries` to `deps`. Further, for forward
compatibility, we introduce `haskell_import` to mirror `cc_import` and
`java_import`. It doesn't do much at the moment beyond reexport the
shared library it was given and do some sanity checking. But it may
need to do in the future.

This PR introduces two new tests for this functionality. They depend
on `rules_nixpkgs`, which in turn depends on Nix. Since repositories
are loaded lazily, it doesn't mean that Nix is required just to use
the rules, only to test them. We could perhaps conditionally disable
Nix in the future (using constraints?), but this is good enough for
now. And better than the alternative: hard coding absolute paths to
libraries in tests.

Fixes #15.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,18 @@
 
 [![CircleCI](https://circleci.com/gh/tweag/rules_haskell.svg?style=svg)](https://circleci.com/gh/tweag/rules_haskell)
 
+To use these rules, you'll need [Bazel >= 0.8.1][bazel-install]. To
+run tests, you'll furthermore need [Nix][nix] installed.
+
 [bazel]: https://bazel.build/
+[bazel-install]: https://docs.bazel.build/versions/master/install.html
+[nix]: https://nixos.org/nix
 
 ## Rules
 
 * [haskell_binary](#haskell_binary)
 * [haskell_library](#haskell_library)
+* [haskell_import](#haskell_import)
 
 ## Setup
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -134,3 +140,52 @@ haskell_library(
     </tr>
   </tbody>
 </table>
+
+### haskell_binary
```
jml: Should read `haskell_import`, I think.

README.md
```diff
@@ -134,3 +140,52 @@ haskell_library(
...
+
+Imports a prebuilt shared library.
```
jml: Maybe add a sentence that begins, "Use this to …"? Suggesting this because I'm not 100% sure of when I need to use this.

WORKSPACE
```diff
@@ -3,4 +3,15 @@ workspace(name = "io_tweag_rules_haskell")
...
   name = "examples",
   path = "examples"
-)
\ No newline at end of file
+)
+
+http_archive(
+    name = "io_tweag_rules_nixpkgs",
+    strip_prefix = "rules_nixpkgs-a300f574885c50430147e457d21ec22a9fe015f4",
+    urls = ["https://github.com/tweag/rules_nixpkgs/archive/a300f574885c50430147e457d21ec22a9fe015f4.tar.gz"],
```
jml: In other repos, I like to add a comment saying why I'm using a particular SHA1. e.g.

```python
# latest master at the time of last update
```

Or

```python
# Must be later than 0.x in order to get $bugfix
```

Worth doing that here?

haskell/toolchain.bzl
```diff
@@ -103,39 +103,43 @@ def link_haskell_bin(ctx, object_files):
...
-  external_libraries = {}
+  external_libraries = depset()
   for dep in ctx.attr.deps:
-    external_libraries.update(
-      dep[HaskellPackageInfo].external_libraries
-    )
+    if HaskellPackageInfo in dep:
+      external_libraries = external_libraries.union(dep[HaskellPackageInfo].external_libraries)
+    else:
+      external_libraries = external_libraries.union(depset(dep.files))
```
jml: What case does this branch handle?

haskell/toolchain.bzl
```diff
@@ -394,31 +396,43 @@ def gather_dependency_information(ctx):
...
+        confs = hpi.confs,
+        caches = hpi.caches,
+        static_libraries = hpi.static_libraries,
+        dynamic_libraries = hpi.dynamic_libraries,
+        interface_files = hpi.interface_files,
+        static_library_dirs = hpi.static_library_dirs,
+        dynamic_library_dirs = hpi.dynamic_library_dirs,
+        prebuilt_dependencies = hpi.prebuilt_dependencies,
+        external_libraries = hpi.external_libraries.union(depset(dep.files)),
+      )
```
jml: Maybe split the two branches of this conditional into two helper functions?

mboes:
@jml thanks for review comments! This patch is all the better for them I think.

PR #27, opened 2017-12-21
Configurable tooling

Fuuzetsu:
This allows us to provide tools to bazel rules to build with. By
default we use whatever `@ghc` and `@binutils` are available in the
workspace.

I'm unsure how to document/deal with this properly: right now it may
be surprising to the user to have to populate those. They are no
longer accepted from environment however which I suspect is what we
want to begin with.

Ideally we would provide some `compiler.bzl` with various GHC versions
out of rules_nixpkgs/official bindists. I'll have to look how to do
that and make it available to user just like we can say `@openjdk`
today without much extra work.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
             apk --no-progress update
             apk --no-progress add bash ca-certificates
             nix-channel --update
-            nix-env -iA nixpkgs.bazel
+            nix-env -iA nixpkgs.bazel nixpkgs.binutils
       - run:
           name: Run tests
           command: bazel test //...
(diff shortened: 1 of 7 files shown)


johnynek:
Have you seen bazel's support for toolchains?

for instance:
https://github.com/bazelbuild/rules_rust/pull/52/files

https://github.com/bazelbuild/rules_scala/pull/364

for examples.

I think you may want to integrate with this.

Fuuzetsu:
@mboes I have added linking fixes to this PR as well. PATAL in case you already have before.

@johnynek I'll certainly have a look.

PR #38, opened 2017-12-25
Turn on --allow_incompatible_changes in CI

mboes:
See https://docs.bazel.build/versions/master/skylark/backward-compatibility.html.

This required a workaround because of bazelbuild/bazel#4354.


Fuuzetsu:
Mostly LGTM but there are some semantics-changing edits in some places. I think they are not serious (don't break builds) but they will produce weird looking commands with repetitions. I'll fix that up then merge.

PR #40, opened 2017-12-26
Use Skylib.

mboes:
Skylib is located at https://github.com/bazelbuild/bazel-skylib. It
turns out that it already contained functions that replace
`drop_path_prefix()`, `replace_ext()` and `path_append()`. So we have
less code to maintain ourselves. In fact the stats on this PR are -100
lines net.

Also includes a related refactor that removes getter functions that
were adding needless indirection.


mboes:
Ah, makes sense.

PR #42, opened 2017-12-27
Allow hsc files in srcs, remove `hscs` rule attribute.

### LINKED ISSUE #26: Merge hscs and srcs
Some source files need preprocessing before being handed over to GHC. One way or another, they are still source files and the good news is that which preprocessor to run is always clear from the name of the source file alone. So we can list them in `srcs` along with the rest and add an extra action for those files that end in `.hsc`. Easier for the users: sources in `srcs`, dependencies in `deps`, etc.

Fuuzetsu:
We can uniquely identify these through an extension anyway.
Technically one could have an arbitrary extension and still want to
treat the file as `hsc` file but we weren't allowing that anyway is
`hscs` field already. This way is friendlier to the user and none of
the code really cared about the distinction anyway.

Fixes #26.

### DIFF
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -75,7 +75,7 @@ _haskell_common_attrs = {
     doc="Directory in which module hierarchy starts."
   ),
   "srcs": attr.label_list(
-    allow_files=FileType([".hs"]),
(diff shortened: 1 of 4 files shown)


mboes:
Yay!

PR #43, opened 2017-12-27
Toolchain support.

### LINKED ISSUE #36: Use bazel's toolchain support
In #27 it was pointed out that bazel has toolchain support. We should use it and provide some default toolchains. This could also solve things like #24 for us as version information could be included out of the box.

mboes:
Instead of a build_tools attribute in rules, we use the standard
toolchains attribute for the same purpose: finding the commands to use
for building.

Resolves #36.


mboes:
Those look to me like separate toolchains.

PR #47, opened 2017-12-28
Initial Haddock support.

### LINKED ISSUE #34: Documentation (haddock) support.
Support builds with documentation. 

As we may want to build _all_ documentation, #29 might be related here. Possibly due to the way haddock works, bazel aspects are actually relevant here.

Fuuzetsu:
I'm able to generate documentation for existing
library (`haddock-lib-b` in tests) that depends on another. They both
automagically get their docs generated. Some notes:

* Linking to direct dependencies seems to work. I'm unsure that we
  need transitive interface file inclusion: I don't think so.
* It's able to skip non-Haskell dependencies so it doesn't try to do
  something weird on system deps for example but see comment in tests.
* Linking to `prebuilt_dependencies` documentation is not solved: we'd
  need to find out the actual paths to these to pass to Haddock. A
  possibility is to get it from GHC itself, basically a little script
  that calls `ghc-pkg`, finds out where packages live and gets the
  paths that way. Possibly we don't care about providing this with
  "you should be defining dependencies in bazel". I think we should
  provide this if it's not too hard though.
* We only output default files for the direct dependencies mentioned
  in `haskell_haddock`. This seems to work OK though I'm not 100% on
  whether we can rely no other docs always being neighbours. See
  comment in code on `--read-interface`. FWIW we could output them all
  if we wanted but it doesn't change the assumption.
* I haven't added `haddock` to the tool check toolchain does but
  I'm relying on it being there. This is because it's there normally
  but it's not inconceivable the user might not care about haddock and
  user's toolchain doesn't actually provide it.
* Various versions of Haddock behave in different ways: for example
  Haddock master doesn't output `haddock-util.js` anymore. Unsure how
  to deal with this except letting the user to specify what Haddock
  version they are using (and getting this info from toolchain
  somehow by default somehow).
* There is a lot of repetition of logic when it comes to GHC flags and
  we risk the same situation that we had with
  `haskell_library/haskell_binary`. This sucks but I'm unsure how to
  deal with this nicely: ideally we'd want some sort of "OK let's just
  use the same flags as the package used" with a caveat that not all
  those flags are relevant: we don't care about `-odir` or `-no-link`
  for example. I suspect we could ‘register’ flags we potentially care
  about at build time and propagate them in `HaskellPackageInfo` for
  Haddock's use later (if any). I don't see an elegant way to do this
  except straight up mutating some builder object throughout the
  program though... Nevertheless this needs to be addressed unless we
  want to re-implement a lot of build logic for Haddock.

Closes #34. If we merge this in its current state, I'll create tickets
for some of the above points. May want to address that last one before
even initial merge.

### DIFF
--- a/haskell/haddock.bzl
+++ b/haskell/haddock.bzl
@@ -1,0 +1,165 @@
+"""Haddock suppport."""
+
+load (":path_utils.bzl",
+      "path_to_module",
+)
+
+load(":tools.bzl",
+     "get_haddock",
+     "get_build_tools_path",
+)
+
+load(":toolchain.bzl",
+     "HaskellPackageInfo"
+)
+
+load("@bazel_skylib//:lib.bzl", "paths")
+
+HaddockInfo = provider(
+  doc = "Haddock information.",
+  fields = {
+    "outputs": "All interesting outputs produced by Haddock.",
+    "interface_file": "Haddock interface file.",
+    "doc_dir": "Directory where all the documentation files live.",
+  }
+)
+
+def _haskell_haddock_aspect_impl(target, ctx):
+  if HaskellPackageInfo not in target:
+    return []
+
+  pkg_id = "{0}-{1}".format(ctx.rule.attr.name, ctx.rule.attr.version)
+
+  doc_dir = ctx.actions.declare_directory("doc-{0}".format(pkg_id))
+  haddock_interface = ctx.actions.declare_file(
+    paths.join(doc_dir.basename, "{0}.haddock".format(pkg_id)),
+
+  )
+  hoogle_file = ctx.actions.declare_file(
+    paths.replace_extension(ctx.rule.attr.name, ".txt"),
+    sibling = haddock_interface
+  )
+  input_sources = [ f for t in ctx.rule.attr.srcs for f in t.files.to_list() ]
+
+  args = ctx.actions.args()
+  args.add([
+    "-D", haddock_interface,
+    "--package-name={0}".format(ctx.rule.attr.name),
+    "--package-version={0}".format(ctx.rule.attr.version),
+    "-o", doc_dir,
+    "--html", "--hoogle",
+    # TODO: We shouldn't have to remember this, I think haskell
+    # targets should expose all GHC flags they were ultimately built
+    # with. That way we can pass exactly the same stuff. Having said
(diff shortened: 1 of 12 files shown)


tests/haddock/BUILD
```diff
@@ -0,0 +1,30 @@
...
+)
+
+haskell_library(
+  name = "haddock-lib-a",
+  srcs = ["LibA.hs", "LibA/A.hs"],
+  prebuilt_dependencies = ["base"],
+)
+
+# TODO: Add system library to deps. I can't do it with something like
+# zlib because bazel freaks out that zlib is a non-test dependency
```
mboes: Actually, `zlib` *is* only for tests. So should be fine to mark it `testonly`.

tests/binary-with-sysdeps/BUILD
```diff
@@ -3,6 +3,7 @@ package(default_testonly = 1)
 load("@io_tweag_rules_haskell//haskell:haskell.bzl",
   "haskell_binary",
   "haskell_import",
+
```
mboes: extraneous whitespace change.

haskell/haddock.bzl
```diff
@@ -0,0 +1,165 @@
...
+
+  # Pass same flags the user set for the build.
+  for ghc_flag in ctx.rule.attr.compiler_flags:
+    args.add("--optghc={0}".format(ghc_flag))
+
+  args.add(input_sources)
+
+  static_haddock_outputs = [
+    ctx.actions.declare_file(f, sibling=haddock_interface)
+    for f in ["doc-index.html", "haddock-util.js",
```
mboes: Any chance we could avoid this hard coded list of Haddock artifacts?

mboes:
> ideally we'd want some sort of "OK let's just use the same flags as the package used" with a caveat that not all those flags are relevant

If the non-relevant flags are also non-harmful, then perhaps it's okay to just blindly reuse all build flags as-is?

haskell/haddock.bzl
Fuuzetsu: I don't really see how. We could list these out per-version after #56 but we need these one way or another.

Fuuzetsu:
I opened multiple tickets, addressed the couple of comments and added couple of non-controversial commits which immediately close few of the tickets while I was doing that. Merging.

Issue #11, opened 2017-12-08
Dynamic libs need GHC version suffix


Fuuzetsu:
Dyn libs in `dynamic-library-dirs` section of ghc package files need to have `-ghcVERSION` name suffix. This is a problem because we don't know the version and we can't generate it dynamically which means we need to provide it as input to the rule. For now I'm just doing a ghcVersion attribute with default value but we need to figure out how to define and use toolchain info instead.

Fuuzetsu:
Was fixed by #49.

PR #49, opened 2017-12-29
Specify compiler version in haskell_toolchain

### LINKED ISSUE #24: ghcVersion should be derived auto-magically
We need this attribute to know what library name GHC will be trying to look for during linking. Currently the user has to specify it and it's error prone: even on master, this is set to `8.2.2` while CI uses GHC 8.0.2: this will give the user non-obvious error messages the moment they want to use shared library dependencies.

mboes:
Instead of in `haskell_library` / `haskell_binary` rules. Also,
document `haskell_toolchain`.

Resolves #24 and #11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ run tests, you'll furthermore need [Nix][nix] installed.
 
 * [haskell_binary](#haskell_binary)
 * [haskell_library](#haskell_library)
+* [haskell_toolchain](#haskell_import)
 * [haskell_import](#haskell_import)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -141,6 +142,69 @@ haskell_library(
   </tbody>
 </table>
 
+### haskell_library
```
Fuuzetsu: Look at rendered version of this README. We now have two `haskell_library` sections.

PR #79, opened 2018-01-03
Allow arbitrary inputs during GHC's compilation.

Fuuzetsu:
This is necessary to accomodate a hack in sparkle without particularly
poluting rules_haskell.

### DIFF
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -54,6 +54,19 @@ _haskell_common_attrs = {
     default="1.0.0",
     doc="Package/binary version"
   ),
+  # Allows us to inject arbitrary files that may be needed at GHC's
+  # runtime.
+  #
+  # TODO: This is not very nice, things just get dumped into the env:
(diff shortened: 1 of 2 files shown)


haskell/haskell.bzl
```diff
@@ -54,6 +54,19 @@ _haskell_common_attrs = {
...
+  # Allows us to inject arbitrary files that may be needed at GHC's
+  # runtime.
+  #
+  # TODO: This is not very nice, things just get dumped into the env:
+  # you have to know exactly what you're passing in to make any real
+  # use of it. Apparently java rules &c. manage to put things from
+  # their data fields into tidy data directories. We should do that if
+  # we stick with this.
+  "data": attr.label_list(
+    mandatory=False,
```
mboes: Redundant defaults should not be restated in code (applies to `mandatory` and `default`).

Fuuzetsu:
Yeah this is mostly here so I could point another PR at it.

Fuuzetsu:
Should not be needed anymore.

Issue #80, opened 2018-01-03
Interop with java rules


Fuuzetsu:
If a Java dependency is provided, add its (transitive?) jar(s?) to CLASSPATH during compilation. This should allow much more natural implementation of things like https://github.com/tweag/sparkle/pull/133

johnynek:
in bazel, you would not add the transitive deps to the compile classpath, but instead the transitive compile classpath (which only includes direct deps and exports).

Skylark does have a pretty good java API, so this is quite doable.

Fuuzetsu:
Right now I'll only be adding direct deps to classpath because bazel throws NPE if I try to access `transitive_compile_time_jars` (which is what I think you mean we should be using) on one of our Java targets. I'll be filing a ticket in a bit after I craft a repro case.

Fuuzetsu:
Relevant upstream ticket: https://github.com/bazelbuild/bazel/issues/4384

mboes:
Close? We can create another ticket to track the upstream issue.

Fuuzetsu:
I don't know. The ticket is just not complete until we have that I would think. We can create one to track upstream I guess.

mboes:
We have Java interop - just sometimes the `CLASSPATH` isn't exactly the one we want. I thin you should claim victory for this ticket and create another one for what remains.

Fuuzetsu:
#96

PR #81, opened 2018-01-03
Introduce interop with Java rules.

Fuuzetsu:
This allows us to list Java dependencies in `deps` of Haskell rules
and have appropriate inputs and CLASSPATH be available during compile
time if for example `javac` is called during build.

Relates to #80 but does not close it: we ideally want transitive
classpath (I think? cc @mboes , maybe we're OK with how things are
now) but accessing the necessary attribute crashes bazel in some
cases. I'll be creating an upstream ticket shortly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ haskell_binary(
       <td><code>deps</code></td>
       <td>
         <p><code>List of labels, required</code></p>
-        <p>List of other Haskell libraries to be linked to this target</p>
+        <p>List of dependencies to of this target</p>
       </td>
     </tr>
   </tbody>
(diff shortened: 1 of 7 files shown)


Fuuzetsu:
Upstream ticket for no transitive: https://github.com/bazelbuild/bazel/issues/4384

Fuuzetsu:
@mboes rule_test doesn't actually run any of the binaries, all it does is check that rule generates promised output… also why should the classpath exist during execution? we care about it during build and that's exactly what the test is doing: it checks that it exists during run; we could check that files exist too I suppose if we wanted to

PR #83, opened 2018-01-04
Add .so to default outputs, drop pointless directories

Fuuzetsu:
Adding produced .so to default outputs allows us to use `haskell_library` rules directly in `srcs` on `cc_library` rules which allows the cc rule to use the shared object. There is an alternative solution where we have `haskell_so` rule that just takes the shared object from the provider and forwards it. This _may_ be still be needed for linking purposes.

Additionally in separate commit I drop static_library_dir and dynamic_library_dir: they are not needed because the library names are already unique enough and they were just pointless noise.


Fuuzetsu:
seems `haskell_so` is needed, will re-open later

PR #84, opened 2018-01-04
Introduce haskell_so, drop useless dirs

Fuuzetsu:
See README for info on haskell_so.

Includes small cleanup removing pointless layer of directories.


Fuuzetsu:
haskell_so outputs transitive .so while presumably we want to only output .so of library itself from haskell_library

mboes:
I agree. But then you seem to be saying that we _don't_ need `haskell_so`? Or did you mean to say something else?

Fuuzetsu:
if you have something like

```
haskell_library(
  name = "foo";
  …
)

haskell_library(
  name = "bar";
  deps = [":foo"];
)

cc_binary(
  name = "cc",
  srcs = ["something.c", ":bar"],
)
```

you are likely to get a link error because `bar` can't find foo: `cc` only sees `bar.so`. If you use haskell_so instead:

```
haskell_so(
  name = "bar-libs",
  dep = ":bar",
)

cc_binary(
  name = "cc",
  srcs = ["something.c", ":bar-libs"],
)
```

`cc` is given `bar.so` and `foo.so` and everything is OK. The way to achieve this without defining `haskell_so` is to stick `.so` outputs of all transitive dependencies in the default outputs of `haskell_library`: that is if you were to say `bazel build //:bar`, you would get both `foo.so` and `bar.so` which seems strange to me. If you want that sure.

mboes:
In principle we should treat transitive dependencies the same way CC does.
It probably passes these around using a provider. But problem is, we can't
currently produce CC provider instances from skylark. In which case
haskell_so might be a temporary necessity if hacking the output list is
problematic.

Fuuzetsu:
Right, I think that'd be the ideal way. So in meantime let's have this?

PR #85, opened 2018-01-04
Do not -package transitive dependencies.

### LINKED ISSUE #72: Builds shouldn't include transitive `-package`s
Currently if you have
```
A, deps = []
B, deps = [A]
exe, deps = [B]
```

you can import modules from `A` in `exe`.

Fuuzetsu:
If A depends on B and B depends on C, previously A could import
directly from C without listing C in dependencies. It should not.

Fixes #72.

### DIFF
--- a/haskell/actions.bzl
+++ b/haskell/actions.bzl
@@ -503,7 +503,7 @@ def gather_dependency_information(ctx):
       new_external_libraries = new_external_libraries.union(pkg.external_libraries)
       hpi = HaskellPackageInfo(
         name = hpi.name,
-        names = hpi.names + pkg.names,
+        names = hpi.names + [pkg.name],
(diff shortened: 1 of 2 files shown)


johnynek:
it would be nice to have a set of negative tests of things that shouldn't work. We have some of those in rules_scala. So, you could have a test that the `A -> B -> C` case would not allow `A` to import `C` without a compile failure.

Fuuzetsu:
I'll check it out. Anywhere in particular I should look?

johnynek:
check out:

https://github.com/bazelbuild/rules_scala/blob/master/test_rules_scala.sh

see all the `test_expect_failure` stuff.

We have a significant number of negative tests.

mboes:
CI is red on this branch. Could you have a look?

PR #88, opened 2018-01-06
Update rule attributes in README

mboes:
Also drop docstrings for attributes. We've so far failed to keep them
in sync with README. It's only going to get worse. Skydoc chokes on
them anyways. We'll revert once we switch to using Skydoc for API
documentation.


Fuuzetsu:
I actually wanted to propose to do the opposite: use Skydoc. I wasn't aware that was broken though.

PR #89, opened 2018-01-07
Introduce haskell_repositories().

mboes:
This brings skylib into scope. That way users don't need to worry
about this. When [recursive workspaces][1] will finally be implemented
in Bazel, this function will become a noop. A `*_repositories` rule is
a standard idiom in Bazelia these days: skydoc uses it, scala uses it,
etc etc.

[1]: https://bazel.build/designs/2016/09/19/recursive-ws-parsing.html


Fuuzetsu:
That's much nicer.

PR #91, opened 2018-01-09
Adding ghc_link_flags attribute to haskell_binary rule.

gdeest:
The main motivation for this PR is to compile binaries with linking-only options such as `-threaded` or `-with-rtsopts=...`.

### DIFF
--- a/haskell/actions.bzl
+++ b/haskell/actions.bzl
@@ -114,6 +114,7 @@ def link_haskell_bin(ctx, object_files):
   # function yet, and need to construct full shell command below as
   # a string.
   link_args = []
+  link_args.extend(ctx.attr.ghc_link_flags)
   link_args.extend(["-o", ctx.outputs.executable.path, dummy_static_lib.path])
 
   for o in object_files:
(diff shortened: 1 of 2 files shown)


Fuuzetsu:
I think this is OK: can you add a test that relies on flags being passed in during linking fails or it fails? If it's easy it'd be nice to do.

mboes:
Is this flag useful? Why aren't we passing `compiler_flags` to the linking phase too? All flags are accepted by GHC regardless of the phase. We could add `linkopts` like the C/C++ rules do, but those are flags for the linker, not for the linking phase of GHC. It'll get messy...

Fuuzetsu:
I must have misread. I agree whe should just pass those: this is what we discussed on Slack so unless there's a good reason for separation, let's not.

gdeest:
Off the top of my head: having separate flags for compilation flags and linking could be useful if we ever want to provide non-empty defaults for either one, while letting the user customize the other. For example, producing threaded binaries out-of-the-box might make sense ; but then, with a single flag, things could break unexpectedly if the flags are changed to, say, `-O2` by the end user. 

However, we may not want to do that. Since this change is easy to re-introduce, I concur that we should just pass `compiler_flags` to both phases for the time being. I'll open a new PR close this one.

mboes:
Cabal doesn't do that, so I'm skeptical that there would be any good reason for us doing that.

gdeest:
Closed in favor of #93 .

PR #93, opened 2018-01-10
Pass compiler_flags to GHC during linking of Haskell binaries

gdeest:
As discussed in #91 , some compiler flags such as `-threaded` MUST be passed during the linking phase. This PR solves this problem by passing `compiler_flags` to GHC during linking.

Do we want to add a test for this change ?

### DIFF
--- a/tests/BUILD
+++ b/tests/BUILD
@@ -41,6 +41,13 @@ rule_test(
 )
 
 rule_test(
+  name = "test-binary-with-link-flags",
+  generates = ["binary-with-link-flags"],
+  rule = "//tests/binary-with-link-flags",
+  size = "small",
+)
+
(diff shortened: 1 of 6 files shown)


mboes:
Can you include a test please?

gdeest:
Sure thing.

tests/BUILD
```diff
@@ -40,6 +40,13 @@ rule_test(
   size = "small",
 )
 
+rule_test(
```
mboes: This rule is redundant. It doesn't test any property that shouldn't already be covered by other tests. It's the `sh_test` below that tests a useful property.

gdeest: Granted.

PR #106, opened 2018-01-29
Refactoring

mrkkrp:
Ready.

### DIFF
--- a/haskell/haddock.bzl
+++ b/haskell/haddock.bzl
@@ -1,6 +1,6 @@
 """Haddock suppport."""
 
-load (":path_utils.bzl", "path_to_module")
+load (":path_utils.bzl", "module_name")
 
 load(":tools.bzl",
   "get_haddock",
@@ -113,7 +113,7 @@ def _haskell_doc_aspect_impl(target, ctx):
   module_htmls = [
     ctx.actions.declare_file(
       paths.replace_extension(
-        path_to_module(ctx, f, sep='-', prefix=ctx.rule.attr.src_strip_prefix),
+        module_name(ctx, f).replace('.', '-'),
         ".html"
       ),
       sibling=haddock_interface
--- a/haskell/path_utils.bzl
+++ b/haskell/path_utils.bzl
@@ -85,8 +85,7 @@ def mk_name(ctx, name_prefix):
   return "{0}-{1}-{2}".format(name_prefix, ctx.attr.name, ctx.attr.version)
 
 def mk_module_name(ctx, source_name, name_prefix):
-  """
-  Make a target-unique and `source_name`-unique name.
+  """Make a target-unique and `source_name`-unique name.
 
   This is quite similar to `mk_name` but also uses a path built from
   `source_name` to prevent clashes with other names produced using the same
@@ -100,7 +99,7 @@ def mk_module_name(ctx, source_name, name_prefix):
   Returns:
     string: Target- and source-unique name.
   """
-  return "{0}-{1}-{2}-{3}__".format(
+  return "{0}-{1}-{2}/{3}__".format(
     name_prefix,
     ctx.attr.name,
     ctx.attr.version,
--- a/haskell/path_utils.bzl
+++ b/haskell/path_utils.bzl
@@ -104,8 +104,11 @@ def _rel_path_to_module(ctx, f):
     paths.join(
       ctx.label.workspace_root,
       ctx.label.package,
+      # Since the src_strip_prefix attribute is always present in rule
+      # attributes, if it's not there, the function is called from aspect
+      # implementation and so we can access ctx.rule.attr.src_strip_prefix.
       ctx.attr.src_strip_prefix if hasattr(ctx.attr, "src_strip_prefix")
-                                else ""
+                                else ctx.rule.attr.src_strip_prefix
     )
   )
 
--- a/haskell/path_utils.bzl
+++ b/haskell/path_utils.bzl
@@ -2,46 +2,71 @@
 
 load("@bazel_skylib//:lib.bzl", "paths")
 
-def path_to_module_path(ctx, hs_file, prefix=None):
-  """Map a source file to a module name.
+def module_name(ctx, f):
+  """Given Haskell source file path, turn it into a dot-separated module name.
 
-  some-workspace/some-package/src/Foo/Bar/Baz.hs => Foo/Bar/Baz
-
-  Args:
-    ctx: Rule context.
-    hs_file: Haskell source file.
-    prefix: Prefix to strip. ctx.attr.src_strip_prefix used if not set.
-
-  Returns:
-    string: Module part of hs_file. See example above.
-  """
-  strip_prefix = prefix if prefix != None else ctx.attr.src_strip_prefix
-
-  # Directory under which module hierarchy starts.
-  pkg_dir = paths.join(ctx.label.workspace_root,
-                       ctx.label.package,
-                       strip_prefix)
-  # Module path without the workspace and source directories, just
-  # relevant hierarchy.
-  path_no_prefix = paths.relativize(hs_file.path, pkg_dir)
-  # Drop extension.
-  return path_no_prefix[:path_no_prefix.rfind(".")]
-
-def path_to_module(ctx, hs_file, sep='.', prefix=None):
-  """Given Haskell source file path, turn it to a module name.
-
-  some-workspace/some-package/src/Foo/Bar/Baz.hs => Foo.Bar.Baz
+  module_name(
+    ctx,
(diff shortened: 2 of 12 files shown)


haskell/haddock.bzl
```diff
@@ -113,7 +113,7 @@ def _haskell_doc_aspect_impl(target, ctx):
   module_htmls = [
     ctx.actions.declare_file(
       paths.replace_extension(
-        path_to_module(ctx, f, sep='-', prefix=ctx.rule.attr.src_strip_prefix),
```
mrkkrp: Is `ctx.rule.attr.src_strip_prefix` different from `ctx.attr.src_strip_prefix` here? If so, we need to test this functionality, so the tests would fail on this change.

mrkkrp:
@mboes, @Fuuzetsu, please see f85ad16fd3668387ba6dd7080abb2b2996fde9ba. If you like this approach I can continue and convert the following fields of the `HaskellPackageInfo` provider to be sets:

* `static_libraries` (this is currently a list, but maybe it should be a set? I'm not sure, I'd like a comment about this)
* `dynamic_libraries` (this and the others are explicitly marked in the source with the phrase "Do _not_ use a depset.", so I guess using sets here is an improvement).
* `interface_libraries`
* `prebuilt_dependencies`
* `external_libraries`

----

Just in case, motivation for this `set.bzl` thing is given in commit message of 809afac858866fdd207827d3d6ca896f0647def8.

mboes:
@Fuuzetsu any thoughts on the matter?

haskell/haddock.bzl
Fuuzetsu: `ctx.rule.attr` accesses attributes of a target we're applying [aspects](https://docs.bazel.build/versions/master/skylark/aspects.html) to. If you're in an aspect context, `ctx.attr` will refer to the attribute of the aspect itself rather than the target.

This makes a bunch of code very awkward sadly so you get keyword args like the one you see above. However now thinking about it maybe we can get rid of that and pass `ctx.rule` instead of `ctx` and it will just work. I would encourage you to try that so we can drop this `None` business.

haskell/path_utils.bzl
```diff
@@ -2,46 +2,71 @@
...
+    ctx,
+    "some-workspace/some-package/src/Foo/Bar/Baz.hs",
+  ) => "Foo.Bar.Baz"
 
   Args:
-    ctx: Rule context.
-    hs_file: Haskell source file.
-    prefix: Prefix to strip. ctx.attr.src_strip_prefix used if not set.
+    ctx:  Rule context.
+    path: Haskell source file.
```
Fuuzetsu: your argument name is `f` but is documented as `path`

haskell/set.bzl
```diff
@@ -0,0 +1,111 @@
...
+  """Return true if `e` is in the set `s`.
+
+  Args:
+    s: The set to inspect.
+    e: The element to search for.
+
+  Result:
+    Bool, true if `e` is in `s`, false otherwise.
+  """
+  e in s._set_items.get(e)
```
Fuuzetsu: why the `get`? `e in s._set_items` is what I'd expect

haskell/set.bzl
```diff
@@ -0,0 +1,111 @@
...
+  Args:
+    s: Set to insert new element into.
+    e: The element to insert.
+
+  Result:
+    A copy of set `s` with `s` element added.
+  """
+  r = dict(s._set_items)
+  r[e] = None
+  return struct(_set_items = r)
```
Fuuzetsu: I think `return struct(_set_items = dict(s._set_items, e = None))` would work if we want to be more concise without explicit binding to `r` and mutation.

haskell/set.bzl
```diff
@@ -0,0 +1,111 @@
...
+    s1: Another set.
+
+  Result:
+    Set, union of the two sets.
+  """
+  r = dict(s0._set_items)
+  r.update(s1._set_items)
+  return struct(_set_items = r)
+
+def _map(s, f):
```
Fuuzetsu: Considering lack of lambda &c., I think this is pretty much useless. I'm not even sure bazel lets you pass a function in to begin with actually.

If we're keeping it, use dict comprehension instead, IIRC it can be a lot faster.

```
return struct(_set_items = { f(x): None for x in s_set_items.keys() })
```

haskell/set.bzl
```diff
@@ -0,0 +1,111 @@
...
+  Args:
+    l: List, source of the elements for the new set.
+
+  Result:
+    Set containing elements from given list.
+  """
+  r = dict()
+  for x in l:
+    r[x] = None
+  return (struct(_set_items = r))
```
Fuuzetsu:
```
return struct(_set_items = { x: None for x in l })`
```

haskell/set.bzl
Fuuzetsu: Ah I see it used below actually, disregard first part then.

haskell/tools.bzl
```diff
@@ -38,7 +47,7 @@ def _get_build_tool(ctx, tool_name):
   Returns:
     File: Build tool with the name user asked for.
   """
-  for tool in get_build_tools(ctx).to_list():
+  for tool in set.to_list(get_build_tools(ctx)):
```
Fuuzetsu: I wonder if we can implement a method that lets us traverse over `set` without explicit to_list? `dict` is traversable I believe (though that may be only with `.items` &c.)

haskell/tools.bzl
```diff
@@ -1,17 +1,19 @@
...
+
 def get_build_tools(ctx):
   """Get the set of all build tools we have available.
 
   Args:
     ctx: Rule context.
 
   Returns:
-    depset of File: All build tools provided to the rule.
+    Set of File: All build tools provided to the rule.
```
Fuuzetsu: `set of File`

haskell/set.bzl
```diff
@@ -0,0 +1,111 @@
...
+  Args:
+    s0: One set.
+    s1: Another set.
+
+  Result:
+    Set, union of the two sets.
+  """
+  r = dict(s0._set_items)
+  r.update(s1._set_items)
+  return struct(_set_items = r)
```
Fuuzetsu: I think `return struct(_set_items = dict(s0._set_items, **(s1._set_items)))` is extremely fast and memory efficient.

Fuuzetsu:
I left some comments, mostly about implementations. I don't see anything obviously wrong.

w.r.t. dict vs depset: I think that yes, we do have a big win even with a to_list kind of thing: depsets are really awful for anything but lazy union but in our case the unions are pretty much never lazy: we consume everything that we put in a set in the next target "up".

Bazel allows you to profile your rules, you should try it, it's quite easy. See [this](https://www.kchodorow.com/blog/2015/09/18/build-y-u-go-slow/) for an example. On my small project few weeks ago, depsets were _the_ biggest time and memory consumer, about 25% of the analysis time. Having said that, it was all still quick because my project was small.

I think we won't see any real benefit until:

* we convert most/all things to dicts (we can do this nearly everywhere, I think only in one place we need extra feature of depsets that allows to keep us ordering of unions)
* we have a big project to try against

mboes:
> depsets are really awful for anything but lazy union but in our case the unions are pretty much never lazy

It's possible, however, that dicts are just as bad if not worse. @mrkkrp could you split the sets stuff out into a separate PR, so that you can provide perf analysis separately and ideally submit it to @faineance for review of the perf impact on their project?

mrkkrp:
@Fuuzetsu Unfortunately passing `ctx.rule` instead of `ctx` does not work. I'll try something another thing that should allow to avoid that `None` business tomorrow.

mrkkrp:
@Fuuzetsu, what do you think about the last commit? Unfortunately, even just checking if there is a `rule` attribute in `ctx` raises that error "rule attributes are only available in aspect implementations" (this limitation doesn't make sense IMO and limits what you can do).

Fuuzetsu:
>It's possible, however, that dicts are just as bad if not worse.

dicts are the recommended structure to use instead of depsets if you're doing anything but union; skylib uses them for collections too; but perf could be nice anyway

@mrkkrp You shouldn't check if there is a `rule` attribute: I simply mean that you can use a function that uses ctx only by passing `ctx.rule` when you know you're in a context. i.e. anywhere where you would have had `foo(ctx, some_attr=ctx.rule.attr.some_attr)` where `foo` uses `some_attr` if set and `ctx.attr.some_attr` if not set, you would now just do `foo(ctx.rule)`.

Anyway last commit seems ok

mrkkrp:
Yeah, I understood you correctly, but we also access `ctx.label` there, and if we pass `ctx.rule` there is no `ctx.rule.label` and the whole thing breaks.

mrkkrp:
@mboes This is ready for merging.

PR #114, opened 2018-01-31
Fix compilation failures involving TH depending on shared libs

mrkkrp:
Ready.

### DIFF
--- a/haskell/actions.bzl
+++ b/haskell/actions.bzl
@@ -179,7 +179,7 @@ def compile_haskell_lib(ctx):
   c = compilation_defaults(ctx)
   c.args.add([
     "-package-name", "{0}-{1}".format(ctx.attr.name, ctx.attr.version),
-    "-static", "-dynamic-too",
+    "-static", "-dynamic-too"
   ])
 
   object_dyn_files = [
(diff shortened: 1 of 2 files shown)


haskell/actions.bzl
```diff
@@ -367,8 +361,10 @@ def compilation_defaults(ctx):
   # Common flags
   args.add([
     "-no-link",
+    "-fPIC",
```
mboes: Is this required?

haskell/path_utils.bzl
```diff
@@ -84,6 +84,11 @@ def declare_compiled(ctx, src, ext, directory=None):
   fp_with_dir = fp if directory == None else paths.join(directory.basename, fp)
   return ctx.actions.declare_file(fp_with_dir)
 
+# def resolve_symlink(ctx, path):
```
mboes: Dead code. Remove.

haskell/actions.bzl
mrkkrp: Cabal always passes this. HVR says:

> however, it's generally a good idea to have PIC compiled code...

PR #120, opened 2018-02-06
Add support for boot files

### LINKED ISSUE #30: .hs-boot file support
We should have support for .hs-boot files (https://wiki.haskell.org/Mutually_recursive_modules). In current state this likely involves no real new work from us: just allowing these files in `srcs` should be enough. This ticket should only be closed after adding tests with .hs-boot files such that any future changes that are likely to break them (such as incremental builds) are caught and we don't lose the ability to build with these.

mrkkrp:
This PR adds support for compiling mutually recursive modules using `.hs-boot` and `.lhs-boot` files.

Close #30.

### DIFF
--- a/haskell/actions.bzl
+++ b/haskell/actions.bzl
@@ -379,12 +379,15 @@ def compilation_defaults(ctx):
 
   # Common flags
   args.add([
-    "-no-link",
+    "-c",
+    "--make",
     "-fPIC",
     "-hide-all-packages",
     "-odir", objects_dir,
     "-hidir", interfaces_dir,
   ])
+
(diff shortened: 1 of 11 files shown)


haskell/actions.bzl
```diff
@@ -379,13 +379,16 @@ def compilation_defaults(ctx):
 
   # Common flags
   args.add([
-    "-no-link",
+    "-c",
```
mrkkrp: `-c` makes it stop after producing object files, this is what we probably wanted with `-no-link`. @Fuuzetsu says that `-no-link` is probably useless. Cabal passes `-c`, so I decided to pass it too, this may make building a bit faster, but I honestly have no idea if that's the case...

PR #121, opened 2018-02-07
Introduce compilation modes

### LINKED ISSUE #29: Compilation modes
We need support for compilation modes. An easy example is profiling build. I have previously evaluated aspects and they are not suited. I have since found out that Go rules support this (https://github.com/bazelbuild/rules_go/blob/master/go/modes.rst#id1) so we may want to do something similar.

mrkkrp:
This introduces two compilation modes: `profiling` and `coverage`. The modes can be enabled either via setting the corresponding `haskell_binary`/`haskell_library` attributes, on the command line by passing `--features=profiling` and/or `--features=coverage`. Let me know what you think.

Close #29.

### DIFF
--- a/haskell/actions.bzl
+++ b/haskell/actions.bzl
@@ -31,6 +31,10 @@ load(":providers.bzl", "HaskellPackageInfo")
 
 load("@bazel_skylib//:lib.bzl", "paths", "dicts")
 
+load(":mode.bzl",
+     "is_profiling_enabled",
+)
+
 _DefaultCompileInfo = provider(
   doc = "Default compilation files and configuration.",
   fields = {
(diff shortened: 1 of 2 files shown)


Fuuzetsu:
Surely this doesn't _actually_ make profiling or coverage work, only exposes modes which just cause failed builds. For example with -prof you get p_o and p_dyn files too and with coverage you need the tix/mix files.

mboes:
Do we have tests for these modes? Note that features can forced on a per-package basis using the `package()` function call.

I'm not a fan of using features to select the compilation mode, because of what the rules_go folks note:

> Features are generic tags that affect *all* rules, not just the ones you specify or even just the go ones, and any feature can be interpreted by any rule. There is also no protections that two different rules will not intepret the same feature in very different ways, and no way for rule authors to protect against that, so it is up to the user when specifying a feature on the command line to know what it's affects will be on all the rules in their build.

We could name the features `haskell_prof`, `haskell_cov` etc, but then if all rules do it that way profiling would need to be enabled for each language separately, rather than uniformly. I much prefer reinterpreting the semantics of existing mechanisms.

In particular, we could make the `dbg` compilation mode always turn on profiling. With any luck future versions of Bazel may have a separate compilation mode defined for profiling, in which case we'll just use that. And coverage already has its own specific mechanism. See the `bazel coverage` command and https://docs.bazel.build/versions/master/skylark/lib/configuration.html#coverage_enabled.

PR #122, opened 2018-02-07
Add ‘ghc_bindist’

### LINKED ISSUE #31: Support using official GHC bindists
#27 allows users to pull in their own tools to use during build. This lets them use GHC with prebuilt packages being available &c. The default there however uses `rules_nixpkgs`: in general we do not want to force nix on user.

We should provide basic targets for various GHC versions that will use official binary distribution. This will be plenty for users which only need vanilla GHC tooling to start building their project.

Likely related to #24

mrkkrp:
Close #31.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ wih the [setup script below](#setup).
 | Rule | Description |
 | ---: | :--- |
 | [`haskell_repositories`](#haskell_repositories) | Declare [external repositories][external-repositories] needed for rules_haskell |
+| [`ghc_bindist`](#import_ghc_bindist) | Setup a binary distribution of GHC |
 
 **BUILD rules:**
 
--- a/haskell/BUILD
+++ b/haskell/BUILD
@@ -12,6 +12,12 @@ load(
 
 skylark_doc(
   name = "docs",
-  srcs = ["haskell.bzl", "cc.bzl", "haddock.bzl", "toolchain.bzl"],
+  srcs = [
+    "haskell.bzl",
+    "cc.bzl",
+    "haddock.bzl",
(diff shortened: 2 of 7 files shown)


mrkkrp:
@qnikst Suggested that we could download from gentoo repos, at least temporarily.

Fuuzetsu:
> we somehow specify a shell script to execute on archive downloading

We can just use `build_file_contents` to perform arbitrary actions.

> we need to pass an absolute path to configure

Does this path matter after the installation is done? I'm guessing it does but if not then we can just give it the absolute path of target output directory and "install" into there.

>GHC also needs other binaries such as gcc to function properly

Unsure if this is a new problem. You still need this even today, the difference is that you just happen to make it available with GHC from your system (maybe?). So GCC can be a requirement. Actually C/C++ bazel rule fragments can be used for this probably like we do with `ld`.

>we create and host an archive with batteries included (with gcc and stuff). Then we just download that and we should be done

I think that's too much effort on our side, we'd have to do it for every platform and every release we want and it's much harder than adding a single line for a release.

haskell/ghc_bindist_build_template.bzl
```diff
@@ -0,0 +1,15 @@
+package(default_visibility = ["//visibility:public"])
```
mboes: This is not a Skylark source file, so should not have the `.bzl` file extension. These kings of templates are typically called `BUILD.ghc` or somesuch.

mboes:
Alternate name for the rule: simply `ghc_bindist`.

mrkkrp:
@mboes Your feedback has been addressed.

PR #131, opened 2018-02-12
Move the doc generating rule into a separate package

mrkkrp:
With current placement, the following happens when user requires `haskell.bzl`:

```
[mark@arch test-bindist]$ bazel build //...
ERROR: /home/mark/projects/tweag/test-bindist/BUILD:9:1: every rule of type _haskell_toolchain implicitly depends upon the target '@io_tweag_rules_haskell//haskell:ghc-version-check.sh', but this target could not be found because of: error loading package '@io_tweag_rules_haskell//haskell': Extension file not found. Unable to load package for '@io_bazel_skydoc//skylark:skylark.bzl': The repository could not be resolved
ERROR: Analysis of target '//:ghc-impl' failed; build aborted: error loading package '@io_tweag_rules_haskell//haskell': Extension file not found. Unable to load package for '@io_bazel_skydoc//skylark:skylark.bzl': The repository could not be resolved
INFO: Elapsed time: 5.917s
FAILED: Build did NOT complete successfully (6 packages loaded)
    currently loading: @io_tweag_rules_haskell//haskell
```

Bazel is confused about the skydoc repo.

### DIFF
--- a/Dockerfile.web
+++ b/Dockerfile.web
@@ -15,11 +15,12 @@ RUN apt-get update && apt-get install -y curl \
 
 COPY WORKSPACE /bazel/
 COPY haskell /bazel/haskell/
+COPY docs /bazel/docs/
 COPY tests /bazel/tests/
 COPY third_party /bazel/third_party/
 
-RUN cd /bazel; bazel build //haskell:docs
+RUN cd /bazel; bazel build //docs:docs
 RUN cp /bazel/site/start /site \
-    && unzip -d /site /bazel/bazel-bin/haskell/docs-skydoc.zip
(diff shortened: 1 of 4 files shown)


mrkkrp:
I better try that instead, something is wrong with putting `load` there.

Dockerfile.web
```diff
@@ -18,8 +18,8 @@ COPY haskell /bazel/haskell/
 COPY tests /bazel/tests/
```
mboes: You'll probably need an extra `COPY` line, or Heroku will fail.

mrkkrp:
It looks like something went wrong anyway, I don't see the new `ghc_bindist` rule in the docs online. Where can I check what happened?

PR #138, opened 2018-02-14
Generate and provide ‘Paths_*’ modules

### LINKED ISSUE #137: Generate and provide Paths_ modules during compilation
Many projects rely on existence of Cabal-generated `Path_<package_name>` modules which we currently don't emulate. This leads to compilation failures. We should generate such a module and provide it during compilation.

mrkkrp:
Close #137.

I request your comments on this one. It looks like the `version` attribute should be made mandatory. Also, an interesting problem is that the `Paths_*` module depends on `base` because it must provide a value of the type `Version` and so we need to import `Data.Version` there. This makes us make every target of the type `haskell_binary`/`haskell_library` depend implicitly on `base`. May be this is undesirable? CI fails exactly because of missing `base`.


Fuuzetsu:
Personally I don't care all that much about version but care about data files.

w.r.t. relying on base: it ships with GHC, I don't think it's a common usecase where base is not available. I'm sure there might be one but we can deal with it if it ever comes up.

haskell/utils.bzl
```diff
@@ -0,0 +1,38 @@
...
+    String, String:
+      * Filename for the module file.
+      * Textual contents of the file.
+  """
+
+  adjusted_pkgname = name.replace("-", "_")
+
+  contents = """
+  {-# OPTIONS_GHC -fno-warn-implicit-prelude #-}
+  module Paths_"""
```
iphydf: (drive-by comment) I'd probably go for a template with formatting here, i.e.:
```python
contents = """
{{-# OPTIONS_GHC ... #-}}
module Paths_{adjusted_pkgname} (
  version
) where

...

version = Version [{version}] []
""".format({
    "adjusted_pkgname": name.replace("-", "_"),
    "version": version.replace(".", ","),
})
```

mrkkrp: Well I'd probably use `ctx.action.expand_tempalte` now.

iphydf:
Re: depending on base. I think the use of `Paths_` is sufficiently rare that you can make it optional via a parameter to `haskell_library`. So far (in my bazel build), I only found bifunctors using it.

iphydf:
BTW, version is effectively mandatory anyway if you want your build to actually work, since the version you put there will end up in MIN_version macros, affecting the build of dependencies. Often, the default of 1.0.0 works if you use the latest version of everything, because many packages are in the 0.x.x range, but for more mature packages that have versions like 2.x, the MIN_version macros will do the wrong thing.

iphydf:
I hope this will get picked up again. Without this, bifunctors doesn't build, and I'm stuck. Right now I'm using https://github.com/iphydf/rules_haskell which has this PR merged, but I'd rather not maintain a fork for long.

PR #141, opened 2018-02-18
Fix grep not found error

mboes:
At least in sandbox mode, the ghc-defs-cleanup.sh script fails with
a `grep: command not found` error, because `PATH` is sanitized by
Bazel. We set `use_default_shell` to reuse the host shell environment.
This is an impurity, because it makes the build result depend on
what's on the host's PATH. But we already borrow up `/bin/sh` from the
host. Arguably it's also reasonable to borrow `grep`.

### DIFF
--- a/haskell/hsc2hs.bzl
+++ b/haskell/hsc2hs.bzl
@@ -62,6 +62,7 @@ def _make_ghc_defs_dump(ctx):
     inputs     = [ghc_defs_dump_raw],
     outputs    = [ghc_defs_dump],
     executable = ctx.file._ghc_defs_cleanup,
+    use_default_shell_env = True,
     arguments  = [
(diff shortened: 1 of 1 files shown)


johnynek:
You could make a repository rule to find the path to grep and use it I think, without totally opening to the path.

mboes:
I see. Something that we'd do as part of the `haskell_repositories` macro? That sounds fair enough.

mrkkrp:
I don't know about `haskell_repositories`, but it seems logical to define a separate toolchain for this, call it something like `binutils`. This would include all the standard executables like `ln` (we use that too, so it potentially may be missing too), `grep`, etc.

mboes:
Merging, but needs a better fix longer term, as discussed.

PR #147, opened 2018-02-21
Resolve the “file name does not match module name” issue

### LINKED ISSUE #139: base-orphans doesn't build: Saw: `Data.Orphans.Prelude' Expected: `Prelude'
WORKSPACE:
```python
new_http_archive(
    name = "haskell_base_orphans",
    build_file = "third_party/haskell/BUILD.base-orphans",
    sha256 = "c7282aa7516652e6e4a78ccdfb654a99c9da683875748ad5898a3f200be7ad0e",
    strip_prefix = "base-orphans-0.6",
    urls = ["https://hackage.haskell.org/package/base-orphans-0.6/base-orphans-0.6.tar.gz"],
)
```

third_party/haskell/BUILD.base-orphans:
```python
load("@io_tweag_rules_haskell//haskell:haskell.bzl", "haskell_library")

haskell_library(
    name = "base-orphans",
    srcs = glob(["src/**/*.hs*"]),
    src_strip_prefix = "src",
)
```

Error:
```
external/haskell_base_orphans/src/Data/Orphans/Prelude.hs:17:8: error:
    File name does not match module name:
    Saw: `Data.Orphans.Prelude'
    Expected: `Prelude'
Target @haskell_base_orphans//:base-orphans failed to build
```

### LINKED ISSUE #145: inline-java fails to build because of "file name does not match module name" error
With the fix from #144, `inline-java` still fails to compile for me with the following error message:

```
ERROR: /home/inline-java/jvm-streaming/BUILD:9:1: Compiling jvm-streaming failed (Exit 1)

jvm-streaming/src/Language/Java/Streaming.hs:23:8: error:
    File name does not match module name:
    Saw: `Language.Java.Streaming'
    Expected: `Streaming'
   |
23 | module Language.Java.Streaming
   |        ^^^^^^^^^^^^^^^^^^^^^^^
INFO: Elapsed time: 271.846s, Critical Path: 10.71s
```

mrkkrp:
Close #145, close #139.

This resolves the issue for the `jvm-streaming` package but I could not reproduce the problem in isolation in our test suite. It makes sense to assume that the problem should have to do with directory structure/file names/module names. I re-created all of this in my test and upon inspection of actual calls to `ghc`, they were virtually identical (sans concrete module names). Yet in the case of `jvm-streaming` we get the error message complaining about the mismatch, while in the other case the test passes.

I further attempted to bring the two cases closer by commenting out `deps` and `prebuilt_dependencies` in the BUILD file of `jvm-streaming`, thinking that since these fields are present in `jvm-steraming` and not in my test, they may somehow make the difference, but that also did not help. `jvm-streaming` fails and my test passes.

I'm out of ideas, it looks like an inconsistent behavior of GHC.

### DIFF
--- a/haskell/actions.bzl
+++ b/haskell/actions.bzl
@@ -4,6 +4,7 @@ load(":path_utils.bzl",
      "declare_compiled",
      "target_unique_name",
      "module_name",
+     "import_hierarchy_root",
 )
 
 load(":set.bzl", "set")
@@ -505,10 +506,9 @@ def compilation_defaults(ctx):
 
   _add_mode_options(ctx, args)
 
-  for idir in set.to_list(set.from_list([f.dirname for f in sources])):
-    items = ["-i{0}".format(idir)]
-    args.add(items)
-    haddock_args.add(items, before_each="--optghc")
+  ih_root_arg = ["-i{0}".format(import_hierarchy_root(ctx))]
+  args.add(ih_root_arg)
+  haddock_args.add(ih_root_arg, before_each="--optghc")
 
   dep_info = gather_dependency_information(ctx)
 
--- a/haskell/path_utils.bzl
+++ b/haskell/path_utils.bzl
@@ -84,6 +84,25 @@ def declare_compiled(ctx, src, ext, directory=None):
   fp_with_dir = fp if directory == None else paths.join(directory.basename, fp)
   return ctx.actions.declare_file(fp_with_dir)
 
+def import_hierarchy_root(ctx):
+  """Return relative path to root of module hierarchy.
+
+  Args:
+    ctx: Rule context.
+
+  Returns:
+    string: Relative path to root of module hierarchy.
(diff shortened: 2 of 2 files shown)


mrkkrp:
This also happens to resolve #139. I'll take another look maybe I can manage to reproduce it in isolation by taking inspiration from 139 maybe...

iphydf:
Fix for #139 confirmed.

mrkkrp:
I have an idea which I got after reading GHC user manual, and that's how I came up with the solution. The idea is that GHC takes import directories and tries to add file names to the end, e.g. `Foo/Bar/` plus `Baz.hs` and if there is such a file, then it thinks that it's `Baz` module. So we need to force it to add the whole sub-path so the full hierarchy goes into module name by specifying only root of source tree with `-i`. This is the best understanding I have so far. Maybe incorrect.

PR #109, opened 2018-01-29
Set experiments

mrkkrp:
Close #77.

PR #149, opened 2018-02-21
Preserve ordering of static libraries to help the linker

### LINKED ISSUE #140: Linked libraries (-optl) not topologically sorted in linking step
I don't have a simple reproducing example at the moment. If necessary I can build one later. The [BUILD.bazel](https://github.com/iphydf/hs-msgpack/blob/bazel-test/BUILD.bazel#L20-L40) file in hs-msgpack contains the test. All the dependencies are in toktok-stack (e.g. [HUnit](https://github.com/iphydf/toktok-stack/blob/testing/WORKSPACE#L47)).

mrkkrp:
Close #140.

Apparently this is a regression introduced in #109. I noticed that the test `two-libs` is not in the test `BUILD` files and so it was essentially redundant because it was not run on `bazel test //...`. I adjusted the test to catch this issue and added it to the test suite.

### DIFF
--- a/haskell/actions.bzl
+++ b/haskell/actions.bzl
@@ -228,12 +228,11 @@ def link_haskell_bin(ctx, object_files):
   # directly rather than doing multiple reversals with temporary
   # lists.
   link_paths = {}
-  static_lib_list = set.to_list(dep_info.static_libraries)
 
-  for lib in static_lib_list:
+  for lib in dep_info.static_libraries:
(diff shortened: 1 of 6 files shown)


Fuuzetsu:
Technically two-libs was a test for something else but if it works for this usecase then great.

mrkkrp:
Well, I just re-used the name, it's a different test now.

PR #153, opened 2018-02-22
Locate binary utilities

mrkkrp:
*Needs an early review before I send more time polishing it, maybe it's not worth it.*

----

OK, this turned out not as nice as I imagined because of certain constraints imposed by Bazel.

We only can detect locations of executables like `ln` and `grep` in `repository_rule` implementation because only this way we can get the context that allows to inspect output of shell commands, and it even has `ctx.which`, which is exactly want we want here.

`repository_rule` necessarily creates a repository. We could put this in `haskell_repositories` but it is not really a good idea because this way we would need to hardcode its name and we'll refer to the repo by its name later, so it looks like it should be configurable. We could of course take that name as an argument of `haskell_repositories` but then why not to provide the repository-creating rule as a separate thing? This is what I do.

Another constraint is that the info about locations of the tools should be put into a toolchain because we have only two ways to communicate with build rules: 1) via attributes 2) via toolchains.

The created repo contains a function that allows us to create a toolchain which provides the info we're interested in. We can't register it automatically in e.g. `haskell_repositories` (making the boilerplate
invisible to the end user), because `register_toolchains` expects target local to current repository, which makes sense. (In general we can't put this in `WORKSPACE` file because it's not a place for normal rules to be in.)

So the solution adds some boilerplate:

* In `WORKSPACE` file user must create a new special repo using `binutils_repository`.
* In `BUILD` file user must import `binutils_toolchain` from that repo and use it to create a toolchain.
* Again in `WORKSPACE` file user must register the toolchain alongside with GHC toolchain.

Now that we rely on that new toolchain (and specify it as required for our rules) the procedure becomes obligatory.

----

The best solution I can imagine is that Bazel adds a new fragment that will provide us with full paths to all standard utilities. This way we just specify that our rules need to access that fragment and we're done. Just in case, I have opened an issue requesting the feature:

https://github.com/bazelbuild/bazel/issues/4681

Maybe it'll be added, although I'm not particularly optimistic about that given how many more serious issues they have.

----

So let me know what you think. The code seems to work, except CI fails a bit, that I'll sure fix if we decide to proceed with this solution.


mboes:
> bazelbuild/bazel#4681
> 
> Maybe it'll be added, although I'm not particularly optimistic about that given how many more serious issues they have.

It turns out, they gave it P1 priority. So it could well happen!

haskell/binutils-repo.bzl
```diff
@@ -0,0 +1,39 @@
+"""Rule for a repo containing an auxiliary toolchain with binutils."""
+
+def _locate_tool(ctx, tool_name):
+  full_path = ctx.which(tool_name)
+  if full_path == None:
+    fail("Could not locale {0}, make sure it's on PATH".format(tool_name))
```
johnynek: do you mean `locate` here?

mrkkrp: Yes, this certainly has nothing to do with locale! :-)

PR #154, opened 2018-02-26
Fix location of shared libraries by test executables

### LINKED ISSUE #151: Dynamic linking failure (at runtime) with cbits dependency
https://travis-ci.org/iphydf/repro140/jobs/344336314#L740

==================== Test output for @haskell_dlist//:test:
/home/travis/.cache/bazel/_bazel_travis/62cd4b4156760703e80e26252af7b00f/bazel-sandbox/1857824427084396710/execroot/repro140/bazel-out/k8-fastbuild/bin/external/haskell_dlist/test.runfiles/repro140/external/haskell_dlist/test: error while loading shared libraries: bazel-out/k8-fastbuild/bin/external/haskell_tf_random/libexternal_Shaskell_Utf_Urandom_Scbits_libcbits.so: cannot open shared object file: No such file or directory

mrkkrp:
Close #151.

This PR makes test executables locate shared libraries (of cbits in particular) correctly so tests can be run. Also we adjust the test suite to catch the issue.

### DIFF
--- a/haskell/actions.bzl
+++ b/haskell/actions.bzl
@@ -135,6 +135,23 @@ def _is_shared_library(f):
   """
   return f.extension == "so" or f.basename.find(".so.") != -1
 
+def _add_external_libraries(args, libs):
+  """Add options to `args` that allow us to link to `libs`.
+
+  Args:
+    args: Args object.
+    libs: set of external shared libraries.
+  """
+  seen_libs = set.empty()
+  for lib in set.to_list(libs):
+    lib_name = _get_lib_name(lib)
+    if not set.is_member(seen_libs, lib_name):
+      set.mutable_insert(seen_libs, lib_name)
+      args.add([
+        "-l{0}".format(lib_name),
+        "-L{0}".format(paths.dirname(lib.path)),
+      ])
+
 def _add_mode_options(ctx, args):
   """Add mode options to the given args object.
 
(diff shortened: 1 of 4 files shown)


mrkkrp:
@iphydf Could you please check if this fixes the problem for you? (Although I'm pretty sure it does, from my tests.)

mrkkrp:
I did not see that they do the same. I have come up with the scheme myself. The trick with setting rpath by itself is pretty common, Cabal does that too.

mboes:
I feel unconfortable *not* duplicating wholesale the approach Bazel used in the CC rules. Our case is exactly the same as their case. If we're not doing the same, we should sensibly make the diff explicit. Code adding `-rpath` in Bazel:

https://github.com/bazelbuild/bazel/blob/f98a7a2fedb3e714cef1038dcb85f83731150246/src/main/java/com/google/devtools/build/lib/rules/cpp/CppActionConfigs.java#L587-L605
https://github.com/bazelbuild/bazel/blob/ae4ba82f58034b1875f918965bf781376f89505a/src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java#L71-L73
https://github.com/bazelbuild/bazel/blob/ae4ba82f58034b1875f918965bf781376f89505a/src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java#L1808

iphydf:
Confirmed: this makes my test run.

PR #156, opened 2018-02-27
Add ‘hidden_modules’ attribute to ‘haskell_library’ rule

### LINKED ISSUE #152: How to prevent exposing modules from a haskell_library?
The aeson package has a private Data.Aeson.Compat, in cabal "other-modules". If exposed, it clashes with the aeson-compat package exposing the same module name.

mrkkrp:
Close #152.

This allows to selectively hide modules, which is necessary in order to compile some existing code from Hackage.

### DIFF
--- a/haskell/actions.bzl
+++ b/haskell/actions.bzl
@@ -447,6 +447,16 @@ def create_ghc_package(ctx, interfaces_dir, static_library, dynamic_library):
   conf_file = ctx.actions.declare_file(paths.join(pkg_db_dir.basename, "{0}.conf".format(get_pkg_id(ctx))))
   cache_file = ctx.actions.declare_file("package.cache", sibling=conf_file)
 
+  # Infer collection of public modules in the library.
+
+  hidden_modules = set.from_list(ctx.attr.hidden_modules)
+  public_modules = []
+
+  for f in _hs_srcs(ctx):
+    mname = module_name(ctx, f)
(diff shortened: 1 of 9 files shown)


haskell/haskell.bzl
```diff
@@ -43,7 +43,7 @@ _haskell_common_attrs = {
   ),
   "srcs": attr.label_list(
     allow_files = FileType([".hs", ".hsc", ".lhs", ".hs-boot", ".lhs-boot", ".h"]),
-    doc = "Haskell source files",
+    doc = "Haskell source files.",
```
mrkkrp: All doc strings end with a full stop except this one, so I could not resist.

haskell/actions.bzl
```diff
@@ -455,8 +465,7 @@ def create_ghc_package(ctx, interfaces_dir, static_library, dynamic_library):
     "id": get_pkg_id(ctx),
     "key": get_pkg_id(ctx),
     "exposed": "True",
-    "exposed-modules":
-      " ".join([module_name(ctx, f) for f in _hs_srcs(ctx)]),
+    "exposed-modules": " ".join(public_modules),
```
Fuuzetsu: Why are we not listing hidden modules in `hidden-modules` field? I think we should have it.

mrkkrp: Good catch, adding.

PR #159, opened 2018-02-28
Add ‘haskell_repl’ rule

### LINKED ISSUE #82: Start an appropriate ghci from Bazel
Cabal and stack have a `ghci` subcommand (_e.g._ `stack ghci`) which loads the appropriate environment (compiler, dependencies, runs within docker or nix when appropriate) as well as the files for the project.

This is used by most IDEs (`intero`, `dante`, `ghcid`) to provide feedback to the programmer (type errors, type informations, completion options, etc…).

In order to provide compatibility with these IDEs, there should be an equivalent invocation of `ghci` with Bazel and rules_haskell, which uses the `WORKSPACE` and `BUILD`.

mrkkrp:
Close #82.

The rule allows to create a script that runs GHCi with all explicit dependencies in scope, similar to `scala_repl` from `rules_scala`.

### DIFF
--- a/docs/BUILD
+++ b/docs/BUILD
@@ -12,6 +12,7 @@ skylark_doc(
 
     "//haskell:haskell.bzl",
     "//haskell:haddock.bzl",
+    "//haskell:ghci-repl.bzl",
     "//haskell:toolchain.bzl",
     "//haskell:cc.bzl",
     "//haskell:repositories.bzl",
--- a/haskell/BUILD
+++ b/haskell/BUILD
@@ -8,8 +8,11 @@ exports_files([
   "haddock.bzl",
   "toolchain.bzl",
   "ghc_bindist.bzl",
+  "ghci-repl.bzl",
 
   "ghc-defs-cleanup.sh",
   "ghc-version-check.sh",
+  "ghci-repl-wrapper.sh",
+  "ghci-script",
 
 ])
--- a/haskell/actions.bzl
+++ b/haskell/actions.bzl
@@ -5,6 +5,7 @@ load(":path_utils.bzl",
      "target_unique_name",
      "module_name",
      "import_hierarchy_root",
+     "get_external_libs_path",
 )
 
 load(":set.bzl", "set")
@@ -39,6 +40,10 @@ load(":mode.bzl",
      "is_profiling_enabled",
 )
 
+load(":utils.bzl",
+     "get_lib_name",
+)
+
 _DefaultCompileInfo = provider(
   doc = "Default compilation files and configuration.",
   fields = {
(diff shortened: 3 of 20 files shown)


Fuuzetsu:
As I understand this will build all dependencies then start GHCi; but does it work for development? For example if I want to hack on something: I don't want to build it fully just to load it in GHCi, I want to change it at some point.

I think what I'm asking for is a blurb in README about this with sample usage for development. Possibly this is just as easy as passing `-I` to source locations.

mrkkrp:
Good point, I need to check if just passing `-I` is enough. Maybe a mixed approach is the best though. Imagine we depend on both `aeson` and `aeson-compat`. If we go just with `-I` there will be conflict between modules. So it looks like `aeson` should be really compiled and loaded via `-package aeson -package-db path/to/cache`. Maybe we should have the `deps` attribute as well as `deps_dev` or something like that.

mrkkrp:
Another feature would be having an option where user could specify modules to import on startup.

Fuuzetsu:
I mean -I to sources we're changing. Dependencies stay as is.

mrkkrp:
@Fuuzetsu I made the REPL more development-friendly. Please take a look.

mrkkrp:
This may also require handling of boot files and shared libraries. But in that case we're going to duplicate a lot of logic from `haskell_library` (which is why the original version with only `deps` may be better in a way, it's easy to make an incomplete module compile after all).

Also: passing `-i` is not enough, we need to pass actual files too or else they won't be available for import. In that case GHCi loads them immediately (but only original `hs` files, not those obtained from `hsc` files through pre-processing). This may be undesirable because it may lead to name clashes.

Fuuzetsu:
Doesn't sound too great ;/ But I'll have to try it before I can make an opinion.

mrkkrp:
OK, here is an update:

* `stack repl` is not much better with respect to end-user experience. It re-compiles everything and then loads all files via a temporary config file with `:add` and `:module +` commands followed by module names. I think this is not good because 1) if something fails to compile, then you'll get partial result (not all modules loaded) and in my case it loaded not the modules that work without the failing file, but rather one seemingly random module 2) loading all modules leads to name clashes, not at the loading time but when you try to actually use ambiguous names. I'd prefer to tell such a tool that I want to interact with such and such packages and then I'd load/import what I really want manually.

* Because of the previous point, and to avoid duplication of the logic from `actions.bzl` I think we should just compile all dependencies specified in `haskell_repl` rule and then load them using `-package` and `-package-db`. This hides/separates the compilation logic from repl and also we don't need to think about boot files, hsc stuff, etc. I think it's the right way to do it: it's easier to implement, easier to maintain, and it does the trick just fine IMO.

* And how to call it. I think we should recommend `bazel run //my/package:my-repl-target`. It's natural and makes sense, better than telling the user to enter a relative path to the created script (not sure why `rules_scala` does it that way).  This also may make the `pwd` hack in the wrapper script unnecessary (right now `bazel run` doesn't work with it, of course). To make it work this way we also need to specify dynamic libraries in `DefaultInfo` returned by `haskell_library` because with current approach we can't get libraries to work as runfiles because we only get conf and cache files but not the other stuff necessary for them to work (interface files, dynamic/static libs).

* I actually expect that some corrections will be necessary in order to teach .so libs how to find cbits with this setup, similar to what we did for test executables (will have to set rpath for .so files of Haskell libraries too). But the same approach should work here just fine.

Opinions?

johnynek:
Rules scala does it that way because bazel closes stdin because it is a c++ program that just relays commands to a java backend.

I wish they would address this and connect the tty over the network, but the best we have is the run_script option (I think it’s called, but I’m on a phone now).

mrkkrp:
@johnynek Ah I keep forgetting about that. OK, then we have no choice, we'll have to tell our users to run the script.

mboes:
@mrkkrp Short term, yes. But I think it's worth putting forth a design doc describing our use case and having that changed. See https://github.com/bazelbuild/bazel/issues/1394.

mrkkrp:
Pushed some updates implementing what I described in the last update. This version also handles dynamic libs better, i.e. it works with Haskell libs with cbits. It makes libraries specified in `deps` available and you then can `:load` or `import` modules of interest. Read for merge imo.

mrkkrp:
Let's consider this ready. @mboes, @Fuuzetsu, should we merge?

mboes:
Can we get a green light from at least one of the reviewers? The basic requirement for this PR is only: all compiled deps (including foreign deps) can successfully be loaded in GHCi. e.g. if my project defines a service, I should be able to start it in an HTTP server from the REPL.

Fuuzetsu:
I was busy. Will try to approve/disapprove today.

Issue #179, opened 2018-03-12
cc_haskell_import should work with haskell_binary


mboes:
Given a `haskell_library`, it's currently possible to export it for consumption by the CC rules using `cc_haskell_import`. But a `haskell_binary` sometimes looks pretty similar to a `haskell_library`. It might not have a `ghc-pkg` database entry associated to it, but if it's compiled with `-dynamic -pie`, then it looks and smells so much like any odd shared library that the `file` command would be fooled. We use this property in sparkle, where we routinely compile PIE binaries and then `dlopen()` them like any other shared object.

This problem perhaps points to a deeper one: `HaskellPackageInfo` is likely too monolithic. It could make sense to split it out into multiple providers focused on specific aspects of the produced artifact, much like the Go rules do (`GoLibrary`, `GoArchive`, `GoArchiveData` etc). See https://github.com/bazelbuild/rules_go/blob/master/go/providers.rst.

mboes:
The original problem was fixed pretty much entirely, as a side effect of #186. There does remain another problem, which is that `cc_haskell_import` should ensure the result is always named as a library. Otherwise `cc_*` rules won't accept the output in the `srcs` attribute.

PR #186, opened 2018-03-15
Provider refactoring, make REPL work with bin targets

### LINKED ISSUE #182: REPL for binaries does not work when the 'Main' file is not 'Main.hs'
Steps to reproduce:

- Create a `haskell_binary` target with a `main_file` different than `Main.hs`.
- Create a `haskell_repl` target for the `haskell_binary` target.
- Build the REPL target and launch the GHCi script.

Example:
```
[nix-shell:~/code/projectx]$ ./bazel-bin/svgs-repl
GHCi, version 8.2.2: http://www.haskell.org/ghc/  :? for help

test/ProjectX/Test/GenerateSVGs.hs:1:1: error:
    File name does not match module name:
    Saw: ‘Main’
    Expected: ‘ProjectX.Test.GenerateSVGs’
  |
1 | {-# LANGUAGE ViewPatterns #-}
  | ^
Failed, no modules loaded.
```

mrkkrp:
Close #182.

That is, binary targets with `Main` module located in files named differently than `Main.hs`. The refactoring also prepares ground for work on #179.


haskell/actions.bzl
```diff
@@ -633,88 +642,121 @@ def _get_library_name(ctx):
...
-        external_libraries = set.mutable_union(hpi.external_libraries, pkg.external_libraries),
-        import_dirs = hpi.import_dirs,
-        exposed_modules = hpi.exposed_modules,
-        hidden_modules = hpi.hidden_modules,
-        haddock_ghc_args = None,
+    if HaskellBuildInfo in dep:
+      binfo = dep[HaskellBuildInfo]
+      package_names = acc.package_names
+      if HaskellBinaryInfo in dep:
+        fail("Target {0} cannot depend on binary".format(ctx.attr.name))
```
Fuuzetsu: I actually think we should allow binaries as input somehow. For example if you need additional build tool available at build time to pre-process some files. Having said that, it has nothing to do with this PR, I was just reminded of it here. We might want to open an issue (or ignore it for now though I on a client project we have precisely such a need so current rules_haskell could not be used there).

mrkkrp: Let's open an issue for this.

Issue #185, opened 2018-03-15
example for depending on a haskell nix package


ghorn:
The only example I could find was in this https://github.com/tweag/rules_haskell/issues/95 ~which doesn't work for me~.

ghorn:
FYI the build error is
```
$ bazel build //:vector
ERROR: /home/greg/butfor/BUILD:8:1: no such package '@vector-toolchain//': Cannot build Nix attribute vector-toolchain. and referenced by '//:vector-toolchain-impl'
ERROR: Analysis of target '//:vector' failed; build aborted: no such package '@vector-toolchain//': Cannot build Nix attribute vector-toolchain.
INFO: Elapsed time: 21.710s
FAILED: Build did NOT complete successfully (7 packages loaded)
```

ghorn:
Sorry I misunderstood the example and the build failure I pasted is irrelevant. I can contribute an example once I finish getting it working.

mboes:
I just added a [rules_haskell_examples](https://github.com/tweag/rules_haskell_examples) repo. Does that help? There's now also a link to this repository from the top-level README.

ghorn:
That new repo doesn't address how to use a prebuilt haskell package, but I made a pull request with the example that I wanted.

I'm brand new to bazel and put it in a new `examples/` folder instead of properly integrating it into `tests/`. Not sure what you want to do with it.

mboes:
Notice that the `vector` example does have `prebuild_dependencies`. You can see this in particular with the  `primitive` and `semigroupoid` packages, which don't ship with GHC, but we create a toolchain with those two packages prebuild by Nixpkgs. I'm open to better ways of presenting that: perhaps as a separate example.

ghorn:
I see, you're totally right. I've been staring at this package for too long.

Maybe the best thing would be waiting for the project to stabilize a bit, and then spending a bit of time on on documentation. I'd be happy to help.

PR #188, opened 2018-03-16
Add example of using prebuilt vector

ghorn:
addresses #185


mboes:
Could you resubmit this to [rules_haskell](https://github.com/tweag/rules_haskell_examples) repo? The reason we have a separate repo for examples is that Bazel does not recognize that your example directory is a separate workspace from the top-level one. So `bazel build //...` in the top-level directory might fail because it will be trying to build targets in the examples folder using the wrong `WORKSPACE` file.

ghorn:
can do

mboes:
OK closing in this repo then.

PR #195, opened 2018-03-22
Document how to load ghc_bindist rule

ghorn:
Not sure if this is the right place to put this, but it took me 15 minutes to find in the issue tracker so may be useful for others.

### DIFF
--- a/haskell/ghc_bindist.bzl
+++ b/haskell/ghc_bindist.bzl
@@ -80,6 +80,7 @@ Example:
 
    ```bzl
    # This repository rule creates @ghc repository.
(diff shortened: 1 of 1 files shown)


haskell/ghc_bindist.bzl
```diff
@@ -80,6 +80,7 @@ Example:
 
    ```bzl
    # This repository rule creates @ghc repository.
+   load("@io_tweag_rules_haskell//haskell:ghc_bindist.bzl", "ghc_bindist")
```
mrkkrp: Well, I'd actually recommend this instead:

```
load("@io_tweag_rules_haskell//haskell:haskell.bzl", "ghc_bindist")
```

mboes:
Thanks!

Issue #194, opened 2018-03-21
is it possible to run doctests?


ghorn:

mrkkrp:
I guess this, in its core, the same sort of thing as linting we've discussed in #181. I did some progress on linting, but so far there is a problem with aspect-based approach: a rule must have outputs (files), and such rules typically do not have outputs (in fact having outputs is undesirable in this case).

mboes:
Right. So the short answer is, "not yet". Once we have #181 sorted out (on which we do know the path forward), the same approach should work here.

PR #196, opened 2018-03-23
Haskell lint

### LINKED ISSUE #181: Passing additional compiler flags to builds
It is sometimes desirable to pass different flags to the compiler depending on context. A good example is `-Werror` (and related options) ; in our previous setup, the CI would use Stack's `--pedantic` flag to turn warning as errors, while we would simply use `stack build` during development.

It would be nice if `rules_haskell` gave us a way to achieve the same thing, either through a command-line parameter, or via environment variables.

mrkkrp:
Close #181.

This PR adds `haskell_lint` rule which allows to lint Haskell targets.

Let's review and merge #191 first. Since these PRs happen to deal with related chunks of code, I could not start this branch from `master` and had to start it from branch of that PR instead so I could continue with newer definitions and avoid emotional trauma.

Notes:

* I could not find how to write a test that expects a certain rule to fail. Must be not very common use case with Bazel. I added a test package for manual testing, see my comment there.
* There is also quite a bit of refactoring that I wanted to do for a while. Now there is only one place where we form set of source files (taking into account pre-processed files from hsc2hs, `_hs_srcs` thus removed) and outputs, so it should be a bit more maintainable.

Now we can actually add support for doctests as requested in #194.


mrkkrp:
Forgot to mention, due to https://ghc.haskell.org/trac/ghc/ticket/14954#ticket, some warnings are not repoted, at least incomplete patterns. I tried to use `-S` as a workaround, but alas, I was told by GHC that `-S` cannot be used with `--make` <shrug>.

Fuuzetsu:
@mboes knew/had an idea on how to test for failures though I don't remember what it was, maybe he can comment

mboes:
rebase?

mrkkrp:
Done.

mboes:
@mrkkrp As concerns the manual test, it would be good to turn that into an automatic test. Have you had a look at https://github.com/tweag/rules_haskell/blob/master/third_party/test_rules.bzl#L54-L87 in our repository? That might be what you're looking for, but have never tried it.

mrkkrp:
Yeah, I looked at all test rules in this file, but concluded that they are inapplicable. The rule seem to create a script that outputs given message and then fails.

PR #204, opened 2018-04-08
start: Fix rules_haskell strip_prefix for 0.4

patrickmn:
`bazel build` would previously fail with

`ERROR: error loading package '': Encountered error while reading extension file 'haskell/repositories.bzl': no such package '@io_tweag_rules_haskell//haskell': Prefix rules_haskell-0.3 was given, but not found in the archive`


mboes:
Thanks!

mboes:
We now check this in CI: 05127f8f6109552f782e8c4cfb92da007d919b9c

PR #210, opened 2018-04-10
Add support for macOS.

judah:
This resolves #165, except that it statically links Haskell dependencies
on macOS (versus dynamically on Linux).  Since everything else seemed
to be working, I figured that was easier to do as a follow-up.

Summary of changes:
- Removed the Linux-specific toolchain constraint.  Not sure if that's the best
  approach; I'm not as familiar with toolchains.
- Don't use `realpath` in the toolchain definition since it's not available
  by default on macOS.
- macOS shared libraries have the externsion "dylib" rather than "so".
- The `cc_haskell_import` test assumed that `Int`/`HsInt` was 32-bit, via use of
  `printf("...%d...")`.  I changed the test to use `Int32` explicitly.
- Some of the binary dependencies (e.g.: ar, gcc) are wrapper scripts
  generated by `bazel` (`ar_wrapper.sh`, `cc_wrapper.sh`) and provided
  via `//tools/defaults:crosstool`.  Additionally,
  `ar_wrapper.sh` depends on another crosstool script, `xcwrapper.sh`.

PR #215, opened 2018-04-12
Fix ghc_bindist, and link to macOS distributions.

judah:
It was broken in #210.

I tested it by modifying `WORKSPACE` and `tests/BUILD` to refer to a
`ghc_bindist` rule, and then running a command which previously failed:
```
//tests/cc_haskell_import:all
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2
 
 jobs:
-  build:
+  build-linux:
     docker:
       - image: nixos/nix
     working_directory: ~/rules_haskell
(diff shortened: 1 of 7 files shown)


judah:
FYI: I was unable to get the bindist on Linux working in the Docker container I was using; the errors did appear unrelated to this change, so it was probably something to do with my setup.  (And I did observe the Bazel code evaluate successfully to the correct download.)

mboes:
Could you rebase atop current master to get a macOS CI check, now that #214 is merged?

PR #222, opened 2018-04-30
Take Bazel package names into account when generating GHC package names

### LINKED ISSUE #219: Conflict between haskell_libraries with the same name in different directories
Example:
https://github.com/tweag/rules_haskell/commit/90390580ed314716bfb532cb3a25ea2262586a4a

A binary depends on two `haskell_library` rules with the same name: `tests/library-deps/sublib:sublib` and `tests/library-deps:sublib`.  This fails:
```
bazel build tests/library-deps:bin
```
It conflicts with tests/library-deps/sublib:sublib; in fact,
renaming one of the `sublib` rules to, e.g., `...:sublibA` lets the build succeed.

The relevant code is here: https://github.com/tweag/rules_haskell/blob/9a2ba48e737ae06e3d448d8e4baf1cdd36dbc28c/haskell/actions.bzl#L784

We should be able to change that function to make a package name that includes the ctx.label's package and workspace_root. (I'm not sure at the moment whether we'd need to escape slashes like we do here: https://github.com/tweag/rules_haskell/blob/9a2ba48e737ae06e3d448d8e4baf1cdd36dbc28c/haskell/actions.bzl#L88)

mrkkrp:
Close #219.

In short, there are many possible approaches to do encoding. The key, it seems, is to take a rare character and use it as escape character remembering to escape the escape character itself to achieve injectivity. With @mboes We discussed several possibilities and I implemented an encoding scheme where escaping character is zero `0`.

So we use `0R` to separate workspace root from (Bazel) package name and `0D` to separate package name from target name. In addition to that we escape zero `0` as `00`, underscore `_` as `0U` and slash `/` as `0S`. This should work.

We also discussed using dash, perhaps something symmetric like `-U-`, but then two underscores in a row would result in an invalid package name, because GHC does not like underscores or more than one dash in a row.

@mboes also proposed to prefix all digits with `0` because this is less-surprising (i.e. we escape 0, and so we should escape all digits for consistency).

@judah What do you think?

### DIFF
--- a/haskell/actions.bzl
+++ b/haskell/actions.bzl
@@ -846,7 +846,7 @@ def _compilation_defaults(ctx):
     ),
   )
 
-def _zero_escape(s):
+def _zencode(s):
   """Zero-escape a given string.
 
   Args:
@@ -855,7 +855,7 @@ def _zero_escape(s):
   Returns:
     string: zero-escaped string.
   """
-  return s.replace("0", "00").replace("_", "0U").replace("/", "0S")
+  return s.replace("Z", "ZZ").replace("_", "ZU").replace("/", "ZS")
 
 def get_pkg_name(ctx):
   """Get package name. Package name includes Bazel package and name of the
@@ -868,10 +868,11 @@ def get_pkg_name(ctx):
   Returns:
     string: GHC package name to use.
   """
-  return "{0}0R{1}0D{2}".format(
-    _zero_escape(ctx.label.workspace_root),
-    _zero_escape(ctx.label.package),
-    ctx.attr.name,
+  return _zencode("%s/%s/%s" %
+                  [ctx.label.workspace_root,
+                   ctx.label.package,
+                   ctx.attr.name,
+                  ]
   )
 
 def get_pkg_id(ctx):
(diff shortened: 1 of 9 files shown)


mboes:
Note for @judah: ghc package names do leak to the user, so if we prefix them and if that requires some encoding, then we need to choose an encoding that's as readable as reasonably possible. These names at least leak in the Haddock HTML indexes, possibly elsewhere. If this weren't the case, any encoding no matter how ugly would do. 

After "sleeping on it", I think it's worth considering these alternative nuances to the bikeshed's colour:

* The rarer the escape character, the more readable because any occurrence of the escape character needs to be double escaped. Dates and version components might appear in folder names, so arguably any capital letter is rarer than any digit. We're restricted to `[a-zA-Z0-9-]` (no underscore) as characters in package names, so using `Z` as the escape character like GHC does could work better.
* I believe the `0R` and `0D` escapes are unnecessary. All fully qualified packages names are of the following form: `external/repo/a/b/c`, or `a/b/c` if not in an external repo. So just escaping the `/`'s as per usual in this fully qualified path should be sufficient.

If we can hide these names to the user in all instances, this all becomes moot.

PR #223, opened 2018-04-30
Fix build breakage on macOS when XCode is not installed.

judah:
If the macOS "command-line developer tools" are installed, but XCode
is not, then Bazel will make `ar_executable` point to `/usr/bin/libtool`
(despite the fact that `ar` and `libtool` are not invoked the same way).
https://github.com/bazelbuild/bazel/blob/71932dd4e25d5e755cb8ce12f4dece438c4b5cb1/tools/cpp/unix_cc_configure.bzl#L101

Fix it by checking for that case and overriding it to point to `/usr/bin/ar`
(which is a path also hard-coded in the above code).  This is not ideal, but I don't
know a good way around it.

Note that `rules_rust` has done a similar workaround:
https://github.com/bazelbuild/rules_rust/blob/df95c3e3cd5afd87a69fa71dc9a56a0d0baa7823/rust/toolchain.bzl#L18

Filed an upstream bug: https://github.com/bazelbuild/bazel/issues/5127

### DIFF
--- a/haskell/toolchain.bzl
+++ b/haskell/toolchain.bzl
@@ -74,8 +74,9 @@ def _haskell_toolchain_impl(ctx):
 
   # If running on darwin but XCode is not installed (i.e., only the Command
   # Line Tools are available), then Bazel will make ar_executable point to
-  # "/usr/bin/libtool".  Since we call ar directly, override it.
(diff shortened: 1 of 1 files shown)


mboes:
Is there an upstream ticket for this Bazel misfeature, one that you could point to in the comment?

PR #225, opened 2018-05-02
Define an implicit REPL target per binary/library target

### LINKED ISSUE #220: Add a "haskell_repl" executable
Currently, in order to use a REPL, we need to write explicit `haskell_repl` rules.  Compare that to the UX of stack or cabal, where you can write `stack repl some-package`, `stack repl some-package:exe:some-binary`, etc., and automatically load a session with all the source files and dependencies for a particular target.

An alternate approach would be for `rules_haskell` to provide a single, generic script that takes a Bazel target (or targets) as a command-line argument, for example:
```
bazel build @io_tweag_rules_haskell//haskell_repl
bazel-bin/external/io_tweag_rules_haskell/haskell_repl/haskell_repl //some/haskell:target
```

One approach would be to write a version of `haskell_repl`  which is an [aspect](https://docs.bazel.build/versions/master/skylark/aspects.html#invoking-the-aspect-using-the-command-line), and then make the above script call `bazel build //some/haskell:target --aspects ...` to generate the script that the explicit rule produces today.

Another approach would be to use an [OutputGroup](https://docs.bazel.build/versions/master/skylark/rules.html#requesting-output-files) for each `haskell_library`/`haskell_binary`, which contains source files and compilation info (similar to what we pass through the existing providers (`HaskellBuildInfo`, `HaskellLibraryInfo`).

mrkkrp:
Close #220.

This PR removes the standalone `haskell_repl` rule and instead adds optional REPL outputs to every `haskell_binary` and `haskell_library` targets. The REPL outputs are not built unless we explicitly ask Bazel to build them and we can do so for any target. See the updated docs for more information how to use it, but I think it should be quite straightforward.


haskell/haskell.bzl
```diff
@@ -138,6 +168,9 @@ def _mk_binary_rule(**kwargs):
         doc = "File containing `Main` module.",
       )
     ),
+    outputs = {
+      "repl": "%{name}-repl",
```
mboes: Bazel conventions seem to pretty consistently favour [snake case](https://en.wikipedia.org/wiki/Snake_case). I think we should stick to Bazel conventions insofar as they exist for auto-generated target names.

.circleci/config.yml
```diff
@@ -22,10 +22,14 @@ jobs:
           command: bazel test //... --config=ci
       - run:
           name: Test REPL for libraries
-          command: bazel-bin/tests/haskell_repl/haskell_lib_repl -e ":quit"
+          command: |
+            bazel build //tests/repl-targets:hs-lib-repl
```
mboes: Does this have to be a two-line sequence, or as a user, can I `bazel run --direct_run //tests/repl-targets:hs-lib-repl`?

haskell/haskell.bzl
```diff
@@ -163,6 +196,11 @@ Example:
       deps = ["//lib:some_lib"]
   )
   ```
+
+Every `haskell_binary` target also defines an optional REPL target that is
+not built by default, but can be built on request. The name of the REPL
+target is the same as the name of binary with `"-repl"` added at the end.
+For example, the target above also defines `main-repl`.
```
mboes: How to load a library or binary into a REPL should be documented (use `bazel run --direct_run`?). Note that in future versions of Bazel, `--direct_run` *might* become the default. See https://github.com/bazelbuild/bazel/commit/1fafc6fd3ac87d2963d08a6b7ddd076edfa2fb94.

haskell/haskell.bzl
mrkkrp: My logic here was:

* Names of (at least) library targets cannot contain underscores because they should be valid GHC package names (maybe now that we escape underscores it's not so important).
* So for consistency let's use dash here.

Snake case is understandable because `foo-bar` is not an identifier in Python/Skylark. But target names shouldn't necesserily be valid Skylark identifiers. In fact, at least in Numero I've seen a lot of targets named using dashes, such as `numero-doc`, etc.

Anyway, the choice of the separator seems not so important to me.

.circleci/config.yml
mrkkrp: I don't have Bazel 0.12 on my machine and CI doesn't have it either, so I think we should use what works with all versions. I don't know what `--direct-run` sets as current working directory and that's important for this script.

haskell/haskell.bzl
mrkkrp: I added something about this.

PR #227, opened 2018-05-02
De-duplicate include flags

mrkkrp:
@judah discovered that the current logic leads to duplication of `-I` flags. This PR should fix that.

### DIFF
--- a/haskell/cc.bzl
+++ b/haskell/cc.bzl
@@ -19,20 +19,29 @@ def cc_headers(ctx):
 
   *Internal function - do not use.*
   """
-  hdrs = [
-    hdr
+  hdrs = depset()
+
+  for dep in ctx.attr.deps:
     # XXX There's gotta be a better way to test the presence of
     # CcSkylarkApiProvider.
-    for dep in ctx.attr.deps if hasattr(dep, "cc")
-    for hdr in dep.cc.transitive_headers.to_list()
-  ] + [
-    hdr
-    # XXX cc_import doesn't produce a cc field, so we emulate it with
-    # a custom provider.
-    for dep in ctx.attr.deps if CcSkylarkApiProviderHacked in dep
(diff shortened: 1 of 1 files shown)


haskell/cc.bzl
```diff
@@ -19,19 +19,23 @@ def cc_headers(ctx):
...
-    for dep in ctx.attr.deps if hasattr(dep, "cc")
-    for hdr in dep.cc.transitive_headers.to_list()
-  ] + [
-    hdr
-    # XXX cc_import doesn't produce a cc field, so we emulate it with
-    # a custom provider.
-    for dep in ctx.attr.deps if CcSkylarkApiProviderHacked in dep
-    for hdr in dep[CcSkylarkApiProviderHacked].transitive_headers.to_list()
-  ]
+  adict = { hdr: None
```
judah: I think these are all already depsets, right?  In that case, would it be simpler to use `depset(transitive=...)` to combine them?

haskell/cc.bzl
```diff
@@ -19,19 +19,23 @@ def cc_headers(ctx):
...
+      # XXX cc_import doesn't produce a cc field, so we emulate it
+      # with a custom provider.
+      for dep in ctx.attr.deps if CcSkylarkApiProviderHacked in dep
+      for hdr in dep[CcSkylarkApiProviderHacked].transitive_headers.to_list()
+    }
+  )
+
+  hdrs = adict.keys()
+
   return hdrs, ["-I" + hdr.dirname for hdr in hdrs]
```
judah: I tried using this PR to help debug a build failure, and realized that l.39 also stands to be deduped since there may be multiple files within the same directory.  Not sure if there's a canonical way to do that; for example:
```
return hdrs, set.to_list(set.from_list(["-I" + hdr.dirname for hdr in hdrs]))
```

mrkkrp:
I've re-written this again.

PR #238, opened 2018-05-07
Follow-up to #211: allow generated `.hi-boot` files.

judah:
It's not perfect, since GHC requires hs-boot files to be in the
exact same directory as the corresponding .hs file; so for example,
you can't have one be generated and one be a raw file.  But that seems
pretty straightforward to work around for now.


haskell/actions.bzl
```diff
@@ -717,7 +717,14 @@ def _compilation_defaults(ctx):
   dep_info = gather_dep_info(ctx)
 
   # Add import hierarchy root.
-  ih_root_arg = ["-i{0}".format(import_hierarchy_root(ctx))]
+  # Note that this is not perfect, since GHC requires hs-boot files
+  # to be in the same directory as the corresponding .hs file.  Thus
+  # the two must both both have the same root; i.e., both plain files,
```
mrkkrp: Repeating "both".

PR #239, opened 2018-05-08
Implement protobuf integration using proto-lens

mrkkrp:
This PR adds support for [protocol buffers](https://blog.bazel.build/2017/02/27/protocol-buffers.html) using [`proto-lens`](https://hackage.haskell.org/package/proto-lens).

### DIFF
--- a/haskell/actions.bzl
+++ b/haskell/actions.bzl
@@ -562,7 +562,7 @@ def link_dynamic_lib(ctx, object_files):
   for package in set.to_list(
       set.union(
         dep_info.package_names,
-        set.from_list(ctx.attr.prebuilt_dependencies)
+        set.from_list(ctx.attr.prebuilt_dependencies),
       )):
     args.add(["-package", package])
 
@@ -636,7 +636,8 @@ def create_ghc_package(ctx, interfaces_dir, static_library, dynamic_library, exp
     "dynamic-library-dirs": "${pkgroot}",
     "hs-libraries": _get_library_name(ctx),
     "depends":
-      ", ".join([ d[HaskellLibraryInfo].package_name for d in ctx.attr.deps if HaskellLibraryInfo in d])
+      ", ".join([ d[HaskellLibraryInfo].package_name for d in
+                  ctx.attr.deps if HaskellLibraryInfo in d])
   }
   ctx.actions.write(
     output=registration_file,
@@ -672,7 +673,6 @@ def _compilation_defaults(ctx):
 
   Args:
     ctx: Rule context.
-    for_binary: We're compiling a binary target.
 
   Returns:
     _DefaultCompileInfo: Populated default compilation settings.
@@ -721,10 +721,10 @@ def _compilation_defaults(ctx):
   # to be in the same directory as the corresponding .hs file.  Thus
   # the two must both have the same root; i.e., both plain files,
   # both in bin_dir, or both in genfiles_dir.
-  root = import_hierarchy_root(ctx)
-  ih_root_arg = ["-i{0}".format(root),
-                 "-i{0}".format(paths.join(ctx.bin_dir.path, root)),
-                 "-i{0}".format(paths.join(ctx.genfiles_dir.path, root))]
+  import_root = import_hierarchy_root(ctx)
+  ih_root_arg = ["-i{0}".format(import_root),
+                 "-i{0}".format(paths.join(ctx.bin_dir.path, import_root)),
+                 "-i{0}".format(paths.join(ctx.genfiles_dir.path, import_root))]
   args.add(ih_root_arg)
   haddock_args.add(ih_root_arg, before_each="--optghc")
 
@@ -758,7 +758,7 @@ def _compilation_defaults(ctx):
   other_sources = []
   modules = set.empty()
   source_files = set.empty()
-  import_dirs = set.singleton(import_hierarchy_root(ctx))
+  import_dirs = set.singleton(import_root)
 
   # Output object files are named after modules, not after input file names.
   # The difference is only visible in the case of Main module because it may
@@ -968,6 +968,7 @@ def gather_dep_info(ctx):
             # files come from haskell_cc_import.
             _mangle_solib(ctx, dep.label, f, CcSkylarkApiProviderHacked in dep)
             for f in dep.files.to_list() if _is_shared_library(f)}
+
           ),
       )
 
--- a/haskell/haskell-impl.bzl
+++ b/haskell/haskell-impl.bzl
@@ -1,0 +1,192 @@
+"""Implementation of core Haskell rules"""
+
+load(":actions.bzl",
+  "compile_haskell_bin",
+  "link_haskell_bin",
+  "compile_haskell_lib",
+  "link_static_lib",
+  "link_dynamic_lib",
+  "create_ghc_package",
+  "gather_dep_info",
+  "get_pkg_id",
+)
+load(":set.bzl",
+  "set",
+)
+load(":providers.bzl",
+  "HaskellBuildInfo",
+  "HaskellBinaryInfo",
+  "HaskellLibraryInfo",
+)
+load(":ghci-repl.bzl",
+  _build_haskell_repl = "build_haskell_repl",
+)
+
+def haskell_binary_impl(ctx):
+  c = compile_haskell_bin(ctx)
+
+  binary = link_haskell_bin(ctx, c.object_dyn_files)
+  dep_info = gather_dep_info(ctx)
+
+  solibs = set.union(
+    set.from_list(dep_info.external_libraries.values()),
+    dep_info.dynamic_libraries,
+  )
+
+  build_info = dep_info # HaskellBuildInfo
+  bin_info = HaskellBinaryInfo(
+    source_files = c.source_files,
+    modules = c.modules,
+    binary = binary,
+  )
+  target_files = depset([binary])
+
+  _build_haskell_repl(
+    ctx,
+    build_info=build_info,
+    target_files=target_files,
+    bin_info=bin_info,
+  )
+
+  return [
+    build_info,
+    bin_info,
+    DefaultInfo(
+      executable = binary,
+      files = target_files,
+      runfiles = ctx.runfiles(
+        files = set.to_list(solibs),
+        collect_data = True,
+      ),
+    ),
+  ]
+
+def haskell_library_impl(ctx):
+
+  c = compile_haskell_lib(ctx)
+
+  static_library = link_static_lib(ctx, c.object_files)
+  dynamic_library = link_dynamic_lib(ctx, c.object_dyn_files)
+
+  dep_info = gather_dep_info(ctx)
+
+  exposed_modules = set.empty()
+  other_modules = set.from_list(ctx.attr.hidden_modules)
+
+  for module in set.to_list(c.modules):
+    if not set.is_member(other_modules, module):
+      set.mutable_insert(exposed_modules, module)
+
+  conf_file, cache_file = create_ghc_package(
+    ctx,
+    c.interfaces_dir,
+    static_library,
+    dynamic_library,
+    exposed_modules,
+    other_modules,
+  )
+
+  build_info = HaskellBuildInfo(
+    package_names = set.insert(dep_info.package_names, get_pkg_id(ctx)),
+    package_confs = set.insert(dep_info.package_confs, conf_file),
+    package_caches = set.insert(dep_info.package_caches, cache_file),
+    # NOTE We have to use lists for static libraries because the order is
+    # important for linker. Linker searches for unresolved symbols to the
+    # left, i.e. you first feed a library which has unresolved symbols and
+    # then you feed the library which resolves the symbols.
+    static_libraries = [static_library] + dep_info.static_libraries,
+    dynamic_libraries = set.insert(dep_info.dynamic_libraries, dynamic_library),
+    interface_files = set.union(dep_info.interface_files, set.from_list(c.interface_files)),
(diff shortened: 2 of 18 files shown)


judah:
Strange, it also works on my machine (macOS Sierra)...

judah:
I managed to reproduce it on SSH, by killing the build steps and then re-running them manually.

It looks like a resource issue:
```
mkdir tmp
cd tmp
cp ../tests/ghc.nix default.nix
/nix/var/nix/profiles/default/bin/nix-build -I nixpkgs=/root/.cache/bazel/_bazel_root/d00c9458fa4b9d008dbcdeefe3e0fc7c/external/nixpkgs -A "" 
...
copying path '/nix/store/3cv3s9lv79qjkqyafbclbk03gz5z813w-ghc-8.2.2' from 'https://cache.nixos.org'...
Killed
```

haskell/haskell.bzl
```diff
@@ -326,6 +332,61 @@ $ bazel-bin/.../hello-lib-repl  # run the script
...
+  return _haskell_library_impl(
+    ctx,
+    use_these_srcs=source_files,
+  )
+
+haskell_proto_library = rule(
+  _haskell_proto_library_impl,
+  attrs = dict(
+    _haskell_common_attrs,
+    hidden_modules = attr.string_list(
```
judah: Is `hidden_modules` still used in this PR?

haskell/protobuf.bzl
```diff
@@ -0,0 +1,246 @@
...
+            protobuf_tools(ctx).protoc_gen_haskell.path,
+  ])
+
+  hs_files = []
+  inputs = []
+
+  args.add(["-I{0}={1}".format(s.short_path, s.path)
+            for s in target.proto.transitive_sources])
+
+  inputs.extend(target.proto.transitive_sources.to_list())
```
judah: This seems like it will generate a single haskell_library out of all the transitive dependencies.  Compare to `cabal2bazel` which only builds the direct_sources: https://github.com/google/cabal2bazel/blob/c208bf179c8e820b0733a4472a5825c4f86e8e60/bzl/def.bzl#L1093
which generates a `haskell_library` out of only the immediate sources, but then makes it depend on the corresponding `proto_library`/aspect deps.

(See below comments on the `BUILD` file.)

tests/haskell_proto_library/BUILD
```diff
@@ -0,0 +1,36 @@
...
+  deps = [":person_proto"],
+)
+
+haskell_library(
+  name = "hs-lib",
+  srcs = ["Bar.hs"],
+  deps = [":person_haskell_proto"],
+  prebuilt_dependencies = ["base"],
+  visibility = ["//visibility:public"],
+)
```
judah: As mentioned, above, I don't think this will work with multiple `haskell_library` rules.  For example:
```
diff --git a/tests/haskell_proto_library/BUILD b/tests/haskell_proto_library/BUILD
index 74efe3a..41526d2 100644
--- a/tests/haskell_proto_library/BUILD
+++ b/tests/haskell_proto_library/BUILD
@@ -10,6 +10,11 @@ proto_library(
   srcs = ["zip_code.proto"],
 )
 
+haskell_proto_library(
+    name = "zip_code_haskell_proto",
+    deps = [":zip_code_proto"],
+)
+
 proto_library(
   name = "address_proto",
   srcs = ["address.proto"],
@@ -30,7 +35,20 @@ haskell_proto_library(
 haskell_library(
   name = "hs-lib",
   srcs = ["Bar.hs"],
-  deps = [":person_haskell_proto"],
+  deps = [
+      ":person_haskell_proto"
+  ],
+  prebuilt_dependencies = ["base"],
+  visibility = ["//visibility:public"],
+)
+
+haskell_library(
+  name = "hs-lib-2",
+  srcs = ["Baz.hs"],
+  deps = [
+      ":person_haskell_proto",
+      ":zip_code_haskell_proto",
+  ],
 )
```

Example run:
```
$ bazel build :hs-lib-2
INFO: Analysed target //tests/haskell_proto_library:hs-lib-2 (0 packages loaded).
INFO: Found 1 target...
ERROR: /Users/judah.jacobson/repos/rules_haskell/tests/haskell_proto_library/BUILD:45:1: Compiling hs-lib-2 failed (Exit 1)

tests/haskell_proto_library/Baz.hs:4:1: error:
    Ambiguous module name ‘Proto.Tests.HaskellProtoLibrary.ZipCode’:
      it was found in multiple packages:
      testsZShaskellZUprotoZUlibraryZSpersonZUhaskellZUproto-1.0.0 testsZShaskellZUprotoZUlibraryZSzipZUcodeZUhaskellZUproto-1.0.0
  |
4 | import Proto.Tests.HaskellProtoLibrary.ZipCode
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

mrkkrp:
Now only Darwin job fails.

judah:
You might  want to make a similar fix as #240 to the Darwin build; that is, change the steps so it installs Nix-1.* . Not sure why it only fails with this PR though.

haskell/haskell.bzl
mrkkrp: I decided that having the same attribute set will allow us to have fewer extra conditionals in `_haskell_library_impl`. If we remove `hidden_modules`, we'll need to guard every lookup of that attribute. On the other hand, if we keep `hidden_modules`, I don't see any problem with that.

mrkkrp:
@judah Please see 6b5d220cd0d0b26f7bbf590a66d0168c32cd9066. Is this what you talked about? What I do is the following:

* Move invocation of `_haskell_library_impl` inside aspect itself. If we want to avoid all-at-once approach, it looks like there is no other way to achieve that, and that's what the code you quote does.

* To invoke `_haskell_library_impl` either the definition of aspect should be moved to `haskell.bzl` or the `_haskell_library_impl` should be moved to a separate internal file we can depend on, and then all protobuf definitions can live in `protobuf.bzl`, which I like more. So I move the functions.

* If I am correct that we want to build a Haskell library per proto target in `_haskell_proto_aspect`, then we'll need to provide enough information for `_haskell_library_impl`. Normally it receives that information via attributes in `ctx`, but in this case we'll have to generate some part of it. Direct Haskell sources we already generate, Haskell dependencies will come as results of invocation of aspect on transitive dependencies. But there are other things like `name` attribute, and quite a few of them. So I decided to check with you before doing major refactoring.

AFAIK when what I started is implemented the error message about duplicate modules should go away.

haskell/haskell.bzl
```diff
@@ -326,6 +332,61 @@ $ bazel-bin/.../hello-lib-repl  # run the script
 ```
 """
 
+def _haskell_proto_library_impl(ctx):
+
+  source_files = depset()
+  for dep in ctx.attr.deps:
+    if HaskellProtobufInfo in dep:
+      source_files = depset(transitive = [
```
mboes: Single depset better than chain of nested depsets.

haskell/actions.bzl
```diff
@@ -268,6 +270,28 @@ def _infer_rpaths(target, solibs):
 
   return r
 
+def _infer_deps(ctx):
```
mboes: I'm concerned that these two functions are too easy to forget to call instead of reaching out for `ctx.attr.deps` etc, leading to subtle bugs down the road. Could this be done differently, e.g. by threading a provider all the way to the use site?

judah: :+1:
As another example where the logic feels a little fragile:
https://github.com/tweag/rules_haskell/pull/239/files/55b131bc543b781126aa11468bfbce7441842e14#diff-656a15c44401b2dbd8f5bfbc145d4f30R808

Rather than having two different ways to call `_haskell_library_impl` (either implicitly through a rule, or explicitly from within `_haskell_proto_aspect_impl`) I think it would be better to structure the code by defining three functions:

- `_haskell_library_impl()`, taking no arguments and used only in the rule definition of `haskell_library`
- `_haskell_proto_aspect_impl()` (taking no arguments)
- `compile_haskell_library(name, srcs, deps, {etc.})` , called by the above two.

haskell/protobuf.bzl
```diff
@@ -0,0 +1,246 @@
...
+    protoc_gen_haskell,
+    haskell_deps=[],
+    haskell_prebuilt_deps=[],
+    protoc=Label("@com_google_protobuf//:protoc"),
+    **kwargs):
+  """Declare a Haskell protobuf toolchain.
+
+  You need at least one of these declared somewhere in your `BUILD` files
+  for the `haskell_proto_library` rules to work. Once declared, you then
+  need to *register* the toolchain using `register_toolchain` in your
```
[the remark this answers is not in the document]
judah: Hm, it looks like most `X_proto_library` rules hard-code `@com_google_protobuf//:protoc`:
https://github.com/bazelbuild/bazel/blob/56eb80b11f475a2e5b6e8803aeb3cf2294e329b4/src/main/java/com/google/devtools/build/lib/rules/proto/ProtoConfiguration.java#L74

Which is also documented here:
https://blog.bazel.build/2017/02/27/protocol-buffers.html

So given that `rules_go` is also hard-coding `protoc` as an attribute default value, I guess it would be simpler for us to just do that too.

(However, I do think the `proto-lens-protoc` executable *should* be provided via a toolchain, since we don't have a consistent way of providing third-party packages.)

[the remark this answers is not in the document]
mrkkrp: But that link says `defaultValue`. We also have the same default value for `protoc` argument of the toolchain.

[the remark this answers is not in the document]
mboes: @judah I think I understand the logic. You're saying that so long as we don't have any standard way to name haskell dependencies, it doesn't make sense to hard-code any label for `proto-lens-protoc`? I can buy that.

But how about this: add `@proto-lens-protoc` to `haskell_repositories()`, effectively vendoring in the code. Many `X_proto_library` rules rely on some amount of language-specific code. Including `rules_go`, which likewise vendors in.

Eventually, we'll want to strip the Nixpkgs dependency for building `@proto-lens-protoc`, to keep the build entirely self-contained (once we know how to do that). Even if the user still wants to rely on Nixpkgs for her own dependencies, it would be fine to build `proto-lens-protoc` in some completely different way. After all it's just a binary. And an internal implementation detail of `rules_haskell`.

haskell/haskell-impl.bzl
```diff
@@ -0,0 +1,192 @@
...
+      executable = binary,
+      files = target_files,
+      runfiles = ctx.runfiles(
+        files = set.to_list(solibs),
+        collect_data = True,
+      ),
+    ),
+  ]
+
+def haskell_library_impl(ctx):
```
judah: This approach still feels fragile; for regular libraries you're passing an actual `ctx`, but for proto libraries you're passing a struct with specific attributes.  At minimum we should document what attributes this function expects.

However, I additionally feel we should make two structural modifications: (1) separate this function from the rules's actual `haskell_library_impl` (as I mentioned in my comment here: https://github.com/tweag/rules_haskell/pull/239#discussion_r187375274), and (2) distinguish the "real" ctx from attributes that are being passed in manually.

For example:

```
def build_haskell_library(ctx, hs_files, attrs):
    # Document the fields expected in attrs,
    # which should only need to be things that are not already available from `ctx.attrs`
    # in both haskell_library and `haskell_proto_library`
    ...

def _haskell_library_impl(ctx):
    build_haskell_library(
        ctx,
       ctx.files.srcs,
        struct(...))
)

def _haskell_proto_library_impl(
    # ...
    hs_srcs = ...
    build_haskell_library(ctx, hs_srcs, struct(...))
)

haskell_library = rule(
    implementation=_haskell_library_impl,
    ...)
haskell_proto_library = rule(
    implementation=_haskell_proto_library_impl,
    ...)
```

Let me know what you think.

haskell/protobuf.bzl
```diff
@@ -0,0 +1,344 @@
...
+      "prebuilt_dependencies": [],
+      "repl_interpreted": True,
+      "repl_ghci_args": [],
+      "version": "1.0.0",
+      "_ghc_defs_cleanup": ctx.attr._ghc_defs_cleanup,
+      "_ghci_script": ctx.attr._ghci_script,
+      "_ghci_repl_wrapper": ctx.attr._ghci_repl_wrapper,
+      "hidden_modules": [],
+    },
+    name = "proto-autogen-" + ctx.rule.attr.name,
```
judah: Why is this using both the positional and the string-based ways of initializing the dictionary?  (For example, `"compiler_flags":...` vs `srcs =...`)

haskell/haskell-impl.bzl
mrkkrp: I agree that this is fragile. I'll do the change you describe, however note that there is no way to forbid a developer accessing `deps` for example from context attributes, like so: `ctx.attr.deps` and not from that additional attribute that we'll be passing around. Test suite may not catch these things and they will lead to bugs. When I proposed the solution with patched `cxt` I kept in mind @mboes's comment that there should be only one way to access something. With the proposed approach I'll always be able to access things in two ways: from context `ctx` and from the additional attribute struct. I'm not even sure how to detect all the places where I need to do changes now.

judah: Yeah, those are fair points; I'm fine doing it either way for now.  (Who knew that dynamically-typed code could be hard to reason about...)

judah:
@mrkkrp , it looks like this change broke the documentation:
https://haskell.build/haskell/haskell.html#haskell_library
Namely, now that haskell_common_attrs is in a different .bzl file, the documentation for those attributes (for example, "srcs" or "deps") isn't being generated for some reason.

mboes:
Likely due to https://github.com/bazelbuild/skydoc/issues/21.

PR #246, opened 2018-05-15
Make package names more similar to the rule name.

judah:
As of #222, both the package name and ID are z-encoded.  However,
GHC only needs the package *ID* to be unique, as long as you
call `-package-id` instead of `-package`.  As a result we can
simplify how package-names are encoded: we can keep them the same
as the rule name, other than fixing underscores.

In addition to (I think) making error messages nicer, this paves the way for
some improvements in Hazel (or similar tools), now that the rule and package
names can be identical:
- It helps support the PackageImports extension.
- GHC >= 8.0 generates the Cabal macros (e.g., `MIN_VERSION_*`) automatically.

Prebuilt dependencies are still passed with `-package` since their IDs are internal details of GHC.

### DIFF
--- a/haskell/actions.bzl
+++ b/haskell/actions.bzl
@@ -412,8 +412,8 @@ def link_haskell_bin(ctx, object_files):
 
   args.add([f.path for f in object_files])
 
-  for package in set.to_list(dep_info.package_names):
-    args.add(["-package", package])
+  for package in set.to_list(dep_info.package_ids):
+    args.add(["-package-id", package])
 
   for cache in set.to_list(dep_info.package_caches):
     args.add(["-package-db", cache.dirname])
@@ -477,9 +477,6 @@ def compile_haskell_lib(ctx):
       import_dirs: import directories that should make all modules visible (for GHCi)
   """
   c = _compilation_defaults(ctx)
-  c.args.add([
-    "-package-name", get_pkg_id(ctx),
-  ])
 
   # This is absolutely required otherwise GHC doesn't know what package it's
   # creating `Name`s for to put them in Haddock interface files which then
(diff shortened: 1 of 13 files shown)


haskell/actions.bzl
```diff
@@ -866,35 +862,41 @@ def _zencode(s):
   return s.replace("Z", "ZZ").replace("_", "ZU").replace("/", "ZS")
 
 def get_pkg_name(ctx):
-  """Get package name. Package name includes Bazel package and name of the
```
mboes: You point out that funny package names won't just be visible in documentation, but much worse, in source code too (if using `-XPackageImports`). Funny names in documentation is less bad because if the names change, nothing breaks. Fortunately, we can separate package-id from package-name, it's just that underscores are still illegal.

If so then I'd lean towards banning underscores outright. That way no encoding necessary, no funny names in package imports. Either that or `s/_/-/`, which yes, is technically a non-injective encoding, but since package names don't need to be unique (only package-id's), that should be ok. And it lets users keep to Bazel-conventional package naming style if desired (i.e. underscores everywhere, no hyphens).

haskell/actions.bzl
```diff
@@ -866,35 +862,41 @@ def _zencode(s):
...
-  component. We can't use just the latter because then two components with
-  the same names in different packages would clash.
+  """Get package name.
+
+  The name is not required to be unique/injective; however it must be a valid
+  GHC package name (i.e., no underscores).  This encoding is not aims to
```
mrkkrp: "does not aim"?

haskell/cc.bzl
```diff
@@ -187,6 +187,7 @@ def _cc_haskell_import(ctx):
     if dbin != None:
       set.mutable_insert(dyn_libs, dbin)
 
+  print(dyn_libs)
```
mrkkrp: Should this be removed?

haskell/providers.bzl
```diff
@@ -15,7 +15,7 @@ HaskellBuildInfo = provider(
 HaskellLibraryInfo = provider(
   doc = "Library-specific information.",
   fields = {
-    "package_name": "Package name, usually of the form name-version.",
+    "package_id": "Package name, usually of the form name-version.",
```
mrkkrp: Let's say "package id" here?

PR #247, opened 2018-05-15
Fix mangling of solibs from different Bazel packages

### LINKED ISSUE #245: Build failure when cc_library deps outside of current package
We have a test that checks that adding a C library to the `deps` of a `haskell_library` works: https://github.com/tweag/rules_haskell/blob/master/tests/c-compiles/BUILD. But imagine the following library:

```python
haskell_library(
  name = "foo",
  deps = ["//tests/c-compiles:c-lib"]
)
```

This will fail with:

```
$ bazel build //:foo
ERROR: /home/mboes/src/tweag/rules_haskell/BUILD:19:1: in haskell_library rule //:foo: 
	File "/home/mboes/src/tweag/rules_haskell/tests/library-with-includes/BUILD", line 19
		haskell_library(name = 'library-with-includes')
[...]
"/home/mboes/.cache/bazel/_bazel_mboes/822aa44857e17f47d08bb198503f3032/external/io_tweag_rules_haskell/haskell/actions.bzl", line 97, in _mangle_solib
		paths.relativize(solib.path, qualsolib.dirname)
	File "/home/mboes/.cache/bazel/_bazel_mboes/822aa44857e17f47d08bb198503f3032/external/bazel_skylib/lib/paths.bzl", line 188, in paths.relativize
		fail(("Path '%s' is not beneath '%s'"...)))
Path 'bazel-out/k8-fastbuild/bin/tests/c-compiles/libc-lib.so' is not beneath 'bazel-out/k8-fastbuild/bin/foo'
```

cc @ghorn @dleprovost @gdeest 

I believe this is a regression introduced by 48c0b07e8837d9dfa1ed75de777706fac0554a9e.

mrkkrp:
Close #245.

### DIFF
--- a/haskell/actions.bzl
+++ b/haskell/actions.bzl
@@ -62,6 +62,31 @@ _DefaultCompileInfo = provider(
   },
 )
 
+def _truly_relativize(target, relative_to):
+  """Return a relative path to `target` from `relative_to`.
+
+  Args:
+    target: string, path to directory we want to get relative path to.
+    relative_to: string, path to directory from which we are starting.
+
+  Returns:
+    string: relative path to `target`.
+  """
+  t_pieces = target.split('/')
+  r_pieces = relative_to.split('/')
+  common_part_len = 0
+
+  for tp, rp in zip(t_pieces, r_pieces):
+    if tp == rp:
+      common_part_len += 1
+    else:
+      break
+
+  result = [".."] * (len(r_pieces) - common_part_len)
+  result += t_pieces[common_part_len:]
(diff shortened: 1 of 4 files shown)


haskell/actions.bzl
```diff
@@ -62,6 +62,31 @@ _DefaultCompileInfo = provider(
   },
 )
 
+def _truly_relativize(target, relative_to):
```
mboes: Don't we already have this function in `haddock.bzl`?

PR #252, opened 2018-05-17
Enable protos from external repositories.

judah:
The `short_path` for an external file contains `..` which messes
up the logic around prefixes and also isn't permitted by `protoc`.
Instead, compute it manually from `path`.  Logic emulated from
rules_go:
https://github.com/bazelbuild/rules_go/blob/67f44035d84a352cffb9465159e199066ecb814c/proto/compiler.bzl#L72

This change required marking the test `haskell_toolchain` as `testonly=0`.
The test that I added works on `@com_google_protobuf//:timestamp_proto`, a
`proto_library` which is not testonly; it looks like the aspect that's
generated also can't be testonly.  If this is an issue we can create a new
local repo for the test; I just wasn't sure it was worth the effort.

### DIFF
--- a/haskell/protobuf.bzl
+++ b/haskell/protobuf.bzl
@@ -42,6 +42,12 @@ def _proto_lens_output_file(path):
 
   return "Proto/" + result
 
+def _proto_path(proto):
+  """A path to the proto file which matches any import statements."""
+  return paths.relativize(
+      proto.path,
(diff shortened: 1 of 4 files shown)


mboes:
> a proto_library which is not testonly; it looks like the aspect that's generated also can't be testonly. If this is an issue we can create a new local repo for the test

That's fine by me.

PR #254, opened 2018-05-18
Fix location of hsc2hs produced sources by haddock

mrkkrp:


mrkkrp:
Hmm, wait a minute, I'll add another similar fix.

mrkkrp:
No, that fix will go to Hazel. This PR is ready.

PR #255, opened 2018-05-18
Add ability to specify hidden haddock modules

mrkkrp:
Let's not rush with this and wait what @judah says. One problem is that we still need to detect hidden modules in Hazel automatically or otherwise we'll have to write rules manually for packages with hidden haddock modules.

### DIFF
--- a/haskell/haddock.bzl
+++ b/haskell/haddock.bzl
@@ -87,7 +87,7 @@ def _haskell_doc_aspect_impl(target, ctx):
       ),
       sibling=haddock_interface
     )
-    for f in input_sources
+    for f in input_sources if f not in ctx.rule.files.hidden_haddock_modules
   ]
 
   self_outputs = [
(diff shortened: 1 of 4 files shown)


judah:
Good call-out.  What package(s) have hidden modules in practice?  I'm trying to wrap my mind around the scope of this and how painful it would be to have to write manual rules in Hazel for this, or perhaps add a configuration parameter marking certain packages.

mrkkrp:
See our Slack channel, so far it's `vector`.

mrkkrp:
Tried to make it work that way but couldn't make it work: https://github.com/tweag/rules_haskell/commit/2b3934f3ef2f929766f0731fe597748fac3324fd. See commit message for a short description.

mboes:
Could the interface files be stored in a separate dir?

mrkkrp:
Probably yes, but it won't solve the main issue with population of the "main" output directory (with HTML files).

judah:
@mrkkrp I tried the "track interface files separately" approach on top of your branch, and it seemed to work fine: https://github.com/tweag/rules_haskell/compare/master...haddock-with-declare-directory
(I didn't get it working with both shallow and deep indexing, but that seems straightforward.)

Does that work for you, or am I missing something?

PR #258, opened 2018-05-21
Standardize handling of executable dependencies in scripts

### LINKED ISSUE #257: RFC: homogenize script actions
We currently have a bunch of bespoke scripts in rules_haskell:

```
$ ls haskell/*.sh
haskell/ghc-defs-cleanup.sh
haskell/ghci-repl-wrapper.sh
haskell/ghc-lint-wrapper.sh
haskell/ghc-version-check.sh
haskell/haddock-wrapper.sh
haskell/make-bin-symlink-which.sh
```

Notice that:

* All of them except `ghci-repl-wrapper.sh` and `haddock-wrapper.sh` are trivially small.
* I guess we expected these scripts to be larger at first, in which case they'd be easier to handle as separate files (proper IDE syntax highlighting etc). But this turned out not to be the case.
* All scripts are called from a single place only.
* The ones that are not trivially small use environment variables to locate executable dependencies. But environment variables are error-prone. They can be a source of subtle bugs (due to shadowing without warning or due to typos in the variable name) that only get noticed at runtime, not graph analysis time.
* We could use runfiles instead of environment variables as a more robust solution. But some of the executable dependencies come from toolchains, so we can't use runfiles because we don't have a label available to identify the tools we want in the runfile tree (related: bazelbuild/bazel#5223).
* `hsc2hs` uses neither conventions. In that case, we set the `PATH` to contain the tools we want (including `gcc`) before executing the action. We have no choice, because `hsc2hs` is not runfile aware and really does want to just find stuff on the `PATH`. In general, commands that we call from scripts will often expect the same thing.

Given that, I propose to:

* Inline trivially small scripts into their call sites in the `.bzl` files.
* For the remaining scripts, set `PATH` to include the tools we wants and use neither runfiles to locate these executable dependencies nor environment variables. Like we already do for `hsc2hs`.

Thoughts?

mboes:
This PR should be just a refactoring with no behaviour changes.

Closes #257.

This sequence of commits removes the trivially small scripts to inline
them instead. Whether inlined or kept separate, all scripts now
standardize on pulling their executable dependencies from the `PATH`.
I also took the opportunity to make one script more declarative by
introducing a `cpp_defines` patterns file.


mboes:
This PR falls short of introducing any kind of `rules_go`-style wrappers as discussed in #257. That could be considered in a second step as part of another PR.

PR #265, opened 2018-05-22
Split actions.bzl into multiple action files.

mboes:
The `actions.bzl` file was getting unwieldy (1k+ lines). Furthermore,
this change is preparation for introducing an ancillary `actions`
struct in all `haskell_*` rules, like `ctx.actions`, but that offers
Haskell-specific actions.

Each action should be largely independent and there should be no
dependencies between them. This is not yet the case, but that's
precisely because there are missing abstractions (to be introduced in
future PR's). With the split, it becomes clear which functions are
used by which actions, and which ones are common and therefore need to
be factored out.


haskell/private/providers.bzl
```diff
@@ -1,3 +1,23 @@
+DefaultCompileInfo = provider(
```
mrkkrp: Maybe rename it to `HaskellCompileInfo`? `DefaultCompileInfo` is a bit too general.

mboes: I agree. I just kept the initial name for now to avoid making this PR bigger than it already is.

PR #273, opened 2018-05-24
Don't use version field to uniquify target names.

mboes:
A rule name is already unique within a package. So we don't need
versions to add extra uniqueness. Dropping versions means we don't
need to thread them across a compilation and linking, when in the end
only packaging cares about versions.

### DIFF
--- a/haskell/lint.bzl
+++ b/haskell/lint.bzl
@@ -75,7 +75,7 @@ def _haskell_lint_aspect_impl(target, ctx):
   args.add(sources)
 
   lint_log = ctx.actions.declare_file(
-    target_unique_name(hs, "lint-log", ctx.rule.attr.version)
+    target_unique_name(hs, "lint-log")
   )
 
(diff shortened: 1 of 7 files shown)


judah:
So, Hazel actually started relying on the `version` attribute as of `FormationAI/hazel#5`, in order to get GHC to generate `MIN_VERSION_*` macros automatically.

This could also in theory break the setup in `TokTok/toktok-stack` for the same reason, though I don't know if they have any packages explicitly relying on it in practice.

Would you be open to me sending a PR that added back the `version` field, just for `haskell_library`, and just for use when creating a per-rule package DB?  (I.e., without influencing id uniqueness either way)?

Otherwise, I could roll back the above Hazel change and/or look into other ways of making it simpler.

mboes:
@judah The `version` field is still there, precisely to support the Hazel use case (I can't think of another good use case for it). It's just not used internally for as many things as before.

PR #274, opened 2018-05-24
Introduce pkg_id struct

mboes:
In the case of libraries, we unfortunately can't use just the label
for everything. We need a version component, otherwise `ghc-pkg`
complains. We used to have an unholy mishmash of stuffs: package
names, package id's and versions, all disconnected from each other.
Now we package everything into a single struct that we pass along as
an argument.

I inlined `get_library_name` because there's no good place to put that
function, and it's a trivially small one anyways.

### DIFF
--- a/haskell/private/pkg_id.bzl
+++ b/haskell/private/pkg_id.bzl
@@ -1,0 +1,50 @@
+"""Package identifiers"""
+
+load("@bazel_skylib//:lib.bzl", "paths")
+
+def _zencode(s):
+  """Z-escape special characters to make a valid GHC package identifier.
+
+  Args:
+    s: string
+  """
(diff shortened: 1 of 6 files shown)


haskell/private/pkg_id.bzl
```diff
@@ -0,0 +1,50 @@
+"""Package identifiers"""
+
+load("@bazel_skylib//:lib.bzl", "paths")
+
+def _zencode(s):
+  """Z-escape special characters to make a valid filename.
```
judah: It's not just for filenames; this also gives us a valid GHC package identifier.  (Otherwise it wouldn't be necessary to escape underlines.) . Clarify in the comment?

mboes: True!

PR #279, opened 2018-05-26
Add mnemonics to most actions

mboes:
The mnemonics follow these principles:

* they all use `Haskell*` as a prefix, to easily distinguish them from
  actions for other rule sets in the real-time status section at the
  bottom of the screen during a build. `rules_go` also follows this
  convention.
* Some outputs had to be reordered because in build logs, Bazel only
  prints the *first* output name together with the mnemonic.
* We prefer mnemonics where they are sufficient. Where there is no
  good output name to print alongside it, we use a `progress_message`
  instead but still define a matching mnemonic.

### DIFF
--- a/haskell/cc.bzl
+++ b/haskell/cc.bzl
@@ -182,6 +182,7 @@ def _cc_haskell_import(ctx):
     ctx.actions.run(
       inputs = [bin],
       outputs = [dyn_lib],
+      mnemonic = "Symlink",
       executable = tools(ctx).ln,
       arguments = ["-s", relative_bin, dyn_lib.path],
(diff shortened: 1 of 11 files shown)


mboes:
I'm still interested in a post merge review by @judah. Maybe he knows about established patterns when it comes to the use of mnemonics, and whether this PR broadly fits these typical patterns.

judah:
LGTM.  The main convention I'm aware of is the prefix, as you've already done.  The Google haskell rules are similar: 
https://github.com/google/cabal2bazel/blob/master/bzl/def.bzl

PR #280, opened 2018-05-27
Add link checks for Skylark code

mboes:
We do this in the form of tests, defined using the `skylark_lint`
macro with a default name. This macro is purely for internal use on
rules_haskell code, not client code. It could be upstreamed to
bazelbuild/buildtools eventually.

The lint tests are for *checking* the formatting. Developers are
expected to use buildifier as a formatter integrated into their
editor. There are instructions in the README to call the formatter
from the command-line.

### DIFF
--- a/skylark/lint.bzl
+++ b/skylark/lint.bzl
@@ -1,0 +1,57 @@
+"""Lint checks for BUILD and *.bzl files."""
+
+# sh_binary(
+#   name = "buildify",
+#   srcs = ["apply.sh"],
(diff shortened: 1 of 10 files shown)


skylark/lint.bzl
```diff
@@ -0,0 +1,57 @@
+"""Lint checks for BUILD and *.bzl files."""
+
+# sh_binary(
```
mrkkrp: Can this be removed?

PR #284, opened 2018-05-30
Add a rule to support ‘c2hs’

mrkkrp:
Doesn't work on Darwin. `clang` complains about missing input files (the way it's called by `c2hs`). Maybe we could pass a dummy file with `-C`. @judah, as you have a machine with MacOS, could you try to see how this can be fixed?


judah:
It works on my Mac if I replace `tools(ctx).cpp` with `tools(ctx).gcc`.  I saw some posts online that suggest this might be a bug with clang that we hit because c2hs passes the `-x` option.  For example: https://stackoverflow.com/questions/36482291/clang-error-no-input-files-in-os-x/36632485#36632485

It might be ok to just use `tools(ctx).gcc` for all OSes, since I observed that c2hs passes `-E` under the hood when I ran `c2hs --dump=trace`.)

tests/c2hs/Bar.chs
```diff
@@ -0,0 +1,4 @@
+module Bar (bar) where
+
+bar :: Int
```
judah: It would be good if these tests could actually pull a definition from a header in a C dependency, to confirm that it's wired up properly.

haskell/c2hs.bzl
```diff
@@ -0,0 +1,100 @@
...
+)
+load("@bazel_skylib//:lib.bzl", "paths")
+
+def _haskell_c2hs_impl(ctx):
+
+  args = ctx.actions.args()
+  common_inputs = []
+
+  # some common options
+  args.add(["-C-E", "-C-I."])
```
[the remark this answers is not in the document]
mrkkrp: Re `-C-E` that's what Google code passes, I just copied that part. Let me check if it works without it.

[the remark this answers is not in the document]
mrkkrp: `-C-E` appears to be necessary. In my tests, without this option I get:

```
[mark@nixos:~/projects/tweag/rules_haskell]$ bazel build //tests/c2hs:c2hs
INFO: Analysed target //tests/c2hs:c2hs (2 packages loaded).
INFO: Found 1 target...
ERROR: /home/mark/projects/tweag/rules_haskell/tests/c2hs/BUILD:9:1: c2hs generating Haskell file for c2hs failed (Exit 1)
/nix/store/27x7pinqdsl9f3rpbm8bsszd9fhwq266-glibc-2.27/lib/crt1.o: In function `_start':
/build/glibc-2.27/csu/../sysdeps/x86_64/start.S:104: undefined reference to `main'
c2hs: Error during preprocessing custom header file
Target //tests/c2hs:c2hs failed to build
INFO: Elapsed time: 0.271s, Critical Path: 0.09s
```

It looks like `gcc` tries to link.

tests/c2hs/Bar.chs
mrkkrp: Done.

haskell/c2hs.bzl
```diff
@@ -0,0 +1,106 @@
...
+)
+
+load("@bazel_skylib//:lib.bzl", "paths")
+
+def _haskell_c2hs_impl(ctx):
+
+  args = ctx.actions.args()
+
+  # some common options
+  args.add(["-C-E"]) # "-C-I."
```
mboes: -C-I. no longer needed?

haskell/c2hs.bzl
```diff
@@ -0,0 +1,106 @@
...
+      mandatory=True,
+      allow_files=True,
+      doc="Input `.chs` files.",
+    ),
+    "deps": attr.label_list(
+      providers=["cc"],
+      doc="List of `cc_library` rules providing headers needed by the `.chs` files.",
+    ),
+    "copts": attr.string_list(
+      doc="List of c options to be passed to c2hs prefixed with `-C`.",
```
mboes: No longer true?

haskell/c2hs.bzl
```diff
@@ -0,0 +1,106 @@
...
+
+    ctx.actions.run(
+      inputs = depset(transitive = [
+        depset(common_inputs),
+        depset([src]),
+      ]),
+      outputs = [output],
+      executable = tools(ctx).c2hs,
+      mnemonic = "HaskellC2Hs",
+      progress_message = "c2hs generating Haskell file for " + ctx.label.name,
```
mboes: Mnemonic is sufficient in this case. It already prints the output name.

haskell/c2hs.bzl
mrkkrp: Why? This is still used: https://github.com/tweag/rules_haskell/blob/add-rule-for-c2hs/haskell/c2hs.bzl#L31

mboes: I misread. I thought  this was saying `-C-I.`. But in this case, I think this is irrelevant low-level detail for user facing documentation.

haskell/private/actions/compile.bzl
```diff
@@ -68,21 +68,20 @@ def _process_hsc_file(hs, cc, ghc_defs_dump, hsc_file):
...
     hs: Haskell context.
+    cc: CcInteropInfo, information about C dependencies.
     ghc_defs_dump: File with GHC definitions.
     hsc_file: hsc file to process.
 
   Returns:
     File: Haskell source file created by processing hsc_file.
   """
 
-  hsc_output_dir = hs.actions.declare_directory(
```
mrkkrp: Creating a per-module unique directory to put a file (which already has unique-enough path) is awkward. I first made this change when I just started and did not better. The proper solution is avoid declaring directory and instead add raw string prefix to each file before declaring them. This way they all have the prefix, but Bazel can't nitpick on the basis that some output path is prefix of another (not true!). This requires a minor change to the `declare_compiled` helper to make it accept strings instead of `File`s.

haskell/private/actions/compile.bzl
```diff
@@ -93,16 +92,59 @@ def _process_hsc_file(hs, cc, ghc_defs_dump, hsc_file):
...
+  # Output a Haskell source file.
+  chs_dir_raw = target_unique_name(hs, "chs")
+  hs_out = declare_compiled(hs, chs_file, ".hs", directory=chs_dir_raw)
+  args.add([chs_file.path, "-o", hs_out.path])
+
+  args.add(["-C-E"])
+  args.add(["--cpp", hs.tools.gcc.path])
+  # XXX c2hs doesn't seem to support direct inclusion of arbitrary headers
+  # args.add(["-C-I{0}".format(ghc_defs_dump.dirname)])
+  # args.add(["-C-i{0}".format(ghc_defs_dump.basename)])
```
judah: I believe that you can do this with `-C-include{0}`.  (The `-include` flag to gcc should have the same effect as `-i` to `hsc2hs`.)

haskell/private/path_utils.bzl
```diff
@@ -79,13 +79,13 @@ def declare_compiled(hs, src, ext, directory=None):
...
     ext: New extension.
-    directory: Directory the new file should live in.
+    directory: String, directory prefix the new file should live in.
 
   Returns:
     File: Declared output file living in `directory` with given `ext`.
   """
   fp = paths.replace_extension(_rel_path_to_module(hs, src), ext)
-  fp_with_dir = fp if directory == None else paths.join(directory.basename, fp)
+  fp_with_dir = fp if directory == None else paths.join(directory, fp)
```
judah: I didn't see anywhere we're calling this function with `directory=None`.  Can we remove that logic?

mrkkrp: This allows to omit the `directory` argument, may be useful in future even though we don't use it right now (used in the past though).

haskell/private/actions/compile.bzl
mrkkrp: Should be fixed now.

WORKSPACE
```diff
@@ -101,6 +99,15 @@ filegroup (
   ]),
   testonly = 1,
 )
+
+filegroup (
```
judah: Is this used?

mrkkrp: I guess it's not used now. Let's try to remove it.

PR #290, opened 2018-06-02
Work around macOS linker limits.

judah:
Previously, the build would fail on macOS if there were too many
transitive Haskell dependencies.  We encountered that situation
in our repo this week:
```
ghc: panic! (the 'impossible' happened)
  (GHC version 8.2.2 for x86_64-apple-darwin):
    Loading temp shared object failed: dlopen(/var/folders/ql/cdls48955v94j_wr38p7kq300000gp/T/ghc39877_0/libghc_16.dylib, 5): no suitable image found.  Did find:
    /var/folders/ql/cdls48955v94j_wr38p7kq300000gp/T/ghc39877_0/libghc_16.dylib: malformed mach-o: load commands size (37032) > 32768
```
This is a known issue that several other builds systems have worked around
(e.g.: Cabal, Stack, Nix).

Luckily, a new fix *just* (two days ago!) landed in GHC HEAD, and is easy to
replicate in `rules_haskell`.  For details, see:
https://phabricator.haskell.org/D4714

I had to add it in a couple places; I saw a bit of other redundant code, but let me know if it makes sense to move this into an existing utility function.

### DIFF
--- a/haskell/private/actions/compile.bzl
+++ b/haskell/private/actions/compile.bzl
@@ -199,6 +199,12 @@ def _compilation_defaults(hs, cc, java, dep_info, srcs, cpp_defines, compiler_fl
 
   add_mode_options(hs, args)
 
+  # Work around macOS linker limits.  This fix has landed in GHC HEAD, but is
+  # not yet in a release; plus, we still want to support older versions of
+  # GHC.  For details, see: https://phabricator.haskell.org/D4714
+  if hs.toolchain.is_darwin:
+    args.add(["-optl-Wl,-dead_strip_dylibs"])
+
   # Add import hierarchy root.
   # Note that this is not perfect, since GHC requires hs-boot files
   # to be in the same directory as the corresponding .hs file.  Thus
--- a/haskell/private/actions/link.bzl
+++ b/haskell/private/actions/link.bzl
@@ -266,6 +266,12 @@ def link_library_dynamic(hs, dep_info, object_files, my_pkg_id):
 
   args.add(["-shared", "-dynamic"])
 
+  # Work around macOS linker limits.  This fix has landed in GHC HEAD, but is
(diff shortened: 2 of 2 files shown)


jamesthompson:
@judah I've just run into this again, or something similar when working with proto-lens v0.6.0.0, but for GHCi. My library links ok with a `build`, but is barfing on a `@repl` call.

```
<command line>: can't load .so/.DLL for:...

...proto-ghc8.6.5.dylib: malformed mach-o: load commands size (43504) > 32768
```

Is there a place where GHCi might need these arguments too?

aherrmann:
@jamesthompson On macOS we are now effectively undoing `-optl-Wl,-dead_strip_dylibs` during linking, the reasoning is explained in [this comment](https://github.com/tweag/rules_haskell/blob/1fedad7e64ea00ecf1bcb4be70ca51a11dc48312/haskell/private/actions/link.bzl#L53-L74). See also https://github.com/tweag/rules_haskell/pull/1238. We have some other workarounds for exceeding MACH-O limits in place, [see here](https://github.com/tweag/rules_haskell/blob/1fedad7e64ea00ecf1bcb4be70ca51a11dc48312/haskell/private/cc_wrapper.py.tpl#L25-L29).

As reported it looks like `...proto-ghc8.6.5.dylib` is the file with the too large MACH-O header. When you say `bazel build` works, is it `bazel build` of that library, or also `bazel build` of another target that depends on that library? IIRC you can create a library with a too large header without error, only trying to load such a library will cause an error.

If this issue only occurs with `@repl` then it could be related to how we call `ghci`. If you `bazel build` the `@repl` target and take a look at the generated output file you will see how we call `ghci`. You can also add `compiler_flags = ["-v", "-keep-tmp-files"]` to your target to get some more insight into what GHC does.

Can you share some code and steps to reproduce the issue?

PR #292, opened 2018-06-04
Extra source files

### LINKED ISSUE #291: Allow to specify non-Haskell source files
I.e. what is called by Cabal "extra source files". These should just be visible during compilation, that's all. This is necessary to get packages like `wai-app-static` to compile. They use TH like this:

```
bazel-out/k8-fastbuild/bin/external/haskell_wai_app_static/gen-srcs-wai-app-static/Network/Wai/Application/Static.hs:234:151: error:
    * Exception when trying to run compile-time code:
        images/folder.png: openBinaryFile: does not exist (No such file or directory)
      Code: embedFile "images/folder.png"
    * In the untyped splice: $(embedFile "images/folder.png")
    |
234 | staticAppPieces _ [".hidden", "folder.png"] _ sendResponse = sendResponse $ W.responseLBS H.status200 [("Content-Type", "image/png")] $ L.fromChunks [$(embedFile "images/folder.png")]
    |                                         
```

`data` files do not help here because these will be available at run time only.

mrkkrp:
Close #291.

This makes it possible to compile TH code that references external files with arbitrary extensions, see the test. Unfortunately this still doesn't work out-of-the-box with Hackage packages because the path to files should include package names. Not a problem for new code though.

### DIFF
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -48,7 +48,7 @@ _haskell_common_attrs = {
   ),
   "extra_srcs": attr.label_list(
     allow_files = True,
-    doc = "Extra (non-Haskell) source files that will be needed at compile time (e.g. by TH).",
+    doc = "Extra (non-Haskell) source files that will be needed at compile time (e.g. by Template Haskell).",
   ),
   "deps": attr.label_list(
     doc = "List of other Haskell libraries to be linked to this target.",
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -46,6 +46,10 @@ _haskell_common_attrs = {
     allow_files = FileType([".hs", ".hsc", ".chs", ".lhs", ".hs-boot", ".lhs-boot", ".h"]),
     doc = "Haskell source files.",
   ),
+  "extra_srcs": attr.label_list(
+    allow_files = True,
+    doc = "Extra (non-Haskell) source files that will be needed at compile time (e.g. by TH).",
+  ),
   "deps": attr.label_list(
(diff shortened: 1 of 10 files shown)


haskell/haskell.bzl
```diff
@@ -46,6 +46,10 @@ _haskell_common_attrs = {
     allow_files = FileType([".hs", ".hsc", ".chs", ".lhs", ".hs-boot", ".lhs-boot", ".h"]),
     doc = "Haskell source files.",
   ),
+  "extra_srcs": attr.label_list(
+    allow_files = True,
+    doc = "Extra (non-Haskell) source files that will be needed at compile time (e.g. by TH).",
```
judah: Suggestion: clarify these docs by expanding "TH" to "Template Haskell".

tests/extra-source-files/FooTH.hs
```diff
@@ -0,0 +1,12 @@
...
+
+module FooTH (embedFile) where
+
+import Language.Haskell.TH
+import Language.Haskell.TH.Syntax
+
+embedFile :: FilePath -> Q Exp
+embedFile path = do
+  str <- runIO (readFile path)
+  addDependentFile path
```
judah: A question: I guess `addDependentFile` is mostly only needed for reloading in the REPL, right?  Otherwise I believe Bazel would track the dependency on the input file automatically.

mrkkrp: Well I just added it because it was the right thing to do. It doesn't hurt in either case, does it?

PR #293, opened 2018-06-05
Fix doctesting of targets with transitive non-haskell deps

mrkkrp:
We need to place the check earlier because if we try get the context on non-haskell transitive deps, they are going to miss some import attributes.


mrkkrp:
Rebased against current master and made lints only lint direct dependencies. @mboes please take a look.

PR #296, opened 2018-06-07
Implement extra_binaries in haskell_toolchain.

### LINKED ISSUE #282: Allow ghc to invoke javac
`inline-java` invokes `javac` in a GHC Core plugin in order to compile generated java stubs. The bytecode is then embedded in the object file of the module.

Bazel, however, requires all inputs to be declared to a rule, or the inputs won't be visible to the compiler. In particular, we would need to somehow say that `javac` is an input to the rule, or the GHC plugin won't be able to invoke it.

I believe that, currently, there is no way to specify programs as inputs to the haskell build rules, so this issue would be about implementing some way to allow it.

`sparkle` and `jvm-batching` used to build with rules_haskell at commit 23d8b0f, probably because of a bug in `rules_haskell`. But this is no longer the case (fcef649). If we can't find a good way to declare `javac` an input to the build rules, it may worth investigating how it did work before and attempting a hack, but I'd prefer not to start like that.

facundominguez:
Now we can make javac visible to ghc to build packages which use inline-java.

Closes #282.

### DIFF
--- a/haskell/toolchain.bzl
+++ b/haskell/toolchain.bzl
@@ -110,6 +110,14 @@ def _haskell_toolchain_impl(ctx):
     targets_r["c2hs"] = ctx.file.c2hs.path
     inputs.append(ctx.file.c2hs)
 
+  extra_binaries_names = set.empty()
+  for binaries in ctx.attr.extra_binaries:
+    for binary in binaries.files:
+      targets_r[binary.basename] = binary.path
+      inputs.append(binary)
+      set.mutable_insert(extra_binaries_names, binary.basename)
+
+  extra_binaries_files = []
   for target in targets_r:
     symlink = ctx.actions.declare_file(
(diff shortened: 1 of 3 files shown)


facundominguez:
Looks like CI fails because of using an old bazel.

mrkkrp:
So maybe we should use `inputs` instead of `tools` for now?

haskell/toolchain.bzl
```diff
@@ -110,6 +110,14 @@ def _haskell_toolchain_impl(ctx):
     targets_r["c2hs"] = ctx.file.c2hs.path
     inputs.append(ctx.file.c2hs)
 
+  extra_binaries_names = set.empty()
+  for binaries in ctx.attr.extra_binaries:
```
mboes: You can use `ctx.files.extra_binaries` as shorthand here.

PR #297, opened 2018-06-07
Single entry point for testing

guibou:
This PR groups all the tests (bazel, failures, starting script) on a single entry point `all_tests.sh`, now used by the continuous integration.

NOTES:

- The output is a bit unreadable by an human, especially for failures tests which print "Failures" in case of success.
- This depends on PR #299


all_tests.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
mboes: Illegal on NixOS. Use `/usr/bin/env bash` instead. POSIX only mandates the presence of `/bin/sh`, not `/bin/bash`.

all_tests.sh
```diff
@@ -0,0 +1,23 @@
...
+set -e
+
+# Bazel tests
+bazel test //... --config=ci
+
+# Test targets that must fail
+tests/targets-that-must-fail/run-tests.sh
```
mboes: Now that we have `all_tests.sh`, let's inline failure tests, rather than keeping the extra level of indirection.

guibou: I was thinking about auto-discovering them using `bazel query`.

mboes: Good idea. They'll need to be marked using the `manual` tag as well, so that `bazel test //...` doesn't fail. This way we don't need a separate WORKSPACE for them.

guibou: Good point!

all_tests.sh
guibou: Fixed

all_tests.sh
guibou: The PR has been updated with:

- Auto detection of failure test
- The separate workspace was removed

PR #300, opened 2018-06-08
Test auto-detection and beautiful test output

guibou:
- Successfull tests are quiet
- Failing test displays its output
- Test (stored in `all_test.sh`) are autodetected

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,0 +1,66 @@
+#!/bin/sh
+
+# This file detects and run all tests
+# USAGE:
+#
+# * Run all tests:
+# $ ./run_tests.sh
+#
+# * Run some tests.sh:
+# $ ./run_tests.sh test_bar test_foo
+
+
+assertBuildFailure() {
+    set +e
+
+    bazel build "$1"
+    if [ $? -eq 0 ]; then
+        echo "bazel build {$1} should have failed."
+        exit 1
+    fi
+
+    set -e
+}
+
+# Color if we are in a terminal, no color otherwise
+if [ -t 1 ]
+then
+    NC='\033[0m'
+    GREEN='\033[0;32m'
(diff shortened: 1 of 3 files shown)


run_tests.sh
```diff
@@ -0,0 +1,66 @@
...
+fi
+
+run_test() {
+  set +e
+  SECONDS=0
+  TEST_ARG=$@
+  echo "running test $TEST_ARG"
+  RES=$($TEST_ARG 2>&1)
+  RESPONSE_CODE=$?
+  DURATION=$SECONDS
```
mboes: Isn't `DURATION` always 0?

run_tests.sh
```diff
@@ -0,0 +1,66 @@
...
+  fi
+}
+
+
+# Load the file which contains the tests
+source ./all_tests.sh
+
+if [ "$#" -eq 0 ]; then
+    # Auto detect tests starting with "test_" and run them
+    TESTS=$(declare -F -p | cut -d " " -f 3 | grep -e '^test_')
```
mboes: This line fails in CI. So the script incorrectly exits with zero return code in this case.

run_tests.sh
mboes: Ah, I didn't know about this shell function.

guibou:
```
       SECONDS
              Each  time this parameter is referenced, the number of seconds since shell invocation is returned.  If
              a value is assigned to SECONDS, the value returned upon subsequent references is the number of seconds
              since  the  assignment plus the value assigned.  If SECONDS is unset, it loses its special properties,
              even if it is subsequently reset.
```

(from `man bash`)

run_tests.sh
guibou: That's unfortunate. I'll find a workaround.

PR #302, opened 2018-06-08
Add a library for looking up runfiles.

judah:
Relevant Bazel issue: bazelbuild/bazel#4460.

I followed the approach documented here:
https://docs.google.com/document/d/e/2PACX-1vSDIrFnFvEYhKsCMdGdD40wZRBX3m3aZ5HhVj4CtHPmiXKDCxioTUbYsDydjKtFDAzER5eg7OjJWs3V/pub
Other than not bothing yet with `RUNFILES_MANIFEST`, which is only needed on
WINDOWS.

The functionality seemed general and simple enough to be worth upstreaming into rules_haskell itself; let me know what you think and/or if there's a better place to put this.


tools/runfiles/Bazel/Runfiles.hs
```diff
@@ -0,0 +1,60 @@
...
+--
+-- For example: @rlocation \"myworkspace/mypackage/myfile.txt\"@
+rlocation :: Runfiles -> FilePath -> FilePath
+rlocation (Runfiles f) g = f </> g
+
+-- | Set environmental variables for locating the given runfiles directory.
+--
+-- Note that Bazel will set these automatically when it runs tests
+-- (@bazel test@).
+runfilesEnv :: Runfiles -> [(String, String)]
```
mboes: What's the use case for this function? Should users set `RUNFILES_DIR` when it's not set? It was my understanding that `RUNFILES_DIR` is set iff the target is a data dependency of another. I don't know about tests.

judah: Added clarification in the comment; it's for non-test binaries to set manually, since from my testing it's only set automatically in "bazel test".

judah:
I switched the module to qualified names and renamed `getRunfiles` to `create`.  (I'm wondering if `create` will be more confusing than helpful; because we're not in an OOP language, it sounds like it's actually constructing the symlink tree or something similar.  But for now it seems fine.)

I avoided renaming the `Runfiles` type to `Handle`, though.  My understanding is that the handle pattern is intended for stateful sequences of IO calls that wrap some form of "service".  In contrast, a `Runfiles` is only ever read from disk once, and then remains a pure, immutable Haskell datatype (and all operations on it are pure).  For example, there's no need for a `close` or `withHandle` function.  So I feel it makes more sense to give it its own type name, like a regular Haskell type, rather than make it seem more stateful than it actually is.  Let me know whether that sounds reasonable to you.

Issue #259, opened 2018-05-21
gcc: error trying to exec 'cc1': execvp: No such file or directory


iphydf:
https://travis-ci.org/iphydf/repro259/jobs/381693316#L1434

Repro: https://github.com/iphydf/repro259

ruuda:
I was able to work around this issue by using GHC from Nixpkgs rather than the bindist.

mboes:
I cannot reproduce. Steps followed:

```
$ docker run -it ubuntu
[...] # install Bazel (as per https://docs.bazel.build/versions/master/install-ubuntu.html#install-on-ubuntu) and make
$ cd /path/to/repro259
$ bazel build //...
$ echo $?
0
```

iphydf:
On Travis it's 100% reproducible: https://travis-ci.org/iphydf/repro259 (e.g. https://travis-ci.org/iphydf/repro259/jobs/393087508#L1382). I can give you push access to that repro repo if it helps?

iphydf:
@ruuda what system are you using where it's reproducible? Do you have a docker image in which it's reproducible?

guibou:
@iphydf I'm interested by the commit right on the repository "iphydf/repro259". I tested on ubuntu / arch linux and I couldn't reproduce the error. However I had to install a few packages on these systems (such as `build-essential` on ubuntu) and perhaps that's just what is missing in your setup.

iphydf:
I added you to the repo as collaborator (you should have received an invitation) yesterday.

The Travis install likely has build-essential. My local install does have that. It has gcc, as, ld, etc.

iphydf:
My local install is Debian 9.

iphydf:
Is there anything I can help with?

iphydf:
It may be related but that pr doesn't fix this issue (I tested this a while ago because I hoped it would). I'll try to reproduce it on a Debian docker image.

iphydf:
I can't do that today. The Travis docker images are here: https://hub.docker.com/u/travisci/ maybe that helps?

guibou:
Full repro at home with this docker file, I will be able to investigate more:

```
FROM travisci/ci-garnet:packer-1512502276-986baf0

RUN apt-get install realpath

RUN git clone https://github.com/iphydf/repro259.git repro259

# install bazel
RUN wget https://github.com/bazelbuild/bazel/releases/download/0.14.1/bazel-0.14.1-installer-linux-x86_64.sh
[...]

PR #304, opened 2018-06-10
Tether haskell toolchain and CC toolchain

### LINKED ISSUE #250: Linking error when using a Template Haskell library with external C++ deps
We are seeing a linking error against `libstdc++.so.6` with the following setup:

- A `WORKSPACE` file declaring an external repository with:
  - A C++ library.
  - A Haskell library depending on the C++ library and exposing Template Haskell functions.
- In the main workspace: a Haskell library including a TH splice the Template Haskell library.

I have pushed a reproducible test case here:
https://github.com/gdeest/repro_link_error

(Note that you will need to change the path to the `other-repo` directory in `repro-bug/WORKSPACE`. That is because `new_local_repository` requires absolute paths.)

Here is the error message we get:
```
ERROR: /home/damien/workspace/repro-archive/repro-bug/BUILD:14:1: Compiling lib failed (Exit 1)
<command line>: can't load .so/.DLL for: bazel-out/k8-fastbuild/bin/external/other_repo/libHSexternalZSotherZUrepoZShslib-1.0.0-ghc8.2.2.so (libstdc++.so.6: cannot open shared object file: No such file or directory)
INFO: Elapsed time: 3.458s, Critical Path: 1.01s
```
Apparently, the transitive runtime deps required by the Template Haskell library can't be seen at compile-time when those deps come from an external location. Removing the TH splice makes the failure disappear.

Additional data points:
- The problem occurs on Ubuntu 18.04 LTS, but not on NixOS. 
- The behavior is the same on 0.4 and latest master. 

cc @dleprovost

mboes:
GHC calls GCC itself when it wants to do some linking, assembly, or
preprocessing. The problem is, GHC might not be calling the same GCC we
expect. While we expect the GCC autodected for the Bazel CC toolchain, GHC
uses whatever GCC is found in a `settings` file that it ships with.

The key is to override the defaults by passing `-pgm*` flags to GHC. In
this way, GHC uses the same GCC as the rest of the Bazel rules, hence
avoiding any potential linker issues.

Fixes #250.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,11 +67,11 @@ jobs:
           shell: /bin/bash -eilo pipefail
           command: |
             nix-build ./tests/protoc_gen_haskell.nix | cachix push tweag
-            nix-shell --run 'bazel build --jobs=2 //...'
+            nix-shell --run 'CC=$(which clang) bazel build --jobs=2 //...'
       - run:
           name: Run tests
           shell: /bin/bash -eilo pipefail
-          command: nix-shell --run ./run_tests.sh
+          command: nix-shell --run 'CC=$(which clang) ./run_tests.sh'
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,17 +62,16 @@ jobs:
           command: |
             brew update
             brew install bazel
-            brew install go
       - run:
           name: Build
           shell: /bin/bash -eilo pipefail
           command: |
             nix-build ./tests/protoc_gen_haskell.nix | cachix push tweag
-            bazel build --jobs=2 //...
+            nix-shell --run 'bazel build --jobs=2 //...'
       - run:
           name: Run tests
           shell: /bin/bash -eilo pipefail
-          command: ./run_tests.sh
+          command: nix-shell --run ./run_tests.sh
 
 workflows:
   version: 2
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -5,8 +5,8 @@ haskell_repositories()
 
 http_archive(
   name = "io_tweag_rules_nixpkgs",
-  strip_prefix = "rules_nixpkgs-0.2.2",
-  urls = ["https://github.com/tweag/rules_nixpkgs/archive/v0.2.2.tar.gz"],
+  strip_prefix = "rules_nixpkgs-0.2.3",
+  urls = ["https://github.com/tweag/rules_nixpkgs/archive/v0.2.3.tar.gz"],
 )
 
 load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl",
@@ -21,7 +21,10 @@ nixpkgs_git_repository(
   # use a version of nixpkgs that is newer than 18.03.
 
   # Keep this value in sync with `nixpkgs.nix`
-  revision = "7c3dc2f53fc837be79426f11c9133f73d15a05c4",
+  revision = "8d5f06cbeedcfb4a88c111754baf2d7e315e74de",
+  # TODO Using a fork with Bazel v0.15. Switch to mainline once
+  # https://github.com/NixOS/nixpkgs/pull/42735 merged.
+  remote = "https://github.com/mboes/nixpkgs",
 )
 
 nixpkgs_package(
--- a/haskell/cc.bzl
+++ b/haskell/cc.bzl
@@ -19,14 +19,22 @@ CcInteropInfo = provider(
   fields = {
     "hdrs": "CC headers",
     "cpp_flags": "Preprocessor flags",
+    "compiler_flags": "Flags for compilation",
+    "linker_flags": "Flags to forward to the linker",
     "include_args": "Extra include dirs",
   }
 )
 
-def cc_headers(ctx):
-  """Bring in scope the header files of dependencies, if any.
+def cc_interop_info(ctx):
+  """Gather information from any CC dependencies.
 
   *Internal function - do not use.*
+
+  Args:
+    ctx: Rule context.
+
+  Returns:
+    CcInteropInfo: Information needed for CC interop.
   """
   hdrs = depset()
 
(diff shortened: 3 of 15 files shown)


mboes:
Looks like the Darwin build failure has been fixed upstream in Clang: https://reviews.llvm.org/D35462. We'll need a workaround meanwhile.

gdeest:
I now see the following build failure with this PR:

```
<command line>: can't load .so/.DLL for: /nix/store/27x7pinqdsl9f3rpbm8bsszd9fhwq266-glibc-2.27/lib/librt.so (/nix/store/27x7pinqdsl9f3rpbm8bsszd9fhwq266-glibc-2.27/lib/librt.so: symbol __close_nocancel, version GLIBC_PRIVATE not defined in file libc.so.6 with link time reference)
```

mboes:
That's great! What minimal repro can I use to test this?

mboes:
The Darwin issue is nasty. The problem is this. The GCC compiler version used in Nixpkgs supports the `-no-pie` flag. GHC is not keen on PIE for some reason so it passes `-no-pie` if it thinks the compiler knows how to do PIE. Whether this is the case is controlled by the `settings` file. This `settings` file is found at an absolute path that cannot be overridden on the CLI or with an environment variable. That's a big problem, because on macOS the CC toolchain detected by Bazel is the system one, i.e. the Clang that ships with Xcode, yet Clang doesn't support the `-no-pie` flag. Not because Clang doesn't do PIE. But because turning off PIE in Clang is done with `-nopie` (without the middle hyphen).

There's no way to prevent GHC from passing `-no-pie`, i.e. a flag Clang doesn't understand, without modifying the `settings` file. That's of course not possible or at the very least a bad idea, since this is a file in a system folder somewhere, so no perms to modify or otherwise read-only. Newer Clang's understand `-no-pie` as an alias to `-nopie`, but I wasn't able to convince Xcode to upgrade itself on the headless macOS VM's we have on CircleCI. See https://reviews.llvm.org/D35462.

The workaround to all this is to make Bazel use the Nixpkgs GCC as the CC toolchain. But the root issue is that GHC should let us provide our own `settings` file, or equivalently make sure that all fields in the settings file are overridable. That's an important feature request for that we should file for future GHC versions.

gdeest:
I can reproduce the `librt` / `libc` issue with the following:

https://github.com/gdeest/repro_librt

I am unsure how useful the `shell.nix` file is, but it should make sure that the GCC versions used by GHC and the one present in the environment are different..

cc @mboes @ghorn

mboes:
@gdeest I think I got to the bottom of the issue you're seeing. I claim that it's an upstream bug.

Let's start with the following observation: after making sure the revisions for Nixpkgs in the `WORKSPACE` and `shell.nix` files match (they didn't), adding `nixpkgs.bazel` to the `buildInputs` in your `shell.nix` file fixes the issue. That's unexpected, but clearly that means bazel is somehow choosing the C compiler not by picking it in the current path, but via some hardcoded path inside Bazel itself. Different bazels find different C compilers. This gave me a clue as to the real culprit:

https://github.com/NixOS/nixpkgs/blob/6f03c349ed4dff3f5d7e3ae7df141b0db954eb38/pkgs/development/tools/build-managers/bazel/default.nix#L121

This is overriding the `PATH` that `bazel` sees to contain the C compiler that was in the `stdenv` when compiling Bazel itself. Hence why you end up with a `gcc` using one nixpkgs, and a `ghc` using another nixpkgs (and therefore potentially a different `glibc`).

mboes:
@guibou PTAL. When you do, you might be surprised to notice that I have not only kept things substantially the same, I've doubled down, in an effort to simultaneously solve #250 and #259.

> I'm now convinced that the correct way of fixing this issue is to let GHC use the GCC it as been built with and to use a similar GCC version to build other files.

I disagree on four counts:
* GHC is quite happy using (most) GCC's you provide it, *by design*. That's why the `-pgm*` flags are there. There are known limitations of the `-pgm*` flags currently (see https://ghc.haskell.org/trac/ghc/ticket/15319), but we can work with that.
* You objected that the GHC and GCC might not be expecting the same Glibc, but inconsistent Glibc's will cause problems with or without this PR, making that a moot point for this discussion (indeed that's what we already see in #250).
* There is no good way in general to let GHC impose its own GCC. Because you can't assume our toolchains come from Nixpkgs.
* #259 brought to light that we must explicitly share the same subset of compiler and linker flags across all rules (see below). This makes it in general unsafe to implicitly rely on a consistent environment, because with shared flags we're explicitly forcing the same versions of the toolchains everywhere anyways (flags are compiler version specific).

Recall that the original impetus for this PR is that in #250, @gdeest and others were observing weird linker errors because the build was inconsistently using the CC toolchains. Native CC rules were using one compiler and GHC was transitively calling another, even with a `shell.nix` to control the version of the compiler. With a *pure* `shell.nix`, the problem would go away altogether, but we should not impose the requirement of pure Nix shells on users. rules_haskell is currently agnostic about the source of toolchains (Nixpkgs or elsewhere). Keeping it that way means that using consistent toolchain versions must be done at the Bazel level, not externally.

A good indication that this is the right thing to is this PR fixes another unrelated issue, #259, practically for free. And does it better than my broken attempt in #320. To fix #259 on older Ubuntu's and on Arch Linux as well (not just Ubuntu 18.04), we need to furthermore ensure not just consistent compiler versions, but that C compiler and linker flags are also consistent across the board. That's straightforward to do if we're explicitly setting the C compiler and the linker GHC must call (and unsafe to do if we don't). I could have kept those changes for a followup PR, but doing it all in one PR means we can have a complete fix for #259 on all linux distributions now.

mboes:
@iphydf @ruuda could you confirm this PR fixes #259 for you even on Ubuntu 16.04 and Arch? Travis seems to say that it has: https://travis-ci.com/iphydf/repro259/builds/77803230.

iphydf:
Almost there:
```
external/haskell_clock/BUILD.bazel:3:1: HaskellBuildLibrary @haskell_clock//:clock failed (Exit 1)

<no location info>: error:
    Warning: Couldn't figure out linker information!
             Make sure you're using GNU ld, GNU gold or the built in OS X linker, etc.
collect2: fatal error: cannot find 'ld'
compilation terminated.
Target @haskell_clock//:clock failed to build
INFO: Elapsed time: 3.938s, Critical Path: 0.86s
```

It gets past assembling, but stops at linking.

mboes:
Gah! #259 is like a hydra that keeps on growing new heads.

@iphydf do you have a repro? Cuz the original one seems to pass now (see link in previous message).

@ruuda could you tell me what passing `-s` to the build command says?

Let's continue both these discussions in #259.

iphydf:
I can confirm that my repro259 works on my machine. I'm investigating the error I get on the real repo now.

PR #305, opened 2018-06-10
Create a nix FHS environment

guibou:
Bazel supposes a lot of hardcoded binaries, such as /usr/bin/{ar,cc,c++},
... which are not available in NixOS.

Nix provided bazel is already hacked to hardcode some of these paths, but
not all of them, such as /usr/bin/ar

This FHS environment contains all the needed binaries at the "usual"
location supposed by bazel.

### DIFF
--- a/fhs.nix
+++ b/fhs.nix
@@ -1,0 +1,19 @@
+/*
+
+This FHS env allows bazel to run on nixos:
+
+$ nix-shell ./fhs.nix
+
+then, `bazel build //something` will work without any other configuration.
+
+NOTE: for an unknown (yet) reason, part of bazel cache is invalidated if the nix-shell is re-run.
+
+*/
+
+{ pkgs ? import <nixpkgs> {} }:
+
+(pkgs.buildFHSUserEnv {
+  name = "rules-haskell-env";
+  targetPkgs = pkgs: (with pkgs;
+    [ bazel nix binutils go which python27]);
+}).env


mboes:
I'm using NixOS as my base system and using Bazel everyday on it just fine. I do think having a `shell.nix` is a good idea, in particular to control the recommended Bazel version during development. But I don't see why this needs to be an FHS environment. Could you explain why? In particular, I'm looking for an anwser in the form, "I tried X but it didn't work for Y reason".

Bazel is available as a Nixpkgs package. If there are remaining hard coded paths, and if they do cause problems, then let's fix that at the source.

guibou:
Actually I'm a bit lost. I'm sure that I tried:

- `nix-shell -p bazel binutils which go python27`
- `nix-env -i bazel binutils which go python27` (as user or as root)
- Adding `bazel binutils which go python27` in my `environment.systemPackages` in `/etc/nixos/configuration.nix`

It was complaining about a dangling `/usr/bin/ar` symlink.

But well, I tried again, and it works, even after a complete bazel cache clean.

*edit*: actually there is a difference between the package name `binutils` and the "attribute name" `binutils`, one is a wrapper which does not contain `ar` and the other is not. My mistake may come from here. Sorry.

mboes:
OK but nevertheless, adding a `shell.nix` as you were proposing is still a good idea. It just doesn't need to be FHS based. See e.g. https://github.com/tweag/inline-java/blob/master/shell.nix for an example of how we probably want to do it: with a companion `nixpkgs.nix` to pin the nixpkgs version. In turn, it would be nice if the `nixpkgs_repository` rule could consume the `nixpkgs.nix` file, to stay in sync.

guibou:
See #307, a simple `shell.nix`. I did something simpler than the `inline-java` one, we may improve it later. I'll have a look at `rules_nixpkgs` to improve `nixpkgs_repository` in the way you described.

mboes:
Simpler in what way?

guibou:
Does not handle anything about `openjdk` or `ghc` dependencies, we'll add this later if we want nix to handle this part of environment, but I had the impression that we wanted this to be done inside bazel.

mboes:
Those parts are irrelevant to rules_haskell. Bazel itself depends on OpenJDK. It's a good question whether Bazel nevertheless wants some `javac` in the `PATH`, but anything to do with controlling the versions of toolchains is something for the `WORKSPACE` file, not the `shell.nix` file.

PR #308, opened 2018-06-12
Propagate runfiles in cc_haskell_import.

facundominguez:
This is necessary so the rules depending on the output of `cc_haskell_import` can see the runfiles.


facundominguez:
cc @mboes

facundominguez:
This is superseded by https://github.com/tweag/clodl/commit/d588f0c1e9b0776ed6558eb89e06e7a62263c6cd

I'm merging it because I suppose it might still be useful to someone in the future. But clodl does no longer depend on it.

PR #310, opened 2018-06-14
Make a separate toolchain for `doctest`.

judah:
For Hazel, we would like `haskell_doctest` to use a `doctest` binary built
by Hazel itself.  However, since `haskell_binary` depends on the
`haskell_toolchain`, which takes the binary as an attribute, doing so would
currently cause a dependency loop.  (The alternative is to pull it from
Nix, which is redundant since Hazel can already build its library
dependencies.)

Making a separate toolchain resolves the loop, and also seems fairly
reasonable since the doctest parameter was already optional in
`haskell_toolchain`.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -81,6 +81,7 @@ nixpkgs_package(
 
 register_toolchains(
   "//tests:ghc",
+  "//tests:doctest-toolchain",
   "//tests:protobuf-toolchain",
 )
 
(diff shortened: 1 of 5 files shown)


judah:
The corresponding change to Hazel: https://github.com/FormationAI/hazel/commit/f41e069a2541dc41f7e6d1eeecafd9059ad39e7f

PR #320, opened 2018-06-28
Fix missing inputs declaration on hs.tools.gcc

### LINKED ISSUE #259: gcc: error trying to exec 'cc1': execvp: No such file or directory

mboes:
Fixes #259


mboes:
@iphydf @ruuda could you confirm that this fixes #259 for you?

iphydf:
Yes, it works for most libraries, but not for e.g. text, which depends on a cc_library. I'll try some things.

iphydf:
Actually no, it doesn't work for clock either. I don't know what I saw that made me think it works. Did you get clock to build?

mboes:
Yes. In @guibou's Docker container, appropriately patched to use a `rules_haskell` that includes this branch.

```
root@2088ef99a88d:/repro259# bazel build @haskell_clock//:clock
Starting local Bazel server and connecting to it...
.
INFO: Analysed target @haskell_clock//:clock (10 packages loaded).
INFO: Found 1 target...
Target @haskell_clock//:clock up-to-date:
  bazel-bin/external/haskell_clock/externalZShaskellZUclockZSclock-0.7.2/externalZShaskellZUclockZSclock-0.7.2.conf
  bazel-bin/external/haskell_clock/externalZShaskellZUclockZSclock-0.7.2/package.cache
INFO: Elapsed time: 2.297s, Critical Path: 0.03s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
```

iphydf:
Investigating.. I'm doing bazel clean --expunge, and also testing the repro259 case. I'll come back with results either later today or tomorrow (UTC).

iphydf:
https://travis-ci.org/iphydf/repro259/jobs/397999549#L1383 <- it fails like this on my debian machine.

iphydf:
Note that it does get further along, so this PR does fix something :).

mboes:
We'll have to workaround https://github.com/bazelbuild/bazel/issues/5494 to make progress on older Ubuntu's.

ruuda:
The build does fail on my Arch system now though, where I have set `CC=clang CXX=clang++ CPP="clang -E"`. Even when I prefix `bazel build` with `CC=gcc CXX=g++ CPP="gcc -E"` I still get the following message, also after first running `bazel shutdown`:

```
ERROR: /home/ruud/.cache/bazel/_bazel_ruud/6e22cf90e77d4f0962fbd130038e840e/external/hackage_time_locale_compat/BUILD.bazel:4:1: HaskellBuildLibrary @hackage_time_locale_compat//:time-locale-compat failed (Exit 1)
ghc: could not execute: clang
Target //:vaultenv failed to build
```

Where does it get `clang` from?

iphydf:
@ruuda try `bazel clean --expunge`. It may have stored `clang` in the "configure" step, i.e. the WORKSPACE building step, which is cached.

ruuda:
@iphydf Thanks, that did the trick. After doing a `bazel clean --expunge` and then building with `CC=gcc CXX=g++ CPP="gcc -E" bazel build`, I now get this error on my Arch system:
```
ERROR: /home/ruud/.cache/bazel/_bazel_ruud/6e22cf90e77d4f0962fbd130038e840e/external/hackage_attoparsec/BUILD.bazel:3:1: HaskellBuildLibrary @hackage_attoparsec//:core failed (Exit 1)
gcc: error trying to exec 'as': execvp: No such file or directory
`gcc' failed in phase `Assembler'. (Exit code: 1)
Target //:vaultenv failed to build
```

mboes:
Right, so the assembler thing is trickier to fix, because of an upstream issue: https://github.com/bazelbuild/bazel/issues/5494. I attempted a workaround on this branch but ultimately, I think merging #304 is going to be a better solution, though we still need a bit more work to also handle the assembler properly. Note that this is a problem specific to the `ghc_bindist` rule. You can get GHC from Nixpkgs as workaround while I fix this.

PR #322, opened 2018-07-02
Fix loading of external shared libraries in REPL

### LINKED ISSUE #312: REPL script fails to load shared library
It looks like REPL scripts for Haskell targets depending on `rules_nixpkgs`-provided libraries fail to find these dependencies on load

**Example:**

This is on latest master.

`WORKSPACE`
```python
workspace(name = "repro")

rules_haskell_rev = "1d3daac33f90b0e2fec8114fd4599782494471d0"
http_archive(
    name = "io_tweag_rules_haskell",
    strip_prefix = "rules_haskell-{}".format(rules_haskell_rev),
    urls = ["https://github.com/tweag/rules_haskell/archive/{}.tar.gz".format(rules_haskell_rev)],
)

load("@io_tweag_rules_haskell//haskell:repositories.bzl", "haskell_repositories")
haskell_repositories()

http_archive(
  strip_prefix = "rules_nixpkgs-0.2",
  urls = ["https://github.com/tweag/rules_nixpkgs/archive/v0.2.tar.gz"],
)

load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl",
  "nixpkgs_git_repository",
  "nixpkgs_package"
)

nixpkgs_package(
  name = "toolchain",
  nix_file_content = """
let config = {};
    pkgs = import <nixpkgs> { config = config; }; in
{
  "islEnv" = rec {
    toolchain = pkgs.buildEnv rec {
      name = "isl-bindings-toolchain";
      extraOutputsToInstall = [ "lib" "dev" ];
      paths = with pkgs; [
        (haskell.packages.ghc841.ghcWithPackages (p: with p;
          [ base
          ]
        ))
        isl
      ];
    };
  };
}
""",
  attribute_path = "islEnv.toolchain",
  build_file_content = """
filegroup(
  name = "bin",
  srcs = glob(["bin/*"]),
)

filegroup(
  name = "isl_so",
  srcs = ["lib/libisl.so"],
)

filegroup(
  name = "isl_headers",
  srcs = glob(["include/isl/**/*.h"]),
)
"""
)

register_toolchains("//:toolchain")
```

`BUILD`:
```python
package(default_visibility = ["//visibility:private"])

load(
  "@io_tweag_rules_haskell//haskell:haskell.bzl",
  "haskell_cc_import",
  "haskell_library",
  "haskell_toolchain",
)

haskell_cc_import(
  name = "isl",
  hdrs = ["@toolchain//:isl_headers"],
  shared_library = "@toolchain//:isl_so",
)

haskell_toolchain(
  name = "toolchain",
  version = "8.4.1",
  tools = "@toolchain//:bin",
)


haskell_library(
  name = "some-lib",
  deps = [":isl"],
  srcs = [ "src/SomeLib.hs" ],
  src_strip_prefix = "src",
  prebuilt_dependencies = [
    "base",
  ]
)
```

`src/SomeLib.hs`:
```haskell
{-# LANGUAGE ForeignFunctionInterface #-}
module SomeLib where

import Foreign.Ptr

foreign import ccall "isl/ctx.h isl_ctx_alloc" c_ctx_alloc :: IO (Ptr ())
```

**Output:**

```
[huginn@frigg:~/code/repl-repro]$ bazel run --direct_run some-lib-repl
INFO: Analysed target //:some-lib-repl (0 packages loaded).
INFO: Found 1 target...
Target //:some-lib-repl up-to-date:
  bazel-bin/some-lib-repl
INFO: Elapsed time: 0.117s, Critical Path: 0.00s
INFO: Build completed successfully, 1 total action
GHCi, version 8.4.1: http://www.haskell.org/ghc/  :? for help
<command line>: user specified .o/.so/.DLL could not be loaded (libisl.so: cannot open shared object file: No such file or directory)
Whilst trying to load:  (dynamic) isl
Additional directories searched:   external/toolchain/lib
```

**Other things tried:**

I have tried specifying versioned `.so` files `lib/libisl.so.15` and `lib/libisl.so.15.2` instead of `lib/libisl.so` (as this fixed issues I had with shared libraries on CXX code). Then linking fails with:

```
[huginn@frigg:~/code/repl-repro]$ bazel build ...
INFO: Analysed 3 targets (0 packages loaded).
INFO: Found 3 targets...
ERROR: /home/huginn/code/repl-repro/BUILD:23:1: HaskellLinkDynamicLibrary libHSsome-lib-1.0.0-ghc8.4.1.so failed (Exit 1)
/nix/store/b0zlxla7dmy1iwc3g459rjznx59797xy-binutils-2.28.1/bin/ld: cannot find -lisl
INFO: Elapsed time: 0.573s, Critical Path: 0.40s
```

mrkkrp:
Close #312.

In order to load various things into GHCi we need to know their location. Since we're calling the script from workspace root we need to know relative path from there to all the components like shared libraries.

There are two problems to solve with this approach:

* Previously we relied on the fact that Bazel creates "convenience symlinks", e.g. `bazel-bin`, `bazel-out`, etc. And so we could get `path` of a `File` and use it directly as a relative path. This does not work for external dependencies (from other repos) because these live under `bazel-<workspace>/external` and there is no "convenience symlink" to that `external` directory.
* The convenience symlinks are not guaranteed to use the `bazel-` prefix, because user can change the prefix to anything he/she links using the `--symlink_prefix` option.

However, due to "internal issues" as mentioned in e.g. bazelbuild/bazel#3796, `bazel-out` is always created no matter what prefix user specifies manually. This gives us a way to reliably locate top-level output directory. (That directory is even harder if at all possible to locate from Skylark because of bazelbuild/bazel#5506.) The idea is to `readlink` of that `bazel-out` symlink and then go up one level. Then we use this as a prefix for various things. It appears thath `LD_LIBRARY_PATH` is quite enough to specify the directories that GHCi should search for shared libraries, so I dropped `-L` flags because they will be quoted with single quotes anyway and the trick of prefixing them with `$RULES_HASKELL_EXEC_ROOT` won't work.


haskell/private/ghci_repl_wrapper.sh
```diff
@@ -6,15 +6,34 @@
...
+fi
+
+# This is a workaround for https://github.com/bazelbuild/bazel/issues/5506
+# and also for the fact that REPL script relies on so-called “convenience
+# links” and the names of those links are controlled by the --symlink_prefix
+# option, which can be set by the user to something unpredictable.
+#
+# It seems that we can't locate the files of interest/build outputs in
+# general. However, due to “internal issues” in Bazel mentioned e.g.
+# https://github.com/bazelbuild/bazel/issues/5506, the directory bazel-out
```
guibou: did you mean `https://github.com/bazelbuild/bazel/issues/3796`?

haskell/private/ghci_repl_wrapper.sh
```diff
@@ -6,15 +6,34 @@
 # likely that the script is invoked incorrectly, so we need to output a
 # warning with instructions how to invoke it.
 
-GHCI_LOCATION="{GHCi}"
-SCRIPT_LOCATION="{SCRIPT_LOCATION}"
+if [ ! -d "bazel-out" ]; then
+  echo "The bazel-out symlink must be present in workspace directory."
```
guibou: As this fix will stop working when bazelbuild/bazel #3796 will be fixed, perhaps an error message with more information can be useful. Especially I can imagine a user with an indirect bazel upgrade ending to this error message. Something such as `"The bazel-out symlink must be present in workspace directory. This is not the case and it means that you upgraded to a bazel version not (yet) supported by rules_haskell. Please report the issue to rules_haskell."

mrkkrp:
@guibou Good catches, I addressed your feedback, PTAL.

guibou:
LGTM.

PR #326, opened 2018-07-04
Remove ‘haddock-util.js’

mrkkrp:
Modern versions of Haddock do not create this file, and Bazel complains about outputs that were not created. A thing to check: whether Haddock that comes with GHC 8.2 works well with this change.

### DIFF
--- a/haskell/haddock.bzl
+++ b/haskell/haddock.bzl
@@ -175,9 +175,12 @@ def _haskell_doc_rule_impl(ctx):
 
   # Do one more Haddock call to generate the unified index
 
+  index_root_raw = paths.join(doc_root_raw, "index")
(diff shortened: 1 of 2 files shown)


mrkkrp:
Actually removing that file has a minor negative effect on users of GHC 8.2: when you switch to viewing sources and hover over a variable name all other occurrences of this variable are not highlighted.

Ideally we'd like to do the right thing for all GHC versions, but I'm not sure how to do it.

mrkkrp:
Actually we could just check GHC version (from toolchain) and include `haddock-util.js` conditionally. I'll push this shortly.

mboes:
Is there no way to avoid having to declare all haddock output files individually?

PR #328, opened 2018-07-09
Add explicit settings for ‘locale’ and ‘locale_archive’

mrkkrp:
This PR allows to avoid issues related to reading Unicode text files at compile time. Previously, due to unset locale, locale defaulted to ASCII and reading Unicode text was impossible. This prevented use of e.g. Dhall's TH helpers.


Profpatsch:
Note that the default `glibc` locale package is >20MB in size iirc. It provides a mechanism to create a database with a subset of locales though, like so: https://gitlab.techcultivation.org/sangha/sangha-deployment/blob/master/containers/helpers/gen-utf8-localedb.nix

haskell/toolchain.bzl
```diff
@@ -271,6 +278,14 @@ _haskell_toolchain = rule(
         doc = "Additional binaries necessary for building",
         default = [],
     ),
+    "locale": attr.string(
+      default = "en_US.UTF-8",
+      doc = "Locale that will be set during compiler invocations.",
+    ),
+    "locale_archive": attr.label(
+      allow_single_file = True,
+      doc = "Label pointing to the locale archive file to use.",
```
judah: Add a note that this is nix-specific, or a link to docs?

haskell/toolchain.bzl
```diff
@@ -271,6 +278,16 @@ _haskell_toolchain = rule(
...
         default = [],
     ),
+    "locale": attr.string(
+      default = "en_US.UTF-8",
+      doc = "Locale that will be set during compiler invocations.",
+    ),
+    "locale_archive": attr.label(
+      allow_single_file = True,
+      doc = """
+Label pointing to the locale archive file to use. Mostly useful on NixOS.
```
mboes: So it turns out `LOCALE_ARCHIVE` is completely nixpkgs specific, not a POSIX thing. They patch glibc to handle that environment variable. Not sure why it's not called `NIX_LOCALE_ARCHIVE` but whatever. Given this, it's a shame that we're polluting rules_haskell with Nixpkgs specifc hacks. A better fix long term will be to submit a patch so that the `ghc` derivation in Nixpkgs includes a locale archive in its closure.

@Profpatsch thoughts?

Profpatsch: Hm, I don’t know. That would be for @peti to decide.

According to `glibc` source code it’s normally fixed at configuration time via the `COMPLOCALEDIR` option, instead of being a variable path at runtime. I’m not sure why it’s solved with an env var instead of a fix filepath (e.g. `/var/lib/locale-archive`) in nixpkgs.

peti: > I’m not sure why it’s solved with an env var instead of a fix filepath (e.g. /var/lib/locale-archive) in nixpkgs.

That path needs super user privileges to set up and it's system-wide. `$LOCALE_ARCHIVE`, on the other hand, can be configured on a per-process basis and needs no privileges to set or unset. See https://github.com/NixOS/nixpkgs/issues/38991 for an example why such concerns may be important.

Issue #324, opened 2018-07-03
Profiling build of rules_haskell no longer works


mboes:
The build using the `dbg` configuration fails currently:

```
$ cd rules_haskell
$ nix-shell --pure
$ bazel build -c dbg //...
INFO: Build options have changed, discarding analysis cache.
DEBUG: /home/mboes/.cache/bazel/_bazel_mboes/822aa44857e17f47d08bb198503f3032/external/io_bazel_rules_go/go/private/rules/binary.bzl:54:5: DEPRECATED: linkstamp, please use x_def for all stamping now main
INFO: Analysed 144 targets (0 packages loaded).
INFO: Found 144 targets...
INFO: From Executing genrule @com_github_bazelbuild_buildtools//build:parse.y.go_yacc:

conflicts: 58 shift/reduce
ERROR: /home/mboes/src/tweag/rules_haskell/tools/runfiles/BUILD:8:1: HaskellBuildLibrary //tools/runfiles:runfiles failed (Exit 1)

<interactive>:1:1: error:
    Failed to load interface for `GHC.Integer.Type'
    Perhaps you haven't installed the "p_dyn" libraries for package `integer-gmp-1.0.1.0'?
    Use -v to see a list of the files searched for.
INFO: Elapsed time: 3.295s, Critical Path: 0.76s
INFO: 3 processes: 1 local, 2 processwrapper-sandbox.
FAILED: Build did NOT complete successfully
```

mrkkrp:
To summarize:

* GHC bindists typically do not contain libraries compiled in `p_dyn` way.
* Thus `-prof` with `-dynamic` or `-dynamic-too` is not going to work.
* Cabal just doesn't compile dynamic versions of things when `-prof` is enabled.

So the solution seems to be: do the same thing as Cabal does and skip creation of dynamic libraries when profiling is enabled.

BTW, @Fuuzetsu do you have any thoughts?

Fuuzetsu:
I'd copy cabal and enquire into why GHC bindist doesn't ship with p_dyn.

mrkkrp:
Resolved by #332.

PR #332, opened 2018-07-11
Add support for profiling

mrkkrp:
This PR adds support for building things with profiling.


mboes:
Fixes #324. cc @iphydf

mrkkrp:
OK, 7 tests fail with profiling. I'll look into this.

iphydf:
My `@haskell_clock` test doesn't seem to work with this, yet. I haven't tested on travis, but it doesn't work locally. https://github.com/iphydf/repro259 is still a valid repro for my issue (with `-c dbg`).

haskell/private/actions/compile.bzl
```diff
@@ -175,7 +174,36 @@ def _process_chs_file(hs, cc, ghc_defs_dump, chs_file, chi_files=[]):
 
   return hs_out, chi_out, idir
 
-def _compilation_defaults(hs, cc, java, dep_info, srcs, extra_srcs, cpp_defines, compiler_flags, main_file = None, my_pkg_id = None):
+def _output_file_ext(base, dynamic, profiling_enabled):
```
mboes: I find this function doesn't pay its own way and could be more harmful than good. The cheap option is to construct the file name extensions manually every time: `.p_o_dyn` etc. It should prove more readable. Also your line breaking below won't need to be as ugly, because the `declare_compiled` stuff will fit on one line.

haskell/private/haskell_impl.bzl
```diff
@@ -95,19 +129,43 @@ def haskell_library_impl(ctx):
...
+  c_p = None
+
+  if with_profiling:
+    c_p = hs.toolchain.actions.compile_library(
+      hs,
+      cc,
+      java,
+      dep_info,
+      srcs = ctx.files.srcs,
+      # NOTE We must make the object files compiled without profiling
```
mboes: Or use `-fext-external-interpreter`, which I believe should resolve this kind of problem. See https://ghc.haskell.org/trac/ghc/wiki/RemoteGHCi. I'm surprised it's still not the default - we should pass that flag systematically if not.

haskell/private/actions/compile.bzl
```diff
@@ -212,8 +250,11 @@ def _compilation_defaults(hs, cc, java, dep_info, srcs, extra_srcs, cpp_defines,
   args.add(compiler_flags)
   haddock_args.add(compiler_flags, before_each="--optghc")
 
-  # Output static and dynamic object files.
-  args.add(["-static", "-dynamic-too"])
```
mboes: If I understand correctly this hack is a workaround to an upstream issue in GHC. It would be good to call that out explicitly, and to look to the relevant upstream issue.

haskell/private/actions/compile.bzl
mrkkrp: Yes and no... I think it reduces the possibility of typos somewhat, but the main reason is the `with_profiling` argument. If we "inline" `_output_file_ext` (with some beta-reduction) there will be a lot of conditionals of the form:

```
"." + ("_p" if with_profiling else "") + "dyn_o"
```

Etc. which is not much better and definitely not shorter. I'd leave the function "as is".

(Well another option would be something like this: `".p_dyn_o" if with_profiling else ".dyn_o"` which is typo-prone I think and not really shorter either.)

haskell/private/haskell_impl.bzl
mrkkrp: You meant `-fexternal-interpreter` I think?

https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/ghci.html?highlight=#ghc-flag--fexternal-interpreter

This looks good, although I'm not sure it makes sense to spend more time switching to that as I've already spent more hours than I expected to spend with this. Also, right now we're doing what Cabal is doing here. Of course Cabal works that way because it needs to support older GHC, but maybe there are other undiscovered issues with `-fexternal-interpreter`? CC @facundominguez, he mentioned something like "there may be issues with that".

Let's open an issue for the future and leave this "as is" so I can spend more time on client's stuff instead of re-working the working solution that replicates what Cabal does?

haskell/private/actions/compile.bzl
mrkkrp: I don't know if there is a ticket for this, actually. CC @hvr

mrkkrp: After some searching, opened https://ghc.haskell.org/trac/ghc/ticket/15394 for now. Even if it's a dupe, we'll know that soon.

mrkkrp: I added a note in the source too.

Issue #75, opened 2017-12-31
Merge deps and prebuilt_dependencies


Fuuzetsu:
We have a problem in that to build anything non-trivial, we need `prebuilt_dependencies`: this is a list of packages that we expect to be available to GHC. It is up to the user to make sure a toolchain where they are present is used.

Sadly this causes a big inconvenience: the user has to *both* list everything in `prebuilt_dependencies` and then manually aggregate all these lists for all packages that toolchain will build back to toolchain definition. This is extremely tedious and goes out of date very quickly. We would like to only list packages in a single place: the dependencies for the package.

There is a way to achieve this and that is to stop using `prebuilt_dependencies` and instead pass honest bazel targets to `deps`. The question is where to get these targets. For purpose of this ticket, I will assume `nix` is still the answer. Note however there is no dependency on `rules_nixpkgs` from `rules_haskell`: in theory the user will be able to use the same mechanism with a different backend. It's up to the user.

Instead of generating a toolchain using `ghcWithPackages (p: [ every-package-we-need-ever ])`, generate package sets. A good candidate for this are LTS packages (but note user is not limited to these: they can cabal2nix/stackage2nix anything they want and use it just the same). The user defines:

```
http_archive(
  name = "lts",
  repository = …,
  strip_prefix = "lts-collection/lts-10.0",
)
```

Inside `lts-10.0` there's a BUILD file with one bazel target per one package in the snapshot. These targets are provided by `lts-10.0/WORKSPACE` and may look a little like

```
nixpkgs_package(
  name = "aeson",
  attribute = "haskellPackages.aeson",
  build_file_content = """
filegroup(
  name = "haskell-outputs",
  …
  srcs = ..., # package db, package conf, libraries, everything needed.
)
"""
)
```

then inside build file we just have

```
something_todo(
  name = "aeson",
  target = "@aeson//:haskell-outputs",
)
```

Finally the user can now say

```
haskell_library(
  name = "my_lib",
  …,
  deps = [
    "@lts//:aeson"
  ],
)
```

Where `rules_haskell` comes in is to make sure it can deal with these dependencies properly: notably that it can find, use and include all necessary files. This for most part will involve `-package-db` and `-package` flags into outputs of those packages. Possibly `something_todo` above can just output a Provider that directly gives us all this information regardless of backend: we can provide default implementation for nix packages somewhere.

[...]

PR #336, opened 2018-07-14
An option to list packages ignored by `haskell_doc`

### LINKED ISSUE #335: haskell_doc fails if sub-package does not provide haddock file
A call to `haskell_doc` will fail if a dependency package does not provide a `.haddock` file.

See #336 for an example.

Most of the time this should be fixed upstream (i.e. the dependency must have documentation), but it may be a pragmatic solution to allow `haskell_doc` to ignore some packages, such as:

```
haskell_doc( ignore_sub_packages = ["mtl"], ...)
```

After a quick look in the internals of `haskell_doc`, I don't know how to do that, any pointer will be appreciated.

guibou:
Fix #335 by providing `haddock_ignore_prebuilts` attribute to `haskell_toolchain`.

### DIFF
--- a/haskell/haddock.bzl
+++ b/haskell/haddock.bzl
@@ -55,9 +55,12 @@ def _haskell_doc_aspect_impl(target, ctx):
     args.add("--read-interface=../{0},{1}".format(
       pid, transitive_haddocks[pid].path))
 
+  hs = haskell_context(ctx, ctx.rule.attr)
+
   prebuilt_deps = ctx.actions.args()
   for dep in set.to_list(target[HaskellBuildInfo].prebuilt_dependencies):
-    prebuilt_deps.add(dep)
+    if dep not in hs.toolchain.haddock_ignore_prebuilts:
+        prebuilt_deps.add(dep)
   prebuilt_deps.use_param_file(param_file_arg = "%s", use_always = True)
 
   ctx.actions.run(
--- a/haskell/toolchain.bzl
+++ b/haskell/toolchain.bzl
@@ -231,6 +231,8 @@ def _haskell_toolchain_impl(ctx):
       compiler_flags = ctx.attr.compiler_flags,
       locale = ctx.attr.locale,
       locale_archive = locale_archive,
+      # XXX: bazel have no builtin set, so we use a dict here for later O(1) query of `haddock_ignore_prebuilts`
+      haddock_ignore_prebuilts = {i:None for i in ctx.attr.haddock_ignore_prebuilts},
       mode = ctx.var["COMPILATION_MODE"],
       actions = struct(
         compile_binary = compile_binary,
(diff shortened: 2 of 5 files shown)


mboes:
This is very ad hoc information for a toolchain to track. Could you tell us more about,

1. the specific scenario in which you originally encountered #335,
2. the alternatives you considered how they (did not) work out.

guibou:
1. I work on a big monorepo with hundreds of prebuilt dependencies provided by `nixpkgs` and `rules_nixpkgs`. A few of theses dependencies have a broken haddock (for reasons which may depend on the package, `nix`, `bazel`, or `rules_haskell`), so it is impossible to use `haskell_doc`. Everything can be fixed upstream (or with local overrides), but that's a lot more work than this ad hoc solution.

For example, consider a developer, with no knowledge of nix, bazel or haddock, who just add a new dependency to one of his `haskell_library`, this may break the documentation for all the repository and his only solution is to fix the documentation of a package he doesn't own using a tool he may not know about (nix). That's a lot to ask for a developer who may have other constraints in his schedules. This PR provides a perfectible, but quick and easy solution.

2. Another solution may be to let haddock fails gracefully and ignore the broken package:

- inside `haddock_wrapper.sh`, loop over the `haddock` call and progressively removes failing package until it converges to a working solution. That's fragile but may work.
- inside the `haskell_doc` rule, test for the presence of the `package.haddock` file for each sub package.

mboes:
> A few of theses dependencies have a broken haddock (for reasons which may depend on the package, nix, bazel, or rules_haskell).

I understand. I'm asking for specific instances. From the test in the PR, should I infer `mtl` is one such package? And if so, is this because of a nixpkgs packaging issue?

guibou:
> From the test in the PR, should I infer mtl is one such package?

I had issues with `spatial-math` for example. Have a `git grep dontHaddock` in a nixpkgs clone and you'll get a non-exhaustive list of packages with are known to have broken haddock. Most of the issues are due to `nixpkgs` packaging. I had a few coming from my client private repositories too.

> IMO, if nix packages things incorrectly that should be fixed upstream or via Nix overrides, on Nix side, because it's a nix-related and temporary in nature issue. This solution, while I understand it would allow you to make progress, is quite ad-doc. I think if we start to introduce such ad-hoc solutions for every little problem now and then, the API will become a monstrosity...

I agree.

>> inside haddock_wrapper.sh, loop over the haddock call and progressively removes failing package until it converges to a working solution. That's fragile but may work.

> This is doable.

Will you consider this for merging?

This being said, I won't champion this PR for long, I'm not stuck because of this issue (I solved my upstream problems), but I wanted to provide an easy solution for users which may encounter this issue in the future. I agree that it is an ad hoc solution for a rare problem, so we can clone as *won`t fix* and instruct our users to fix their upstream problems.

mboes:
If you don't need this, then let's close for now. The `prebuilt_dependencies` mechanism is a short-term hack anyways, that needs fixing (see #75).

That said, making `haddock_doc` more tolerant of missing haddocks in dependencies, by checking whether the `.haddock` file exists, sounds like a good idea to me.

mboes:
> That said, making haddock_doc more tolerant of missing haddocks in dependencies, by checking whether the .haddock file exists, sounds like a good idea to me.

@guibou could you file that as a ticket and close this PR?

guibou:
@mboes I think that #335 is appropriate.

PR #337, opened 2018-07-16
Add a new rule `haskell_prebuilt_library`.

judah:
This rule forwards a prebuilt dependency, so that it may be
specified directly in `deps` rather than `prebuilt_dependencies`.

This gives progress on #75.  Similarly, it simplifies the use of Hazel,
which can now generate repositories for the prebuilt dependencies
just like for regular packages.

Relevant Hazel change that would depend on this one: https://github.com/FormationAI/hazel/commit/143d8f71bf110c3f39bffb7bbef86832d89c68dc#diff-c40927b6657800717d4b0b498e1bad55

Let me know what you think of this change. I'm open to modifying the API/names.


haskell/private/dependencies.bzl
```diff
@@ -95,6 +96,19 @@ def gather_dep_info(ctx):
...
+    elif HaskellPrebuiltPackageInfo in dep:
+      pkg = dep[HaskellPrebuiltPackageInfo].package
+      acc = HaskellBuildInfo(
+        package_ids = acc.package_ids,
+        package_confs = acc.package_confs,
+        package_caches = acc.package_caches,
+        static_libraries = acc.static_libraries,
+        dynamic_libraries = acc.dynamic_libraries,
+        interface_files = acc.interface_files,
+        prebuilt_dependencies = set.insert(acc.prebuilt_dependencies, pkg),
```
mrkkrp: We could use `set.mutable_insert` here.

haskell/haskell.bzl
```diff
@@ -226,6 +227,42 @@ $ bazel-bin/.../hello-lib-repl  # run the script
...
+def _haskell_prebuilt_library(ctx):
+  return [HaskellPrebuiltPackageInfo(
+            package = ctx.attr.package)]
+
+"""Wrap a prebuilt dependency as a rule.
+
+Example:
+  ```bzl
+  haskell_prebuilt_library(
+      name = "base_pkg",
```
mboes: There seems to be typo in this example, since you name the target `"base_pkg"` by the rule below depends on `":base"`.

haskell/haskell.bzl
```diff
@@ -226,6 +227,42 @@ $ bazel-bin/.../hello-lib-repl  # run the script
...
+supplied by Bazel.  Depending on the wrapped rule eliminates the need to list
+the package name in prebuilt_dependencies.
+
+Often, rules of this type will be generated automatically by frameworks such
+as Hazel.
+"""
+haskell_prebuilt_library = rule(
+  _haskell_prebuilt_library,
+  attrs = dict(
+    package = attr.string(doc = "A non-Bazel-supplied GHC package name."),
```
mboes: It would be good for the `package` to have `name` as the default value. In fact, could `name` alone be sufficient? The list of allowed characters in a package is a subset of the legal target names. Although people might want to import `ghc-prim` as `ghc_prim`.

judah: Good suggestion; I've set it default to name.  (I agree that `ghc-prim` vs `ghc_prim` means we should keep the separate attribute as an option.)

haskell/haskell.bzl
judah: Done.

judah:
Renamed to `haskell_import_library`; PTAL.

mboes:
@judah FYI I know I proposed `haskell_import_library` earlier, but that's because I didn't remember that in the CC rules this is actually called `cc_import` (no `_library). So just pushed a commit to have `haskell_import` here too. Holler if you disagree and we can revert.

PR #342, opened 2018-07-17
Improve doctest support

mrkkrp:
This PR redesigns the `haskell_doctest` rule to make it capable to do everything normal doctest test suites do. In particular now it's possible to:

* Specify which modules to test (`doctest` allows that, and in fact if we always go for all modules it's not going to work because some modules may contain things that doctest doesn't like). Specifying module names is the sanest approach here because filenames of modules may change if pre-processing is involved, etc. If the `modules` attribute is not specified we default to testing all modules.
* Only direct dependencies are checked, and the rule has nothing to do with aspects now.
* cbits and hopefully all other cases are supported, approach similar to what we are doing with Haddocks is used: record important flags, then pass them to the tool that invokes GHC(i).
* It's now possible to pass arbitrary flags to `doctest` executable, because that's what one can do without Bazel.


judah:
I was able to reproduce the darwin build failure on my laptop, and fix it by prefixing "-optl" to the arguments:

```
diff --git a/haskell/doctest.bzl b/haskell/doctest.bzl
index ee491b2..7213ae8 100644
--- a/haskell/doctest.bzl
+++ b/haskell/doctest.bzl
@@ -112,9 +112,9 @@ def _haskell_doctest_single(target, ctx):
     if not set.is_member(seen_libs, lib_name):
       set.mutable_insert(seen_libs, lib_name)
       args.add([
-        "-l{0}".format(lib_name),
+        "-optl-l{0}".format(lib_name),
         # For some reason plain -L doesn't seem to work.
-        "-L{0}".format(paths.dirname(lib.path)),
+        "-optlL{0}".format(paths.dirname(lib.path)),
       ])
```
Not sure whether that also works on Linux or if we need to special-case it.

judah:
Actually, that probably should be `-optl-L{0}`, not `-optlL{0}`.  Interestingly, though, if I deleted that line altogether it still worked.

mrkkrp:
Doesn't seem to work without `-L`. Will try `-optl-L` now.

judah:
Quick clarification: the `-opt-l` (lower-case "l") change in the diff I posted *did* seem to be required in order to get the build working on Darwin.

haskell/doctest.bzl
```diff
@@ -0,0 +1,213 @@
...
+  attrs = {
+    "deps": attr.label_list(
+      doc = "List of Haskell targets to lint.",
+    ),
+    "doctest_flags": attr.string_list(
+      doc = "Extra flags to pass to doctest executable.",
+    ),
+    "modules": attr.string_list(
+      doc = """List of names of modules that will be tested. If the list is
+omitted, all modules provided by `deps` will be tested.
```
judah: Clarify that this is all *exposed* modules provided by deps? (Assuming I'm reading the code correctly above)

haskell/doctest.bzl
```diff
@@ -0,0 +1,213 @@
...
+  toolchains = [
+      "@io_tweag_rules_haskell//haskell:toolchain",
+      "@io_tweag_rules_haskell//haskell:doctest-toolchain",
+  ],
+)
+"""Run doctest test on targets in `deps`.
+
+Note that your toolchain must be equipped with `doctest` executable, i.e.
+you should specify location of the executable using the `doctest` attribute
+when you invoke `haskell_toolchain`.
```
judah: It's now configured by `doctest_toolchain` rather than `haskell_toolchain`.  (Sorry, looks like I forgot to update this doc when I made that change earlier.)

haskell/private/actions/compile.bzl
```diff
@@ -210,7 +210,7 @@ def _compilation_defaults(hs, cc, java, dep_info, srcs, extra_srcs, cpp_defines,
     DefaultCompileInfo: Populated default compilation settings.
   """
   args = hs.actions.args()
-  haddock_args = hs.actions.args()
+  ghc_args = []
```
judah: Here's a refactoring suggestion; up to you whether it's worth it: consider moving l.272-282 below.  Nearly all of args and ghc_args seem duplicated; it seems like we should be able to structure the code like:

```
ghc_args = []
ghc_args += ["-hide-all-packages"]
ghc_args += ...
...

args = hs.actions.args()
args.add(ghc_args)
# then add -hidir, -odir, etc to args
```

In theory it might even be possible to extract the computation of ghc_args into a separate function.  Again, up to you whether it's worth it at this point.

Issue #340, opened 2018-07-16
nix: test-binary-simple does not build, ghc-with-packages fails


Profpatsch:
I set `NIX_PATH` to the empty string, so the provided `nixpkgs_git_repository` should be used if I read the code correctly. I’m not sure where the magic `repository = "@nixpkgs"` is resolved, though. 

```
$ env NIX_PATH= bazel build //tests:test-binary-simple --verbose_failures
these derivations will be built:
  /nix/store/lh944ba5x97w83lf6w1c64hm5rhgchhd-proto-lens-0.3.1.0.drv
  /nix/store/0n99dz505jpzk3i7z4cirkni5nvbbq9w-ghc-8.2.2-with-packages.drv
building '/nix/store/lh944ba5x97w83lf6w1c64hm5rhgchhd-proto-lens-0.3.1.0.drv'...
setupCompilerEnvironmentPhase
Build with /nix/store/i2d6l6w99qlwhz6gbz4hxl4d9238drfj-ghc-8.2.2.
ignoring (possibly broken) abi-depends field for packages
ignoring (possibly broken) abi-depends field for packages
unpacking sources
unpacking source archive /nix/store/85w6agkmzy6cy606h2bs42ry7azczn2m-proto-lens-0.3.1.0.tar.gz
source root is proto-lens-0.3.1.0
setting SOURCE_DATE_EPOCH to timestamp 1525626655 of file proto-lens-0.3.1.0/src/Data/ProtoLens/TextFormat.hs
patching sources
compileBuildDriverPhase
setupCompileFlags: -package-db=/build/setup-package.conf.d -j1 -threaded
[1 of 1] Compiling Main             ( Setup.hs, /build/Main.o )
Linking Setup ...
configuring
[...]

PR #343, opened 2018-07-17
Create only one dummy archive for ghc linking

### LINKED ISSUE #278: Many dummy archives created when only one is sufficient
When linking a binary, we use a hack stolen from Buck, which consists in providing GHC with at least one static archive, it's just that it's empty. The thing is, we create as many empty archives as we have targets to link, when we could do with only one (empty archives always have the same content...).

```
$ bazel clean
$ bazel build //tests/... 2>&1 | grep HaskellDummyObjectGhc | wc -l
24
```

Profpatsch:
Before, each run of the Haskell rule would create a static dummy library, which
the ghc linker needs. Factor out into a default target.

Have only tested the analyze phase locally, the build [fails with some strange nix-based error](https://github.com/tweag/rules_haskell/issues/340).

Fixes #278

### DIFF
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -117,7 +117,14 @@ def _mk_binary_rule(**kwargs):
       main_file = attr.label(
         allow_single_file = FileType([".hs", ".hsc", ".lhs"]),
         doc = "File containing `Main` module.",
-      )
+      ),
+      _dummy_static_lib = attr.label(
+        default = Label("@io_tweag_rules_haskell//haskell:dummy_static_lib"),
+        allow_single_file = True,
+        doc = """
+A dummy library needed for the GHC linking process.
+""",
+      ),
     ),
     outputs = {
       "repl": "%{name}-repl",
--- a/haskell/private/create_dummy_archive.bzl
+++ b/haskell/private/create_dummy_archive.bzl
@@ -1,0 +1,58 @@
+load("@bazel_skylib//:lib.bzl", "paths")
+load(":private/context.bzl", "haskell_context")
+
+def _impl(ctx):
+  """Create empty archive so that GHC has some input files to work on during
+  linking.
+
+  See: https://github.com/facebook/buck/blob/126d576d5c07ce382e447533b57794ae1a358cc2/src/com/facebook/buck/haskell/HaskellDescriptionUtils.java#L295
+
+  Returns:
+    File, the created dummy archive.
+  """
+
+  # TODO: this is a duplication from ./haskell_impl.bzl
+  hs = haskell_context(ctx)
+  cc_toolchain = ctx.toolchains["@bazel_tools//tools/cpp:toolchain_type"]
+
+  dummy_raw = "BazelDummy.hs"
+  dummy_input = hs.actions.declare_file(dummy_raw)
+  dummy_object = hs.actions.declare_file(paths.replace_extension(dummy_raw, ".o"))
+
+  hs.actions.write(output=dummy_input, content="""
+{-# LANGUAGE NoImplicitPrelude #-}
+module BazelDummy () where
+""")
+
+  hs.toolchain.actions.run_ghc(
+    hs,
+    inputs = [dummy_input],
+    outputs = [dummy_object],
+    mnemonic = "HaskellDummyObjectGhc",
+    arguments = [
+      "-optc" + f for f in cc_toolchain.compiler_options()
+    ] + ["-c", dummy_input.path],
+  )
+
+  ar_args = hs.actions.args()
+  ar_args.add(["qc", ctx.outputs.archive, dummy_object])
+
+  hs.actions.run(
+    inputs = [dummy_object] + hs.tools_runfiles.ar,
+    outputs = [ctx.outputs.archive],
+    mnemonic = "HaskellDummyObjectAr",
+    executable = hs.tools.ar,
+    arguments = [ar_args]
+  )
+
+
+create_dummy_archive = rule(
+  implementation = _impl,
+  toolchains = [
+    "@io_tweag_rules_haskell//haskell:toolchain",
+    "@bazel_tools//tools/cpp:toolchain_type",
+  ],
+  outputs = {
+    "archive": "libempty.a"
+  }
+)
--- a/haskell/BUILD
+++ b/haskell/BUILD
@@ -1,7 +1,7 @@
 load("//skylark:lint.bzl", "skylark_lint")
-load(":private/create_dummy_archive.bzl", "create_dummy_archive")
+load(":private/dummy_linker_archive.bzl", "dummy_linker_archive")
 
-create_dummy_archive(
+dummy_linker_archive(
   name = "dummy_static_lib",
   visibility = ["//visibility:public"]
 )
--- a/haskell/BUILD
+++ b/haskell/BUILD
@@ -1,4 +1,10 @@
 load("//skylark:lint.bzl", "skylark_lint")
+load(":private/create_dummy_archive.bzl", "create_dummy_archive")
+
+create_dummy_archive(
+  name = "dummy_static_lib",
+  visibility = ["//visibility:public"]
+)
 
 exports_files(
   glob(["*.bzl"]) + [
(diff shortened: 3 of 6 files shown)


haskell/toolchain.bzl
```diff
@@ -287,7 +287,13 @@ _haskell_toolchain = rule(
       doc = """
 Label pointing to the locale archive file to use. Mostly useful on NixOS.
 """,
-    )
+    ),
+    "_dummy_static_lib": attr.label(
```
mrkkrp: I think you can't put this here because to generate a dummy archive you need a toolchain and a toolchain needs a dummy archive now. This creates a dependency cycle which leads to the failure you can see on CI.

Profpatsch: Yes, I also got this:

```
$ bazel query 'deps(//tests:test-binary-with-lib)'
//tests:test-binary-with-lib
@bazel_tools//tools/test:test_setup
@bazel_tools//tools/test:runtime
@bazel_tools//tools/test:test-setup.sh
@bazel_tools//tools/test:collect_coverage
//tools/defaults:coverage_support
@bazel_tools//tools/test:coverage_support
@bazel_tools//tools/test:collect_coverage.sh
//tools/defaults:coverage_report_generator
@bazel_tools//tools/test:coverage_report_generator
@bazel_tools//tools/coverage:coverage_report_generator
@bazel_tools//tools/coverage:dummy_coverage_report_generator
//tests/binary-with-lib:binary-with-lib
//tests/binary-with-lib:lib
@io_tweag_rules_haskell//haskell:private/ghci_repl_wrapper.sh
@io_tweag_rules_haskell//haskell:assets/ghci_script
@io_tweag_rules_haskell//haskell:assets/cpp_defines
//tests/binary-with-lib:src/Lib.hs
//tests/binary-with-lib:Main.hs
```

Which doesn’t include the target as I would expect.

mrkkrp: So to make it work you'll need to remove this from the toolchain and perhaps reference that dummy archive from `compile_binary` like so: `@io_tweag_rules_haskell//haskell:dummy_static_lib`. That may work.

Profpatsch: Ah, I think it have to add it to `_mk_binary_rule` instead of the toolchain.

Profpatsch:
This looks better:

```
$ bazel query 'deps(//tests:test-binary-with-lib)'
//tests:test-binary-with-lib
@bazel_tools//tools/test:test_setup
@bazel_tools//tools/test:runtime
@bazel_tools//tools/test:test-setup.sh
@bazel_tools//tools/test:collect_coverage
//tools/defaults:coverage_support
@bazel_tools//tools/test:coverage_support
@bazel_tools//tools/test:collect_coverage.sh
//tools/defaults:coverage_report_generator
@bazel_tools//tools/test:coverage_report_generator
@bazel_tools//tools/coverage:coverage_report_generator
@bazel_tools//tools/coverage:dummy_coverage_report_generator
//tests/binary-with-lib:binary-with-lib
@io_tweag_rules_haskell//haskell:dummy_static_lib
//tests/binary-with-lib:lib
@io_tweag_rules_haskell//haskell:private/ghci_repl_wrapper.sh
@io_tweag_rules_haskell//haskell:assets/ghci_script
@io_tweag_rules_haskell//haskell:assets/cpp_defines
//tests/binary-with-lib:src/Lib.hs
//tests/binary-with-lib:Main.hs
```

haskell/private/create_dummy_archive.bzl
```diff
@@ -0,0 +1,60 @@
...
+  linking.
+
+  See: https://github.com/facebook/buck/blob/126d576d5c07ce382e447533b57794ae1a358cc2/src/com/facebook/buck/haskell/HaskellDescriptionUtils.java#L295
+
+  Returns:
+    File, the created dummy archive.
+  """
+
+  # This is a bit of a hack to make `cc_interop_info` accept this rule’s `ctx`
+  ctx.attr.deps = []
```
Profpatsch: I thought I could just stub out the attribute without creating an extra field in `attrs`, but that is forbidden:

```
ERROR: /Users/distiller/project/haskell/private/create_dummy_archive.bzl:16:3: cannot assign to 'ctx.attr.deps'
```

haskell/private/create_dummy_archive.bzl
```diff
@@ -0,0 +1,63 @@
...
+    arguments = [ar_args]
+  )
+
+
+create_dummy_archive = rule(
+  implementation = _impl,
+  attrs = {
+    # This is a bit of a hack to make `cc_interop_info` accept this rule’s `ctx`
+    "deps": attr.label_list(),
+  },
```
Profpatsch: Is this okay?

haskell/private/create_dummy_archive.bzl
```diff
@@ -0,0 +1,63 @@
...
+
+  See: https://github.com/facebook/buck/blob/126d576d5c07ce382e447533b57794ae1a358cc2/src/com/facebook/buck/haskell/HaskellDescriptionUtils.java#L295
+
+  Returns:
+    File, the created dummy archive.
+  """
+
+  # TODO: this is a duplication from ./haskell_impl.bzl
+  hs = haskell_context(ctx)
+  cc = cc_interop_info(ctx)
```
Profpatsch: Also this, but I think that can’t be circumvented in this case.

haskell/private/create_dummy_archive.bzl
```diff
@@ -0,0 +1,63 @@
...
+    # This is a bit of a hack to make `cc_interop_info` accept this rule’s `ctx`
+    "deps": attr.label_list(),
+  },
+  toolchains = [
+    "@io_tweag_rules_haskell//haskell:toolchain",
+    "@bazel_tools//tools/cpp:toolchain_type",
+  ],
+  outputs = {
+    "archive": "libempty.a"
+  }
```
Profpatsch: I’m not sure whether I need the output, but it feels more explicit this way.

haskell/private/create_dummy_archive.bzl
mrkkrp: I don't see where this `cc` thing is used at all, drop it?

haskell/private/create_dummy_archive.bzl
mrkkrp: If we drop `cc = cc_interop_info(ctx)` this should go away too.

haskell/private/create_dummy_archive.bzl
mrkkrp: We're generating a dummy module and then compile it, why should we be concerned with cc interop in this case?

haskell/toolchain.bzl
```diff
@@ -287,7 +287,7 @@ _haskell_toolchain = rule(
       doc = """
 Label pointing to the locale archive file to use. Mostly useful on NixOS.
 """,
-    )
+    ),
```
mrkkrp: Avoid such changes?

mrkkrp:
Please rebase against current master.

haskell/private/create_dummy_archive.bzl
Profpatsch:
```
  hs.toolchain.actions.run_ghc(
    hs,
    inputs = [dummy_input],
    outputs = [dummy_object],
    mnemonic = "HaskellDummyObjectGhc",
    arguments = [
      "-optc" + f for f in cc.compiler_flags
    ] + ["-c", dummy_input.path],
  )
```

is the part where it’s used. Didn’t look like something I can just remove.

haskell/haskell.bzl
```diff
@@ -126,7 +126,14 @@ def _mk_binary_rule(**kwargs):
       main_file = attr.label(
         allow_single_file = FileType([".hs", ".hsc", ".lhs"]),
         doc = "File containing `Main` module.",
-      )
+      ),
+      _dummy_static_lib = attr.label(
+        default = Label("//haskell:dummy_static_lib"),
```
mboes: I _think_, but I'm not sure, that this needs to be a fully qualified label, in the style of https://github.com/tweag/rules_haskell/pull/343/files#diff-f479d5c4fb32903b900b580306254f64R89, to work correctly in downstream repos that import rules_haskell as an external repo.

Profpatsch: Ah, because they might see it as `//` of their project. I think @guibou mentioned something along those lines.

mboes:
@mrkkrp PTAL

Profpatsch:
Done.

PR #350, opened 2018-07-19
Expose extra_srcs during linking.

facundominguez:
Enables using -optl-Wl,--dynamic-list=file


facundominguez:
The test doesn't work in darwin because --dynamic-list is not implemented there, it seems.
I'll submit the tests after I figure out how to run them in linux only.

PR #352, opened 2018-07-20
Make REPLs not to depend on their targets

### LINKED ISSUE #289: REPL targets depend on their corresponding target library/executable
rules_haskell generates REPL targets for Haskell libraries and executables. However, these depend on the library/executable they are for, and will fail to build if that library/executable is in a state where it doesn't build.

Say I have a library `mylib` and want to work on it using the corresponding REPL target. But, I'm in the middle of developing a feature, so somewhere in `mylib` is some type error, or maybe a typed hole. If I issue `bazel run //mylib:mylib-repl` bazel will first try to build `mylib`. That will fail and I cannot open the REPL. I first have to undo any unfinished work and bring `mylib` back to a state where it builds. Then I can build and open the REPL, and then I can bring back my unfinished work and `:reload` in the REPL to continue working. It would be nice if the REPL would build and start straight-away.

mrkkrp:
Close #289.

There seems to be no way to tell GHCi "keep going and load as many modules as possible even if some fail", but now we can load REPL with *some* modules in it, and the rest can be done from the running REPL. As good as you can get, apparently, stack works the same way.

----

Also remove the `repl_interpreted` attribute which was an attribute I introduced on my own and which I'm sure none is using. Further, the bug that we're trying to fix is unavoidable when `repl_interpreted = False`, hence it makes sense to always assume it to be `True` satisfying the expectations that users have for Haskell REPLs (as established by other tooling such as Cabal and Stack).

### DIFF
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -68,17 +68,6 @@ _haskell_common_attrs = {
     "prebuilt_dependencies": attr.string_list(
         doc = "Non-Bazel supplied Cabal dependencies.",
     ),
-    "repl_interpreted": attr.bool(
-        default = True,
-        doc = """
-Whether source files should be interpreted rather than compiled. This allows
-for e.g. reloading of sources on editing, but in this case we don't handle
-boot files and hsc processing.
-
-For `haskell_binary` targets, `repl_interpreted` must be set to `True` for
-REPL to work.
-""",
-    ),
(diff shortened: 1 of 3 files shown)


mboes:
What feature should we ask for upstream? You should file a ticket if none exists already.

mrkkrp:
I filed this one: https://ghc.haskell.org/trac/ghc/ticket/15424#ticket.

PR #355, opened 2018-07-21
Add deprecation notes for the ‘prebuilt_dependencies’ attribute

mrkkrp:


mboes:
Would be good to adjust tests as well. Or the test suite will be generating a ton of warnings.

mboes:
Should also file a feature request upstream for attribute-level deprecation warnings.

mrkkrp:
Opened an issue with a feature request: https://github.com/bazelbuild/bazel/issues/5654.

PR #356, opened 2018-07-21
Move back to nixpkgs checkout

mboes:
We no longer need to pull bazel from cachix, now that
https://github.com/NixOS/nixpkgs/pull/42735 has been merged and has
hit the cache.nixos.org cache.


tests/ghc.nix
```diff
@@ -1,11 +1,11 @@
-{ pkgs ? import <nixpkgs> {} }:
+{ pkgs ? import ../nixpkgs.nix {} }:
 
 with pkgs;
 
 let haskellPackages = pkgs.haskell.packages.ghc822.override {
       overrides = with pkgs.haskell.lib; self: super: rec {
-        lens-labels = super.lens-labels_0_2_0_1;
```
mrkkrp: Why not to delete the whole override if we don't actually change anything with it?

tests/protoc_gen_haskell.nix
```diff
@@ -4,8 +4,8 @@ with pkgs;
 
 let haskellPackages = pkgs.haskell.packages.ghc822.override {
       overrides = with pkgs.haskell.lib; self: super: rec {
-        lens-labels = super.lens-labels_0_2_0_1;
-        proto-lens = super.proto-lens_0_3_1_0;
+        lens-labels = super.lens-labels;
```
mrkkrp: The same here.

tests/ghc.nix
mboes: Good point. Wasn't thinking!

PR #362, opened 2018-07-23
Ignore missing .haddock files when generating documentation

### LINKED ISSUE #335: haskell_doc fails if sub-package does not provide haddock file

Profpatsch:
Fixes #335.

For now only a test reproducing the issue:

```
$ bazel test //tests/haddock
INFO: Analysed target //tests/haddock:haddock (0 packages loaded).
INFO: Found 1 target and 0 test targets...
ERROR: /home/philip/code/bazel/rules_haskell/tests/haddock/BUILD:46:1: HaskellHaddock //tests/haddock:haddock-lib-b failed (Exit 1)
haddock: internal error: /nix/store/9ab4cswajxzxgy8s6khkr99sdp96xm9k-libc-0.1.0.0/share/doc/x86_64-linux-ghc-8.2.2/libc-0.1.0.0/html/libc.haddock: openBinaryFile: does not exist (No such file or directory)
Target //tests/haddock:haddock failed to build
```

### DIFF
--- a/tests/ghc.nix
+++ b/tests/ghc.nix
@@ -2,7 +2,15 @@
 
 with pkgs;
 
-haskell.packages.ghc822.ghcWithPackages (p: with p; [
+let haskellPackages = pkgs.haskell.packages.ghc822.override {
+      overrides = with pkgs.haskell.lib; self: super: rec {
+        libc = import ./haddock/libC.nix self pkgs;
+      };
+    };
+
+in haskellPackages.ghcWithPackages (p: with p; [
+
+  # haskell_proto_library inputs
   bytestring
   containers
   data-default-class
@@ -10,4 +18,8 @@ haskell.packages.ghc822.ghcWithPackages (p: with p; [
   lens-labels
   proto-lens
   text
-  ])
+
+  # test inputs
+  libc
+
+])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
+## [0.7] - ???
+
+* Don’t crash on inputs missing `.haddock` interface files. See
+  [#362](https://github.com/tweag/rules_haskell/pull/362)
+
 ## [0.6] - 2018-07-21
 
 ### Added
--- a/haskell/private/haddock_wrapper.sh
+++ b/haskell/private/haddock_wrapper.sh
@@ -11,9 +11,29 @@ extra_args=()
 
 for pkg in $(< $PREBUILT_DEPS_FILE)
 do
+    # Assumption: the `haddock-interfaces` field always only contains
+    # exactly one file name. This seems to hold in practice, though the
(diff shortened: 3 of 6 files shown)


tests/ghc.nix
```diff
@@ -6,6 +6,7 @@ let haskellPackages = pkgs.haskell.packages.ghc822.override {
       overrides = with pkgs.haskell.lib; self: super: rec {
         lens-labels = super.lens-labels_0_2_0_1;
         proto-lens = super.proto-lens_0_3_1_0;
+        libc = import ./haddock/libC.nix self pkgs;
```
Profpatsch: This needs to be solved differently; my suggestion is creating a different `ghcWithPackages` for the tests.

Profpatsch:
yay

screenshot

Profpatsch:
This is ready for review, but it still needs a solution for how to handle the `ghc.nix` issue I commented on above.

mboes:
@guibou do you have a good solution to @Profpatsch's question above?

Profpatsch:
I’d use a superset of packages for the tests, where we can add more stuff when we need it.

guibou:
Well done. I'm OK with your "hack" in the `ghc.nix`, that's part of the testing environment.

However I'll had more documentation inside `ghc.nix` and I'll do something simpler. You can have a look at the original PR I did for this issue (with a test similar to yours), where I only put the `mtl` packages available with `dontHaddock` flag. 

https://github.com/tweag/rules_haskell/pull/336/commits/df74df518670c630233f5b77cc865a04958fc2e7#diff-8c75769c4d497fd08fde7536e40fa727

Profpatsch:
> I'm OK with your "hack" in the ghc.nix, that's part of the testing environment.

Yes, after studying the source code a bit more it is now clear to me that the `ghc.nix` file is only used locally, for the protobuf library and the tests.

> and I'll do something simpler. You can have a look at the original PR I did for this issue (with a test similar to yours), where I only put the mtl packages available with dontHaddock flag.

We can switch to an existing `haskellPackages` library for the example, but it won’t be as self-contained and reusable for other tests. Also it might give a compilation overhead (`libC.nix` is basically the most minimal cabal package input one can give to `haskellPackages.callPackage`).

Profpatsch:
Added two small comments to `ghc.nix`.

guibou:
LGTM. You convinced me with the "small" example (which is, in fact, bigger than a oneliner pulling a package from hackage ;)

Profpatsch:
@mboes If you have no further corrections, we can merge.

Profpatsch:
Added a changelog entry, being a good citizen. :)

PR #364, opened 2018-07-23
Fix c2hs_library in external repositories.

judah:
Previously the rule didn't work when `workspace_root` was nontrivial.


mboes:
Weird... CircleCI is running in the FormationAI org, using their settings (and failing).

judah:
Strange...I think I managed to get CircleCI to behave by rebasing the branch onto latest master (maybe just by poking it).

PR #373, opened 2018-07-27
Allow rules to override -O2 with their own compiler_flags.

judah:
Currently, `-c opt` will turn on optimization for all Haskell
rules with no way to disable it.  Previously, it was possible
to do so with `compiler_flags = ["-O0"]`, but that behavior was
changed in #342:
https://github.com/tweag/rules_haskell/pull/342/files#diff-33a8c46955e2a9917090b5258346aee2L247

Some Haskell packages take a *lot* of memory and a lot longer to compile with
optimizations.  So it's useful for us to be able to control this tradeoff.

### DIFF
--- a/haskell/private/actions/compile.bzl
+++ b/haskell/private/actions/compile.bzl
@@ -133,6 +133,11 @@ def _compilation_defaults(hs, cc, java, dep_info, srcs, import_dir_map, extra_sr
     )
 
     # Default compiler flags.
+
+    # Compilation mode.  Allow rule-supplied compiler flags to override it.
+    if hs.mode == "opt":
+      ghc_args += ["-O2"]
+
     ghc_args += hs.toolchain.compiler_flags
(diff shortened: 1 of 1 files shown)


haskell/private/actions/compile.bzl
```diff
@@ -133,6 +133,11 @@ def _compilation_defaults(hs, cc, java, dep_info, srcs, import_dir_map, extra_sr
     )
 
     # Default compiler flags.
+
+    # Compilation mode.  Allow rule-supplied compiler flags to override it.
+    if hs.mode == "opt":
```
mrkkrp: This way `-O2` will be passed to `haddock` and `doctest`, I guess it'll be ignored and there is nothing harmful in that though?

judah: Good point!  I've reorganized the change so it's not passed to them; let me know what you think.

PR #374, opened 2018-07-27
Override TMPDIR in haddock_wrapper.sh

### LINKED ISSUE #229: Haddock fails on parallel executions
If `haskell_doc` is called with more than one job, the related instances of Haddock will concurrently try to create the same temporary directory, and fail the job.

By now `-j 1` has to be passed to `bazel build`. It should be handled by default in `haskell_doc` (until `haddock` evolution on parallel executions).

Note: We noticed this behavior with bazel version 0.12.0, but not with 0.11.0.

thumphries:
This fixes #229 by overriding `TMPDIR` on each Haddock execution.

### DIFF
--- a/haskell/private/haddock_wrapper.sh
+++ b/haskell/private/haddock_wrapper.sh
@@ -36,4 +36,8 @@ do
     fi
 done
 
+# Override TMPDIR to prevent race conditions on certain platforms.
+mkdir haddock-tmp
+TMPDIR=haddock-tmp
 haddock "${extra_args[@]}" "$@"
+rmdir haddock-tmp
--- a/haskell/private/haddock_wrapper.sh
+++ b/haskell/private/haddock_wrapper.sh
@@ -37,7 +37,9 @@ do
 done
 
 # Override TMPDIR to prevent race conditions on certain platforms.
-mkdir haddock-tmp
-TMPDIR=haddock-tmp
-haddock "${extra_args[@]}" "$@"
-rmdir haddock-tmp
+# BSD and GNU mktemp are very different; attempt GNU first
+TEMP=$(mktemp -d 2>/dev/null || mktemp -d -t 'haddock_wrapper')
+trap cleanup 1 2 3 6
+cleanup() { rmdir "$TEMP"; }
+TMPDIR=$TEMP haddock "${extra_args[@]}" "$@"
+rmdir "$TEMP"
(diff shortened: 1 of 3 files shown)


haskell/private/haddock_wrapper.sh
```diff
@@ -36,4 +36,8 @@ do
     fi
 done
 
+# Override TMPDIR to prevent race conditions on certain platforms.
+mkdir haddock-tmp
+TMPDIR=haddock-tmp
```
mrkkrp: I do not understand the solution and you do not explain it in the description. You set `TMPDIR` to a constant name, `haddock-tmp`, how does it help to avoid name clashes?

Another question: what happens when `haddock-tmp` already exists?

And finally, is this code exception-safe? What if while Bazel is working I hit control-C and shut it down. I don't know exactly, but perhaps the script will receive a signal of some sort. So that signal could arrive after `mkdir` but before `rmdir` and so the directory won't be cleaned up.

I suggest we use `mktemp` which can randomize names by itself and `trap` for signal handling perhaps? What do you think?

thumphries: AIUI this script is always run in a unique directory per-target. It seemed to be different on each bazel build, so I assumed it to be a clean temporary directory. I don't really know bazel, so maybe that is not quite correct, but this fixed the bug without dealing with cross-platform `mktemp` invocation

`mktemp` is fine by me if you prefer it. I'll amend the PR soon.

haskell/private/haddock_wrapper.sh
```diff
@@ -36,4 +36,10 @@ do
     fi
 done
 
-haddock "${extra_args[@]}" "$@"
+# Override TMPDIR to prevent race conditions on certain platforms.
+# BSD and GNU mktemp are very different; attempt GNU first
+TEMP=$(mktemp -d 2>/dev/null || mktemp -d -t 'haddock_wrapper')
+trap cleanup 1 2 3 6
+cleanup() { rmdir "$TEMP"; }
+TMPDIR=$TEMP haddock "${extra_args[@]}" "$@"
```
Profpatsch: There should be a link to an upstream haddock issue here.

Profpatsch: Sorry, this should have been a comment, not a blocker. Thanks to @judah for catching it and merging!

PR #376, opened 2018-07-28
Fix missing as

### LINKED ISSUE #259: gcc: error trying to exec 'cc1': execvp: No such file or directory

mrkkrp:
Close #259.

@iphydf Here is a solution to the remaining problems described in #259. See descriptions of the problems in the readme and the workaronds. I did not include the fixes in the code because I felt that we should not hardwrie these workarounds into `rules_haskell`, but @m may have a different opinion, let me know what you think too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,64 @@ published on Hackage, or part of Stackage snapshots, using Bazel.
 
 [hazel]: https://github.com/formationai/hazel
 
-## Troubleshooting `rules_haskell` development
+## Troubleshooting
+
+### No such file or directory
+
+If you see error messages complaining about missing `as` (`ld` or indeed
+some other executable):
+
+```
+cc: error trying to exec 'as': execvp: No such file or directory
+`cc' failed in phase `Assembler'. (Exit code: 1)
+```
+
+It means that your `gcc` cannot find `as` by itself. This happens only on
+certain operating systems which have `gcc` compiled without `--with-as` and
+`--with-ld` flags. We need to make `as` visible manually in that case:
+
+```bzl
+# Create a symlink to system executable 'as'
+genrule(
+    name = "toolchain_as",
+    outs = ["as"],
+    cmd = "ln -s /usr/bin/as $@",
+)
+
+# Make it visible to rules_haskell rules:
+haskell_toolchain(
+    name = "ghc",
+    tools = "@ghc//:bin",
+    version = "8.4.1",
+    extra_binaries = [":toolchain_as"], # <----
+)
+```
+
+### `__STDC_VERSION__` does not advertise C99 or later
+
+If you see an error message like this:
+
+```
+/root/.cache/bazel/_bazel_root/b8b1b1d6144a88c698a010767d2217af/external/ghc/lib/ghc-8.4.1/include/Stg.h:29:3: error:
+     error: #error __STDC_VERSION__ does not advertise C99 or later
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -200,6 +200,60 @@ published on Hackage, or part of Stackage snapshots, using Bazel.
...
+  ```
+
+  It means that your `gcc` cannot find `as` by itself. This happens only on
+  certain operating systems such as Ubuntu. We need to make `as` visible
+  manually in that case:
+
+  ```bzl
+  # Create a symlink to system executable 'as'
+  genrule(
+      name = "as",
```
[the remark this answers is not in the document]
iphydf: It fixes the problem. I have other problems now, like linking errors, but this seems to work. I'll provide repros for my issues once this is merged.

README.md
```diff
@@ -200,6 +200,60 @@ published on Hackage, or part of Stackage snapshots, using Bazel.
...
+
+* If you see error messages complaining about missing `as` executable:
+
+  ```
+  cc: error trying to exec 'as': execvp: No such file or directory
+  `cc' failed in phase `Assembler'. (Exit code: 1)
+  ```
+
+  It means that your `gcc` cannot find `as` by itself. This happens only on
+  certain operating systems such as Ubuntu. We need to make `as` visible
```
mboes: We should be more precise. Recent ubuntu's work fine. If anything it's Fedora that's more of a problem. In any case, what she should say is, this happens on certain platforms where GCC was built without the `--with-as` and `--with-ld` configure flags.

README.md
```diff
@@ -200,6 +200,60 @@ published on Hackage, or part of Stackage snapshots, using Bazel.
 
 [hazel]: https://github.com/formationai/hazel
 
+## Troubleshooting
```
mboes: We should merge the two troubleshooting sections. No need to have two separate ones. Use the same format as the troubleshooting section below this one.

mrkkrp:
@m PTAL.

mrkkrp:
@mboes PTAL.

README.md
```diff
@@ -200,7 +200,63 @@ published on Hackage, or part of Stackage snapshots, using Bazel.
...
+```
+
+It means that your `gcc` cannot find `as` by itself. This happens only on
+certain operating systems which have `gcc` compiled without `--with-as` and
+`--with-ld` flags. We need to make `as` visible manually in that case:
+
+```bzl
+# Create a symlink to system executable 'as'
+genrule(
+    name = "as",
```
iphydf: I've named this "toolchain_as". I'd recommend putting warning-free code in the example.

README.md
```diff
@@ -200,7 +200,63 @@ published on Hackage, or part of Stackage snapshots, using Bazel.
 
 [hazel]: https://github.com/formationai/hazel
 
-## Troubleshooting `rules_haskell` development
+## Troubleshooting
+
+### `as`: no such file or directory
+
+If you see error messages complaining about missing `as` executable:
```
iphydf: Or `ld', if you like, or anything really (I had to put `touch` in there as well for `language-c` to compile - I don't know why).

mrkkrp:
@iphydf Thanks for your feedback, I've amended the commit.

Issue #357, opened 2018-07-22
Support reexporting modules


mboes:
GHC supports reexporting modules from upstream dependencies, since 7.10.3. See https://www.haskell.org/cabal/users-guide/developing-packages.html#pkg-field-library-reexported-modules. This will be useful for Backpack support. We could implement this with an API like the following:

```python
haskell_library(
  name = "foo",
  deps = [":bar", ":baz"],
  exports = {":bar": "Data.Bar1 as Data.Bar1_2, Data.Bar2 as Data.Bar2_2"},
)
```

I propose the `exports` attribute name to be consistent with the Java rules. The syntax inside the strings is the same as Cabal and GHC.

lunaris:
I've started hacking at this (PR shortly hopefully). Basics are there but current set of compromises/challenges is:

* The best we have for dictionaries is `attr.label_keyed_string_dict`; I think we actually want the nonexistent `attr.label_keyed_string_list_dict` so that we can map from dependency to list of reexports. However I can work around this for now with string joining/splitting and helpers and am of course open to better suggestions on how to tackle this. (For the record we want a label key to resolve the target and its `package_id`, which we need for the registration file)

* On the above point, prebuilt (GHC-included) dependencies don't have `package_ids` and I need a version for the reexport. I can get one from `ghc-pkg list <name>` so just have to hack that into `haskell_import` as an output (`%{name}-package-id`, say) and use that in the other rule. Still getting my head around this since I can't just run `ghc-pkg` at analysis time and stick the version in the provider.

lunaris:
OK I think I've got this working. However, in doing so I've made some design decisions that I think it might be worth discussing here before I open some PRs (not sure who other than @mboes I should tag). The thing that generally underpins all of these is the problems created by packages that come pre-installed with GHC:

* Working backwards, the first job is to amend `package` (`actions/package.bzl`) so that it extends `exposed-modules` with reexports of the form `Module.Name from package-id:Original.Module.Name`. 

* This is fine for `HaskellLibraryInfo`-type targets since they have `package_id`s and we can pull `Module.Name` and `Original.Module.Name` from the (arbitrarily-specified) `string`s we have in the `attr.label_keyed_string_dict` that is the `exports` attribute.

[...]

PR #381, opened 2018-07-30
Support reexporting modules from `haskell_library`

lunaris:
Cabal's `reexported-modules` field allows one to reexport and optionally
rename modules from a library. This commit adds an `exports` attribute
to `haskell_library` which accomplishes the same, accepting a dictionary
mapping dependencies to strings using the same syntax. For example:

```
haskell_library(
    ...,
    exports = {
        ":dependency": "A, B as BNew",
    },
    ...
)
```

Addresses https://github.com/tweag/rules_haskell/issues/357

### DIFF
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -185,6 +185,9 @@ haskell_library = rule(
         hidden_modules = attr.string_list(
             doc = "Modules that should be unavailable for import by dependencies.",
         ),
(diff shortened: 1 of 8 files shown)


lunaris:
I've got some failing tests locally I've missed since making `version` mandatory; will address these.

lunaris:
Closing this for now as it needs more work I think.

PR #383, opened 2018-08-02
Make Bazel no longer manage modules

### LINKED ISSUE #168: Make main_file redundant where possible
`main_file` is sometimes necessary, but currently we're relying on it being specified by the user in cases where we could as well have guessed the file with the `Main` module. The only time we don't need the `main_file` argument is when the file is called `Main.hs`. But if the source file's name starts with a lower-case letter, then it's not a valid module name and must surely be the Main module. So in this case, we shouldn't need the user to specify `main_file`.

### LINKED ISSUE #375: Be as lenient as GHC about module file names
As noted  in https://github.com/tweag/rules_haskell/issues/168#issuecomment-394060386 by @judah, GHC is quite lenient about where modules should be located in the filesystem. In fact GHC happily accepts the following baroque layout, even if `A`, `B`, `C` depend on each other:

```
a/A.hs
b/B.hs
c/C.hs
```

But it goes further. GHC is fine with me defining module `C` in `A.hs`, and `A` in `C.hs`. The problem is, in this case the output objects and interface files don't match the names we expect, so we bomb out. The fix should be the same as the solution to #168: use `declare_directory()` liberally rather than guessing the exact output files given any input file.

mrkkrp:
Close #375, close #168.

It used to be that we would infer modules from the names of source
files, and then create a dependency graph based on these module names.
But this is,

1. unnecessary and somewhat more complicated,
2. incorrect in the general case, because GHC knows module names
   better than we do.

Both #375 and #168 are cases where module names don't match what we
expect. The solution is to stay out of the business of managing
modules, and manage only packages. Rather than create individual
outputs for each module, we can use `declare_directory` to create
a single output for all the interface files of a library, another for
all the object files etc.


mrkkrp:
There are still a couple of things to check experimentally though, so don't merge just yet.

mrkkrp:
OK, this is good to go.

mboes:
Resolve conflicts and merge.

PR #385, opened 2018-08-05
Reexported modules, take 2

mboes:
This PR is effectively a reopen of #381. Starting from the same code,
I made three changes:

1. rebased on top of current `master`,
2. revert the stuff that tries to ensure haskell_imports provide
   a pkg_id like a haskell_library does,
3. switch from `ghc-pkg register` to `ghc-pkg cache`.

The key observation that made this PR possible is that while `ghc-pkg
register` is very picky about how we name packages e.g. in `depends`
and `exposed-modules` fields, GHC itself is much less so. GHC is quite
happy for us to *not* specify a version number, let alone a package
hash. It's probably a bug that `ghc-pkg register` is this picky. See
https://ghc.haskell.org/trac/ghc/ticket/15478.

We side step this bug by simply not using `ghc-pkg register`. This is
a temporary workaround, because that command does some useful sanity
checks we're not getting the benefit of anymore.

### DIFF
--- a/haskell/haddock.bzl
+++ b/haskell/haddock.bzl
@@ -33,12 +33,19 @@ def _haskell_doc_aspect_impl(target, ctx):
     html_dir = ctx.actions.declare_directory(html_dir_raw)
     haddock_file = ctx.actions.declare_file(_get_haddock_path(package_id))
 
+    # XXX Haddock really wants a version number, so invent one from
+    # thin air. See https://github.com/haskell/haddock/issues/898.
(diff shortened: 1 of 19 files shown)


mboes:
Yes, the exported-modules list needs some post processing now to make it cromulent to the REPL.

lunaris:
Looks good to me. I still don't think it'd be the end of the world to have `ghc-pkg` resolve the names to unit IDs at some point, but this works now :smiley:

PR #394, opened 2018-08-09
Link indirect external dependencies

### LINKED ISSUE #170: Transitive C dependencies are not linked

```python

cc_library(
    name = "cbits-indirect",
    srcs = ["cbits/impl.c"],
)

cc_library(
    name = "cbits",
    deps = ["cbits-indirect"],
)

haskell_library(
    name = "mypkg",
    srcs = ["src/MyModule.hs"],
    src_strip_prefix = "src",
    deps = [":cbits"],
)

haskell_binary(
    name = "mytest",
    srcs = ["test/Main.hs"],
    src_strip_prefix = "test",
    deps = [":mypkg"],
)
```

guibou:
This should fix #170.

`cc_library` are not linked with their dependencies (`.so` too, I still wonder why). So linking only with the top level `cc_library` is not enough, we need to link also with the transitive dependencies. They appears inside `dep.cc.libs`, so I'm just adding them to `external_libraries`.

### DIFF
--- a/haskell/private/dependencies.bzl
+++ b/haskell/private/dependencies.bzl
@@ -21,6 +21,17 @@ def _is_shared_library(f):
     """
     return f.extension in ["so", "dylib"] or f.basename.find(".so.") != -1
 
+def _is_static_library(f):
+    """Check if the given File is a static library.
+
+    Args:
+      f: The File to check.
+
+    Returns:
+      Bool: True if the given file `f` is a static library, False otherwise.
+    """
+    return f.extension in ["a"]
+
 def _mangle_solib(ctx, label, solib, preserve_name):
     """Create a symlink to a dynamic library, with a longer name.
 
@@ -114,6 +125,16 @@ def gather_dep_info(ctx):
                 direct_prebuilt_deps = set.mutable_insert(acc.direct_prebuilt_deps, pkg),
             )
         else:
+            transitive_deps = {}
+
+            # Transitives static dependencies
+            if hasattr(dep, "cc"):
+                transitive_deps = {
+                    name: _mangle_solib(ctx, dep.label, name, CcSkylarkApiProviderHacked in dep)
+                    for name in dep.cc.libs.to_list()
+                    if _is_static_library(name)
+                }
+
             # If not a Haskell dependency, pass it through as-is to the
             # linking phase.
             acc = HaskellBuildInfo(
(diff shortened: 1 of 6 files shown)


guibou:
I've added a maximum of person to the review, @Fuuzetsu who wrote the initial code and @judah who gives good advice in #170.

guibou:
Issue fixed on my codebase.

Fuuzetsu:
Looks OK to me though leaving merging to someone else.

guibou:
> This is all very confusing. Could we take a step back and write down the spec of what want, first?

Totally, as I said on Slack, I was not really happy with this PR and I was opening it just to start a discussion. I should have stated this in the PR body too.

haskell/private/dependencies.bzl
```diff
@@ -21,6 +21,17 @@ def _is_shared_library(f):
...
+def _is_static_library(f):
+    """Check if the given File is a static library.
+
+    Args:
+      f: The File to check.
+
+    Returns:
+      Bool: True if the given file `f` is a static library, False otherwise.
+    """
+    return f.extension in ["a"]
```
Profpatsch: Isn’t the information how a target has been linked propagated somehow?

guibou: I don't know, I'll investigate.

mboes:
Any progress on this front?

mboes:
>  I'll look at this on monday.

@guibou what's the new ETA on this?

iphydf:
See https://github.com/tweag/rules_haskell/issues/429 for a use case that doesn't work with this PR.

haskell/private/dependencies.bzl
guibou: Not I'm aware of.

iphydf:
Is there anything I can do to help move this forward? I'm stuck on this commit because it solves my problem, but that means I can't use master, and can't test new developments.

guibou:
@iphydf I really need to focus an that PR but I had a lot other task to do.

In the meantime, I rebased this branch on top of master.

guibou:
I'm closing this PR for now, the current version introduces a regression and does not fit the quality needed. I will ping bug #170 once I'll come with a satisfying solution.

mboes:
@iphydf To be clear, #170 is still very much a priority.

guibou:
I'm reopening this PR. #170 can be fixed by this code and the regression detailed in #429 is an unrelated problem. I'll detail #429 reasons and workaround in its own issue.

Code have been rebased and cleaned a bit following @mboes advices.

guibou:
I'll reopen it in another PR, for a reason I don't understand, the "Reopen" button is disabled.

guibou:
see #483

PR #396, opened 2018-08-09
Generate uniq haskell packages name

### LINKED ISSUE #395: Link issue when deps contains two rules with the same prefix
See 36d3de9944f1a4b70bf1de22588474db0c1bef06  for a repo.

```
haskell_binary(
    name = "foo",
    srcs = ["Foo.hs"],
    deps = [
    "//tests/same_name/a:type",
    "//tests/same_name/b:type",
    ":base",
    ],
)
```

The two references to `//...:type` leads to a link issue:

```shell
$ bazel build //tests/same_name/bin:foo
INFO: Analysed target //tests/same_name/bin:foo (4 packages loaded).
INFO: Found 1 target...
ERROR: /home/guillaume/tweag/rules_haskell/tests/same_name/bin/BUILD:5:1: error executing shell command: '${1+"$@"} $(< bazel-out/k8-fastbuild/bin/tests/same_name/bin/_obj/foo.manifest)' failed (Exit 1)
bazel-out/k8-fastbuild/bin/tests/same_name/bin/_obj/foo/Main.dyn_o:s1m8_info: error: undefined reference to 'testsZZSsameZZUnameZZSaZZStype_AziType_foo_closure'
bazel-out/k8-fastbuild/bin/tests/same_name/bin/_obj/foo/Main.dyn_o(.data.rel.ro+0x28): error: undefined reference to 'testsZZSsameZZUnameZZSaZZStype_AziType_foo_closure'
Target //tests/same_name/bin:foo failed to build
INFO: Elapsed time: 0.637s, Critical Path: 0.26s
INFO: 0 processes.
```

guibou:
Closes #395 

Haskell package name was setup with `bazel_rule.name`, which is not unique, for example `//a:foo` and `//b:foo` have the same rule `name` but different rule `package`.

We add the package name (here `a` and `b`) to the haskell package name to avoid package name conflicts.

### DIFF
--- a/haskell/private/pkg_id.bzl
+++ b/haskell/private/pkg_id.bzl
@@ -11,7 +11,7 @@ def _zencode(s):
     """
     return s.replace("Z", "ZZ").replace("_", "ZU").replace("/", "ZS")
 
-def _to_string(my_pkg_id):
+def _to_string(label):
     """Get a globally unique package identifier.
 
     The identifier is required to be unique for each Haskell rule.
@@ -21,9 +21,9 @@ def _to_string(my_pkg_id):
     """
     return _zencode(
         paths.join(
-            my_pkg_id.label.workspace_root,
-            my_pkg_id.label.package,
-            my_pkg_id.name,
+            label.workspace_root,
+            label.package,
+            label.name,
         ),
     )
 
@@ -43,7 +43,7 @@ def _new(label, version = None):
     """
     return struct(
         label = label,
-        name = label.name.replace("_", "-"),
+        name = _to_string(label),
         version = version,
     )
 
@@ -55,13 +55,12 @@ def _library_name(hs, my_pkg_id, prof_suffix = False):
       my_pkg_id: pkg_id struct.
       prof_suffix: whether to automatically add profiling suffix.
     """
-    library_name = "HS" + _to_string(my_pkg_id)
+    library_name = "HS" + my_pkg_id.name
     if is_profiling_enabled(hs) and prof_suffix:
         library_name += "_p"
     return library_name
 
 pkg_id = struct(
     new = _new,
-    to_string = _to_string,
     library_name = _library_name,
 )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [0.7] - ???
 
+* Package name change. See
+  [#396](https://github.com/tweag/rules_haskell/pull/396)
+
 * Don’t crash on inputs missing `.haddock` interface files. See
   [#362](https://github.com/tweag/rules_haskell/pull/362)
 
--- a/haskell/private/actions/compile.bzl
+++ b/haskell/private/actions/compile.bzl
@@ -100,7 +100,7 @@ def _compilation_defaults(hs, cc, java, dep_info, srcs, import_dir_map, extra_sr
         # package directory.
         interfaces_dir = hs.actions.declare_directory(
             paths.join(
-                pkg_id.to_string(my_pkg_id),
+                my_pkg_id.name,
                 interface_dir_raw,
             ),
         )
@@ -184,8 +184,8 @@ def _compilation_defaults(hs, cc, java, dep_info, srcs, import_dir_map, extra_sr
     if my_pkg_id != None:
         unit_id_args = [
             "-this-unit-id",
-            pkg_id.to_string(my_pkg_id),
-            "-optP-DCURRENT_PACKAGE_KEY=\"{}\"".format(pkg_id.to_string(my_pkg_id)),
+            my_pkg_id.name,
+            "-optP-DCURRENT_PACKAGE_KEY=\"{}\"".format(my_pkg_id.name),
         ]
         ghc_args += unit_id_args
 
--- a/haskell/private/actions/package.bzl
+++ b/haskell/private/actions/package.bzl
@@ -18,7 +18,7 @@ def package(hs, dep_info, interfaces_dir, interfaces_dir_prof, static_library, d
     Returns:
       (File, File): GHC package conf file, GHC package cache file
     """
-    pkg_db_dir = pkg_id.to_string(my_pkg_id)
+    pkg_db_dir = my_pkg_id.name
     conf_file = hs.actions.declare_file(
         paths.join(pkg_db_dir, "{0}.conf".format(pkg_db_dir)),
     )
@@ -42,8 +42,8 @@ def package(hs, dep_info, interfaces_dir, interfaces_dir_prof, static_library, d
     metadata_entries = {
         "name": my_pkg_id.name,
         "version": my_pkg_id.version,
-        "id": pkg_id.to_string(my_pkg_id),
-        "key": pkg_id.to_string(my_pkg_id),
+        "id": my_pkg_id.name,
+        "key": my_pkg_id.name,
         "exposed": "True",
         "hidden-modules": " ".join(other_modules),
         "import-dirs": " ".join([import_dir, import_dir_prof]),
--- a/haskell/private/haskell_impl.bzl
+++ b/haskell/private/haskell_impl.bzl
@@ -273,7 +273,7 @@ use the 'haskell_import' rule instead.
         )
 
     build_info = HaskellBuildInfo(
-        package_ids = set.insert(dep_info.package_ids, pkg_id.to_string(my_pkg_id)),
+        package_ids = set.insert(dep_info.package_ids, my_pkg_id.name),
         package_confs = set.insert(dep_info.package_confs, conf_file),
         package_caches = set.insert(dep_info.package_caches, cache_file),
         # NOTE We have to use lists for static libraries because the order is
@@ -288,7 +288,7 @@ use the 'haskell_import' rule instead.
         external_libraries = dep_info.external_libraries,
     )
     lib_info = HaskellLibraryInfo(
-        package_id = pkg_id.to_string(my_pkg_id),
+        package_id = my_pkg_id.name,
         version = version,
         import_dirs = c.import_dirs,
         ghc_args = c.ghc_args,
--- a/tests/package-name/BUILD
+++ b/tests/package-name/BUILD
@@ -8,8 +8,6 @@ load(
 )
 
 haskell_library(
-    # The character "Z" should be untouched in the GHC package name.
-    # However, underscores (which are not legal) should be turned into dashes.
     name = "lib-a_Z",
     srcs = ["Lib.hs"],
     version = "1.2.3.4",
--- a/tests/package-name/Main.hs
+++ b/tests/package-name/Main.hs
@@ -5,17 +5,17 @@ module Main (main) where
 import Control.Monad (when)
 
 -- Test the PackageImports extension.
-import "lib-a-Z" Lib
+import "testsZSpackage-nameZSlib-aZUZZ" Lib
 
 -- Test macros that GHC creates automatically based on the package version.
 versionHighEnoughTest, versionTooLowTest :: Bool
-#if MIN_VERSION_lib_a_Z(1,2,3)
+#if MIN_VERSION_testsZSpackage_nameZSlib_aZUZZ(1,2,3)
 versionHighEnoughTest = True
 #else
 versionHighEnoughTest = False
 #endif
 
-#if MIN_VERSION_lib_a_Z(1,2,4)
+#if MIN_VERSION_testsZSpackage_nameZSlib_aZUZZ(1,2,4)
 versionTooLowTest = False
 #else
 versionTooLowTest = True
@@ -27,7 +27,7 @@ check x y = when (x /= y) $ error $ "Failed check: " ++ show (x, y)
 main :: IO ()
 main = do
     check foo 42
-    check VERSION_lib_a_Z "1.2.3.4"
-    check libPackageKey "testsZSpackage-nameZSlib-a-ZZ"
+    check VERSION_testsZSpackage_nameZSlib_aZUZZ "1.2.3.4"
+    check libPackageKey "testsZSpackage-nameZSlib-aZUZZ"
     check versionHighEnoughTest True
     check versionTooLowTest True
--- a/tests/same_name/a/A/Type.hs
+++ b/tests/same_name/a/A/Type.hs
@@ -1,0 +1,3 @@
+module A.Type where
+
+foo = "A"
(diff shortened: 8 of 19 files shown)


guibou:
Note: `//foo/bar:baz` and `//foo:bar-baz` will have the same name `foo-bar-baz`. I'm updating the PR with `foo-bar-PKG-baz`.

haskell/private/pkg_id.bzl
```diff
@@ -43,7 +43,12 @@ def _new(label, version = None):
     """
     return struct(
         label = label,
-        name = label.name.replace("_", "-"),
+        # the haskell package name is the concatenation of the bazel
+        # package name and the rule name
+
+        # for example:
+        # //things/animal:dog -> things-animal-dog
+        name = (label.package.replace('/', '-') + '-' + label.name).replace("_", "-"),
```
judah: @guibou the other methods (`to_string` and `library_name`) already "uniquify" the name by prepending the package and workspace.  GHC allows nonunique package names as long as the package *id* is unique.  So we shouldn't need to change the package name too. 

Furthermore, the package name is user-visible (GHC error messages, haddock index, maybe others) so we should prefer to keep it simple.

We have tests for this in `//tests/package-id-clash`, but only at the level of a haskell_library; I suspect there's somewhere in the code for `haskell_binary`  that's using the name when it should be using the id.

judah:
On the other hand, if I'm wrong and GHC requires different package *names* when linking, we should reuse the zencoding process that we do for the id, rather than reimplementing it here.

guibou:
@judah This is what `rules_haskell` generates:

```
-package-id testsZSsameZUnameZSaZStype
-package-id testsZSsameZUnameZSbZStype
-package-db bazel-out/k8-fastbuild/bin/tests/same_name/a/testsZSsameZUnameZSaZStype
-package-db bazel-out/k8-fastbuild/bin/tests/same_name/b/testsZSsameZUnameZSbZStype
```

So perhaps the order is important, I'll read the doc.

guibou:
The `Z` encoding does not fix the issue for `//foo:bar-baz` and `//foo/bar:baz` no?

guibou:
@judah you are right:

```
[nix-shell:~]$ ghci -hide-all-packages -package base -package-id resourcet-1.1.11-E7qV1UZhQ4cHKFY4kMd5uD -package-id resourcet-1.2.1-IOZGXO0zxUVq6tnDL3HEG
GHCi, version 8.4.3: http://www.haskell.org/ghc/  :? for help
Prelude> import Control.Monad.Trans.Resource

<no location info>: error:
    Ambiguous module name ‘Control.Monad.Trans.Resource’:
      it was found in multiple packages: resourcet-1.1.11 resourcet-1.2.1
Prelude> 
Prelude> 
Leaving GHCi.

[nix-shell:~]$ ghci -hide-all-packages -package base -package resourcet -package resourcet
GHCi, version 8.4.3: http://www.haskell.org/ghc/  :? for help
Prelude> import Control.Monad.Trans.Resource
Prelude Control.Monad.Trans.Resource> 
```

judah:
@guibou for z-encoding we make it unambiguous by joining the package name and rule name with "/" rather than "-".  For example:

- `//foo:bar-baz` becomes `zencode("foo/bar-baz")`
- `//foo/bar:baz` becomes `zencode("foo/bar/baz")`

guibou:
Well, if I'm changing the order of `-package-db`, it does not change anything. However if I'm changing the order of `-package-id`, it changes the link error.

guibou:
@mboes: 

`haskell_library` generates haskell packages, identified by a package `name` and a package `id`. For example, the `haskell_library` `@toto//foo/bar:baz` will have a package `name` `baz` and a package `id` which is a zencoding of `@toto//foo/bar:baz`.

- The package `id` is unique in a bazel project
- The package `name` is not unique. For example `@titi/foo/bar/baz:types` and `@toto/dog/rat/animals:types` have the same package name `types`.
- The codebase is supposed to use package `id` to discriminate packages, however #395 shows an issue (with a repro) where if two `haskell_library` have the same package `name`, they conflict and only one is linked to the final `binary`. For example the following will leads to link issue where the symbols of `//foo:types` are not available:

```
haskell_library(
   name = "binary",
   deps = [
       "//foo:types",
       "//bar:types",
    ],
)
```

The failure doesn't give any clue about the reasons of the failure and it can appears in any codebase during a mass name refactoring. For the example, my client hit this bug during a refactoring, this cost two engineers days to understand it.

## Observations

- If I flip the order of `deps`, it changes which symboles are not seen. Actually the symbols available are the one of the last `//xx:types` dependency in `deps`.
- If I flip the order of `-package-id` call in the command line used to link the executable, this changes the behavior similarly. Actually the only linked dependency is the last specified by `-package-id`.

# Proposed solutions

- Understanding why `-package-id` is not working as expected. We may hit a bug in `rules_haskell`, or a bug in GHC itself, or a bug in our understanding of `-package-id`, this needs more investigations. @Profpatsch  suggested that it may be the merging of the `-package-db` which results in only one package available.
- Fix the package `name` to something unique in a bazel project (i.e. get something like the package `id`).

I'm in favor of the second proposal, it is simple, can be fixed now without much effort and provides an additional benefit: the package `name` will become unambiguous. In a huge repository you can easily ends with many `haskell_library` with the same bazel `name`, however GHC will only see their unqualified package `name`, and will give feedbacks based on that name. On my current codebase I have 36 `haskell_library` which package name is `types`, that's awful for debuging. 

Please note that the current status of this PR is not functional, the current `package` name does not include the `workspace` name. But it will be a trivial fix once we agreed (or not) the the solution.

guibou:
@Profpatsch saw that in the documentation:

Package version selection: When selecting a package, GHC will search for packages in all available databases. If multiple versions of the same package are available the latest non-broken version will be chosen.

> Version conflict resolution: If multiple instances of a package version chosen by GHC are available then GHC will choose an *unspecified* instance.

(emphasis is mine).

Profpatsch:
So, this here is perfectly fine for GHC:

```
> nix-build -E 'with import <nixpkgs> {}; haskellPackages.ghcWithPackages (h: [h.aeson h.aeson_1_4_0_0])'
/nix/store/hp69gyvihgq6xjlq3ylzbbpndhbw1l88-ghc-8.4.3-with-packages
> ghc-pkg list --package-db result/lib/ghc-8.4.3/package.conf.d/ -v
…
    Cabal-2.2.0.1 (Cabal-2.2.0.1)
    aeson-1.3.1.1 (aeson-1.3.1.1-D1yTdtIN0v9XC16FCyHJW)
    aeson-1.4.0.0 (aeson-1.4.0.0-BOChrCsaZK2EZpXl1QePbU)
…
```

It also works with the same version number:

```
    aeson-1.2.4.0 (aeson-1.2.4.0-4cLVQkQqqOiDavYWPkGtAL)
    aeson-1.2.4.0 (aeson-1.2.4.0-INEBSJz6uawE1FDpdKs0Tx)
```

Profpatsch:
Even when I request the exact package IDs (with help by @hvr)

```
$ ghci -package-db result/lib/ghc-8.4.3/package.conf.d/ \
    -package-id aeson-1.3.1.1-D1yTdtIN0v9XC16FCyHJW \
    -package-id aeson-1.4.0.0-BOChrCsaZK2EZpXl1QePbU
Prelude> :show linker
----- Linker state -----
Pkgs: [aeson-1.4.0.0, vector-0.12.0.1, uuid-types-1.0.3,
…
```

Now I turn the `-package-id`s around:

```
> ghci -package-db result/lib/ghc-8.4.3/package.conf.d/ \
    -package-id aeson-1.4.0.0-BOChrCsaZK2EZpXl1QePbU \
    -package-id aeson-1.3.1.1-D1yTdtIN0v9XC16FCyHJW
Prelude> :show linker
----- Linker state -----
Pkgs: [aeson-1.3.1.1, vector-0.12.0.1, uuid-types-1.0.3,
…
```

So GHC only picks up the last `-package-id`, and always by package name.

I guess the solution is: have unique package names in your whole project, or GHC will silently drop your packages.

guibou:
So if we agree to change the package name to something more qualified, what should we do to keep something user friendly and robust?

- Zencoding is ugly, if user are using the `CPP` defined package name constants, this is not really readable
- Another solution? Keep in mind that `//foo/bar:baz` and `//foo:bar-baz` should not results in the same value. So simply replacing the `/` and `/` by `-` is not correct.

lunaris:
Could we perhaps replace special characters by `-` and append a hash of the original package/target name to the end? So e.g. `//foo/bar:baz` would become `foo-bar-baz-<HASH OF //foo/bar:baz>` and `//foo:bar-baz` would become `foo-bar-baz-<HASH OF //foo:bar-baz>`. If we wanted to make it a bit more obvious in most cases we could favour `_` before the `:` separating package and target and `-` afterwards (so `foo_bar-baz-HASH` and `foo-bar-baz-HASH`, maybe). Package IDs do also support `+` I believe, if we think that could be used to improve readability.

guibou:
@lunaris the issue with `HASH` at the end (for package name at least) is that user won't be able to easily know the hash. And that's used in, for example, GHC `VERSION_pkgname` macro.

This being said, I realized a bug I had since a long time with conflicting GHC `VERSION_pkgname` in my codebase was due to that issue ;)

lunaris:
Apologies; good point. Is this somewhat destined to be painful whatever happens then? E.g. if one uses Hazel to pull in a package from Hackage that uses `CPP` and `VERSION_Cabal`, will that break in any case where the package name is not "Cabal"? I.e. do we risk implementing something that hurts a lot of pre-existing code?

guibou:
@lunaris Realy good point. `haskell_import` should not change the package `name`. (Or should it, we may actually want to use different version of the same package with a different name? But what about the `VERSION` macros?)

guibou:
I pushed an updated version, which uses Z encoding for package name. At least we'll get the CI green before continuing the discussion.

guibou:
@lunaris I checked, `haskell_import` uses the `package` attribute to name the Haskell package, so it won't break any GHC macro.

judah:
Unfortunately Hazel needs to generate MIN_VERSION_* macros not just for built-in packages but also haskell_library rules.  For example, Hazel generates a `haskell_library(name = "text", ...)` rule, and any other package that depends on it needs to get the MIN_VERSION_* macro, e.g.: 
http://hackage.haskell.org/package/data-default-instances-text-0.0.1/docs/src/Data-Default-Instances-Text.html

Maybe at this point we should give up on GHC autogenerating these macros and move Hazel back to generating them manually.  (Basically, undoing FormationAI/hazel#5.)  That would also let us remove the `version` attribute from `haskell_library`.  @mboes , any thoughts?

guibou:
@judah I'm not aware of hazel internal, but I don't understand the issue you are depicting.

The situation here is:

- `haskel_import(package = "foo", name = "bar")` will expose a package named `foo` and expose the `VERSION_foo` macro. This is *unchanged* by the PR.
- `haskell_library(name = "foo")` in `@ws//pkg:foo` will expose a package named `foo` and expose the `VERSION_foo` macro. The PR *changes* that to package named `wsZCpkgZCfoo` and expose the `VERSION_wsZCpkgZCfoo` macro.

In which case is this incompatible with the way hazel behaves?

judah:
Example failure: in a checkout of the hazel repository, I changed the rules_haskell SHA to point to this PR.  I then ran:

    bazel build @haskell_chunked_data//:chunked-data

And got:
```
ERROR: /private/var/tmp/_bazel_judah.jacobson/0dd0c61b72eb0880ee09f3b1c25cfb68/external/haskell_chunked_data/BUILD:17:1: error executing shell command: '${1+"$@"}' failed (Exit 1)

bazel-out/darwin-fastbuild/bin/external/haskell_chunked_data/gen-srcs-chunked-data/Data/IOData.hs:20:22: error:
     error: token is not a valid binary operator in a preprocessor subexpression
   |
20 | #if !MIN_VERSION_text(0, 11, 3)
   |                      ^
#if !MIN_VERSION_text(0, 11, 3)
    ~~~~~~~~~~~~~~~~~^

bazel-out/darwin-fastbuild/bin/external/haskell_chunked_data/gen-srcs-chunked-data/Data/IOData.hs:66:5: error:
     error: function-like macro 'MIN_VERSION_text' is not defined
   |
66 | #if MIN_VERSION_text(0, 11, 3)
   |     ^
#if MIN_VERSION_text(0, 11, 3)
    ^
2 errors generated.
`cc' failed in phase `C pre-processor'. (Exit code: 1)
```

The `chunked-data` package depends on `text`, and contains `MIN_VERSION_text` within its source code.  In Hazel, both `chunked-data` and `text` are provided via `haskell_library` rules.  Currently Hazel relies on GHC autogenerating the macro based on the package *name*; if that changes then the name of the generated macro changes and the build fails.

guibou:
@judah Indeed, that's an issue.

guibou:
This discussion is complicated, and users are still facing unexpected build errors with unclear reasons. I propose to introduce a new error in `rule_haskell` in this case, see #402.

lunaris:
Sorry if this is the wrong place to put this or I end up rehashing things that have already been discussed, but does this work:

* We add an explicit `package_name` (or similar) attribute to `haskell_library`, `haskell_binary`, etc.
* The understanding is that this new attribute works exactly like Cabal target names work today -- they have to be globally unique (or suffer the consequences)
* We use this name verbatim as that which we register with GHC (and from which GHC generates macros)

It creates some user hassle (you have a `name` and a `package_name`), but this might be OK (Hazel takes care of your Hackage/external package scenarios and in your codebase you have full control over neat Bazel target partitioning whilst still have responsibility and control of the global namespace).

I feel like I've missed something glaring but I guess I'm basically suggesting "can we get around it the same way Stack/Cabal do -- by just having two entries"?

guibou:
@lunaris I agree with you, I'm adding:

- By default, if `package_name` is not provided, it uses the "encoded" fully qualified name. This way, most users won't have to care about setting `package`, most use cases will work out of the box.
- We add a warning / error (see #402) if a rule uses two dependencies with the same package name.

Drawbacks of this approach, any tool / repository which use the implicit unqualified package name will have to be fixed. This means hazel.

mboes:
> Maybe at this point we should give up on GHC autogenerating these macros and move Hazel back to generating them manually. (Basically, undoing FormationAI/hazel#5.) That would also let us remove the version attribute from haskell_library. @mboes , any thoughts?

OK by me, since it would (very slightly) simplify rules_haskell internals. But it would be a shame because we'd be getting a small amount of simplicity in rules_haskell for quite a bit larger complexity added to Hazel.

After further investigation, it looks like the root cause is a very strange upstream bug. I observed that Cabal does not exhibit linking issues when two packages have the same name but different package id's. It turns out that if we do linking during the `--make` phase, we likewise don't see any issues. It's when we do the linking in a separate command (as Facebook's Buck does and the GHC user guide documents) that the problem arises.

PR #401, opened 2018-08-14
Add a tutorial to rules_haskell.

mboes:
This is an adaptation of the official Bazel C++ tutorial [1] to
Haskell. It is located in `//docs` and uses Sphinx to generate HTML
from ReST source. The API documentation that used to be in `//docs`
has now moved to `//haskell`.

[1]: https://docs.bazel.build/versions/master/tutorial/cpp.html


lunaris:
Really like this effort. Especially valuable if/when we bring Hazel in.

docs/haskell.rst
```diff
@@ -0,0 +1,357 @@
...
+Notice that for the build to succeed, we make the ``//lib:booleans``
+target in ``lib/BUILD`` explicitly visible to targets in
+``main/BUILD`` using the ``visibility`` attribute. This is because by
+default targets are only visible to other targets in the same
+``BUILD`` file. (Bazel uses target visibility to prevent issues such
+as libraries containing implementation details leaking into public
+APIs.)
+
+You have built the project as two packages with three targets and
+understand the dependencies between them.
```
mrkkrp: "...and *understood* the dependencies"?

docs/haskell.rst
```diff
@@ -0,0 +1,357 @@
...
+
+Use labels to reference targets
```
mrkkrp: It's a bit strange that the tutorial ends on this section, it looks out-of-place. Also you several times write "you'll learn about X later", maybe this section could be moved closer to the beginning?

PR #403, opened 2018-08-17
Fix handling of packages with clashing names

### LINKED ISSUE #395: Link issue when deps contains two rules with the same prefix

mboes:
In #396, @guibou proposes to resolve #395 by making not just package
id's unique but also package names. That's a really big hammer, indeed
one that would cause collateral damage. Hazel would need to be
modified, because CPP macros generated by GHC would no longer be using
the names that upstream code pulled from Hackage expects. Also, the
names in Haddock would look worse.

Package names should not need to be unique. After all, that's the very
reason for the package-name / package-id distinction. It turns out
that the root cause for the original problem was very subtle, but very
simple to fix. The problem was that while we were passing
`-hide-all-packages` to GHC during compilation, we were not doing so
during linking. This flag makes sense, and indeed not passing it
causes bizarre `module not found` errors on import. It turns out that
not passing it at link time also causes the bizarre link errors we
were seeing (which only show up when two packages share the same
name).

I can't explain why `-hide-all-packages`, or lack thereof, has this
effect. But I do know that using the same flags during linking as
during compilation is the right thing to do. With this one line change
to the source code, we can continue to have arbitrary non-unique
package names, relying on package id's generated by rules_haskell
internally to be unique.

cc @guibou @profpatsch @judah @lunaris

Closes #395.

Closes #396.


haskell/private/actions/link.bzl
```diff
@@ -124,6 +124,13 @@ def link_binary(
 
     args.add(["-o", compile_output.path, dummy_static_lib.path])
 
+    # In compile.bzl, we pass this just before all -package-id
```
guibou: This comment should be copy pasted in front of all similar lines. This will motivate us to factorise and document more.

mboes: Added a reference. Duplication in comments is worse than in code, because comments are not checked.

PR #405, opened 2018-08-19
Add additional filegroup srcs in the GHC `new_local_repository` declaration

jin:
`hsc2hs` and `haddock` are required during compilation, but they're not globbed by `ghc*`. Let's add them in manually.


mboes:
What we have in the `WORKSPACE` for rules_haskell itself is:

```python
filegroup(
    name = "bin",
    srcs = glob(["bin/*"]),
)
```

More general without being too general, and simpler?

PR #413, opened 2018-08-22
WIP: CPP macro conflict

guibou:
This PR adds an unit test to reproduce a conflit in CPP macro which arise when:

- `LANGUAGE CPP` is enable in a module
- This module depends on two haskell packages with the same name. The package name is based on the bazel rule name, so `//tests/cpp_macro_conflict:bytestring` and `//tests:bytestring` have the same name `bytestring`.

See the associated unit test for details.

### DIFF
--- a/tests/cpp_macro_conflict/BS.hs
+++ b/tests/cpp_macro_conflict/BS.hs
@@ -1,0 +1,1 @@
+module BS where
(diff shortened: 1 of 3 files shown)


guibou:
Closed in favor of the discussion inside: https://github.com/tweag/rules_haskell/issues/414

PR #421, opened 2018-08-27
Fix haskell_proto_library dependency list

palmerlao:


haskell/protobuf.bzl
```diff
@@ -201,12 +201,14 @@ _haskell_proto_aspect = aspect(
...
-    dep = ctx.attr.deps[0]  # FIXME
-    return [
-        dep[HaskellBuildInfo],
-        dep[HaskellLibraryInfo],
-        DefaultInfo(files = dep[HaskellProtobufInfo].files),
-    ]
+    res = []
+    defaultInfoFilesDepset = depset()
+    for dep in ctx.attr.deps:
+        res.append(dep[HaskellBuildInfo])
```
judah: I don't think this will work like you intend.  As far as I know, you can only return a single provider from a rule.  For example, you'd need to return a single HaskellLibraryInfo that merged all of the dependencies' HaskellLibraryInfos into one.

tests/haskell_proto_library/BUILD
```diff
@@ -48,6 +53,16 @@ haskell_library(
     ],
 )
 
+haskell_library(
+    name = "hs-lib-b",
+    srcs = ["Bar.hs"],
```
judah: It looks like `Bar.hs` only imports `Proto...Person.hs`.  If you make it import `Address.hs` as well I think the build will fail:
https://github.com/tweag/rules_haskell/blob/110721c4434c842a683a0c6f38d9e59327b04b44/tests/haskell_proto_library/Bar.hs#L3

(I'm having difficulty building the repo on my laptop so haven't managed to check it myself yet.)

palmerlao: This is correct, I'll read up and try again

haskell/protobuf.bzl
```diff
@@ -201,12 +201,14 @@ _haskell_proto_aspect = aspect(
...
+        res.append(dep[HaskellLibraryInfo])
+        defaultInfoFilesDepset = depset(transitive = [defaultInfoFilesDepset, dep[HaskellProtobufInfo].files])
```
palmerlao: Also, I looked at some docs for `DefaultInfo` [0] which indicate that it has many fields, is it bad to only set `files`?

0: https://docs.bazel.build/versions/master/skylark/lib/DefaultInfo.html

palmerlao:
Thanks for the ping, and apologies for the stale PR. I no longer plan to work on this.

PR #423, opened 2018-08-28
Allow directly using the lint aspect

### LINKED ISSUE #384: Add a way to lint all visible targets
The `haskell_lint` target allows to lint a number of `haskell_*` targets, however the list of targets we want to lint has to be hardcoded in the `deps` parameter of the target.

It would be much more convenient for most use-cases to have a command to lint all the visible targets the same way `bazel build //...` allows us to build everything.

This can currently be done with `bazel build //... --aspects @io_tweag_rules_haskell//haskell:lint.bzl%haskell_lint_aspect`, however the `haskell_lint_aspect` isn't part of the interface of `rules_haskell` so there's no guaranty that this won't break anytime soon.

Can we provide an officially supported way of doing that, either by making `haskell_lint_aspec` part of the exported interface or providing an alternative method?

/cc @mboes

regnat:
Expose the `haskell_lint_aspect` in `haskell.bzl` and adds the (dummy)
generated file to it so that we can directly use it from the
command-line without having to write a `haskell_lint` target, with

```console
$ bazel build //my/haskell:target --aspects @io_tweag_rules_haskell//haskell:haskell.bzl%haskell_lint_aspect
```

Fixes #384

### DIFF
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -27,6 +27,7 @@ load(
 load(
     ":lint.bzl",
     _haskell_lint = "haskell_lint",
+    _haskell_lint_aspect = "haskell_lint_aspect",
 )
 load(
     ":toolchain.bzl",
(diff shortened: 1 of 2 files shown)


regnat:
I've merged #424 into this PR and added some documentation (in the "use cases" section). I also mention the `haskell_doc` and `haskell_lint` target in it since it wouldn't make much sense to mention one without the other.

PR #424, opened 2018-08-28
Allow directly using the haddock aspect

regnat:
Similar in spirit to #423, allows building the haddock for any
`haskell_library` or `haskell_binary` rule with

```console
$ bazel build //my/haskell:target --aspects @io_tweag_rules_haskell//haskell:haskell.bzl%haskell_haddock_aspect
```


mboes:
You could alternatively document this directly in the API documentation. It
might even be better. The important is that this should be documented. You
choose how.

regnat:
Afaik there is no way to expose an aspects in the API documentation (see <https://github.com/bazelbuild/skydoc/issues/33>), so the "use case" section seems like a good fit for me (and it's a good opportunity to start filling it)

PR #427, opened 2018-09-03
Compatibility with python3

### LINKED ISSUE #426: decoding failure when haskell package contains non-ascii symbols and python3 is used
rules_haskell does not work if the `python` provided by the environment is python3 and if a haskell package database contains non-ascii symbols.

For a simple repro, change `python` to `python3` in `tests/ghc.nix` and run the testsuite again. It should fail, because the `base-orphans` package contains non ascii chars in the name of the authors "João Cristóvão", you can see the faulty file at `bazel-out/host/bin/tests/ghc-global-pkgdb`:

```
[nix-shell:~/tweag/rules_haskell]$ head -n 1012 bazel-out/host/bin/tests/ghc-global-pkgdb | tail -n 1
           (c) 2014-2017 João Cristóvão,
```

This file comes from `ghc-pkg dump` earlier in the process.

# Error discussion

rules_haskell depends on `python` from the environment for the script `./haskell/private/ls_modules.py` which is used for compile rules.

This script is compatible python2 and 3, but with a different behavior when it comes to character encoding. One of the main difference between python 2 and 3 is that file system functions (i.e. `read` / `open`) uses array of bytes in python2  and array of opaque unicode characters in python 3. This means that all the file operations in python3 take an `encoding` attribute to specify how to decode the file (stored in bytes) to output "unicode" string.

# Possible solutions

- Continue to use byte and write the python script to use bytes in python3. This is the solution of maximum compatibility, because we keep the currently used behavior of python2.
- Start using "unicode" even in python2. The documentation of `ghc-pkg dump` states that the file is encoded in utf-8:

>  ghc-pkg dump
    Dump the registered description for every package.  This is like
    "ghc-pkg describe '*'", except that it is intended to be used
    by tools that parse the results, rather than humans.  The output is
    always encoded in UTF-8, regardless of the current locale.

- rewrite this logic in skylark. This is the more robust solution, and we'll drop the python dependency, but also the more complex.

guibou:
- files are now opened with io.open(..., encoding = 'utf8'). This is
  python 2.7 and 3 compatible and returns "unicode"
- from __future__ import unicode_literals make string literals behaves
  similarly in python 2 and 3 (i.e. "unicode")

fixes #426 (see issue for bug description and more design informations)

Rules haskell running with python3 was failing because python3 was trying to "decode" package database using an ascii codec. Some haskell package database contains name with non-ascii symbols, hence a runtime failure.

This fix specifies the decoding codec in python3 and is compatible with python2.

*NOTE*: this is a change of behavior compared to the previous version (python2) which was working on bytes only, which now works using "unicode".

### DIFF
--- a/haskell/private/ls_modules.py
+++ b/haskell/private/ls_modules.py
@@ -6,12 +6,15 @@
 # dependencies). The exposed modules are filtered using a provided
 # list of hidden modules, and augmented with reexport declarations.
 
+from __future__ import unicode_literals
+
 import collections
 import fnmatch
 import itertools
 import os
 import re
 import sys
+import io
 
 if len(sys.argv) != 6:
     sys.exit("Usage: %s <DIRECTORY> <GLOBAL_PKG_DB> <HIDDEN_MODS_FILE> <REEXPORTED_MODS_FILE> <RESULT_FILE>" % sys.argv[0])
@@ -22,7 +25,7 @@ hidden_modules_file = sys.argv[3]
 reexported_modules_file = sys.argv[4]
 results_file = sys.argv[5]
 
-with open(global_pkg_db_dump, "r") as f:
+with io.open(global_pkg_db_dump, "r", encoding='utf8') as f:
     names = [line.split()[1] for line in f if line.startswith("name:")]
     f.seek(0)
     ids = [line.split()[1] for line in f if line.startswith("id:")]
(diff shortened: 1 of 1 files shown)


guibou:
> Is there no way to not interpret the input in any encoding (like Python2 was doing)? GHC does not mandate that the pkg db files are encoded using UTF-8.

That's the second solution, it will be a bit more ugly and error prone: all string literal `"foo"` have to be converted to `b"foo`. Actually it won't cause any harm to python2, so the patch can evolve in this direction without breaking anything.

Issue #440, opened 2018-09-13
Get rid of `HaskellPrebuiltPackageInfo`


regnat:
I've been thinking a bit lately around #75, #367 and other related issues, and
my conclusion is that we should merge the `HaskellPrebuiltPackageInfo` and
`HaskellLibraryInfo` providers, but I'm not sure this is feasible in practice
so I want to gather a bit of feedback before trying to implement this.
Here's the interface I imagine:

# Low-level interface

We extend the `haskell_import` workspace rule to take as arguments
(more or less) all the arguments required by `HaskellLibraryInfo`, symlinks all
the required files into the workspace and returns a `HaskellLibraryInfo` for
the given package.

For example, instead of having say `haskell_import(name = "bytestring")`, we would
have something like

```python
haskell_import(
  name = "bytestring",
  package_id = "bytestring-0.10.8.2",
  version = "0.10.8.2",
  import_dirs = "/nix/store/a0dr5x063v0zm8z0z5fwh8dldg10yl07-ghc-8.4.3/lib/ghc-8.4.3/bytestring-0.10.8.2",
  header_files = ["fpstring.h"],
  boot_files = [],
  source_files = [],
  extra_source_files = [],
  ghc_args = [],
  exposed_modules = [
    "Data.ByteString Data.ByteString.Builder",
    "Data.ByteString.Builder.Extra Data.ByteString.Builder.Internal",
    "Data.ByteString.Builder.Prim Data.ByteString.Builder.Prim.Internal",
    "Data.ByteString.Char8 Data.ByteString.Internal Data.ByteString.Lazy",
    "Data.ByteString.Lazy.Builder Data.ByteString.Lazy.Builder.ASCII",
    "Data.ByteString.Lazy.Builder.Extras Data.ByteString.Lazy.Char8",
    "Data.ByteString.Lazy.Internal Data.ByteString.Short",
    "Data.ByteString.Short.Internal Data.ByteString.Unsafe",
  ],
)
```

which would Generate a `bytestring` repository containing:
- A symlink `import_dir` to `/nix/store/a0dr5x063v0zm8z0z5fwh8dldg10yl07-ghc-8.4.3/lib/ghc-8.4.3/bytestring-0.10.8.2`
- A file `exposed_modules_file` containing the list of `exposed_modules`
- A `BUILD` file, with a `package` rule which returns a `HaskellLibraryInfo`
  with the same arguments as the rule, except `import_dirs` which points
[...]

Issue #367, opened 2018-07-24
doctest does not find packages


guibou:
This commit: 
https://github.com/tweag/rules_haskell/commit/4b8abee80a797127c19ef17bd418bcd9d9197968

fails:

```
[nix-shell:~/tweag/rules_haskell]$ bazel build //tests/haskell_doctest:doctest-lib
INFO: Analysed target //tests/haskell_doctest:doctest-lib (1 packages loaded).
INFO: Found 1 target...
ERROR: /home/guillaume/tweag/rules_haskell/tests/haskell_doctest/BUILD:53:1: error executing shell command: '
    external/doctest/bin/doctest "$@" > bazel-out/k8-fastbuild/bin/tests/haskell_doctest/doctest-log-doctest-lib-testsZShaskellZUdoctestZSlib-b-1.0.0 2>&1 || rc=$? && cat bazel-out/k8-fastbuild/bi...' failed (Exit 1)
doctest: <command line>: cannot satisfy -package lens-labels
    (use -v for more information)
Target //tests/haskell_doctest:doctest-lib failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 0.544s, Critical Path: 0.09s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
```

I used `lens-label` because that's not a package which comes builtin with GHC by default. All the other example are using packages such as `bytestring` or `base` which are builtin with GHC.

guibou:
`doctest` needs ghc options, such as `--package-db` to find the package. For now they are not passed. I'm working on it.

Profpatsch:
Hm, guibhui, I suspect you will duplicate a lot of functionality that is used to generate ghci inputs: https://github.com/tweag/rules_haskell/blob/master/haskell/private/actions/repl.bzl#L51

guibou:
I'm closing. I'm still experimenting a similar issue, but the current repro and issue description is not up to date.

PR #442, opened 2018-09-19
Remove `HaskellPrebuiltPackageInfo`

regnat:
(depends on tweag/rules_nixpkgs#29)

Solve #440 by provinding an alternative `haskell_import` rule in `haskell/import.bzl` which directly generates an `HaskellLibraryInfo`, `HaskellBuildInfo` and `HaddockInfo` for the dependency.

There is also a new set of rules in `haskell/nix.bzl` to generate these rules from a nixpkgs definition with just a bit of boilerplate (see the documentation of `haskell_nixpkgs_packages`).

Apart from the testsuite, I've succesfully tried it on a mid-sized haskell repository (apart from doctest, but it wasn't working before either because of #367, so nothing new on this side)

@mrkkrp since you proposed some help, there's probably a lot to say regarding this PR as it is my first real skylark work and I'm not familiar with the internals of `rules_haskell`


regnat:
Hum, an unpleasant side-effect of this (more precisely of the `gen_packages_list` rule) is that we now need `nix-build` in path to compute the build graph (because we `load` the result of a `nixpkgs_package` in the WORKSPACE file). I don't know how much this in undesirable.

(the `gen_packages_list` function is technically optional as we could replace its result by a hand-written list of the haskell packages needed, but this list needs to be closed by transitive dependency so maintaining it by hand is probably gonna be quite awful)

regnat:
@Profpatsch thanks to the early review.
I've rebased the PR to have more meaningful commits (I tried not to touch the three first too much since you already reviewed them).

haskell/import.bzl
```diff
@@ -0,0 +1,91 @@
...
+        static_libraries_prof = [],
+        dynamic_libraries = set.empty(),
+        interface_dirs = set.empty(),
+        prebuilt_dependencies = set.empty(),
+        external_libraries = {},
+        direct_prebuilt_deps = set.empty(),
+    )
+    return [buildInfo, libInfo]
+
+haskell_import = rule(
```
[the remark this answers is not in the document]
regnat: Eventually yes, though we probably should have a way to generate the correct (new) `haskell_import` statements for other inputs than nix (like ghc bindist or an external ghc) first.

haskell/nix.bzl
```diff
@@ -0,0 +1,166 @@
...
+    repositories = {
+        "bazel_haskell_wrapper": "@io_tweag_rules_haskell//haskell:nix/default.nix",
+    } + repositories
+
+    nixpkgs_args = dict(
+        name = name,
+        attribute_path = attribute_path,
+        build_file_content = build_file_content,
+        nix_file_deps = nix_file_deps + ["@io_tweag_rules_haskell//haskell:nix/default.nix"],
+        repositories = repositories,
```
lunaris: In playing with this I've realised (I believe) that it depends on the `allow-multiple-repositories` branch of `rules_nixpkgs` being merged first; is this right?

haskell/nix.bzl
```diff
@@ -0,0 +1,166 @@
...
+
+    Example:
+
+      In `haskellPackages.nix`:
+
+      ```nix
+      with import <nixpkgs> {};
+
+      let wrapPackages = callPackage <bazel_haskell_wrapper> { }; in
+      { haskellPackages = genBazelBuild haskell.packages.ghc822; }
```
lunaris: This should be `let genBazelBuild = ...` I think.

haskell/nix.bzl
regnat: Oh you're right, I've forgotten to write it explicitely. I'll add it to the description of the PR

haskell/nix.bzl
regnat: Indeed, thanks

lunaris:
Can't quite get this to work. With `WORKSPACE`:

```
workspace(name = "test")

# allow-multiple-repositories branch's most-recent commit at time of writing
http_archive(
    strip_prefix = "rules_nixpkgs-b81967b6ccac855f71151785dbb9fe56f81a0562",
    urls = ["https://github.com/tweag/rules_nixpkgs/archive/b81967b6ccac855f71151785dbb9fe56f81a0562.tar.gz"],
)

load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl", "nixpkgs_git_repository", "nixpkgs_package")

# Commit in this PR before documentation fix
http_archive(
  name = "io_tweag_rules_haskell",
  strip_prefix = "rules_haskell-4666061507b88f6e0934e9d1db04d95fccd4b2b1",
  urls = ["https://github.com/tweag/rules_haskell/archive/4666061507b88f6e0934e9d1db04d95fccd4b2b1.tar.gz"],
)

load("@io_tweag_rules_haskell//haskell:repositories.bzl", "haskell_repositories")
haskell_repositories()

# Recent Nixpkgs commit
nixpkgs_git_repository(
  name = "nixpkgs",
  revision = "d22e22e2dd83510ca11ab6966cacd2ee929a2d73",
)

nixpkgs_package(
  name = "ghc",
  repositories = {"nixpkgs": "@nixpkgs"},
  attribute_path = "haskell.compiler.ghc822",
)

register_toolchains("//:ghc")

load("@io_tweag_rules_haskell//haskell:nix.bzl", "haskell_nixpkgs_packages", "gen_packages_list")

gen_packages_list(
  name = "hackage_packages",
  repositories = {"nixpkgs": "@nixpkgs"},
  nix_file = "//:haskell-packages.nix",
)

load("@hackage_packages//:all-haskell-packages.bzl", "packages")

haskell_nixpkgs_packages(
  name = "hackage",
  packages = packages,
  repositories = {"nixpkgs": "@nixpkgs"},
  nix_file = "haskell-packages.nix",
)
```

`BUILD`:

```
load("@io_tweag_rules_haskell//haskell:haskell.bzl", "haskell_toolchain")

haskell_toolchain(
  name = "ghc",
  version = "8.2.2",
  tools = "@ghc//:bin",
)
```

`haskell-packages.nix`:

```
with import <nixpkgs> {};

let
  genBazelBuild = callPackage <bazel_haskell_wrapper> { };
in {
  haskellPackages = genBazelBuild haskell.packages.ghc822;
}
```

`shell.nix`:

```
{ pkgs ? import ./nixpkgs.nix {} }:

with pkgs;
with darwin.apple_sdk.frameworks;

# XXX On Darwin, workaround
# https://github.com/NixOS/nixpkgs/issues/42059. See also
# https://github.com/NixOS/nixpkgs/pull/41589.
let
  cc = stdenv.mkDerivation {
    name = "cc-wrapper-bazel";
    buildInputs = [ stdenv.cc makeWrapper ];
    phases = [ "fixupPhase" ];
    postFixup = ''
      mkdir -p $out/bin
      makeWrapper ${stdenv.cc}/bin/clang $out/bin/clang \
        --add-flags "-isystem ${llvmPackages.libcxx}/include \
                    -F${CoreFoundation}/Library/Frameworks \
                    -F${CoreServices}/Library/Frameworks \
                    -F${Security}/Library/Frameworks \
                    -F${Foundation}/Library/Frameworks \
                    -L${libcxx}/lib \
                    -L${darwin.libobjc}/lib"
    '';
  };
  mkShell = pkgs.mkShell.override {
    stdenv = if stdenv.isDarwin then overrideCC stdenv cc else stdenv;
  };
in
  mkShell {
    buildInputs = [
      binutils
      git
      go
      nix
      which
      python
    ]
    # TODO use Bazel from Nixpkgs even on Darwin. Blocked by
    # https://github.com/NixOS/nixpkgs/issues/30590.
    ++ lib.optional stdenv.isLinux bazel;
  }
```

`nixpkgs.nix`:

```
# Keep this value in sync with `WORKSPACE`
import (fetchTarball "https://github.com/NixOS/nixpkgs/archive/d22e22e2dd83510ca11ab6966cacd2ee929a2d73.tar.gz")
```

I get:

```
➜  test git:(master) ✗ nix-shell --pure --run "bazel build --verbose_failures @hackage//:containers"                      
INFO: Build options have changed, discarding analysis cache.
ERROR: /.../.cache/bazel/_bazel_.../.../external/hackage/BUILD:15239:1: no such package '@hackage-containers//': Not a regular file: /.../test/external/haskell-packages.nix and referenced by '@hackage//:containers'
ERROR: Analysis of target '@hackage//:containers' failed; build aborted: Analysis failed
INFO: Elapsed time: 6.827s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (3 packages loaded)
➜  test git:(master) ✗ nix-shell --pure --run "bazel build --verbose_failures @hackage-containers//..."
ERROR: Not a regular file: /.../test/external/haskell-packages.nix
INFO: Elapsed time: 0.208s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
```

The issue seems to be the `/external/` being placed before `haskell-packages.nix` when it's being referenced, but digging through the code in this PR I can't see where that might be happening/whether I'm misnaming something to cause this. Any pointers?

regnat:
@lunaris That's another typo in the example, it should be `nix_file = "//:haskell-packages.nix"` and not `nix_file = "haskell-packages.nix` at line 51 of the `WORKSPACE` (I just tried your example with this change and it's enough to make it build)

lunaris:
Awesome -- it works! I'll put some testing in now but I think this is sweet and it's just a matter of what the final version should be (e.g. the `_new` stuff and bikeshedding -- should `gen_packages_list` be `haskell_gen_packages_list` or the like). And perhaps (though I've built locally so this may already be there) some more documentation in the Skydoc vein.

haskell/nix/default.nix
```diff
@@ -0,0 +1,113 @@
...
+          fi
+        done
+        echo -n "]"
+      }
+
+      mkdir -p $out
+      cat <<EOF > $out/BUILD.bzl
+      load("@io_tweag_rules_haskell//haskell:import.bzl", haskell_import_new = "haskell_import")
+      deps_ids = $(query_list depends)
+      deps_repos = $(get_deps)
```
lunaris: So this has the potential to cause a dependency cycle it seems. When building `haddock-library` (e.g. with `bazel build @hackage-haddock-library//...`), this list includes `haddock-library` itself. I've not worked out if this is a quirk of the documentation generation process or the Nix derivation for `haddock-library`, but we may need a workaround here.

regnat: There is something strange going on there indeed. It looks like the `haddock-library` package has several libraries, and that apparently perturbs the script.
I can't get the nix package to build though, which prevents me from testing it further, which nixpkgs revision are you using (or maybe you have a custom override for it?)

regnat: The issue seems to come from [here](https://github.com/tweag/rules_haskell/pull/442/files/8fadda46c00a4d6fbcf7cca7e9d96b073b666ccf#diff-0415aa1dd12e257be8d4ff76b830094fR45): I've assumed that package-ids were always in the form `{package_name}-{version}-{some_random_garbage}`, but this isn't true when package define "internal libraries" whose ids are of the form `{package_name}-{version}-{some_random_garbage}-{library_name}` (in this example, `haddock-library-1.5.0.1-919WkNWGnku6Kvc4vlhbv1` depends on `haddock-library-1.5.0.1-919WkNWGnku6Kvc4vlhbv1-attoparsec`, but the script doesn't make any difference between both).

Since the internal libraries are in the same package as their main library, there is no need to distinguish them so the TL;DR; is that we can just filter-out self-dependencies and things should work smoothly

regnat: My last commit should fix this, though I couldn't test it on `haddock-library` as I can't build it

lunaris: Nice -- I noticed this when inspecting the `ghc-pkg` DB for that derivation but didn't clock the subpackage. The latest commit resolves this for me :smile:

lunaris:
Alas, it seems there is something more to the subpackage issue. If you take the files I posted earlier and add something like this to `BUILD`:

```

load("@io_tweag_rules_haskell//haskell:haskell.bzl", "haskell_library")
load("@io_tweag_rules_haskell//haskell:haskell.bzl", "haskell_toolchain")

haskell_toolchain(
  name = "ghc",
  version = "8.4.3",
  tools = "@ghc//:bin",
)

haskell_library(
  name = "pandoc-test",
  src_strip_prefix = "pandoc-test",
  srcs = glob(["pandoc-test/**/*.hs"]),
  deps = [
    "@hackage//:base",
    "@hackage//:pandoc",
  ],
)
```

with a `pandoc-test/Main.hs` containing anything, e.g.:

```
module Main (main) where

main :: IO ()
main = putStrLn "Hello"
```

Then this will fail to build due to not being able to satisfy our old friend `haddock-library`. Adding `@hackage//:haddock-library` to the `deps` doesn't help either -- the key is that it can't satisfy `haddock-library-...-attoparsec`.

regnat:
I still can't build your example (because `haddock-libaryr` doesn't build), but I can reproduce the bug by replacing `ghc822` by `ghc843`. This comes from the fact that multi-library packages export one package.conf per library, and I was only importing one into bazel. I've pushed a commit which fixes it (at least your example builds fine with `ghc843`) and I'll add a regression test for it too

lunaris:
Yes, works for me too!

mrkkrp:
So this is still blocked at least on https://github.com/tweag/rules_nixpkgs/pull/29 ?

regnat:
@mrkkrp yes, still. Though  tweag/rules_nixpkgs#29 is a much simpler change so should be easy to merge if we think that this one is good-to-go

haskell/nix.bzl
```diff
@@ -48,3 +48,119 @@ targets()
...
+    name = "{package}",
+    actual = "@{base_repo}-{package}//:pkg",
+)
+        """.format(
+            package = package,
+            base_repo = repository_ctx.attr.base_repository,
+        )
+    repository_ctx.file("BUILD", build_file_content)
+
+bundle = repository_rule(
```
Profpatsch: Should have a docstring/`docs` attribute. I opened an issue upstream: https://github.com/bazelbuild/bazel/issues/6271, so we can reference that.

haskell/nix.bzl
```diff
@@ -48,3 +48,119 @@ targets()
...
+    )
+
+def gen_packages_list(name, base_attribute_path, repositories = {}, *args, **kwargs):
+    """Generate a list of all the available haskell packages"""
+    repositories = {
+        "@io_tweag_rules_haskell//haskell:nix/default.nix": "bazel_haskell_wrapper",
+    } + repositories
+    nixpkgs_package(
+        name = name,
+        attribute_path = base_attribute_path + ".packageNames",
```
Profpatsch: Hm, how does this work?

```
nix-repl> haskellPackages.packageNames
error: attribute 'packageNames' missing, at (string):1:1
```

haskell/nix.bzl
```diff
@@ -48,3 +48,119 @@ targets()
...
+            **kwargs
+        )
+    bundle(
+        name = name,
+        packages = packages,
+        base_repository = name,
+    )
+
+def gen_packages_list(name, base_attribute_path, repositories = {}, *args, **kwargs):
+    """Generate a list of all the available haskell packages"""
```
Profpatsch: I think this needs a bit of documentation itself, for what the attributes represent.

haskell/nix/default.nix
```diff
@@ -0,0 +1,113 @@
...
+          )
+      EOF
+    '';
+  genAllBuilds = pkgSet:
+    let newSet =
+      lib.mapAttrs (package_name: package:
+      let
+        # Some nix packages are actuall `null` because the haskell package is
+        # bundled with ghc (so it doesn't have a custom derivation of its own).
+        # For these, we simply pass the ghc derivation instead of their.
```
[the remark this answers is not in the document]
lunaris: Or `theirs`. Also `actually` in the opening sentence.

haskell/nix.bzl
lunaris: I also think its name should be reconsidered so that it is namespaced to `rules_haskell` appropriately. E.g. `haskell_...`. That said, is there a use case where I wouldn't immediately tee this up into `haskell_nixpkgs_packages`? If not, we could combine them into a single function which would arguably be a lower learning curve?

haskell/haddock.bzl
```diff
@@ -25,6 +25,8 @@ def _get_haddock_path(package_id):
 def _haskell_doc_aspect_impl(target, ctx):
     if HaskellBuildInfo not in target or HaskellLibraryInfo not in target:
         return []
+    if HaddockInfo in target:
+        return target[HaddockInfo]
```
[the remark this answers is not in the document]
regnat: The `haskell_import` rule also returns an `HaddockInfo` provider (because the haddocks are already here in the imported package) which is the reason for this change. I'll add a comment to explain it

haskell/import.bzl
```diff
@@ -4,32 +4,71 @@ load(":private/context.bzl", "haskell_context")
...
-    recache_args.add(["-D"] + list(ctx.attr.package_conf.files) + [
-        "--target-directory",
-        local_package_conf,
-    ])
+    local_haddock_html = ctx.actions.declare_directory("haddock-html")
+    copy_all(
+        ctx = ctx,
+        srcs = ctx.attr.haddock_html.files,
+        dest = local_haddock_html,
+    )
```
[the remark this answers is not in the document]
regnat: I don't understand why (probably related to the internal cooking of the `haddock` aspect), but just symlinking these caused me some permission errors when trying to build the haddocks, hence the copying

haskell/nix.bzl
regnat: Agreed for the name and the documentation. Would `haskell_nixpkgs_package_list` be OK? That doesn't sound awesome but I can't come up with anything better.

For the merge with `haskell_nixpkgs_packages`, I'd like to have it, but `load` statements can't be placed in the body of functions in bazel, so to the top of my knowledge it isn't possible.

haskell/nix.bzl
regnat: This requires the `bazel_haskell_wrapper` wrapper to work (which is also needed by `haskell_import_packages` so not a big deal here)

haskell/import.bzl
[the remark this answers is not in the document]
Profpatsch: Hm, interesting. Can you paste the error messages? I didn’t manage to change it to a symlink semantics (naïvely switching to `link_forest` doesn’t work).

Maybe this also warrants a code comment.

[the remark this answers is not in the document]
regnat: By changing the `copy_all` rule as

```diff
diff --git a/haskell/private/path_utils.bzl b/haskell/private/path_utils.bzl
index 7126118..8fd2e2f 100644
--- a/haskell/private/path_utils.bzl
+++ b/haskell/private/path_utils.bzl
@@ -237,11 +237,12 @@ def copy_all(ctx, srcs, dest):
         )
     else:
         args = ctx.actions.args()
-        args.add_all(srcs)
+        real_srcs = [ _truly_relativize(src.path, dest.path) for src in srcs]
+        args.add_all(real_srcs)
         ctx.actions.run_shell(
             inputs = depset(srcs),
             outputs = [dest],
             mnemonic = "Copy",
-            command = "mkdir -p {dest} && cp -L -R \"$@\" {dest}".format(dest = dest.path),
+            command = "mkdir -p {dest} && ln -s \"$@\" {dest}".format(dest = dest.path),
             arguments = [args],
         )
```

(so which effectively does a symlinking of everything), I get

```
ERROR: /home/regnat-dev/.cache/bazel/_bazel_regnat-dev/8da26023ce322171e002355d3467f67c/external/hackage-template-haskell/BUILD:6:1: Error while validating output Tree
Artifact File:[[<execution_root>]bazel-out/k8-fastbuild/bin]external/hackage-template-haskell/haddock-html : /home/regnat-dev/.cache/bazel/_bazel_regnat-dev/8da26023ce
322171e002355d3467f67c/execroot/io_tweag_rules_haskell/bazel-out/k8-fastbuild/bin/external/hackage-template-haskell/haddock-html/template-haskell-2.12.0.0 (Read-only f
ile system)
```

I must admit I didn't try to dig deep into this since copying is cheap enough not to be a real problem

haskell/import.bzl
```diff
@@ -0,0 +1,104 @@
...
+        dynamic_libraries = set.from_list(ctx.attr.dynamic_libraries),
+        interface_dirs = set.empty(),
+        prebuilt_dependencies = set.empty(),
+        external_libraries = {},
+        direct_prebuilt_deps = set.empty(),
+    )
+    html_files = list(ctx.attr.haddock_html.files)
+    transitive_html = {ctx.attr.package_id: local_haddock_html}
+    interface_files = list(ctx.attr.haddock_interfaces.files)
+    transitive_haddocks = {ctx.attr.package_id: interface_files[0]} if interface_files != [] else {}
```
Profpatsch: Adding an explanatory comment to this line is the last thing from my side, then +1 for the merge.

.netlify/install.sh
```diff
@@ -7,3 +7,18 @@ V=0.16.0
 curl -LO https://github.com/bazelbuild/bazel/releases/download/$V/bazel-$V-installer-linux-x86_64.sh
 chmod +x bazel-$V-installer-linux-x86_64.sh
 ./bazel-$V-installer-linux-x86_64.sh --user
+
+# XXX: Hack to prevent the `gen_packages_list` rule from crashing
```
lunaris: If we rename `gen_packages_list` we should update this comment. Might also be worth explaining what about the situation causes it to crash if this hack isn't present.

haskell/haddock.bzl
```diff
@@ -216,10 +221,11 @@ def _haskell_doc_rule_impl(ctx):
     if ctx.attr.index_transitive_deps:
         # Include all packages in the unified index.
         for package_id in html_dict_copied:
-            args.add("--read-interface=../{0},{1}".format(
-                package_id,
-                haddock_dict[package_id].path,
-            ))
+            if package_id in haddock_dict:
```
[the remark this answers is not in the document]
regnat: That was needed to avoid an error for prebuild packages without haddock, but I've changed the way their `HaddockInfo` is generated so that the `transitive_html` and `transitive_haddocks` dicts have the same shape (though I'm not really fond of the fact that we manipulate two different dicts with the implicit invariant that they have the same keys, but that's out of scope for this PR)

haskell/nix/default.nix
```diff
@@ -0,0 +1,120 @@
...
+  - Its documentation
+  - A bazel file ready to be loaded from the `BUILD` file and containing the
+  right call to `haskell_import`
+*/
+{ runCommand, lib, writeTextDir, symlinkJoin }:
+let
+  /* Generate the BUILD file for the package */
+  genBuildFile =
+    { package_name, package, ghc }:
+    runCommand "${package_name}-BUILD" {
```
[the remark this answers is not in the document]
regnat: Your guess is correct regarding the arguments.
The only `"$@"` in the script are inside bash functions and are the arguments of these functions.

`runCommand` is similar to bazel's `genrule` and generates a store path (referenced as `$out` in the script), so `genBuildFile` returns a store path containing the `BUILD.bzl` file generated in the script.

.circleci/config.yml
```diff
@@ -18,7 +18,9 @@ jobs:
           name: Build
           command: |
             # XXX Workaround https://github.com/NixOS/nix/issues/1969.
-            nix-build nixpkgs.nix -A haskell.compiler.ghc822
+            echo "y: x: x" > dummy_wrapper.nix
```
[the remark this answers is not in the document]
regnat: Right that would have deserved a bit of explanation,

I've actually changed this on the `tests/ghc.nix` file so that we don't need this anymore

tests/binary-with-prebuilt/BUILD
```diff
@@ -3,14 +3,22 @@ package(default_testonly = 1)
...
-        "//tests:base",
-        "//tests:template-haskell",
+        "@hackage//:base",
+        "@hackage//:streaming",
+        "@hackage//:template-haskell",
+        "@hackage//:void",
     ],
 )
+
+haskell_doctest(
```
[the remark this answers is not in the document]
regnat: This should actually get away, it was a mistake on my side to commit it

lunaris:
Is this blocked on the "next step" mentioned in the previous comment? I.e. do we need to provide a script/document an existing one that generates `haskell_import` calls for a system GHC before merging this?

haskell/nix.bzl
```diff
@@ -0,0 +1,172 @@
...
+)
+
+def haskell_nixpkgs_package(
+        name,
+        attribute_path,
+        nix_file_deps = [],
+        repositories = {},
+        build_file_content = None,
+        build_file = None,
+        *args,
```
mboes: I don't think we need to support non-keyword arguments here. It would be bad style if the user did that.

haskell/import.bzl
```diff
@@ -0,0 +1,106 @@
...
+        package_id = ctx.attr.package_id,
+        transitive_html = transitive_html,
+        transitive_haddocks = transitive_haddocks,
+    )
+    return [buildInfo, libInfo, haddockInfo]
+
+haskell_import = rule(
+    _haskell_import_impl,
+    attrs = dict(
+        package_id = attr.string(doc = "Workspace unique package identifier"),
```
mboes: Do we really need this? I think we ought to derive the `package_id` from the target name.

haskell/import.bzl
```diff
@@ -0,0 +1,106 @@
...
+        deps_ids = attr.string_list(),
+        deps = attr.label_list(doc = "Haskell dependencies for the package"),
```
mboes: Why do we need both `deps` and `deps_ids`?

haskell/import.bzl
```diff
@@ -0,0 +1,106 @@
...
+        dynamic_libraries = attr.label_list(),
+        version = attr.string(doc = "Package version."),
+        deps_caches = attr.label(),
```
mboes: This attribute should be documented.

haskell/import.bzl
```diff
@@ -0,0 +1,106 @@
...
+        haddock_interfaces = attr.label(doc = "List of haddock interfaces"),
+        haddock_html = attr.label(doc = "List of haddock html dirs"),
+        package_confs = attr.label(),
```
mboes: This attribute should be documented.

haskell/nix.bzl
```diff
@@ -0,0 +1,172 @@
...
+
+bundle = repository_rule(
+    attrs = {
+        "packages": attr.string_list(),
+        "base_repository": attr.string(),
+    },
+    implementation = _bundle_impl,
+)
+
+def haskell_nixpkgs_packages(name, base_attribute_path, packages, *args, **kwargs):
```
mboes: The naming looks redundant. We could as well call this `haskell_nixpkgs`, a repository rule that imports every Haskell package found in Nixpkgs.

haskell/import.bzl
regnat: Indeed we don't need it.
I didn't try at all to minimize the interface for this rule since it isn't intended to be hand-written anyways so having a bit too much information in exchange of making its implementation easier isn't a big deal, but in this case it's trivial to extract it, so we can remove the `deps_ids` argument

haskell/import.bzl
regnat: It actually isn't even used in the implementation, I'll drop it

regnat:
@mboes, thanks for the review

>   Finish up haskell_import(). It's awkward to have two rules with the same name that do different things. The new rule should be a drop-in replacement for the old one, and importable from the same location.

I've bee a bit schizophrenic here because I didn't really know initially how this would look like eventually and I hoped this could become a drop-in replacement for `haskell_import`, but I now think the name `haskell_import` is misleading because although it has the same purpose as the old `haskell_import` one, it is intended to be used in a quite different way. So what about renaming it to `haskell_external_import` or some other name to clearly state that this isn't `haskell_import`?

>  The haskell_nixpkgs_packages{,_list} macros look awkward to me. Why do I as a user need to plumb packages through myself? That's just boilerplate that will always be the same in all projects. I also have to repeat a bunch of Nix related arguments in the invocations to both macros. Would it be possible to generate a .bzl that we just include verbatim in the WORKSPACE?

It's not possible to have one macro do the `haskell_nixpkgs_package_list` + `haskell_nixpkgs_packages` plumbing because of the `load` statement which has to stay between them − and we can't `load` inside a macro − but we might indeed reduce this to one macro call + one `load`. I'll have to think about it further, but it would be a great improvement since this looks quite awkward to me too.

> PR claims to remove HaskellPrebuiltPackageInfo but doesn't actually do that. Is this intended?

That's also a consequence of me not knowing what I was doing at the begining.
There isn't a way yet to generate the new `haskell_import` calls for a non nix-provided GHC (i.e. using a GHC from the environment or a bindist), so these still need to use the old `haskell_import` rule and `HaskellPrebuiltPackageInfo`. I plan on actually removing these in a further PR

regnat:
I have reworked a bit the end-user API which now looks like [this](https://github.com/tweag/rules_haskell/blob/a1525a66e1a7c21b3ea00f906f03fb7f4a0b9ac7/WORKSPACE#L113-L120) − I find this much nicer, but at the cost of a more convoluted implementation.

Regarding the CI failure, I don't really understand what it is due to: It looks like the bazel daemon starts eating all the memory, leading to nix-build calls being OOM-killed, but

- I can't reproduce this locally
- Shutting it down (with `bazel shutdown`) and restarting the build makes it build successfully

regnat:
Rebased it on top of `master` and fixed the CI failures.

@mboes do you like the [new api](https://deploy-preview-442--api-haskell-build.netlify.com/haskell/nix.html#haskell_nixpkgs_packageset) better?

lunaris:
@mboes This just blocked on your approval?

lunaris:
That said, on rebasing and testing here, does this interact with https://github.com/tweag/rules_haskell/commit/b58050c284a676d64c4ea082ec44899aee837906 so that we don't get CPP macros generated any more?

mboes:
@lunaris correct. Large PR's like this require large amounts of contiguous high-concentration time, which has been hard to come by lately. But I have discussed this PR with @regnat offline and I think we're pretty much there, at least for a first iteration. @regnat could you rebase on top of current `master` and resolve the conflicts?

lunaris:
@mboes Apologies for my terseness; please don't misinterpret -- I didn't
mean to come across as pestering and completely understand the work
required. I was more wondering if there's more testing or the like to be
done since I'm already semi-happily using the work in this branch and am
happy to lend more eyes :smile:

mboes:
Oh, sorry I should have been clearer: my intent was to provide some much needed clarity on the status of this PR, which I have failed adequately provide previously. Also, sorry to @regnat that this process has taken so long. No worries at all about any pestering, and thanks for the ping!

regnat:
I've just rebased on top of master, there wasn't any major conflict, just some package renamings in tests.

> I'm wondering if there's any possibility that it affects Bazel's remote caching support

I don't see what in this change could cause it, but who knows…

> does this interact with b58050c so that we don't get CPP macros generated any more

In theory it shouldn't since the choice of generating or not the macro is done via a ghc flag passed during the build of the package and we don't touch these. Did you have any issue with it?

lunaris:
After rebasing, two packages that had built before that use `-XCPP` (`bloodhound` and `hs-webdriver`) refused to do so as CPP cited it didn't understand `#if !MIN_VERSION_base(4,8,0)` and the like. Naturally when I rebased I pulled in stuff from `master` too, but the commit in question seems to indicate macros are only generated if a version is passed and I wondered what with pinning things to Nixpkgs that maybe we'd stopped doing that in this PR.

regnat:
Hm… it's actually surprising that it worked before because I forgot to forward the `include-dirs` which are needed for these macros.

If it's really the problem, it should be easy to fix. Do you by any chance have a small repro handy?

lunaris:
@regnat Reproducing repo for the CPP problem at https://github.com/lunaris/bazel-example (`master`); hopefully that helps.

PR #443, opened 2018-09-20
Fix doctest invocation without a `module` argument

regnat:
Probably due to #383, calling `haskell_doctest` without a `module`
argument resulted in doctest being run on no module at all instead of
all the modules.

This PR fixes it.

### DIFF
--- a/haskell/doctest.bzl
+++ b/haskell/doctest.bzl
@@ -132,14 +132,12 @@ def _haskell_doctest_single(target, ctx):
     )
 
     if not ctx.attr.modules:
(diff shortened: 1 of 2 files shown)


lunaris:
LGTM; will merge when the checks pass.

mboes:
@regnat could you have a look at the failing checks?

regnat:
Oh sorry, I haven't seen the CI failed (I though github was sending notification emails for failed CI runs, but it apparently isn't the case and I can't find a way to enable this in the settings :/). Anyays, I've pushed a fix, it should be okay now

Issue #425, opened 2018-08-30
allow passing command line flags to haddock


ghorn:
I would like to be able to pass `--ignore-all-exports` to haddock in order to display haddock comments on internal functions. Could a field be added to `haskell_doc` for passing additional flags? Or maybe just a boolean field for this flag if you're afraid of users trying to unsafely override flags.

ghorn:
from https://www.haskell.org/haddock/doc/html/invoking.html:

> --ignore-all-exports
> Causes Haddock to behave as if every module has the ignore-exports attribute ([Section 3.7, “Module Attributes”](https://www.haskell.org/haddock/doc/html/module-attributes.html)). This might be useful for generating implementation documentation rather than interface documentation, for example.

guibou:
A bit of followup on that one.

This is a real pain. It is possible to set attributes to a rule which will be passed to the aspects (used by `haskell_doc`). However these attributes are limited: they must be a `string` from a finite list of possible values:

> Public aspect attributes are of type string and are called parameters. Parameters must have avalues attribute specified on them. In this case we have a parameter called extension that is allowed to have ‘*’, ‘h’, or ‘cc’ as a value.

(from https://docs.bazel.build/versions/master/skylark/aspects.html)

That's by *design* in bazel.

PR #444 was a PR with a proof of concept of a limited set of arguments. It was closed because the solution was a HACK.

mboes:
Yet another reason to change the design in the following way?

```
$ bazel build //some/target@doc
```

i.e. use implicit targets for API documentation, like we currently do for REPL's?

PR #444, opened 2018-09-21
haskell_doc: add 'flags' attribute

guibou:
This PR introduces `flags` to `haskell_doc`, this allows the user to pass flags to haddock
to change the haddock behavior, for exemple, to insert unicode symbols.

```python
haskell_doc(name = "foo", deps = [...], flags = "-U --ignore-all-exports")
```

The implementation is an art of hacking because bazel allows top level
rule to pass arguments to dependency tree of aspect. However the API
only alows string and forces us to provide a known in advance list of
possibles string values.

So for now, limitations are:

- Can only pass a string of flags, separated by space. Formatting
  issues will result in an error
- The list of possible flags in hardcoded (there is two of them, -U
  and --ignore-all-exports). Future update may add more.

# Discussion

That's the WORSE pull request I hope I'll do in my life, this is ugly, but I cannot see any other solution with the current bazel API.

This should provide a temporary fix for #425.

### DIFF
--- a/haskell/haddock.bzl
+++ b/haskell/haddock.bzl
@@ -53,6 +53,10 @@ def _haskell_doc_aspect_impl(target, ctx):
         "--title={0}".format(package_id),
         "--hyperlinked-source",
     ])
+    if ctx.attr.flags:
+        # XXX: hack: bazel API constraints flags to be a string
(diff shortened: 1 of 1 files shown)


guibou:
Closed because this solution is ugly.

PR #445, opened 2018-09-25
Refactor duplicated code to expose packages

### LINKED ISSUE #406: Refactoring: avoid duplication in package argument code
We need to pass a bunch of package (`-hide-all-packages -package-id foo` etc) related arguments to GHC every time we call it, be it for compiling or for linking. Currently, compiling, linking binaries and linking libraries all have their own snippets of code to do this.

Duplication in this kind of code is best avoided. It turns out that #395, where the symptoms were an obscure link error, was caused by the linking arguments for packages being slightly different from the arguments used during compilation.

guibou:
This is a blind refactor, it mostly moves every `-package` duplicated code in the `expose_packages` functions.

There is a special case for `lib_info` which is sometimes `None`, sometimes has a value. I wonder if the two cases where it is fixed to `None` are not "copy/paste" errors and if they should be fixed.

It closes #406

### DIFF
--- a/haskell/private/packages.bzl
+++ b/haskell/private/packages.bzl
@@ -1,0 +1,56 @@
+"""Package list handling"""
+
+load(":private/set.bzl", "set")
+
+def expose_packages(build_info, lib_info, use_direct, use_my_pkg_id, custom_package_caches):
+    """
+    Returns the list of command line argument which should be passed
+    to GHC in order to enable haskell packages.
+
+    build_info: is common to all builds
+
+    All the other arguments are not understood well:
+
+    lib_info: only used for repl and linter
+    use_direct: only used for one specific task in compile.bzl
+    use_my_pkg_id: only used for one specific task in compile.bzl
+    custom_package_caches: override the package_caches of build_info, used only by the repl
+    """
+
+    args = [
+        # In compile.bzl, we pass this just before all -package-id
+        # arguments. Not doing so leads to bizarre compile-time failures.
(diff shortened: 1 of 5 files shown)


guibou:
It also closes #406

guibou:
I'm investigating the CI failure.

Profpatsch:
Huh, the CI failures seem unrelated.

guibou:
@Profpatsch Actually it was ;) I missed a small detail on the code, sometimes it uses `prebuilt_dependencies` and sometimes it uses `direct_prebuilt_dependencies`. It is fixed now.

Note: this fix also reorder the different loop. On 5 occurrences of theses loop, we manage to write 3 different orders ;)

There is a lot of `XXX` in the new refactored code, mostly because I blindly refactored without any understanding of what I'm doing (This should be harmless).

haskell/private/packages.bzl
```diff
@@ -0,0 +1,47 @@
...
+        # https://github.com/tweag/rules_haskell/issues/395.
+        "-hide-all-packages",
+    ]
+
+    # Expose all prebuilt dependencies
+
+    # We have to remember to specify all (transitive) wired-in
+    # dependencies or we can't find objects for linking
+
+    # XXX This should be really dep_info.direct_prebuilt_deps, but since we
```
philderbeast: The pattern elsewhere with the comment tokens is `XXX:`.

guibou: Good catch, will be corrected on nix iteration.

PR #446, opened 2018-09-25
CPP macro are only generated when 'version' attribute is specified.

### LINKED ISSUE #414: CC Macro name conflict when two packages have the same name
580ea82b4eda211c252bd60f4a5f4333658bc5c0 is a unit test reproducing the issue

# Problem statment

The haskell package name of an `haskell_library` is based on the bazel rule name, so `//tests:bytestring` and `//tests/cpp_macro_conflict:bytestring` have the same name `bytestring`.

When enabling haskell extension `CPP`, ghc is exporting macros `VERSION_pkgname` (e.g. `VERSION_bytestring`), this results in macro name conflicts when a `haskell_library` or `haskell_binary` have two `deps` with the same bazel name.

Using code from 580ea82b4eda211c252bd60f4a5f4333658bc5c0, I got this error when building with `bazel build //tests/cpp_macro_conflict/...`

```
/run/user/1000/ghc8953_0/ghc_2.h:13:0: error:
     error: "VERSION_bytestring" redefined [-Werror]
     #define VERSION_bytestring ""
     
   |
13 | #define VERSION_bytestring ""
   | ^

/run/user/1000/ghc8953_0/ghc_2.h:7:0: error:
     note: this is the location of the previous definition
     #define VERSION_bytestring "0.10.8.2"
     
  |
7 | #define VERSION_bytestring "0.10.8.2"
  | ^

/run/user/1000/ghc8953_0/ghc_2.h:14:0: error:
     error: "MIN_VERSION_bytestring" redefined [-Werror]
     #define MIN_VERSION_bytestring(major1,major2,minor) (\
     
   |
14 | #define MIN_VERSION_bytestring(major1,major2,minor) (\
   | ^

/run/user/1000/ghc8953_0/ghc_2.h:8:0: error:
     note: this is the location of the previous definition
     #define MIN_VERSION_bytestring(major1,major2,minor) (\
     
  |
8 | #define MIN_VERSION_bytestring(major1,major2,minor) (\
  | ^
```

Note: this is an error due to `-Warn`, but usually that's just annoying warnings.

One *simple* workaround is to tell user to avoid package name conflict when using `CPP`.

# Discussion

Version name is useful for library from the outside world (i.e. hackage) because they may implement different behaviors depending on some version of their dependencies, which are also coming from the outside world. These external libraries can still be built inside our bazel repository (i.e. using Hazel). These external libraries depends on these macro and on the correct package name. For them we need to export the version macro and keep the official package name. External libraries have a version number (and hazel exposes it using `version` attribute).

Libraries defined in a bazel repository may not be concerned by version number, that the point of  a mega repo, you control all your dependencies and everything must be coherent inside the mega repo. Hence users won't need these macros for local libraries and we can disable them.

The undocumented option `-fno-macro-version` (https://ghc.haskell.org/trac/ghc/ticket/11763) works for all the macros during one build operation. This does not allow a per-package choice. For example, if a rule depends on both the "hackage" `bytestring` and a local `bytestring`, we cannot enable macro generation for the former and not for the later.

## Design proposal

Each package controls if its macro version will be exported when the package is used as a dependency. Using for example an `export_macros` attribute. The exported macro depends on the `version` attribute.

Users will still have to change their naming scheme but only if they want to export the version macro of two packages with the same name.

For example:

```python
# This is the hackage "bytestring" library
haskell_library(
   name = "bytestring",
   version = "1.2.3",
   export_macros = True, # can be deduced from the presence of `version`
)

# This is a local library
haskell_library(
   name = "foo",
   export_macros = False, # can be deduced from the absence of `version` 
)

haskell_library(
   name = "bar",
   deps = [":bytestring", ":foo"],
)
```

Here, `bar` will be built with version macro for `bytestring` but not for `foo`.

### Implementation details

We'll have to use `-fno-macro-version` for all build and then manually export the needed macros. 

We'll have to manually replicate the macro export mechanisms, AFAIK there is no flag to tell GHC to export the version macro per package.

## Alternative design 0

We can tell user that, if they want to use CPP macro, they need to avoid package name conflict. They will need to change their naming scheme, but there is no work from our side (except documentation).

Simple, but forces the user to change its naming scheme.

## Design 1: disable all macros on a rule

This is the solution with the simplest implementation, using `-fno-macro-version`, we can disable all the macro generation during the rule execution. For example:

```python
# This is the hackage "bytestring" library, it needs the version macro to build
haskell_library(
   name = "bytestring",
   deps = [":text", ":lens"], # others dependencies from hackage
   enable_version_macro = True,
)

# This is a local library which does not need versino macro to build
haskell_library(
   name = "foo",
   enable_version_macro = False
)
```

Here `bytestring` will be built with the macro version for `lens` and `text`. However `foo` will be built without any macro version.

I don't like this approach, because disabling all the macro for a build changes nothing compared to the current status of rules_haskell, except that it will disable the warnings.

guibou:
This fix #414.

In #414, we observed that CPP version macro generated by GHC can leads to conflict if two bazel targets have the same name, hence the same GHC package name. For example `//foo:bar` and `//frob:bar` will have package name `bar` and will generated `VERSION_bar` macros.

This PR deactivate macro generation if rule does not have a `version` attribute. see #414 for discussion.

### DIFF
--- a/haskell/private/packages.bzl
+++ b/haskell/private/packages.bzl
@@ -2,12 +2,13 @@
 
 load(":private/set.bzl", "set")
 
-def expose_packages(build_info, lib_info, use_direct, use_my_pkg_id, custom_package_caches):
+def expose_packages(build_info, lib_info, use_direct, use_my_pkg_id, custom_package_caches, version):
     """
     Returns the list of command line argument which should be passed
     to GHC in order to enable haskell packages.
 
     build_info: is common to all builds
+    version: if the rule contains a version, we will export the CPP version macro
 
     All the other arguments are not understood well:
 
@@ -16,7 +17,6 @@ def expose_packages(build_info, lib_info, use_direct, use_my_pkg_id, custom_pack
     use_my_pkg_id: only used for one specific task in compile.bzl
     custom_package_caches: override the package_caches of build_info, used only by the repl
     """
-
     args = [
         # In compile.bzl, we pass this just before all -package-id
         # arguments. Not doing so leads to bizarre compile-time failures.
@@ -26,11 +26,21 @@ def expose_packages(build_info, lib_info, use_direct, use_my_pkg_id, custom_pack
         "-hide-all-packages",
     ]
 
-    # Expose all prebuilt dependencies
+    has_version = version != None and version != ""
 
+    if not has_version:
+        args.extend([
+            # Macro version are disabled for all packages by default
+            # and enabled for package with version
+            # see https://github.com/tweag/rules_haskell/issues/414
+            "-fno-version-macros",
+        ])
+
+    # Expose all prebuilt dependencies
+    #
     # We have to remember to specify all (transitive) wired-in
     # dependencies or we can't find objects for linking
-
+    #
     # XXX: This should be really dep_info.direct_prebuilt_deps, but since we
     # cannot add prebuilt_dependencies to the "depends" field on package
     # registration (see a comment there), we have to pass all transitive
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -119,6 +119,9 @@ def _mk_binary_rule(**kwargs):
                 allow_single_file = FileType([".hs", ".hsc", ".lhs"]),
                 doc = "File containing `Main` module (deprecated).",
             ),
+            version = attr.string(
+                doc = "Executable version. If this is specified, CPP version macros will be generated for this build.",
+            ),
             _dummy_static_lib = attr.label(
                 default = Label("@io_tweag_rules_haskell//haskell:dummy_static_lib"),
                 allow_single_file = True,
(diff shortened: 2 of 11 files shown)


haskell/private/packages.bzl
```diff
@@ -26,10 +26,20 @@ def expose_packages(build_info, lib_info, use_direct, use_my_pkg_id, custom_pack
...
+
+    if not has_version:
+        args.extend([
+            # Macro version are disabled for all packages by default
+            # and enabled for package with version
+            # see https://github.com/tweag/rules_haskell/issues/414
+            "-fno-version-macros",
+        ])
+
+        # Expose all prebuilt dependencies
```
guibou: This should not be indented, that's `buildifier` fault ;(

guibou:
#445 was merged, and this PR works perfectly on a client huge codebase.

guibou:
@mrkkrp @Profpatsch @mboes ping?

Profpatsch:
I can’t say anything about the validity of the change per se, but the code cooks sensible.
Some function calls are getting very long, I suggest we split them to multi-line.

@regnat does this interferes with your changes in #442?

PR #448, opened 2018-10-03
Support building mostly-static binaries

### LINKED ISSUE #378: Mostly static binaries
We already support building static libraries. In fact we already know how to static binaries as well: when profiling is enabled we use static linking because profiling and dynamic linking don't mix. This ticket is about exposing to the user the functionality, so that users can *force* static linking.

This ticket is about static linking in the sense of GHC: in Bazel terms, this corresponds to "mostly static" linking. That is, Haskell code is statically linked, but system libraries (like libgmp, glibc, libpthread etc) are still dynamically linked.

cc @lunaris

lunaris:
Closes #378.

* Adds a `linkstatic` attribute to the `haskell_binary` rule (and `haskell_test` through use of `_mk_binary_rule` -- but I think this is OK?).
* `linkstatic` defaults to `False`, which produces the behaviour shown currently, in which all code is dynamically linked (unless profiling is enabled).
* When `linkstatic` is set to `True`, GHC will produce _mostly-static_ binaries (in which all Haskell code is statically linked where possible but system and foreign libraries are still dynamically linked).
* While not matching other attributes in `snake_case`, `linkstatic` is the attribute used in e.g. the C/C++ rules shipped with Bazel (and also corresponds to _mostly-static_ linking in those cases), so it was chosen here for consistency.

### DIFF
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -107,6 +107,10 @@ def _mk_binary_rule(**kwargs):
         executable = True,
         attrs = dict(
             _haskell_common_attrs,
+            linkstatic = attr.bool(
+                default = False,
+                doc = "If enabled, this option tells GHC to link statically whenever possible. While this captures all Haskell code built, some system libraries may still be linked dynamically, as are libraries for which there is no static library. So the resulting executable will still be dynamically linked, hence only mostly static.",
+            ),
             generate_so = attr.bool(
                 default = False,
                 doc = "Whether to generate also a .so version of executable.",
--- a/haskell/private/actions/link.bzl
+++ b/haskell/private/actions/link.bzl
@@ -122,14 +122,6 @@ def link_binary(
     if not linkstatic:
         if hs.toolchain.is_darwin:
             args.add(["-optl-Wl,-headerpad_max_install_names"])
-
-            # Nixpkgs commit 3513034208a introduces -liconv in NIX_LDFLAGS on
-            # Darwin. We don't currently handle NIX_LDFLAGS in any special
-            # way, so a hack is to simply do what NIX_LDFLAGS is telling us we
-            # should do always when using a toolchain from Nixpkgs.
-            # TODO remove this gross hack.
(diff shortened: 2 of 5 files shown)


guibou:
I'm not sure your code is at fault. `master` is already broken on darwin with the same error and this appears after I merged one of my branch, with successful CI.

I wonder if something changed on the darwin environment to generate this error.

guibou:
@lunaris you should rebase on top of `master` which contain a fix for the darwin CI, this may fix your build.

lunaris:
Seems to be in better shape now -- @guibou are you happy to review or should I rope in others?

lunaris:
Sure; will sort that shortly. I've shied away because I can't get the tests to run locally (I believe an issue due to how Go looks for its SDKs which I discussed in another PR) but I can probably guess and push in the worst case. Thanks for the look over :+1:

guibou:
@lunaris I've done it. You can do that by clicking "Details" and then there is a button on the top right of the UI. (You can also "rerun with SSH" if you want to remotely debug, that's a great feature ;)

lunaris:
I'm not sure I have permission to re-run unfortunately, I was wondering if a message in this PR like `ci build` would do it, but that's good to know. I'll have a dig now and see if CI produces a set of Skydocs, because I'd like to verify I've documented it correctly, but if it all looks good then I'll merge.

lunaris:
*Sigh* Something fishy about `iconv` on Darwin it seems. Will see if maybe I can replicate on my Mac but I'm still struggling to get the tests running there too.

lunaris:
Yep, I'm an idiot. There is a workaround to ensure that `iconv` is linked correctly on Darwin but my changes meant that was only being applied when not statically linking; it needs to be applied regardless I think. CircleCI will hopefully confirm.

lunaris:
CI passed; good to merge?

lunaris:
Easy to revert if something goes wrong; merging.

Profpatsch:
> (I believe an issue due to how Go looks for its SDKs which I discussed in another PR)

We can fix that. Let me prepare a PR.

PR #449, opened 2018-10-03
Remove the dummy archive

guibou:
This is a *DISCUSION* PR, do not merge without discussion.

(edited: I proposed that PR because I initially thought that dummy archive where responsible for a break in CI, described in #450, however it is not the case. This PR may continue its life as a discussion for code simplification). 

I realized that I don't see the point of this dummy archive.


guibou:
The commit introducing this dummy is https://github.com/tweag/rules_haskell/commit/44cf931308b9f699e68e422dc67a2ce42c170491 and it points to a comment in `buck`: https://github.com/facebook/buck/blob/126d576d5c07ce382e447533b57794ae1a358cc2/src/com/facebook/buck/haskell/HaskellDescriptionUtils.java#L295

@Fuuzetsu do you have more informations about why you introduced this piece of code? I have no issue if I remove it, so something may have changed.

Fuuzetsu:
It was stolen (cargo-culted) from buck on like day 1 of these rules then never confirmed necessary. If the reason mentioned in buck commit is not something we care about or is just not necessary anymore, we can just drop it.

guibou:
Apparently, even `ghc 8.6` have the issue of failing if no input file are given. And yet, this PR does not break the tests.

All the linked objects are passed using parameter files, see, https://github.com/tweag/rules_haskell/blob/master/haskell/toolchain.bzl#L47 so I think that we will always have input files and hence the code that this PR is removing is useless.

Tests are passing, I also used this PR an a client big codebase without issue.

Profpatsch:
Awesome! I think we can merge, there’s not a lot to review if the CI is green.

PR #453, opened 2018-10-03
Use bazel from nixpkgs instead of brew for darwin CI

### LINKED ISSUE #451: Pin circleCI darwin bazel to a fixed version
Current circle CI darwin target use `brew` to install `bazel`, without any version pinning. Hence the build is not hermetic and can fail in `bazel` version change.

guibou:
This way, bazel version is pinned and build is more reproducible. Fix #451.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,11 +34,6 @@ jobs:
           command: |
             curl https://nixos.org/nix/install | sh
       - run:
-          name: System dependencies
-          command: |
-            brew update
-            brew install bazel
-      - run:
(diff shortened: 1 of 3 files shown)


Profpatsch:
I finally got around to updating the pinned nixpkgs version: https://github.com/tweag/rules_haskell/pull/456
So after that is merged, this here should work as well.

mboes:
Build failing. @guibou or @Profpatsch ?

guibou:
I updated the PR, we'll see if the CI pass.

PR #456, opened 2018-10-15
Update nixpkgs version & fix compiler version check

Profpatsch:
Moves the nixpkgs version to something recent, to incorporate various upstream improvements.
We also get bazel 0.17.1 and GHC 8.4.3.

@judah the new test for static builds fails on my machine (let’s see what the CI says). Maybe something changed with how GHC static linking works?

@mboes I *think* we also need to update `./start`, maybe that can be automated a bit.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -15,25 +15,12 @@ load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl",
     "nixpkgs_package",
 )
 
-nixpkgs_git_repository(
-    name = "nixpkgs",
-    # To make protobuf support work we need packages such as
-    # lens-labels_0_2_0_0 to be available in nixpkgs. This means we need to
-    # use a version of nixpkgs that is newer than 18.03.
-
-    # You need to be in an environment that is on the same commit
-    # as the one below. Use `nix-shell shell.nix`.
-
-    # Keep this value in sync with `nixpkgs.nix`
-    revision = "9a787af6bc75a19ac9f02077ade58ddc248e674a",
-    # TODO Using a fork with Bazel v0.15. Switch to mainline once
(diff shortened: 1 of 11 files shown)


WORKSPACE
```diff
@@ -7,32 +7,20 @@ http_archive(
...
-    revision = "9a787af6bc75a19ac9f02077ade58ddc248e674a",
-    # TODO Using a fork with Bazel v0.15. Switch to mainline once
-    # https://github.com/NixOS/nixpkgs/pull/42735 merged.
-    remote = "https://github.com/mboes/nixpkgs",
-)
-
 nixpkgs_package(
     name = "ghc",
-    repository = "@nixpkgs",
+    # rules_nixpkgs assumes we want to read from `<nixpkgs>` implicitely
```
philderbeast: Typo.
```suggestion
    # rules_nixpkgs assumes we want to read from `<nixpkgs>` implicitly
```

tests/haddock/libC.nix
```diff
@@ -17,7 +17,7 @@ let
   cabal = pkgs.writeText "libc.cabal" ''
     name: libc
     version: 0.1.0.0
-    build-type: simple
+    build-type: Simple
```
mrkkrp: Huh?

Profpatsch: My mistake I think.
Now, newer cabal seems to think it matters?

WORKSPACE
Profpatsch: Haha, I can never remember how that is written.

tests/haddock/libC.nix
mrkkrp: I think it doesn't care, this just looks like noise in the diff :)

mrkkrp: Ah OK, then it's fine.

PR #457, opened 2018-10-16
Use golang from nixpkgs

Profpatsch:
Shares some commits with https://github.com/tweag/rules_haskell/pull/456, so it’s best if that one is merged first.

Last commit is the interesting part.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -155,29 +155,43 @@ skydoc_repositories()
 
 # For buildifier
 
-# XXX Need a patched version of rules_go to workaround warnings fixed
-# by https://github.com/NixOS/nixpkgs/pull/28029 on NixOS. Revert to
-# official release once fix hits Nixpkgs master.
 http_archive(
     name = "io_bazel_rules_go",
-    strip_prefix = "rules_go-6a2b1f780b475a75a7baae5b441635c566f0ed8a",
-    urls = ["https://github.com/mboes/rules_go/archive/6a2b1f780b475a75a7baae5b441635c566f0ed8a.tar.gz"],
+    strip_prefix = "rules_go-0.15.1",
+    urls = ["https://github.com/bazelbuild/rules_go/archive/0.15.1.tar.gz"],
+    sha256 = "168330e27e30dc1344cc49dfe7383e4566f67d555365ebd494d1ce8f48812f3e",
 )
 
 http_archive(
     name = "com_github_bazelbuild_buildtools",
-    strip_prefix = "buildtools-588d90030bc8054b550967aa45a8a8d170deba0b",
-    urls = ["https://github.com/bazelbuild/buildtools/archive/588d90030bc8054b550967aa45a8a8d170deba0b.tar.gz"],
+    strip_prefix = "buildtools-0.17.2",
+    urls = ["https://github.com/bazelbuild/buildtools/archive/0.17.2.tar.gz"],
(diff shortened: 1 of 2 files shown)


WORKSPACE
```diff
@@ -7,32 +7,20 @@ http_archive(
...
-    # https://github.com/NixOS/nixpkgs/pull/42735 merged.
-    remote = "https://github.com/mboes/nixpkgs",
-)
-
 nixpkgs_package(
     name = "ghc",
-    repository = "@nixpkgs",
+    # rules_nixpkgs assumes we want to read from `<nixpkgs>` implicitely
+    # if `repository` is not set, but our nix_file uses `./nixpkgs/`.
+    # TODO(Profpatsch)
```
lunaris: Typo here I think -- should be `implicitly`.

lunaris:
Seems Netlify still isn't happy; anything to be done?

Profpatsch:
> Seems Netlify still isn't happy; anything to be done?

Yep, pulling `go` from nixpkgs introduces nix as indirect dependency to bazelify, which is run by the netlify build. As you can see in the commit, I’m trying to install nix locally, though it’s quite a mess. Sorry for the notification spam, I have to force-push a few times to test building with netlify.

We could leave out the buildify check, though it would be quite nice to be able to install nix in an arbitrary directory (without access to `/nix`), that’s why I am investigating.

skylark/go_wrap_sdk_unless_env.bzl
```diff
@@ -0,0 +1,37 @@
...
+        "root_file": attr.label()
+    }
+)
+
+def go_wrap_sdk_unless_env(name, **kwargs):
+    """Same as go_wrap_sdk, but checks whether the environment
+    variable `DONT_WRAP_GO_SDK` is set, otherwise uses the default
+    toolchain provided by rules_go."""
+    _go_wrap_sdk_unless_env(name = name, **kwargs)
+    print("hi")
```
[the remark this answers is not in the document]
Profpatsch: cc @guibou can you take a look? I don’t understand.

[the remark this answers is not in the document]
guibou: `go_wrap_sdk_unless_env` is used by rule `go_sdk` which is never used because you removed `go_rules_dependencies()` and `go_register_toolchains()`.

WORKSPACE
```diff
@@ -185,13 +187,12 @@ nixpkgs_package(
 # Use the go distribution from nixpkgs instead of letting rules_go
 # download some arbitrary binary distribution. This fixes buildifier
 # on both NixOS and Darwin.
-go_wrap_sdk(
+# If the environment variable DONT_WRAP_GO_SDK is set, it will
```
guibou: I rather prefer a positive variable name, such as `WRAP_GO_SDK` which default to one, instead of the inverted logic.

WORKSPACE
```diff
@@ -185,13 +187,12 @@ nixpkgs_package(
...
+# instead use the normal rules_go mechanism and download a go sdk.
+go_wrap_sdk_unless_env(
     name = "go_sdk",
     # root_file should point to a file in the go distribution folder,
     # which is `share/go` in the nixpkgs package.
     root_file = "@go//:share/go/README.md",
 )
 
-go_rules_dependencies()
```
guibou: Should not delete those ones

Profpatsch:
After a bit of discussion, we agreed that it’s not worth the extra workaround just to support `netlify`. Best to keep it implicit by using `host` for now, even though that might lead to reproducibility problems.

PR #467, opened 2018-10-26
Document using Nix to expose prebuilt Cabal packages

mboes:
This includes the following two new sections:

- Picking a compiler
- Building code with Hackage dependencies (using Nix)

### DIFF
--- a/docs/haskell-use-cases.rst
+++ b/docs/haskell-use-cases.rst
@@ -150,13 +150,17 @@ construct a compiler with all the packages you depend on in scope::
     text
   ])
 
+Each package mentioned in ``ghc.nix`` can then be imported using
+`haskell_import`_ in ``BUILD`` files.
+
+.. _haskell_import: http://api.haskell.build/haskell/haskell.html#haskell_import
+
 Building code with Hackage dependencies (using Hazel)
 -----------------------------------------------------
 
 .. todo
 
    Explain how to use Hazel instead of Nix
-
 
 Generating API documentation
 ----------------------------
--- a/docs/haskell-use-cases.rst
+++ b/docs/haskell-use-cases.rst
@@ -3,6 +3,64 @@
 Common Haskell Build Use Cases
 ==============================
 
+Picking a compiler
+------------------
+
+Unlike Bazel's native C++ rules, rules_haskell does not auto-detect
+a Haskell compiler toolchain from the environment. This is by design.
+We require that you declare a compiler to use in your ``WORKSPACE``
+file.
+
+There are two common sources for a compiler. One is to use the
+official binary distributions from `haskell.org`_. This is done using
+the `ghc_bindist`_ rule.
+
+The compiler can also be pulled from Nixpkgs_, a set of package
+definitions for the `Nix package manager`_. Pulling the compiler from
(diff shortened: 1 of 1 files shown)


docs/haskell-use-cases.rst
```diff
@@ -3,6 +3,64 @@
...
+the `ghc_bindist`_ rule.
+
+The compiler can also be pulled from Nixpkgs_, a set of package
+definitions for the `Nix package manager`_. Pulling the compiler from
+Nixpkgs makes the build more hermetic, because the transitive closure
+of the compiler and all its dependencies is precisely defined in the
+``WORKSPACE`` file. Use `rules_nixpkgs`_ to do so (where ``X.Y.Z``
+stands for any recent release)::
+
+  http_archive(
```
regnat: We could avoid the repetition of the release name with something like

```python
nixpkgs_release = "X.Y.Z"
http_archive(
    strip_prefix = "rules_nixpkgs-{}".format(nixpkgs_release),
    urls = ["https://github.com/tweag/rules_nixpkgs/archive/v{}.tar.gz".format(nixpkgs_release)],
)

Comment (Profpatsch) [reply docs/haskell-use-cases.rst]: For the example it’s easier to see with the dummy, it acts like a “meta”-variable already, non?

File: docs/haskell-use-cases.rst [hunk 2]
```diff
@@ -3,6 +3,64 @@
...
+  )
+
+  nixpkgs_git_repository(
+      name = "nixpkgs",
+      revision = "18.09", # Any tag or commit hash
+  )
+
+  nixpkgs_package(
+      name = "ghc",
+      repositories = { "nixkgs": "@nixpkgs//:default.nix" }
```
Profpatsch: typo `nixkgs`

docs/haskell-use-cases.rst
```diff
@@ -38,5 +96,63 @@ This works for any ``haskell_binary`` or ``haskell_library`` target.
...
+detailed rationale.
+
+To use Nixpkgs in Bazel, we need `rules_nixpkgs`_. See `Picking
+a compiler`_ for how to import Nixpkgs rules into your workspace and
+how to use a compiler from Nixpkgs. To use Cabal packages from
+Nixpkgs, replace the compiler definition with the following::
+
+  nixpkgs_package(
+      name = "ghc",
+      repositories = { "nixkgs": "@nixpkgs//:default.nix" }
```
Profpatsch: same typo

docs/haskell-use-cases.rst
mboes: @Profpatsch agreed.

mboes:
@regnat but `haskell_import` _is_ covered. Just not as part of this PR, because it's already in `master` (see the tutorial).

regnat:
> haskell_import is covered. Just not as part of this PR, because it's already in master

My bad, I didn't check long enough. Though I think it would be worth mentioning it again here because by just reading the use-case I have the impression that there isn't anything more to do and I can either directly refer to them in my `BUILD` files (but how?) or they will be implicitly added as a dependency to all my packages. That could be just one sentence such as

```rst
These packages may now be imported into bazel with the `haskell_import`_ macro.
```

mboes:
@regnat PTAL

PR #468, opened 2018-10-27
docs: expand Bazel rationale

mboes:
This expansion of the "Rationale" section explains why key Bazel
features matter in practice.

### DIFF
--- a/docs/why-bazel.rst
+++ b/docs/why-bazel.rst
@@ -54,6 +54,38 @@ build caching, test result caching or distributed builds for faster
 build and test times. Those features start to matter for larger
 projects, and become essential for very large monorepos_.
 
+Why exactly do these features matter?
+
+* **Hermetic builds** are builds that do not take any part of the
+  host's system configuration (set of installed system libraries and
+  their versions, content of ``/etc``, OS version, etc) as an input.
+  If all build actions are deterministic, hermeticity guarantees that
+  builds are reproducible anywhere, anytime. More developers on
+  a project means more subtly different system configurations to cope
+  with. The more system configurations, the more likely that the build
+  will fail in one of these configurations but not in others... Unless
+  the build is completely hermetic.
+* **Sandboxing build actions** guarantees that all inputs to all build
+  actions are properly declared. This helps prevent build system
+  correctness bugs, which are surprisingly and exceedingly common in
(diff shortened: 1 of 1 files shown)


johnynek:
As a lurker here, I think this is really great.

People without experience in large code bases usually have a hard time believing this explanation in my experience but it is still well written in my opinion.

mboes:
Thanks for the feedback. :)

docs/why-bazel.rst
```diff
@@ -54,6 +54,38 @@ build caching, test result caching or distributed builds for faster
...
+  the build is completely hermetic.
+* **Sandboxing build actions** guarantees that all inputs to all build
+  actions are properly declared. This helps prevent build system
+  correctness bugs, which are surprisingly and exceedingly common in
+  most non-sandboxing build systems, especially as the build system
+  becomes more complex. When a build system *might* be incorrect,
+  users regularly have to wipe the entire build cache to work around
+  issues. As the codebase becomes very large, rebuilding from scratch
+  can cost a lot of CPU time.
+* **Distributed build caches** make building the code from a fresh
```
regnat: I fear that a lot of people might consider that the two first points are nice-to-have features, but not something ground-breaking at all − so a lot of trouble for a minor benefit. It would be nice to also highlight that regardless of their intrinsic value they are a prerequisite for the three last ones

docs/why-bazel.rst
```diff
@@ -54,6 +54,38 @@ build caching, test result caching or distributed builds for faster
...
+  checkout trivially fast. Continuous integration populates the build
+  cache at every branch push, so that building all artifacts from
+  fresh checkouts seldom needs to actually build anything at all
+  locally. In the common case, builds become network-bound instead of
+  CPU-bound.
```
Profpatsch: “For fast build artifacts, a local build can be forced to cut down on network roundtrips.”

Profpatsch:
Two more possible topics:

* Lazy collection of build targets:
  The minimum transitive closure of required build steps is calculated. Even the calculation of build steps only accesses the minimum required build description files, making very large monorepos feasible (for example in combination with [Git VFS](https://github.com/Microsoft/VFSForGit).
* Separation of build artifacts and source files:
  (something about how even “modern” build systems like Shake don’t cleanly separate the location of generated build artifacts and source folders)

PR #474, opened 2018-11-04
Use Bazel 0.18 for development

mboes:
Includes switching to a newer Nixpkgs and a small compiler version
update.


nixpkgs/default.nix
```diff
@@ -1,4 +1,4 @@
 import (fetchTarball {
-   url = "https://github.com/NixOS/nixpkgs/archive/9f768ff43cece413e922f93cf861a28eaa969c41.tar.gz";
-   sha256 = "0dnfqh1hm0b3wwghqxhj358jhlprdrx9da1kd20g9635g7j0qs10";
+   url = "https://github.com/NixOS/nixpkgs/archive/ef4d78adedcfbb69392e7634bc6912c019b88bcc.tar.gz";
```
guibou: I suggest to add a comment with the date of the nixpkgs checkout.

mboes: Good idea.

Profpatsch:
I think the darwin CI instance is stuck.

guibou:
@Profpatsch yes, it was building `ghc` `8.2.2` from source ;)

PR #477, opened 2018-11-06
Tests cleanup

Profpatsch:
A bit of cleanup of our test suite.

First commit puts all C files that were the same into one target, second one fixes a problem with the test runner.


Profpatsch:
Darwin build timed out, should we merge nonetheless?

PR #483, opened 2018-11-18
Link indirect static dependencies

### LINKED ISSUE #170: Transitive C dependencies are not linked

guibou:
Fix #170.

A static archive only contains code which generate the archive, but not the code of direct or indirect dependencies.

For example, suppose we have `libA` which contains function `a` and `libB` contains function `b` which uses `a`. An executable who want to use `b` must link to `libB`. If `libB` is linked as a dynamic library, it is self contained: all symbols are contained OR available in another dynamic library, hence we only need to link to `libB`. However, if `libB` is linked as a static library, it only contains the symbols created by `libB`, no the symbols from any other dependencies (such as `libA`). In that case, we need to link to the static `libA` and `libB`.

The current version of `rules_haskell` was only linking with direct cxx dependencies, hence a few missing symbol during link.

This PR:

- Introduce a test with indirect dependencies
- Pick the list of indirect dependencies from the `CC` provider `cc.libs` and inject it to the list of dependencies used for the link.

### DIFF
--- a/haskell/private/dependencies.bzl
+++ b/haskell/private/dependencies.bzl
@@ -21,37 +21,48 @@ def _is_shared_library(f):
     """
     return f.extension in ["so", "dylib"] or f.basename.find(".so.") != -1
 
-def _mangle_solib(ctx, label, solib, preserve_name):
-    """Create a symlink to a dynamic library, with a longer name.
+def _is_static_library(f):
+    """Check if the given File is a static library.
 
-    The built-in cc_* rules don't link against a shared library
+    Args:
+      f: The File to check.
+
+    Returns:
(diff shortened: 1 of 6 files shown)


guibou:
It is a reopening of #394.

PR #487, opened 2018-11-20
Support linkstatic attribute in libraries

mboes:
PR #448 introduced the `linkstatic` attribute for `haskell_binary()`.
This change introduces `linkstatic` for `haskell_binary()`. Like for
CC rules, while the attributes have the same name, they have
a different meaning. In the case of libraries, `linkstatic` means
"don't bother creating a shared library".

To support this attribute in libraries, we had to make changes in the
binary code path as well. There are two distinct notions:

* Dynamically linked code.
* Shared libraries (which are dynamically linked).

Binaries can have code built in a special way so that it can call code
in shared Haskell libraries. That is to say, the code is `dynamic` or
not. Library rules can produce (dynamically linked) shared libraries.
That is, we compile `with_shared` or not (both internal variable
names not exposed to the user).


haskell/private/haskell_impl.bzl
```diff
@@ -103,6 +105,10 @@ use the 'haskell_import' rule instead.
                 depset([c.objects_dir]),
             ]),
             compiler_flags = ctx.attr.compiler_flags,
+            # NOTE We can't have profiling and dynamic code at the
+            # same time, see:
+            # https://ghc.haskell.org/trac/ghc/ticket/15394
+            dynamic = False,
```
lunaris: Until that Trac issue is resolved, do we think it's worth having the rule print out a warning when `dynamic and with_profiling`?

haskell/private/haskell_impl.bzl
```diff
@@ -223,6 +231,10 @@ use the 'haskell_import' rule instead.
                 depset([c.objects_dir]),
             ]),
             compiler_flags = ctx.attr.compiler_flags,
+            # NOTE We can't have profiling and dynamic code at the
+            # same time, see:
+            # https://ghc.haskell.org/trac/ghc/ticket/15394
+            with_shared = False,
```
lunaris: Pedantry on my part but is there a good reason we use the term "shared" here and "dynamic" elsewhere? Appreciate we have the duality of static/dynamic to contend with but perhaps there's an opportunity to add clarity if not.

lunaris:
Minor comments on my part but I think this looks broadly good :+1:

haskell/private/haskell_impl.bzl
Profpatsch: `Shared` = whether a library is built in a way in that it can be linked dynamically
`Dynamic` = whether a library/binary links against `shared` libraries

Important distinction for the code, but the user is not exposed to the fact (there’s just `linkstatic`). I don’t know whether that’s a good thing or not, but upstream kind of dictates this API (at least if we want to share it, which we want).

tests/library-static-only/BUILD
```diff
@@ -0,0 +1,17 @@
...
+haskell_library(
+    name = "library-static-only",
+    srcs = ["Lib.hs"],
+    visibility = ["//visibility:public"],
+    deps = [
+        "@hackage//:base",
+        "@hackage//:bytestring",
+    ],
+    linkstatic = True,
+)
```
Profpatsch: We could also add a `sh_test` that checks whether the resulting library is actually linked (semi-)statically.

haskell/private/haskell_impl.bzl
mboes: I considered this. But my understanding of Bazel philosophy is that all warnings should either be an error or are useless (when you're working with large code bases you'll just be drowned in warnings making them not actionable). So the choice is between making this an error and silently proceeding. I at first made this an error, but this project's own test suite quickly made it clear that making it an error was going to be very annoying. I dunno. Maybe we should make this an error? We could use tags to filter out tests that are known to fail when compiling in debug mode.

tests/library-static-only/BUILD
mboes: We could. But I ran out of brain power when trying to figure out how to do it very simply. It *might* be worth it if you know how to verify `ldd` output from Bazel without too much fuss.

haskell/private/actions/link.bzl
```diff
@@ -55,20 +55,27 @@ def _fix_linker_paths(hs, inp, out, external_libraries):
...
+        objects_dir.basename + suffix,
         sibling = objects_dir,
     )
 
+    if dynamic:
+        ext = "dyn_o"
+    elif with_profiling:
+        ext = "p_o"
+    else:
+        ext = "o"
```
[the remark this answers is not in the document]
mboes: We can't have both (see link to GHC issue in comment). And I'm not that enums exist in Starlark.

haskell/private/haskell_impl.bzl
mboes: Yeah. I'm not sure why the Bazel devs chose to use the same attribute name for two quite different meanings. Historical reasons probably.

haskell/private/haskell_impl.bzl
mboes: So I went with a warning in the end, as you suggested. We'll see how it goes.

PR #488, opened 2018-11-22
Linkstatic shell tests

Profpatsch:
Building upon https://github.com/tweag/rules_haskell/pull/487 this adds two tests to check whether the generated ELF binaries are actually linked as we expect them to be.

Adds a primitive to read a Skylark string into a single-file target (candidate for some bazel-focused library, akin to `build-support` in nixpkgs?) and a `shell_inline_test` wrapper around `shell_test` to make small test scripts less cluttered to write and read.


Profpatsch:
Meh, `skydoc` barfs at the `provides` attribute, similar to how it [didn’t like `doc`](https://github.com/bazelbuild/skydoc/issues/59) until it was [fixed by Mathieu](https://github.com/mboes/skydoc/commit/c467c4813a195a4f5040bd551ad1141d40f1c1c5) (but that is not upstreamed yet it looks like).

I can just revert that part of the change, it’s not crucial.

haskell/c2hs.bzl
```diff
@@ -8,7 +8,7 @@ load(
     "target_unique_name",
 )
 load(
-    ":private/providers.bzl",
+    "@io_tweag_rules_haskell//haskell:private/providers.bzl",
```
mboes: Necessary?

mboes: Could you add a note inline about this at the top of `providers.bzl` about this?

Issue #489, opened 2018-11-22
Fix Bazel incompatible changes


laurentlb:
I'd like to be able to build rules_haskell with `--all_incompatible_changes`:
  `bazel test //tests:all --all_incompatible_changes`

It may not be possible to fix every issue now due to transitive dependencies, but we should try to fix the issues from this repository.

mboes:
That would be great. But how does one go about doing this? I tried pretty hard recently, but quickly got blocked by upstream dependencies not being compliant. And since I don't know of a way to tell Bazel to ignore the upstream issues and tell me about the issues in this repository, I gave up. Is there a way to do that?

laurentlb:
My PR is fixing multiple issues already. I'm iterating over multiple repositories, we should get to a relatively clean state soon.

I've noticed you have your own fork of skydoc. Did you send patches upstream or filed bugs? It's probably simpler if everyone is using the same documentation generator.

mboes:
For sure. Not too happy with forking. I think the only patch in the fork is CSS styling customizations, which in skydoc last we checked was not configurable.

Profpatsch:
What exactly is there still to be done for this repository @laurentlb? The mentioned PRs from above are already merged.

When I run `bazel test //tests:all --all_incompatible_changes`, it crashes in the builtin(!!) `bazel_tools` repository:

```
ERROR: /home/philip/.cache/bazel/_bazel_philip/cccb2d2da8416067607b8c5cff088df0/external/bazel_tools/tools/build_rules/test_rules.bzl:308:17: Traceback (most recent call last):
	File "/home/philip/.cache/bazel/_bazel_philip/cccb2d2da8416067607b8c5cff088df0/external/bazel_tools/tools/build_rules/test_rules.bzl", line 306
		rule(attrs = {"file": attr.label(mand...))}, <3 more arguments>)
	File "/home/philip/.cache/bazel/_bazel_philip/cccb2d2da8416067607b8c5cff088df0/external/bazel_tools/tools/build_rules/test_rules.bzl", line 308, in rule
		attr.label(mandatory = True, allow_files = Tr..., ...)
'single_file' is no longer supported. use allow_single_file instead. You can use --incompatible_disable_deprecated_attr_params to temporarily disable this check.
ERROR: error loading package 'tests': Extension file 'tools/build_rules/test_rules.bzl' has errors
```

Is there a way to ignore that?

laurentlb:
[...]

PR #490, opened 2018-11-22
Cleanups for compatibility with future Bazel releases

laurentlb:
- Update dependencies
- Stop using Skylib lib.bzl (use more specific files)
- Avoid dictionary concatenation

Progress towards #489


haskell/nix.bzl
```diff
@@ -192,9 +191,9 @@ def haskell_nixpkgs_packageset(name, base_attribute_path, repositories = {}, **k
       `@hackage//:{your-package-name}`
     """
 
-    repositories = {
-        "bazel_haskell_wrapper": "@io_tweag_rules_haskell//haskell:nix/default.nix",
-    } + repositories
+    repositories = dict(repositories)
+    repositories.setdefault("bazel_haskell_wrapper", "@io_tweag_rules_haskell//haskell:nix/default.nix")
```
guibou: I had never thought about this usage of `setdefault` and I find it rather difficult to read. Why not:

```
{ "bazel_haskell_wrapper": "@io_tweag_rules_haskell//haskell:nix/default.nix"}.update(repositories)
￼```

Comment (laurentlb) [reply haskell/nix.bzl]: `dict.update` doesn't return any value.
I've updated the code to use the skylib function `dicts.add`

**Comment by laurentlb** [comment]
The failure on Linux is:
```
error cloning protobuf:
+ cd /root/.cache/bazel/_bazel_root/d00c9458fa4b9d008dbcdeefe3e0fc7c/external
+ rm -rf /root/.cache/bazel/_bazel_root/d00c9458fa4b9d008dbcdeefe3e0fc7c/external/protobuf /root/.cache/bazel/_bazel_root/d00c9458fa4b9d008dbcdeefe3e0fc7c/external/protobuf
+ git clone https://github.com/protocolbuffers/protobuf.git /root/.cache/bazel/_bazel_root/d00c9458fa4b9d008dbcdeefe3e0fc7c/external/protobuf
bash: line 5: git: command not found
```

Is it an option to install `git` on the CI?

Profpatsch:
> Is it an option to install `git` on the CI?

Sure, let me push the change.

Profpatsch:
This should do the trick.

laurentlb:
Thanks, tests pass now!

haskell/nix.bzl
guibou: Indeed about `dict.update`, I wrote this too quickly ;) Good solution with `dicts.add`.

PR #495, opened 2018-11-30
Add a bindist target for Ghc 8.6.2

aspiwack:
I skipped 8.6.1 altogether because it is not recommended to use anyway. I didn't make 8.6.2 the default because it's still very fresh.

I should mention that I haven't tested that the darwin download is correct, I just downloaded it, and used its `sha256sum` in the hash field.

Do I need to sign the contribution agreement or do I get to piggyback on a general Tweag one?

### DIFF
--- a/haskell/ghc_bindist.bzl
+++ b/haskell/ghc_bindist.bzl
@@ -1,6 +1,10 @@
 """Workspace rules (GHC binary distributions)"""
 
 _GHC_BINS = {
+    "8.6.2": {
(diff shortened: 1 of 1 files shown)


aspiwack:
I must confess: I don't understand the CI failure. Can someone provide guidance?

mboes:
Freak failure. OOM kicked in, likely due to too much concurrency.

mboes:
It's been very rare. And I think @regnat knows the fix.

PR #498, opened 2018-12-03
Add missing args to the `haskell_lint` rule

### LINKED ISSUE #390: `haskell_lint` doesn't forward the needed flags to GHC
A lot of packages require passing a number of flags to GHC, like `-I` or `-L` flags for system dependencies. While `haskell_library` and `haskell_binary` do handle that, the `haskell_lint` rule doesn't, which makes it impossible to use.

For example, trying to build the following target (on this repo) will fail with `fatal error: main_definition.h: No such file or directory` because the required `-Itests/textual-hdrs/include` isn't passed to ghc during the `lint` job:

```python
haskell_lint(
    name = "lint",
    deps = ["//tests/textual-hdrs"],
)
```

Likewise, the target `//tests/library-with-sysincludes` builds fine, but the following target will fail because of an unresolved reference to `TH_foo_closure`:

```python
haskell_lint(
    name = "lint",
    deps = ["//tests/textual-hdrs"],
)
```

regnat:
Fix #390

### DIFF
--- a/haskell/lint.bzl
+++ b/haskell/lint.bzl
@@ -14,6 +14,12 @@ load(
     "target_unique_name",
 )
 load(":private/set.bzl", "set")
+load(
+    ":private/path_utils.bzl",
+    "get_external_libs_path",
+    "get_lib_name",
+)
+load("@bazel_skylib//lib:paths.bzl", "paths")
 
 def _collect_lint_logs(deps):
     lint_logs = set.empty()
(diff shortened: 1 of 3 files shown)


haskell/lint.bzl
```diff
@@ -53,14 +59,31 @@ def _haskell_lint_aspect_impl(target, ctx):
...
-        build_info,
-        lib_info,
-        use_direct = False,
-        use_my_pkg_id = None,
-        custom_package_caches = None,
-        version = ctx.rule.attr.version,
-    ))
+    args.add(lib_info.ghc_args if lib_info != None else bin_info.ghc_args)
+
+    # External libraries.
```
mboes: This section of code looks like a duplicate of code in `link.bzl`. Any chance this can be factored out?

regnat: Parts of this code seems to be duplicated in several places through the codebase (like [here](https://github.com/tweag/rules_haskell/blob/master/haskell/private/actions/link.bzl#L222-L230), [here](https://github.com/tweag/rules_haskell/blob/master/haskell/private/actions/repl.bzl#L60-L65) or [here](https://github.com/tweag/rules_haskell/blob/master/haskell/doctest.bzl#L112-L126)), I'd like to take some time to understand what can be factored out and how − possibly taking nixpkgs as a reference since they manage to get most of the tooling to work out-of-the box without having to pass a complex command-line to each of them.

Profpatsch: I added an issue about the copypaste: https://github.com/tweag/rules_haskell/issues/690

Doesn’t mean we should add new duplicates though. :P

haskell/lint.bzl
```diff
@@ -53,14 +59,31 @@ def _haskell_lint_aspect_impl(target, ctx):
...
+            if hs.toolchain.is_darwin:
+                args.add([
+                    "-optl-l{0}".format(lib_name),
+                    "-optl-L{0}".format(paths.dirname(lib.path)),
+                ])
+            else:
+                args.add([
+                    "-l{0}".format(lib_name),
+                    "-L{0}".format(paths.dirname(lib.path)),
+                ])
```
Profpatsch: Yeah, this should definitely be factored out (if it is necessary at all). I believe the code in `link.bzl` has changed in the meantime as well. Then it would be out of date before it was even merged …

Profpatsch:
Ping, @regnat do you want to finish this?

Profpatsch:
Closing, because there is no easy way to merge anymore. @regnat will have to re-do the changes on current master.

PR #500, opened 2018-12-03
Disable clang warning on linux too

guibou:
GHC passes `-pie` to any compiler and `clang` warns about `unused
argument`.

Previous code was disabling this warning on darwin system, where clang
is the default. However the issue can also happen on linux if clang is
used by GHC.

This patch disable the warning for everyone.

### DIFF
--- a/haskell/private/actions/link.bzl
+++ b/haskell/private/actions/link.bzl
@@ -162,18 +162,16 @@ def link_binary(
         dep_info.dynamic_libraries,
     )
 
+    # XXX: Suppress a warning that Clang prints due to GHC automatically passing
+    # "-pie" or "-no-pie" to the C compiler.
+    # This is linked to https://ghc.haskell.org/trac/ghc/ticket/15319
+    args.add([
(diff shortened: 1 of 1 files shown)


guibou:
Discussion:

The real problem is more complicated and GHC should not passe `-pie` to compiler which can't understand it.

However solving the real problem is a complicated task.

This PR is a pragmatic answer because it does nothing more that what is currently done for "darwin" user. Said otherwise, it solves a problem for some linux users AND remove a special case in the code base.

The only issues I can imagine with this PR are:
- if a new compiler version stops supporting some flags we are using AND cause a regression, we won't know it. This is an issue, however I hope that the test suite will catch this.
- if, during development, we add a new wrong flag, we won't know it. However we will not observe the expected change of behavior. It may slow the development a bit, but won't cause any regression.

Profpatsch:
I wonder if it is feasible (in skylark) to create a dispatch layer for different compilers that maps the semantic meaning of option to their actual implementations?

PR #509, opened 2018-12-11
Update GHC bindists

jkachmar:
This just updates the GHC binary distribution workspace rules with download links and hashes for the GHC 8.4 and 8.6 release series; it also sets the default bindist to 8.4.4.

I haven't gotten a chance to test this locally yet (I need to roll back `bazel-0.20`), but this change was fairly mechanical.

Is such a thing desirable, or would you prefer that users maintain these rules on their own?


mboes:
Thanks!

Issue #470, opened 2018-10-29
`run foo-repl` should not pick up anything outside of the bazel project


Profpatsch:
(… apart from the user’s `.ghci` file, and even that is debatable)

@regnat had a broken package database in his home folder, and the `-repl` target’s imports got broken in the process, because ghci used the package database from the home folder.

We need to make sure our ghci scripts exclude all paths and information from outside the bazel workspace.

regnat:
In the case at hand, https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/packages.html#package-databases suggests that `-no-user-package-db` would fix it, but maybe there are other flags to pass too (`-no-global-package-db`? I'm not sure what these dbs correspond to)

mboes:
That sounds bad. Yes - we need to tell GHCi never to look in the home folder, or for that matter also insulate it from [this pesky stuff](https://github.com/haskell/cabal/issues/4542).

mboes:
Fixed in #514.

PR #514, opened 2018-12-14
Hide external dependencies from ghci

SebastianKG:
Per https://github.com/tweag/rules_haskell/issues/470, ghci is currently not correctly isolated from external depedencies. This aims to resolve that, through ghci flags used on its initialization.


Profpatsch:
This sounds like the right change to make.
You could test it by adding a `sh_test` which sets `$HOME` to a fake home folder with some broken package database. Though I’m not sure how much work it is to mock that, and whether it’s worth the effort to reproduce it.

PR #517, opened 2018-12-18
Limit the parallelism inside nix

regnat:
This is #501 reopened since it had been merged too eagerly and then reverted.

Depends on https://github.com/tweag/rules_nixpkgs/pull/57 so that we have a new nixpkgs release with the `nixopts` option available

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -7,7 +7,7 @@ haskell_repositories()
 
 http_archive(
     name = "io_tweag_rules_nixpkgs",
-    # sha256 = "e08bfff0e3413cae8549df72e3fce36f7b0e2369e864dfe41d3307ef100500f8",
+    sha256 = "4f924d839d3a5896e3e22ba1282c686b2c078fd4c98d564ec427ac83ec66302d",
     strip_prefix = "rules_nixpkgs-0.5.1",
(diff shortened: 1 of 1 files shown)


regnat:
The netlify failure was just a race condition because the right tag wasn't pushed to `rules_nixpkgs`, but I don't know how to restart it (if I even have the rights to do it).

WORKSPACE
```diff
@@ -7,9 +7,9 @@ haskell_repositories()
 
 http_archive(
     name = "io_tweag_rules_nixpkgs",
-    sha256 = "e08bfff0e3413cae8549df72e3fce36f7b0e2369e864dfe41d3307ef100500f8",
-    strip_prefix = "rules_nixpkgs-0.4.1",
-    urls = ["https://github.com/tweag/rules_nixpkgs/archive/v0.4.1.tar.gz"],
+    # sha256 = "e08bfff0e3413cae8549df72e3fce36f7b0e2369e864dfe41d3307ef100500f8",
```
Profpatsch: Would be best to re-add the new sha. bazel should display it when it fetches dependencies.

regnat: Indeed, done

Issue #428, opened 2018-09-04
`link.bzl` uses `external_dependencies` as list, but it’s a dict


Profpatsch:
`link.bzl` uses `external_libraries` as a list everywhere, not as dict.

From the skylark repl:

```
>>> d = { "key": "val" }
>>> [el for el in d]
["key"]
```

`external_libraries` is constructed thusly:
```
{
    f:
    # If the provider is CcSkylarkApiProviderHacked, then the .so
    # files come from haskell_cc_import.
    _mangle_solib(ctx, dep.label, f, CcSkylarkApiProviderHacked in dep)
    for f in dep.files.to_list()
    if _is_shared_library(f)
}
```
[in `dependencies.bzl`](https://github.com/tweag/rules_haskell/blob/master/haskell/private/dependencies.bzl#L130-L137).

Interpreting it as a list works in list comprehensions, but it will return the key, not the value. For `external_libraries` this means it uses the original, unmangled file instead of the one put through `_mangle_solib`.

`_mangle_solib` was introduced in commit [448efd1d4f4b138a1cd4ed8a18518e782cd6899f](https://github.com/ghorn/rules_haskell/commit/448efd1d4f4b138a1cd4ed8a18518e782cd6899f) to prevent name clashes.

In commit [4b5117d1b64aa87476bedb02df078cc20b316d70](https://github.com/ghorn/rules_haskell/commit/4b5117d1b64aa87476bedb02df078cc20b316d70) `external_libraries` was turned to a dict and all list uses correctly converted by using `.values()`. It was done in order to support MacOS and the keys were used in [exactly one place](https://github.com/ghorn/rules_haskell/commit/4b5117d1b64aa87476bedb02df078cc20b316d70#diff-656a15c44401b2dbd8f5bfbc145d4f30R367). However, it was found out that this wasn’t working with transitive libraries, so the one use of the keys was removed again in commit [182df778143d47dbd2dc77a08fb40074fc4e80b5](https://github.com/ghorn/rules_haskell/commit/182df778143d47dbd2dc77a08fb40074fc4e80b5), yet `external_libraries` was not converted back to a set.
In the following commits many new usages of `external_commits` did not take into account that it is a `dict`, some treating it as a list, some as a set.

Since there is not reason left to keep it as a dict, it should be converted back to a set.

Bugs that could be caused by this:
- https://github.com/tweag/rules_haskell/issues/370
- https://github.com/tweag/rules_haskell/issues/170
[...]

PR #519, opened 2018-12-18
HaskellBuildInfo: revert external_libraries to a set

Profpatsch:
In commit 4b5117d1b64aa87476bedb02df078cc20b316d70 external_libraries
was turned to a dict and all list uses correctly converted by using
.values(). It was done in order to support MacOS and the keys were
used in exactly one place. However, it was found out that this wasn’t
working with transitive libraries, so the one use of the keys was
removed again in commit 182df778143d47dbd2dc77a08fb40074fc4e80b5, yet
external_libraries was not converted back to a set. In the following
commits many new usages of external_commits did not take into account
that it is a dict, some treating it as a list, some as a set.

Fixes https://github.com/tweag/rules_haskell/issues/428

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -117,8 +117,11 @@ haskell_nixpkgs_packageset(
     name = "hackage-packages",
     base_attribute_path = "haskellPackages",
     nix_file = "//tests:ghc.nix",
+    nixopts = [
+        "-j",
+        "1",
+    ],
     repositories = {"nixpkgs": "@nixpkgs"},
-    nixopts = ["-j", "1"],
 )
 
 load("@hackage-packages//:packages.bzl", "import_packages")
--- a/haskell/doctest.bzl
+++ b/haskell/doctest.bzl
@@ -5,8 +5,8 @@ load("@bazel_skylib//lib:paths.bzl", "paths")
 load(":private/context.bzl", "haskell_context")
 load(
     ":private/path_utils.bzl",
-    "get_external_libs_path",
     "get_lib_name",
+    "make_external_libs_path",
 )
 load(":private/set.bzl", "set")
 load(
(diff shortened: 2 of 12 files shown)


Profpatsch:
Oh, great, linking on MacOS fails. I have no idea how that works, but it looks suspicious to me that it tries to load a `.so` file:

```
ERROR: /Users/distiller/project/tests/BUILD:329:1: Executing genrule //tests:run-bin-with-c-lib failed (Aborted): bash failed: error executing command /nix/store/5rnvfmysc6v1qzzsy46fb084g67blnx1-bash/bin/bash -c ... (remaining 1 argument(s) skipped)

dyld: Library not loaded: bazel-out/host/bin/tests/data/libourclibrary.so
  Referenced from: /private/var/tmp/_bazel_distiller/bedaa68a8664d1b29e96b826d058247f/sandbox/darwin-sandbox/270/execroot/io_tweag_rules_haskell/bazel-out/host/bin/tests/c-compiles/c-compiles
  Reason: image not found
sh: line 2: 51778 Abort trap: 6           bazel-out/host/bin/tests/c-compiles/c-compiles
```

cc @guibou @judah

Profpatsch:
Okay, it turned out that the unmangled library name was indeed still used by the darwin fixups, just implicitly. `for f in somedict` will iterate over the keys implicitly (what a horrible misfeature), so some call-sites introduced bugs by assuming it’s a list.

I restructured the commit so that it explicitly uses a struct with two fields instead of carrying information through the keys of a dict.

WORKSPACE
```diff
@@ -117,8 +117,11 @@ haskell_nixpkgs_packageset(
     name = "hackage-packages",
     base_attribute_path = "haskellPackages",
     nix_file = "//tests:ghc.nix",
+    nixopts = [
```
guibou: What for?

Profpatsch: Same as https://github.com/tweag/rules_haskell/pull/533#discussion_r243332618

PR #529, opened 2018-12-19
New buildifier

guibou:
Close #528 

Buildifier can now be called using `bazel run //:buildifier` and fix (for all the workspace) with `bazel run //:buildifier-fix`.

PR #533, opened 2018-12-20
CircleCI: run linters at the beginning of CI

Profpatsch:
We don’t need to build protobuf to know whether a bracket was on the
wrong line (at least I hope we don’t).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,8 @@
 version: 2
 
 jobs:
+
+  # ATTN: when you change anything here, don’t forget to copy it to the build-darwin section
   build-linux:
     docker:
       - image: nixos/nix:2.1.3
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -14,6 +14,10 @@ jobs:
             apk --no-progress update
             apk --no-progress add bash ca-certificates
             nix-channel --update
+      - run:
+          name: Bazel linter
+          command: |
+            nix-shell --pure --run "bazel test //... --build_tests_only --test_tag_filters=lint"
```
guibou: Linter had changed since #529

.circleci/config.yml
```diff
@@ -14,18 +16,23 @@ jobs:
...
+      - run:
+          name: Bazel linter
+          command: |
+            nix-shell --pure --run 'bazel run //:buildifier'
       - run:
           name: Build
           command: |
-            nix-shell --pure --run "bazel build --jobs=2 //... @haskell_zlib//... --config=ci"
-            nix-shell --pure --run "bazel build -c dbg --jobs=2 //... --config=ci"
+            nix-shell --pure --run 'bazel build --jobs=2 //... @haskell_zlib//... --config=ci'
```
guibou: Why the quotes changes?

.circleci/config.yml
```diff
@@ -14,18 +16,23 @@ jobs:
...
+            nix-shell --pure --run 'bazel build -c dbg --jobs=2 //... --config=ci'
       - run:
           name: Run tests
           # bazel does not support recursive bazel call, so we
           # cannot use bazel run here because the test runner uses
           # bazel
           command: |
-            nix-shell --pure --run "bazel build //tests:run-tests && ./bazel-bin/tests/run-tests"
+            # skip running the linter, because it was already done separately, above
```
guibou: This can be discussed, but it will optimise something like 1s on a really long process. I don't think this optimisation worse the cost of a comment line and a complicated call.

WORKSPACE
```diff
@@ -117,8 +117,11 @@ haskell_nixpkgs_packageset(
     name = "hackage-packages",
     base_attribute_path = "haskellPackages",
     nix_file = "//tests:ghc.nix",
+    nixopts = [
```
guibou: What is this?

.circleci/config.yml
Profpatsch: You are right, I’ll remove it again.

.circleci/config.yml
Profpatsch: Some used `'` and some `"`, so I unified them.

WORKSPACE
Profpatsch: The buildify run to fix CI.

Profpatsch:
@guibou I think this is ready for merge.

Issue #494, opened 2018-11-29
Doc request: How do I point to a local version of rules_haskell


aspiwack:
In the [rules haskell developer README section](https://github.com/tweag/rules_haskell/#for-rules_haskell-developers), it would be useful to add a way to point a WORKSPACE to a local rules_haskell checkout, so that changes can be tested.

PR #535, opened 2019-01-02
README: reference a local checkout of rules_haskell

Profpatsch:
Fixes https://github.com/tweag/rules_haskell/issues/494

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,36 @@ See [rules_haskell_examples][] for examples of using these rules.
 
 ## For `rules_haskell` developers
 
+### Reference a local checkout of `rules_haskell`
+
+When you develop on `rules_haskell`, you usually do it in the context
+of a different project that has `rules_haskell` as a `WORKSPACE`
+dependency, like so:
+
+```
+http_archive(
+    name = "io_tweag_rules_haskell",
+    strip_prefix = "rules_haskell-" + version,
+    sha256 = …,
+    urls = …,
+)
+```
+
+To reference a local checkout instead, use the
+[`--override_repository`][override_repository] command line option:
+   
+```
+bazel build/test/run/sync \
+  --override_repository io_tweag_rules_haskell=/path/to/checkout
+```
+   
+If you don’t want to type that every time, [temporarily add it to
+`.bazelrc`][bazelrc].
+
+[override_repository]: https://docs.bazel.build/versions/master/command-line-reference.html#flag--override_repository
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -135,8 +135,34 @@ See [rules_haskell_examples][] for examples of using these rules.
...
+
+```
+http_archive(
+    name = "io_tweag_rules_haskell",
+    strip_prefix = "rules_haskell-" + version,
+    sha256 = …,
+    urls = …,
+)
+```
+
```
guibou: Why not using the bazel flag `override_repository`?

```
--override_repository \
   io_tweag_rules_haskell=/path/to/local/checkout
```

Profpatsch: Oh, that might be even better on occasions. I’ll add a reference to that as well.

Profpatsch: We should write a blog post about the most helpful bazel command line flags …

guibou: Add the ref and I approve.

Profpatsch: Done.

README.md
```diff
@@ -135,8 +135,47 @@ See [rules_haskell_examples][] for examples of using these rules.
 
 [rules_haskell_examples]: https://github.com/tweag/rules_haskell_examples
 
+
```
mboes: Spurious whitespace change.

README.md
```diff
@@ -135,8 +135,47 @@ See [rules_haskell_examples][] for examples of using these rules.
...
+1) Use the [`--override_repository`][override_repository]
+   command line option:
+   
+   ```
+   bazel build/test/run/sync \
+     --override_repository io_tweag_rules_haskell=/path/to/checkout
+   ```
+   
+2) Edit your `WORKSPACE` to use [`local_repository`][local_repository]
```
mboes: Remove this option. This is the wrong way to do it.

mboes: But bazelrc is even more right. It's the only mechanism that allows you to override on a per user basis without fearing committing the dirty changes in the WORKSPACE by mistake. See https://docs.bazel.build/versions/master/best-practices.html#bazelrc.

I'm not a fan of overlong README's. The bar should be high as to what content goes there. Let's avoid repeating the documentation there for things are bad practices anyways.

Profpatsch: Okay, replaced it with “If you don’t want to type that every time, temporarily add it to
   if you want a more permanent setup (just don’t commit it):	`.bazelrc`.”

mboes: Add the link I pointed you to.

Profpatsch: Oh, right. Done.

Issue #503, opened 2018-12-07
`generate_so` attr not used anywhere


Profpatsch:
https://github.com/tweag/rules_haskell/search?q=generate_so&unscoped_q=generate_so

It should have been completely superseded by `linkstatic`, and should be removed in the next major version bump.

PR #539, opened 2019-01-03
Remove `generate_so` attribute, it is a no-op

Profpatsch:
The changes to `linkstatic` removed any use of `generate_so`.

Fixes https://github.com/tweag/rules_haskell/issues/503


Profpatsch:
I see there’s more stuff deprecated already, so let me prepare a PR that removes everything.

Issue #541, opened 2019-01-03
toolchain_type does not exist


iphydf:
https://github.com/tweag/rules_haskell/blob/master/haskell/toolchain.bzl#L427 references @io_tweag_rules_haskell//haskell:toolchain, which does not exist according to https://github.com/tweag/rules_haskell/blob/master/haskell/BUILD. How can I make this work? I'm building with bazel 0.21.0.

```
ERROR: /home/iphydf/code/toktok-stack/BUILD.bazel:77:1: no such target '@io_tweag_rules_haskell//haskell:toolchain': \
  target 'toolchain' not declared in package 'haskell' (did you mean \
  'toolchain.bzl'?) defined by /usr/local/home/iphydf/.cache/bazel/_bazel_iphydf/87c462f2b35b8d3dfe6f6c21308fd127/external/io_tweag_rules_haskell/haskell/BUILD \
  and referenced by '//:my-ghc'
ERROR: While resolving toolchains for target //c-toxcore/testing:Messenger_test: \
  invalid registered toolchain '//:my-ghc': Analysis failed
```

PR #544, opened 2019-01-04
toolchain_type: Actually add a toolchain_type target

Profpatsch:
It was referenced but didn’t exist, and bazel never checked it before
0.21 it looks like.

Fixes https://github.com/tweag/rules_haskell/issues/541

@iphydf, can you check whether that fixes the bug for you? I [haven’t gotten bazel 0.21 running on my system yet](https://github.com/NixOS/nixpkgs/pull/53412), with 0.20 bazel doesn’t even check the attribute.

### DIFF
--- a/haskell/BUILD
+++ b/haskell/BUILD
@@ -18,3 +18,9 @@ py_binary(
     srcs = ["private/ls_modules.py"],
     visibility = ["//visibility:public"],
 )
+
+# toolchains must have a valid toolchain_type from bazel 0.21
+toolchain_type(
+    name = "toolchain",
+    visibility = ["//visibility:public"],
+)


haskell/BUILD
```diff
@@ -18,3 +18,6 @@ py_binary(
     srcs = ["private/ls_modules.py"],
     visibility = ["//visibility:public"],
 )
+
+# toolchains must have a valid toolchain_type from bazel 0.21
+toolchain_type(name = "toolchain")
```
iphydf: I think it needs visibility public.

iphydf:
`in toolchain rule //:ghc: target '@io_tweag_rules_haskell//haskell:toolchain' is not visible from target '//:ghc'. Check the visibility declaration of the former target if you think the dependency is legitimate`

Profpatsch:
> in toolchain rule //:ghc: target '@io_tweag_rules_haskell//haskell:toolchain' is not visible from target '//:ghc'. Check the visibility declaration of the former target if you think the dependency is legitimate

Top 10 anime betrayals.

PR #547, opened 2019-01-07
debug: Add python linking debugging tools

Profpatsch:
`rules_haskell` development features a large amount of varying fun
with different dynamic linking problems. These python helpers shall
make debugging them less noisy, and more programmatic and to the
point.

We can’t put them in the `tools/` directory because of the problem
mentioned now in tools/README.md.

### DIFF
--- a/debug/linking_utils/BUILD
+++ b/debug/linking_utils/BUILD
@@ -1,0 +1,5 @@
+py_library(
+    name = "linking_utils",
+    srcs = ["ldd.py"],
+    visibility = ["//visibility:public"],
+)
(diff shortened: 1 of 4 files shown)


guibou:
@mboes I think that @Profpatsch implementation of "lddtree" is more versatile for us because it allows the filtering of stuffs such as unimportant rpath such as nix ones.

mboes:
Could you provide an example use case? I am worried that this is a lot of extra code to maintain just to get the equivalent of `lddtree | grep`. So this solution would have to be more powerful. For that matter, more so than

```
$ lddtree | sed "s/ l/- l/" > /tmp/ldd.org
$ emacs /tmp/ldd.org
```

Emacs org-mode will let you fold/unfold subtrees and filter according to whatever criteria you like.

Issue #540, opened 2019-01-03
Remove deprecated prebuilt_dependencies 


Profpatsch:
In theory, `prebuilt_dependencies` should be deprecated and thus unnecessary, but going through the usage, non-trivial parts of `rules_haskell` still depend on it:

- `HaskellHaddock` passes it as argument to `haddock_wrapper.sh`. Does that mean `haskell_import` is not yet supported by the haddock generator, @mrkkrp?
 
https://github.com/tweag/rules_haskell/blob/2b163f714dda6611d07875df236e6ba3bdf11ad0/haskell/haddock.bzl#L117-L124

- `HaskellProtoc` uses it internally

https://github.com/tweag/rules_haskell/blob/2b163f714dda6611d07875df236e6ba3bdf11ad0/haskell/protobuf.bzl#L148

Concerning `haskell_import`, I’m confused as to which version should actually be used. We have the new one in `import.bzl`, but it’s not exported as far as I can see. But since `haskell_import` should be the replacement for `prebuilt_dependencies`, we should define which one to use (or throw one out). cc @regnat

Profpatsch:
Okay, when going through the code with Théophane, we found that `prebuilt_dependencies` actually has two uses:

1. attribute in all haskell rules
2. internal mechanism to pass prebuilt stuff, also for `haskell_import`

Since we already deprecated the user-facing part, we can remove the attribute, but keep the internal usage for now.

The next step would be to finish new-style `haskell_import` and get rid of all that internal wiring and the `HaskellPrebuiltPackageInfo` provider as well, but that’s out of scope for this issue.

Issue #504, opened 2018-12-08
Start script: don't use prebuilt_dependencies


aspiwack:
Since we consider it deprecated:

> DEBUG: <bazel_cache>/external/io_tweag_rules_haskell/haskell/private/haskell_impl.bzl:176:9: The attribute 'prebuilt_dependencies' has been deprecated,
use the 'haskell_import' rule instead.

aspiwack:
(unrelated to previous message)

I was wondering how we could add a test for the start script in CI. I don't know how genrules work in Bazel (I assume the tests are all Bazel-driven), but I'm expecting something like

- Call the start script in some sandbox
- Create a `src/Test.hs` module with basically nothing in it
- Run Bazel, `tee`-ing its output
- If it Bazel fails, fail the test
- If `grep -v deprecated` fails on the recorded output, fail the test (or maybe we want to catch all `DEBUG` messages?)

mboes:
> I was wondering how we could add a test for the start script in CI. 

We already have one. And it's actually creating a problem. The `start` script uses the latest rules_haskell release. But when we update Bazel to a version that is not backwards compatible, CI breaks until we cut a new release.

aspiwack:
Ah good point. In a way, the `start` script should point at the commit which is is taken from. Could this be achievable with git substitutions, maybe? (if there is an appropriate substitution, it would work fine for distributed versions. But not so clear for CI…)

mboes:
I hadn't thought of that. It might well work!

PR #548, opened 2019-01-08
Remove deprecated prebuilt_dependencies attribute.

Profpatsch:
The user-facing parts of `prebuilt_dependencies` have been superseeded
by `haskell_import`. Changes all examples to the new API and removes
the attribute, which has been deprecated since 0.6.

Fixes https://github.com/tweag/rules_haskell/issues/504
Fixes https://github.com/tweag/rules_haskell/issues/540

### DIFF
--- a/haskell/private/dependencies.bzl
+++ b/haskell/private/dependencies.bzl
@@ -85,10 +85,10 @@ def gather_dep_info(ctx):
         static_libraries_prof = [],
         dynamic_libraries = set.empty(),
         interface_dirs = set.empty(),
-        prebuilt_dependencies = set.from_list(ctx.attr.prebuilt_dependencies),
+        prebuilt_dependencies = set.empty(),
         # a set of struct(lib, mangled_lib)
         external_libraries = set.empty(),
-        direct_prebuilt_deps = set.from_list(ctx.attr.prebuilt_dependencies),
+        direct_prebuilt_deps = set.empty(),
     )
 
     for dep in ctx.attr.deps:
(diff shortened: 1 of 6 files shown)


haskell/private/dependencies.bzl
```diff
@@ -85,10 +85,10 @@ def gather_dep_info(ctx):
         static_libraries_prof = [],
         dynamic_libraries = set.empty(),
         interface_dirs = set.empty(),
-        prebuilt_dependencies = set.from_list(ctx.attr.prebuilt_dependencies),
+        prebuilt_dependencies = set.empty(),
```
guibou: Is this field still needed?

haskell/private/dependencies.bzl
```diff
@@ -85,10 +85,10 @@ def gather_dep_info(ctx):
...
         # a set of struct(lib, mangled_lib)
         external_libraries = set.empty(),
-        direct_prebuilt_deps = set.from_list(ctx.attr.prebuilt_dependencies),
+        direct_prebuilt_deps = set.empty(),
```
guibou: Same, is this field still needed?

haskell/private/dependencies.bzl
Profpatsch: Yes, it’s used internally to carry everything added via `haskell_import`.
When https://github.com/tweag/rules_haskell/issues/440 is tackled we should be able to get rid of this field.

haskell/private/dependencies.bzl
Profpatsch: See above.

PR #567, opened 2019-01-14
Fix static linking with indirect C library

dajmaki:
When building a static haskell binary that depends on a C library via a Haskell library, the
resulting C compiler command-line as generated by GHC has the wrong order for linking
libraries. When statically linking a library A depended on B has to be specified last: "-lB -lA", 
as otherwise the linker won't see the symbols (it forgets them when loading the archives if nothing
already depends on them).

The only way to fix this order that I could think of was to specify the dependencies in the 
extra-libraries field of the package database.

This fixes part of the issue of static linking with rules_haskell, with the following caviats:
- @repl targets won't work in the presence of linkstatic = True (due to shared libraries not part of repl target depset. I'll follow up on that).
- the static library dependencies are mentioned twice on the command-line. We need the original "-l" with shared libraries.


dajmaki:
This messes with dynamic linking so it's not quite there yet.

dajmaki:
hopefully I fixed the above issue with the dynamic linking. seems to work on our bigger project up to the point of getting stuck with this issue: https://github.com/tweag/rules_haskell/pull/569. @aherrmann is now giving a go if these two combined fixes things.

PR #568, opened 2019-01-14
Drop file extension when checking for presence of GHC binaries.

mboes:
Windows binaries usually have a file extension.

### DIFF
--- a/haskell/toolchain.bzl
+++ b/haskell/toolchain.bzl
@@ -73,15 +73,14 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, params_file = None, e
     return args
 
 def _haskell_toolchain_impl(ctx):
-    for tool in _GHC_BINARIES:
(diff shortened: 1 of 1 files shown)


haskell/toolchain.bzl
```diff
@@ -73,15 +73,14 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, params_file = None, e
...
-        if tool not in [t.basename for t in ctx.files.tools]:
-            fail("Cannot find {} in {}".format(tool, ctx.attr.tools.label))
-
     # Store the binaries of interest in ghc_binaries.
     ghc_binaries = {}
-    for tool in ctx.files.tools:
-        if tool.basename in _GHC_BINARIES:
-            ghc_binaries[tool.basename] = tool
+    for tool in _GHC_BINARIES:
+        for file in ctx.files.tools:
```
guibou: What can be the size of this list? Should you precompute a set?

PR #569, opened 2019-01-15
Fix static linking with Haskell prebuilt packages

aherrmann:
Currently, static linking (`linkstatic = True`) of Haskell binary targets fails if the target has an indirect dependency on a prebuilt package imported using `haskell_import`. The reason is that `haskell_import`  has no access to the package id and the package id of prebuilt package dependencies is not inserted into the `depends:` section of the package configuration file of Haskell targets. GHC then does not add linker flags for the corresponding static libraries and liking fails on missing symbols. See https://github.com/FormationAI/hazel/issues/64 .

This PR fixes that by inserting the package id of prebuilt packages into the `depends:` section of Haskell targets.

The corresponding package id is obtained using `ghc-pkg field id` on the prebuilt package. The outcome is stored in a file and a reference to that file is stored in the `HaskellPrebuiltPackageInfo` provider. This way `ghc-pkg` only has to be executed once per prebuilt package.

The implementation of `haskell_import`, which is now a little more complex, has been moved into `haskell/private/haskell_impl.bzl` to follow the convention of other more complex rule implementations.

This change also allows to only reference direct prebuilt dependencies on linking, resolving an `XXX` comment in `haskell/private/packages.bzl`.

Finally, this PR adds a regression test.


haskell/private/actions/package.bzl
```diff
@@ -70,18 +69,39 @@ def package(hs, dep_info, interfaces_dir, interfaces_dir_prof, static_library, d
...
+        inputs = [metadata_file, exposed_modules_file, prebuilt_package_ids],
         outputs = [conf_file],
         command = """
-        cat $1 > $3
-        echo "exposed-modules: $(< $2)" >> $3
+        cat $1 > $4
+        echo "exposed-modules: $(< $2)" >> $4
+        echo -n "depends: " >> $4
+        cat $3 >> $4
+        echo >> $4
```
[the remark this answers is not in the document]
Profpatsch: I wonder whether the variables need more quoting. I’d add quotes just to be sure.

haskell/private/haskell_impl.bzl
```diff
@@ -358,11 +359,33 @@ def haskell_library_impl(ctx):
...
     else:
         package = ctx.label.name
-    return [HaskellPrebuiltPackageInfo(package = package)]
+
+    id_file = hs.actions.declare_file(target_unique_name(hs, "id"))
+    hs.actions.run_shell(
+        inputs = [hs.tools.ghc_pkg],
+        outputs = [id_file],
+        command = """
+        $1 --simple-output -v1 field $2 id > $3
```
Profpatsch: I’d add quotes around the variables here as well.

haskell/private/actions/package.bzl
```diff
@@ -96,9 +96,7 @@ def package(hs, dep_info, interfaces_dir, interfaces_dir_prof, static_library, d
         command = """
         cat $1 > $4
         echo "exposed-modules: $(< $2)" >> $4
-        echo -n "depends: " >> $4
-        cat $3 | tr "\\n" " " >> $4
-        echo >> $4
+        echo "depends: $(cat $3 | tr "\\n" " ")" >> $4
```
Profpatsch: Squash into last change?

Profpatsch: (This is on the very last commit, Github’s review feature is very bad with multi-commit PRs)

haskell/haddock.bzl
```diff
@@ -76,7 +76,7 @@ def _haskell_doc_aspect_impl(target, ctx):
```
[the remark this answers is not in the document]
aherrmann: Good point, I've added an explanation to the commit message of the relevant commit.

haskell/private/actions/package.bzl
aherrmann: > (This is on the very last commit, Github’s review feature is very bad with multi-commit PRs)

It is.

> Squash into last change?

Done

haskell/private/haskell_impl.bzl
aherrmann: Done

haskell/private/actions/package.bzl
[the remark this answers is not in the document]
aherrmann: Not strictly necessary. All those are file paths generated by rules_haskell.

aherrmann:
@Profpatsch 
> Also cc @regnat, is that change relevant for your newer haskell_import and has to be added there as well?

If you're referring to `haskell_import` implemented in `haskell/import.bzl`, then no. That one takes `package_id` as an attribute and generates a full `HaskellBuildInfo`. The `@hackage//:XYZ` targets mentioned [above](https://github.com/tweag/rules_haskell/pull/569#issuecomment-454352535) are generated that way.

PR #574, opened 2019-01-16
Migrate rules_haskell to Bazel 0.21

NinjaTrappeur:
Prerequisite of #489

# Fixed Toolchain implicit types

Toolchain-type labels should now point to an actual rule.

# Docs build tools cleanup

The host path is not forwarded anymore to the sandbox in charge of
executing the actions. This means we now need to create some specific
rules for each tool using during the build process of the documentation.

# Protolens configuration update

- Updating the protolens flags.
- Protolens is now using deepseq in its generated files, we need to add
  this library as a toolchain dependency.

### DIFF
--- a/haskell/protobuf.bzl
+++ b/haskell/protobuf.bzl
@@ -116,7 +116,7 @@ def _haskell_proto_aspect_impl(target, ctx):
         ))
 
     args.add([
-        "--haskell_out=no-reexports:" + paths.join(
+        "--haskell_out=no-runtime:" + paths.join(
             hs_files[0].root.path,
             src_prefix,
         ),
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,9 @@ jobs:
             apk --no-progress update
             apk --no-progress add bash ca-certificates
             nix-channel --update
+            # CircleCI and Nix sandboxing don't play nice. See
+            # https://discourse.nixos.org/t/nixos-on-ovh-kimsufi-cloning-builder-process-operation-not-permitted/1494/5
+            mkdir -p /etc/nix && echo "sandbox = false" > /etc/nix/nix.conf
       - run:
           name: Build
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,7 +49,9 @@ jobs:
           name: Run tests
           shell: /bin/bash -eilo pipefail
           command: |
-            nix-shell --arg docTools false --pure --run 'bazel build //tests:run-tests && ./bazel-bin/tests/run-tests'
+            # XXX 2019-01-22 Disable start script checking on Darwin
+            # due to a clash between binutils and clang.
+            nix-shell --arg docTools false --pure --run 'bazel build //tests:run-tests && ./bazel-bin/tests/run-tests --skip "/startup script/"'
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,21 +17,16 @@ jobs:
             apk --no-progress add bash ca-certificates
             nix-channel --update
       - run:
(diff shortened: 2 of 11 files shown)


NinjaTrappeur:
**NOTE**: I'm wondering if this could break some already deployed configurations.

This nixpkgs pin is fetching a new version of proto-lens. The Haskell code generated by this new version now depends on deepseq. I'm wondering if this could lead to a missing dependency error if deepseq is not already added as dependency to the toolchain.

Is there any way to add some "default" dependencies to a toolchain?

NinjaTrappeur:
There's definitively an issue with the way I handle the proto-lens cli flag migration (see https://github.com/google/proto-lens/wiki/Migration-Guide#command-line-flags ).

We need to find a way to use a different flag if proto-lens < 0.4

NinjaTrappeur:
Current status:

There is a problem with the circleci build:

```
ERROR: error loading package '': Encountered error while reading extension file 'packages.bzl': no such package '@hackage-packages//': no such package '@hackage-packages-packages-list//': Traceback (most recent call last):
	File "/root/.cache/bazel/_bazel_root/d00c9458fa4b9d008dbcdeefe3e0fc7c/external/io_tweag_rules_nixpkgs/nixpkgs/nixpkgs.bzl", line 136
		_execute_or_fail(repository_ctx, nix_build, failure_m...), <3 more arguments>)
	File "/root/.cache/bazel/_bazel_root/d00c9458fa4b9d008dbcdeefe3e0fc7c/external/io_tweag_rules_nixpkgs/nixpkgs/nixpkgs.bzl", line 299, in _execute_or_fail
		fail("\n{failure_message}\nCommand: {...))

```

Basically, bazel 0.21 fails to call rules_nixpkgs in a circle-ci container. I am slowly making progress on this. I am now pretty sure the problem comes from a wrong env value contained in repository_ctx.

TL;DR: wip, don't merge this yet.

haskell/protobuf.bzl
```diff
@@ -285,6 +297,13 @@ _protobuf_toolchain = rule(
             mandatory = True,
             doc = "proto-lens-protoc plugin for protoc",
         ),
+        # We need to explicitely get the proto-lens-protoc version
+        # to handle some cli flag changes.
+        # TODO: remove when we can assume that proto-lens-protoc > 0.4
+        "proto_lens_protoc_version": attr.string(
+            doc = "proto-lens-protoc plugin version",
+            default = "0.3",
```
Profpatsch: Do we need the default here when we also pass one down below?

haskell/protobuf.bzl
```diff
@@ -300,6 +319,7 @@ def haskell_proto_toolchain(
         deps = [],
         prebuilt_deps = [],
         protoc = Label("@com_google_protobuf//:protoc"),
+        proto_lens_protoc_version = "0.3.1.2",
```
Profpatsch: Use `0.3` from above?

NinjaTrappeur: oops, indeed. Half finished refactoring here...

NinjaTrappeur:
Current status:

- CI's almost fixed. We still have an error triggered during the tests but that should be an easy fix.
- We are waiting for the PR we made on nixpkgs to land on nixos-unstable ( http://howoldis.herokuapp.com/ ) to leverage the nix cache during the CI ran (last run took ~1.5 hour).

mboes:
Just support 0.4 from now. It's not worth spending time supporting older versions.

mboes:
>  What's the problem here? Is there any way to force-build the missing libraries on the Nixpkgs's Hydra ?

Question for @grahamc.

NinjaTrappeur:
Ahhh, the lack of cache was coming from the use of an old GHC version... I should have thought about it. Thanks @mboes !

PR #576, opened 2019-01-16
Fix repl rule with indirect cc_library deps

dajmaki:
The repl rule adds "-l<somelib>" from external_libraries, but that can contain static libraries, which
won't have the mangled .so/.dylib linked in, and ghci will then fail to find this library. This changes
repl.bzl to only add "-l"s for shared libraries (which come in via haskell_cc_import).

Interestingly this bug can be only triggered if we have a haskell_library depending on a cc_library
that in turn depends on another cc_library. With that we get the following error:
``
<command line>: user specified .o/.so/.DLL could not be loaded (libtests_Slibrary-with-cbits_Sourclibrary-indirect_libourclibrary.so: cannot open shared object file: No such file or directory)
Whilst trying to load:  (dynamic) tests_Slibrary-with-cbits_Sourclibrary-indirect_libourclibrary
``

### DIFF
--- a/tests/binary-indirect-cbits/BUILD
+++ b/tests/binary-indirect-cbits/BUILD
@@ -7,16 +7,17 @@ load(
 haskell_binary(
     name = "binary-indirect-cbits",
     srcs = ["Main.hs"],
+    linkstatic = False,
     deps = [
         "//tests/library-with-cbits",
         "@hackage//:base",
     ],
-    linkstatic = False
 )
 
 haskell_binary(
     name = "binary-indirect-cbits-partially-static",
     srcs = ["Main.hs"],
+    linkstatic = False,
     deps = [
         "//tests/library-with-cbits:library-with-cbits-static",
         "@hackage//:base",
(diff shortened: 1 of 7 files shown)


tests/binary-indirect-cbits/BUILD
```diff
@@ -7,17 +7,16 @@ load(
...
         "//tests/library-with-cbits",
         "@hackage//:base",
     ],
+    linkstatic = False
 )
 
 haskell_binary(
     name = "binary-indirect-cbits-partially-static",
     srcs = ["Main.hs"],
-    linkstatic = False,
```
Profpatsch: Can we actually set `linkstatic` to `True` here, doesn’t that change the meaning of the test?

guibou: AFAIK `True` is the default value.

dajmaki: Indeed, this shouldn't be removed. I'll add it back and rebase the PR.

PR #581, opened 2019-01-16
Use a GHC bindist on Windows

### LINKED ISSUE #559: Select a ghc_bindist as the haskell toolchain for tests on Windows
The rules_haskell test suite unconditionally uses a GHC provided by Nixpkgs. On Windows, Nixpkgs is unavailable. So we should define the toolchain in such a way that if we are on Windows, a `ghc_bindist()` rule is used instead of the current `nixpkgs_package()`. See [select] and [platforms](https://docs.bazel.build/versions/master/platforms.html) for this (see [example here](https://github.com/bazelbuild/rules_go/blob/master/go/core.rst#platform-specific-dependencies)).

### LINKED ISSUE #578: Provide a Go toolchain on Windows CI
The `WORKSPACE` specifies that the Go toolchain should be provided by the host: https://github.com/tweag/rules_haskell/blob/d2abf00e76fe055c19cdb35dc30c16478e06d072/WORKSPACE#L222-L224

This is a bit annoying since `rules_go` provides bindists for Go, which would be much simpler to use on Windows (no nix-shell there).

nmattia:
Fixes #559
Fixes #578 

This makes sure that GHC can be used on Windows:

* Add a test that invokes GHC with `--help` (no compilation)
* Ensure that a Go SDK is present on Windows (needed by some of our tools)
* Disable `glibc_locales` and `c2hs` on Windows
* Prepare and select a GHC from a bindist on Windows

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 * The evaluation of the `WORKSPACE` on Windows is successful. The `nix-build`
   calls are bypassed.
+* The evaluation of the `buildifier` on Windows is successful. The Go SDK
+  is downloaded.
 
 ### Removed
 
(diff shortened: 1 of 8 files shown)


gdeest:
You should squash the history a bit before merging.

PR #587, opened 2019-01-17
Link C libraries statically if possible on linkstatic = True binaries

### LINKED ISSUE #580: Link C libraries statically if possible on linkstatic = True binaries
Currently, a `haskell_binary` target with `linkstatic = True` (the default) will link C library dependencies dynamically, even if a static library is available. Only Haskell library dependencies are linked statically. 

It would be preferable to link C libraries statically if possible. For faster executable startup and smaller MACH-O headers on MacOS. Also, Hazel builds of Hackage packages define any cbits as `cc_library` targets. Not linking these statically means that all these cbits become dynamic library dependencies, which can cause large numbers of dynamic library dependencies.

As a workaround it is possible to enforce that C library dependencies are linked statically by setting `linkstatic = True` on the corresponding `cc_library` rule. However, this is not a viable solution: It will break Template Haskell and `@repl` targets which require shared objects to be available, and it is not applicable for `cc_library` targets that are imported from external workspaces.

Instead, rules_haskell should treat C library dependencies analogously to Haskell library dependencies and link statically where possible.

The following [test case](https://github.com/tweag/rules_haskell/commit/2e61beca5fa7b5e59d6f57db118a12c5f4804c72) illustrates the issue.

aherrmann:
Closes #580

Changes rules_haskell to prefer linking C library dependencies statically over dynamically on Haskell binary targets with `linkstatic = True`.

GHC's default behaviour (following from the linker's behaviour) is to prefer dynamic libraries over static libraries for C library dependencies, even when linking statically against Haskell libraries.

With this change only static versions of C library dependencies are passed as inputs to the linking action - unless no static library is available. This forces the linker to link against static versions of C library dependencies.

Adds a regression test for #580.


mboes:
What do the CC rules do here? If you `cc_import` something that has both a dynamic and static version, is the static version preferred in the default "mostly static" mode?

mboes:
I think I have my answer in example 5 of this section: https://docs.bazel.build/versions/master/be/c-cpp.html#cc_import.

So this PR is ultimately bringing us in line with the CC semantics for `linkstatic`?

aherrmann:
@mboes 
> So this PR is ultimately bringing us in line with the CC semantics for linkstatic?

Yes, `haskell_binary` and `haskell_test` will prefer static linking over dynamic linking for C library dependencies when a static library is available.

tests/binary-linkstatic-flag/BUILD
```diff
@@ -0,0 +1,118 @@
...
+    }),
+    data = [
+        ":binary-dynamic",
+    ],
+    script = """
+    set -eo pipefail
+    binary="$1"
+    mode="$2"
+    if [[ $mode = dbg ]]; then
+        # Skip test in debug builds. Debug mode forces static linking.
```
Profpatsch: Nice trick!

PR #588, opened 2019-01-17
Create wrapper bash scripts

### LINKED ISSUE #573: Work around quoting issues on Windows
Here is the upstream ticket:

https://github.com/bazelbuild/bazel/issues/7122

Potential strategies include:
- Adding dummy spaces to trigger proper quoting.
- Passing command-line arguments via `params_file` or similar (if those are still needed).

nmattia:
Fixes #573 

This is @gdeest 's work.

Works around some quoting issues on Windows by dumping the command to a bash script as opposed to creating a bash command that simply reads its args.

It also tweaks the binary and library builds' mnemonics. This seems to be a good idea in general but in particular here it fixes issues where several parameter files would get the same name (if both prof and non-prof builds are enabled). This fixes the issues but we should think about whether it's useful to have the param files be actually exported and whether we can work around that.


haskell/toolchain.bzl
```diff
@@ -72,7 +86,7 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, params_file = None, e
         mnemonic = mnemonic,
         progress_message = progress_message,
         env = env,
-        arguments = [args] + arguments,
+        arguments = [],
```
aherrmann: What's the reasoning for using `run_shell` instead of `run` with a parameters file like e.g. [here](https://github.com/bazelbuild/rules_dotnet/pull/47/files#diff-72490c7e47e72e5c8662d0f6408b43e1R204)?

gdeest: This is not possible here. `args.use_params_file(…)` instructs Bazel to dump all arguments into an auto-generated file, and replace them with a formatted string containing the path of the file. It works because `csc` can read arguments from so-called "response files", which are specified as file paths prefixed by the `@` character. It would be a good solution if GHC had a `--use-params-from-file=…` command-line parameter, but such is, to my knowledge, not the case.

aherrmann: Thanks for explaining!

PR #589, opened 2019-01-17
Windows: build simple binary

### LINKED ISSUE #584: Build a simple Haskell binary on Windows
This command should succeed on Windows:

> bazel build //tests/binary-simple:binary-simple

nmattia:
Fixes #584 

* Build `//tests/binary-simple` on CI on Windows
* Discard all `cc` args on Windows
* Fix the executable's output name on Windows (haskell name + `.exe`)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,5 +11,5 @@ jobs:
     displayName: 'Install Bazel'
   - bash: |
       # Simple haskell binary
-      /c/bazel/bazel.exe build "///tests/binary-simple" # first '/' gets eaten up
+      /c/bazel/bazel.exe build --compiler=msys-gcc "///tests/binary-simple" # first '/' gets eaten up
     displayName: 'Run Bazel'
--- a/haskell/private/actions/compile.bzl
+++ b/haskell/private/actions/compile.bzl
@@ -29,8 +29,6 @@ def _process_hsc_file(hs, cc, hsc_flags, hsc_file):
       (File, string): Haskell source file created by processing hsc_file and
          new import directory containing the produced file.
     """
-    if hs.toolchain.is_windows:
-        cc = "Bazel's CC shouldn't be used on Windows"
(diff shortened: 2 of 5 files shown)


azure-pipelines.yml
```diff
@@ -10,6 +10,6 @@ jobs:
       mv bazel.exe /c/bazel
     displayName: 'Install Bazel'
   - bash: |
-      # We make sure that GHC can be invoked
-      /c/bazel/bazel.exe build "///tests/ghc:ghc_help" # first '/' gets eaten up
+      # Simple haskell binary
+      /c/bazel/bazel.exe build "///tests/binary-simple" # first '/' gets eaten up
```
gdeest: What about `binary-with-main`, for example ? Does this one pass ?

nmattia: no idea, will look tomorrow

haskell/private/actions/compile.bzl
```diff
@@ -29,6 +29,8 @@ def _process_hsc_file(hs, cc, hsc_flags, hsc_file):
       (File, string): Haskell source file created by processing hsc_file and
          new import directory containing the produced file.
     """
+    if hs.toolchain.is_windows:
+        cc = "Bazel's CC shouldn't be used on Windows"
```
mboes: Shouldn't this be a call to `fail()`? But also, where are we using "Bazel's CC" here? What does this refer to? the `CC` environment variable?

gdeest:
Arguments were dropped so that we don't need to pass `--compiler=msys-gcc`. I am not sure of the consequences.

haskell/private/actions/compile.bzl
gdeest: It seems to override the function's `cc` argument. I think it's here to ensure that we don't actually use it.

gdeest:
@mboes PTAL.

PR #590, opened 2019-01-17
Make c2hs a separate toolchain.

mboes:
The GHC toolchain could previously optionally be provided an extra
"tool", c2hs. But

This is a **breaking change**. But few people use c2hs and the change
is only in the toolchain declaration code, which happens only once per
workspace.

### DIFF
--- a/haskell/c2hs.bzl
+++ b/haskell/c2hs.bzl
@@ -17,6 +17,7 @@ def _c2hs_library_impl(ctx):
     hs = haskell_context(ctx)
     cc = cc_interop_info(ctx)
     args = hs.actions.args()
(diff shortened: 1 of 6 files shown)


haskell/c2hs.bzl
```diff
@@ -98,5 +99,71 @@ c2hs_library = rule(
...
+      ```bzl
+      nixpkgs_package(
+          name = "c2hs",
+          attribute_path = "haskell.packages.ghc822.c2hs",
+      )
+
+      register_toolchains("//:c2hs")
+      ```
+
+    """
```
guibou: A new line between the doc and the implementation?

mboes: This is the default Emacs style for docstrings in Python. Can't say I'm a fan. But we already have this in a few places in the code.

PR #594, opened 2019-01-18
Binary with custom main on Windows

### LINKED ISSUE #585: Build a Haskell binary with custom main on Windows
This command should succeed:

> bazel build //tests/binary-custom-main

gdeest:
Closes #585 .

Easy enough.

### DIFF
--- a/tests/binary-custom-main/BUILD
+++ b/tests/binary-custom-main/BUILD
@@ -1,13 +1,16 @@
 load(
     "@io_tweag_rules_haskell//haskell:haskell.bzl",
+    "haskell_import",
     "haskell_test",
 )
 
 package(default_testonly = 1)
 
+haskell_import(package = "base")
+
 haskell_test(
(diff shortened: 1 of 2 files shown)


tests/binary-custom-main/BUILD
```diff
@@ -1,13 +1,19 @@
 load(
     "@io_tweag_rules_haskell//haskell:haskell.bzl",
+    "haskell_import",
     "haskell_test",
 )
 
 package(default_testonly = 1)
 
+haskell_import(
+    name = "base_pkg",
```
aherrmann: Out of curiosity, why not
```
haskell_import(
    name = "base",
)
```
?

gdeest: I have copied what @nmattia did for `binary-simple`, really, so no particular reason.

nmattia: and I copied it from the docs...

PR #595, opened 2019-01-18
Compile binary against simple (static) library on Windows

### LINKED ISSUE #586: Build a Haskell library on Windows
Building a library with no dependencies should succeed on Windows.

gdeest:
Closes #586 .

### DIFF
--- a/haskell/private/actions/link.bzl
+++ b/haskell/private/actions/link.bzl
@@ -486,11 +486,17 @@ def link_library_static(hs, cc, dep_info, objects_dir, my_pkg_id, with_profiling
             static_library,
             "@" + objects_dir_manifest.path,
         ])
+
+        if hs.toolchain.is_windows:
+            ar_exe = hs.toolchain.tools.ar
+        else:
+            ar_exe = cc.tools.ar
+
         hs.actions.run(
             inputs = inputs,
             outputs = [static_library],
             mnemonic = "HaskellLinkStaticLibrary",
-            executable = cc.tools.ar,
+            executable = ar_exe,
             arguments = [args],
         )
 
--- a/haskell/private/actions/link.bzl
+++ b/haskell/private/actions/link.bzl
@@ -486,17 +486,11 @@ def link_library_static(hs, cc, dep_info, objects_dir, my_pkg_id, with_profiling
             static_library,
             "@" + objects_dir_manifest.path,
         ])
-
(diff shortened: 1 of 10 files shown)


gdeest:
@mboes PTAL.

mboes:
And FWIW I'm not able to reproduce locally. For me Bazel just hangs.

gdeest:
Roughly:

I had to work around another `run_shell` quoting issue, which I handled by moving the generation of package config to an external script, and using param_files ; I think I should be able to revert to `run_shell` without damage now.

Agreed regarding `ar`.

About your other questions / remarks, could you detail:

- Which problems you are having on your Azure VM ?
- At which step Bazel is hanging for you ?

gdeest:
I believe 1. is unrelated to this PR and should be addressed in another one.

Regarding 2., I haven't witnessed that myself, but I'll investigate.

haskell/private/actions/link.bzl
```diff
@@ -381,11 +381,17 @@ def link_library_static(hs, cc, dep_info, objects_dir, my_pkg_id, with_profiling
             static_library,
             "@" + objects_dir_manifest.path,
         ])
+
+        if hs.toolchain.is_windows:
```
gdeest: @mboes I don't know how to get rid of this. Ideally, I'd like to set `hs.toolchain.tools.ar` as `cc.tools.ar` on anything but Windows ; but I can't find a way to get my hand the `cc` context within toolchain configuration.

mboes:
Yes. I think (1) is likely an upstream issue. I'll find some time later to work around it on my local VM.

haskell/private/actions/link.bzl
mboes: `cc_interop_info` in `cc.bzl` is the function creating the `cc` context. It gets called by rules that all have a `//haskell:toolchain` toolchain dependency. So you can get at `hs.tools.ar` from there. That said, I'm not convinced this is necessary. Are you sure `ar` fails if it's not the GHC one? `ar` is a very simple program, whose inputs are all explicit, and whose output format hasn't changed in years or even decades...

gdeest: Well, for starters we don't have a proper MSYS2 toolchain on the Windows CI. We could add one, but with current plans to tether Bazel's CC toolchain on GHC's on Windows, ( cf #304 ) it seems a bit silly…

PR #596, opened 2019-01-19
Introduce our own config settings for platforms to match GHC naming.

mboes:
GHC uses its own naming convention for architectures and for operating
systems. Rather than use Bazel's `@bazel_tools//conditions/*`
directly, which often have different names than GHC uses, we use GHC
names (that way, "mingw32" means we're in an environment that the
mingw32 bindist works on). This change is for now only internally
visible. The user does not have any reason to use these config
settings just yet.

### DIFF
--- a/haskell/platforms/BUILD
+++ b/haskell/platforms/BUILD
@@ -1,0 +1,11 @@
+# This file declares constraint values for each platform supported by
+# GHC. These rules follow the GHC naming convention, for example,
+# //haskell/platform:linux and //haskell/platform:x86_64. See the
+# config.guess in any GHC source distribution for possible platforms.
+#
+# These can be used in select expressions to choose platform-specifc
+# sources and dependencies.
+
+load(":list.bzl", "declare_config_settings")
+
+declare_config_settings()
--- a/haskell/platforms/list.bzl
+++ b/haskell/platforms/list.bzl
@@ -1,0 +1,44 @@
+OS = {
+    "aix": None,
+    "darwin": "@bazel_tools//platforms:osx",
+    "dragonfly": None,
+    "freebsd": "@bazel_tools//platforms:freebsd",
+    "haiku": None,
+    "hpux": None,
+    "ios": "@bazel_tools//platforms:ios",
+    "linux_android": "@bazel_tools//platforms:android",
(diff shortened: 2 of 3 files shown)


haskell/platforms/BUILD
```diff
@@ -0,0 +1,10 @@
+# This file declares constraint values for each platform supported by
+# GHC. These rules follow the GHC naming convention, for example,
```
guibou: Please add a link to GHC naming convention.

haskell/platforms/list.bzl
```diff
@@ -0,0 +1,44 @@
...
+    "aix": None,
+    "darwin": "@bazel_tools//platforms:osx",
+    "dragonfly": None,
+    "freebsd": "@bazel_tools//platforms:freebsd",
+    "haiku": None,
+    "hpux": None,
+    "ios": "@bazel_tools//platforms:ios",
+    "linux_android": "@bazel_tools//platforms:android",
+    "linux": "@bazel_tools//platforms:linux",
+    "mingw32": "@bazel_tools//platforms:windows",
```
guibou: I don't know ghc naming convention, but I'm surprised to find `mingw32` in `OS` instead of `mingw` and 32bits later in `ARCH`. If that't correct, please add a one line comment.

haskell/platforms/list.bzl
```diff
@@ -0,0 +1,44 @@
...
+    "powerpc64le": None,
+    "powerpc": "@bazel_tools//platforms:ppc",
+    "rs6000": None,
+    "sparc": None,
+    "x86_64": "@bazel_tools//platforms:x86_64",
+}
+
+def declare_config_settings():
+    for os, constraint_value in OS.items():
+        if constraint_value:
```
guibou: I don't see the point of listing entries which have no `bazel` associated plateform name.

mboes: Because at least this way the list is exhaustive. And can be mapped to user defined "config values" later. This is also what rules_go does.

haskell/platforms/BUILD
mboes: This is not a GHC specific convention, nor is canonical documentation available. I added a link to `config.guess`, which is essentially standard autoconf stuff.

PR #597, opened 2019-01-19
Don't hardcode platforms in toolchain anymore

mboes:
Our toolchain implementation was incomplete. We used to assume x86_64 as
the architecture, and never set any platform constraints. We now make it
possible to do so in new attributes to `haskell_toolchain`. This has
important implications. It means that we no longer should define a single
toolchain that states different constraints depending on the host platform.
We now instead have multiple toolchains in scope, one for each operating
system. Which properly declared constraints, the toolchain resolution
mechanism can select the right one.

As a side effect, `ghc_bindist` is now easier to use. Users don't need
to write a separate toolchain declaration and toolchain registration
in addition to the declaring the bindist anymore. The next step after
this PR will be to introduce convenience functions to add external
dependencies for all known bindists, and all known GHC's in Nixpkgs.
That sounds overkill, but the toolchain resolution mechanism makes
this fine. And it's how we'll finally make it seamless for some
developers to use bindists while others on the same project can use
Nixpkgs without having to make edits to the `WORKSPACE`.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -28,7 +28,34 @@ load(
 haskell_nixpkgs_package(
     name = "ghc",
     attribute_path = "haskellPackages.ghc",
-    build_file = "//haskell:ghc.BUILD",
+    build_file_content = """
+package(default_visibility = ["//visibility:public"])
+
+filegroup(
+    name = "bin",
+    srcs = glob(["bin/*"]),
+)
+
+cc_library(
+    name = "threaded-rts",
+    srcs = glob(
+        ["lib/ghc-*/rts/libHSrts_thr-ghc*." + ext for ext in [
+            "so",
+            "dylib",
+        ]] +
+        # dependency of `libHSrts_thr_ghc*`
+        # globbing on the `so` version to stay working when they update
+        [
+            "lib/ghc-*/rts/libffi.so.*",
+        ],
+    ),
+    hdrs = glob(["lib/ghc-*/include/**/*.h"]),
+    strip_include_prefix = glob(
+        ["lib/ghc-*/include"],
+        exclude_directories = 0,
+    )[0],
+)
+    """,
     nix_file = "//tests:ghc.nix",
     # rules_nixpkgs assumes we want to read from `<nixpkgs>` implicitly
     # if `repository` is not set, but our nix_file uses `./nixpkgs/`.
@@ -59,11 +86,13 @@ ghc_version = "8.4.4"
 # A GHC from a bindist for Windows
 ghc_bindist(
     name = "ghc_windows",
+    target = "windows_amd64",
     version = ghc_version,
 )
 
 register_toolchains(
-    "//tests:ghc",
+    "//tests:ghc_linux",
+    "//tests:ghc_osx",
     "//tests:c2hs-toolchain",
     "//tests:doctest-toolchain",
     "//tests:protobuf-toolchain",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -28,34 +28,7 @@ load(
 haskell_nixpkgs_package(
     name = "ghc",
     attribute_path = "haskellPackages.ghc",
-    build_file_content = """
-package(default_visibility = ["//visibility:public"])
-
-filegroup(
-    name = "bin",
-    srcs = glob(["bin/*"]),
-)
-
-cc_library(
-    name = "threaded-rts",
-    srcs = glob(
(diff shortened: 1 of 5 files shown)


WORKSPACE
```diff
@@ -28,7 +28,34 @@ load(
...
+filegroup(
+    name = "bin",
+    srcs = glob(["bin/*"]),
+)
+
+cc_library(
+    name = "threaded-rts",
+    srcs = glob(
+        ["lib/ghc-*/rts/libHSrts_thr-ghc*." + ext for ext in [
+            "so",
```
guibou: no `.dll`?

WORKSPACE
```diff
@@ -59,11 +86,13 @@ ghc_version = "8.4.4"
...
 ghc_bindist(
     name = "ghc_windows",
+    target = "windows_amd64",
     version = ghc_version,
 )
 
 register_toolchains(
-    "//tests:ghc",
+    "//tests:ghc_linux",
+    "//tests:ghc_osx",
```
guibou: no `ghc_window`?

tests/BUILD
```diff
@@ -17,39 +17,45 @@ package(default_testonly = 1)
...
+            "-UTESTS_TOOLCHAIN_REPL_FLAGS",
+            # The repl test will need OverloadedString
+            "-XOverloadedStrings",
+        ],
+        target_compatible_with = ["@bazel_tools//platforms:" + os],
+        tools = tools,
+        version = ghc_version,
+    )
+    for os, tools, locale_archive in [
+        ("linux", "@hackage-ghc//:bin", "@glibc_locales//:locale-archive"),
```
guibou: No `"window"` either?

haskell/ghc.BUILD
```diff
@@ -2,9 +2,7 @@ package(default_visibility = ["//visibility:public"])
 
 filegroup(
     name = "bin",
-    srcs = glob([
-        "bin/*",
-    ]),
+    srcs = glob(["bin/*"]),
```
guibou: Cosmetic only, or I missed something?

WORKSPACE
```diff
@@ -28,7 +28,34 @@ load(
 haskell_nixpkgs_package(
     name = "ghc",
     attribute_path = "haskellPackages.ghc",
-    build_file = "//haskell:ghc.BUILD",
```
guibou: Does this mean that `//haskell:ghc.BUILD` is now only used for `ghc_bindist`? Shouldn't the name change?

WORKSPACE
mboes: No. See the commit messages as for why. Windows uses `ghc_bindist` and that rule self registers.

WORKSPACE
mboes: This is a copy/paste of the existing ghc.BUILD. Any improvement here would be an orthogonal change.

WORKSPACE
mboes: With the latest commit in the series, we no longer need to duplicate much of the content of this file. So we're back to using `//haskell:ghc.BUILD` here as before.

tests/BUILD
mboes: Nope. See commit messages as for why.

haskell/ghc_bindist.bzl
```diff
@@ -107,42 +111,51 @@ ghc_bindist = repository_rule(
...
+       load("@io_tweag_rules_haskell//haskell:haskell.bzl", "ghc_bindist")
+
+       # This repository rule creates @ghc repository.
+       ghc_bindist(
+         name    = "ghc",
+         version = "8.2.2",
+       )
+
+       # Register the toolchain defined locally in BUILD file:
+       register_toolchains("//:ghc")
```
aherrmann: The docs seem outdated. As I understand the changes, this should no longer be required of the user.

haskell/ghc_bindist.bzl
```diff
@@ -107,42 +111,51 @@ ghc_bindist = repository_rule(
...
+       ```
+
+       In `BUILD` file:
+
+       ```bzl
+       # Use binaries from @ghc//:bin to define //:ghc toolchain.
+       haskell_toolchain(
```
aherrmann: Same here.

haskell/ghc_bindist.bzl
aherrmann: (resolved in later commit)

haskell/ghc_bindist.bzl
aherrmann: (resolved in later commit)

PR #609, opened 2019-01-22
Fix quoting regression in run_ghc

### LINKED ISSUE #605: Fix quoting regression
According to this comment: 

https://github.com/tweag/rules_haskell/commit/5848c722d52c22b629a1311733025be21e0f45ec#commitcomment-32005874

Commit 5848c722d52c22b629a1311733025be21e0f45ec breaks `--with-rtsopts` flag with multiple options on Linux.

Reproduce and fix.

gdeest:
Closes #605 .

### DIFF
--- a/haskell/toolchain.bzl
+++ b/haskell/toolchain.bzl
@@ -67,13 +67,19 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, params_file = None, e
     if params_file:
         command = """
         export PATH=${PATH:-} # otherwise GCC fails on Windows
-        $(< %s) $(< %s) $(< %s)
+        readarray -t ghc_args < %s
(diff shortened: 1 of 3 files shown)


tests/binary-simple/BUILD
```diff
@@ -14,6 +14,10 @@ haskell_import(
 haskell_test(
     name = "binary-simple",
     srcs = ["Main.hs"],
+    compiler_flags = [
+        "-threaded",
```
nmattia: May be worth having a test with something that only compiles with e.g. `-XOverloadedStrings` to ensure the flags are actually passed

gdeest: We already have a `binary-with-compiler-flags`, it just didn't have any whitespace. I could move those arguments there, actually

PR #610, opened 2019-01-22
Toolchain registration macros

mboes:
It was previously hard to let the user select between a Nixpkgs based
toolchain and a binary distributions based one. Most
projects (including rules_haskell) were essentially hardcoding this
choice for the user. It is now much easier to do. The trick is to
supply three new workspace macros:

```
haskell_register_toolchains()    # Same as the second one.
haskell_register_ghc_bindists()
haskell_register_ghc_nixpkgs()
```

The second function above defines and registers one toolchain per
binary distribution. So e.g. one for Windows, one for Linux etc. The
third one currently defines exactly one toolchain, which matches the
constraints of the host platform. It might define more in the
future (in particular cross-compilers).

As a project author, you are free to just the second one or just the
third one, or both. While the Nixpkgs toolchain was always selected if
it existed previously, now it will only be selected if the user
explicitly sets the host platform in a certain way. See the
documentation about the details.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -51,6 +51,42 @@ nixpkgs_local_repository(
     nix_file = "//nixpkgs:default.nix",
 )
 
+test_ghc_version = "8.6.3"
+
+test_compiler_flags = [
+    "-XStandaloneDeriving",  # Flag used at compile time
+    "-threaded",  # Flag used at link time
+
+    # Used by `tests/repl-flags`
+    "-DTESTS_TOOLCHAIN_COMPILER_FLAGS",
+    # this is the default, so it does not harm other tests
+    "-XNoOverloadedStrings",
+]
+
+test_haddock_flags = ["-U"]
+
+test_repl_ghci_args = [
+    # The repl test will need this flag, but set by the local
+    # `repl_ghci_args`.
+    "-UTESTS_TOOLCHAIN_REPL_FLAGS",
+    # The repl test will need OverloadedString
+    "-XOverloadedStrings",
+]
+
+load(
+    "@io_tweag_rules_haskell//haskell:nixpkgs.bzl",
+    "haskell_register_ghc_nixpkgs",
+)
+
+haskell_register_ghc_nixpkgs(
+    compiler_flags = test_compiler_flags,
+    haddock_flags = test_haddock_flags,
+    locale_archive = "@glibc_locales//:locale-archive",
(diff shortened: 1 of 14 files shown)


haskell/ghc_bindist.bzl
```diff
@@ -190,3 +190,24 @@ def ghc_bindist(name, version, target):
         target = target,
     )
     native.register_toolchains("@{}//:toolchain".format(toolchain_name))
+
+def haskell_register_ghc_bindists(version):
+    """Register all GHC binaries distributions as toolchains.
```
nmattia: (minor)

```suggestion
    """Register GHC binary distributions for all platforms/OSs as toolchains.
```
"all" isn't very clear

haskell/nixpkgs.bzl
```diff
@@ -229,3 +229,118 @@ exports_files(["all-haskell-packages.bzl"])
             **kwargs
         ),
     )
+
+def _ghc_nixpkgs_toolchain_impl(repository_ctx):
+    # These constraints might look tautological, because they always
+    # match the host platform if it is the same as the target
+    # platform. But they are important to state because Bazel
```
nmattia: :man_facepalming:

tests/BUILD
```diff
@@ -264,6 +226,9 @@ rule_test(
     rule = "//tests/java_classpath",
 )
 
+# Keep in sync with test_ghc_version in WORKSPACE.
```
guibou: Conflict with #614, be careful.

WORKSPACE
```diff
@@ -51,24 +51,46 @@ nixpkgs_local_repository(
...
+haskell_register_ghc_nixpkgs(
+    compiler_flags = test_compiler_flags,
+    haddock_flags = test_haddock_flags,
+    locale_archive = "@glibc_locales//:locale-archive",
+    nix_file = "//tests:ghc.nix",
+    repl_ghci_args = test_repl_ghci_args,
+    version = test_ghc_version,
 )
 
+haskell_register_ghc_bindists(version = test_ghc_version)
```
guibou: No `compiler_flags` or `haddock_flags` for that one?

mboes: These should be added, for symmetry with `haskell_register_ghc_nixpkgs`. They weren't needed for the MVP though (since we currently only use bindists on windows and on windows we currently don't want to set any special flags - until we can propertly support them).

PR #614, opened 2019-01-22
Redo -rpath rewiring

Profpatsch:
A long time in the making, this is the second try of rewriting relative `-rpath` directories to stay inside of bazel runfile directories.

The first part of this work is in https://github.com/tweag/rules_haskell/pull/436, but the crucial part had to be reverted because it introduced a big regression. This regression is resolved in this PR. 

There’s a lot going on in these commits, but later commits don’t revert work from previous ones, so stepping through one at a time is the best strategy. The first commit is not new work, so its changes are mostly outdated by the commits that follow. I tried to make everything as self-explanatory as possible, especially for the less obvious parts.

If some comments don’t provide enough context to understand the reason for a change, please mark them. The logic is confusing, because bazel’s behavior is extremely confusing when it comes to paths, external repositories and sandboxing.

@guibou this didn’t require your `-rpath` patch in the end, because it there is a more general solution to the problem.

@aherrmann please test against hazel and the mac stuff, it should shrink the header a bit.

### DIFF
--- a/haskell/private/actions/link.bzl
+++ b/haskell/private/actions/link.bzl
@@ -67,7 +67,12 @@ def _get_target_parent_dir(target):
         return (False, parent_dir)
 
 # tests in /tests/unit_tests/BUILD
-def create_rpath_entry(binary, dependency, keep_filename, prefix = ""):
+def create_rpath_entry(
+        binary,
+        dependency,
+        keep_filename,
+        comes_from_haskell_cc_import,
+        prefix = ""):
     """Return a (relative) path that points from `binary` to `dependecy`
     while not leaving the current bazel runpath, taking into account weird
     corner cases of `.short_path` concerning external repositories.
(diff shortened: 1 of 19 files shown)


tests/constants.bzl
```diff
@@ -0,0 +1 @@
+ghc_version = "8.6.3"
```
guibou: Conflict with #610, be careful.

aherrmann:
@Profpatsch Not sure about the exact issue. I've seen similar things happen if only `libz.so` is added to the runfiles, but if the linker references the library by its soname `libz.so.1`. Then the dynamic load will fail because the library is missing. See https://github.com/tweag/rules_nixpkgs/issues/60

tests/unit-tests/BUILD
```diff
@@ -0,0 +1,121 @@
...
+    binary_short_path = "../foo/a.so",
+    # external dep
+    dependency_short_path = "../bar/b.so",
+    keep_filename = False,
+    # stay in `external`
+    output = "../bar",
+    prefix = "",
+)
+
+# we have no idea how to handle internal dots, should they arise
```
aherrmann: Why is this here? Is this something that should be resolved in future? Or is it unclear if the need will ever arise?

Profpatsch: Yeah, completely unclear. I added a check that `fail()`s, should something like that every appear. The commented out test documents that case.

haskell/private/actions/link.bzl
```diff
@@ -344,18 +524,53 @@ def _add_external_libraries(args, ext_libs):
...
+    # `<dir>` that is added with `-L` (`-dynload=sysdep`).
+    # These paths are of no value to bazel, because the `lib.path`
+    # dirs we add below are only valid at link-time. We add the
+    # correct load-time paths (`lib.short_path`) below manually,
+    # and use `-dynload=deploy` so GHC doesn’t add any.
+    # See https://downloads.haskell.org/~ghc/7.6.1/docs/html/users_guide/using-shared-libs.html#finding-shared-libs-unix
+    # The difference between link-time and load-time comes from
+    # bazel sandboxing the runtime.
+    # (see: bazel runfiles).
+    args.add("-dynload=deploy")
```
aherrmann: Unfortunately, this breaks dependencies on prebuilt packages or the rts on MacOS. E.g. `//tests/binary-dynamic` fails, because it depends on `@rpath/libHSbase-4.12.0.0-ghc8.6.3.dylib`, `@rpath/libHSghc-prim-0.5.3-ghc8.6.3.dylib`, `@rpath/libHSrts_thr-ghc8.6.3.dylib`, but there's no `rpath` entry for any of them.

I think this requires additional work to forward `library-dirs` fields of prebuilt dependencies and add them to `rpath`. But, maybe that's better done in a separate PR, and `-dynload=deploy` is dropped for now.

Profpatsch: removed it from this PR.

haskell/private/actions/link.bzl
```diff
@@ -162,7 +261,12 @@ def _fix_darwin_linker_paths(hs, inp, out, external_libraries):
                 # at execution time.
                 "/usr/bin/install_name_tool -change {} {} {}".format(
                     f.lib.path,
-                    paths.join("@loader_path", backup_path(out), f.lib.short_path),
+                    create_rpath_entry(
```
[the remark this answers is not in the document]
Profpatsch: I don’t know, I didn’t touch the semantics of the original code.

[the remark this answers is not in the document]
Profpatsch: Comes from another change.

Profpatsch:
It’s greeeeen. Halleluja.

From my side we can merge and brace for regressions :shield:.

haskell/private/actions/link.bzl
```diff
@@ -82,42 +87,60 @@ def create_rpath_entry(binary, dependency, keep_filename, prefix = ""):
...
+        # backup the full path to the binary
+        # This abuses the fact that the linker resolves symlinks,
+        # meaning the `binary` path is not actually the path in the
+        # runfiles folder, rather the one in the bazel execpath.
+        bin_backup = [".."] * len(parent_dir_path(binary.path))
+
+        # append the full path of our dependency
+        path_segments = bin_backup + parent_dir_path(dependency.path)
+
+        # external repositories live in `target.runfiles/external`,
```
aherrmann: Is this comment in the right place?

Profpatsch: Good catch, must have been moved by buildifier.

Profpatsch:
The build failure on darwin is transient, I only moved one comment.

Issue #620, opened 2019-01-23
Doctest failure when running all tests at once


nmattia:
The following error shows up somewhat regularly:

```
[nix-shell:~/projects/tweag/rules_haskell]$ bazel test //tests/...
INFO: Invocation ID: 87a10bac-b772-461b-93c6-e0e5e0199bb8
INFO: Analysed 204 targets (0 packages loaded, 0 targets configured).
INFO: Found 132 targets and 72 test targets...
ERROR: /home/nicolas/projects/tweag/rules_haskell/tests/haskell_doctest/BUILD:40:1: error executing shell command: '/nix/store/pawkx6zfnj59zvpa0hkrrqin9f9iwika-bash/bin/bash -c
        export LANG=en_US.UTF-8
export LOCALE_ARCHIVE=external/glibc_locales/lib/locale/locale-archive
        external/hackage-doctest...' failed (Exit 1) bash failed: error executing command /nix/store/pawkx6zfnj59zvpa0hkrrqin9f9iwika-bash/bin/bash -c ... (remaining 57 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
doctest: /run/user/1001/.doctest-6: createDirectory: already exists (File exists)
```

It can be worked around by specifying `--jobs=1` but this makes running the tests take close to forever.

CC @gdeest

nmattia:
Note: another work around is to comment out those lines: https://github.com/tweag/rules_haskell/blob/f7b355a9c5da4e0b81fc60bc963cd78f0dc4fe0b/tests/BUILD#L174-L181

gdeest:
It looks like `doctest` uses the process' ID to generate its temp directory:

https://github.com/sol/doctest/blob/1cc1902b602ed3c6cf9a1c8c2144acb27435dc17/src/Extract.hs#L146

and brackets around that with createDirectory / removeDirectoryRecursive, so it's a bit confusing that multiple tests (seemingly) attempt to create the same directory.

gdeest:
Hmmm, Linux sandboxing support a mechanism of PID namespacing. This might explain it.

gdeest:
@nmattia Could you try the `fix-doctest` branch and tell me how it fares ?

Profpatsch:
Fixed by #622. If not, please reopen.

Issue #616, opened 2019-01-23
buildifier-fix broken On NixOS


gdeest:
`buildifier-fix` is currently broken for me on `master`:

```
[huginn@frigg:~/code/rules_haskell]$ git rev-parse HEAD
f7b355a9c5da4e0b81fc60bc963cd78f0dc4fe0b

[huginn@frigg:~/code/rules_haskell]$ nix-shell --pure

[nix-shell:~/code/rules_haskell]$ bazel run //:buildifier-fix
INFO: Invocation ID: 8358a809-d09f-4a46-9398-2628f8615597
ERROR: /home/huginn/.cache/bazel/_bazel_huginn/ef11f29d626b5c3fd8622d2529b44540/external/go_sdk/BUILD.bazel:36:12: in libs attribute of go_sdk rule @go_sdk//:go_sdk: '@go_sdk//:libs' does not produce any go_sdk libs files (expected .a)
ERROR: /home/huginn/.cache/bazel/_bazel_huginn/ef11f29d626b5c3fd8622d2529b44540/external/go_sdk/BUILD.bazel:37:15: in headers attribute of go_sdk rule @go_sdk//:go_sdk: '@go_sdk//:headers' does not produce any go_sdk headers files (expected .h)
ERROR: Analysis of target '//:buildifier-fix' failed; build aborted: Analysis of target '@go_sdk//:go_sdk' failed; build aborted
INFO: Elapsed time: 4.525s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (6 packages loaded, 8 targets configured)
FAILED: Build did NOT complete successfully (6 packages loaded, 8 targets configured)

[nix-shell:~/code/rules_haskell]$ uname -a
Linux frigg 4.14.82 #1-NixOS SMP Wed Nov 21 08:24:18 UTC 2018 x86_64 GNU/Linux
```

cc @nmattia

mboes:
Cannot reproduce. You're in a `nix-shell`, and I'm trying the exact same commit as you. That's worrying... I know this should not be a solution, but try `bazel clean --expunge`? If that works then it's a bug, but likely a bug upstream in rules_go (hence would need to file a ticket there). I really don't see how this kind of hermeticity issue would creep in, however.

gdeest:
Indeed, `clean --expunge` seems to have solved it. Might be hard to reproduce …

PR #622, opened 2019-01-23
Disable sandboxing on doctest tests to prevent race condition

gdeest:
There are occasional race conditions issues on Linux with doctest tests. The issue seems related to sandboxing (cf. #620 ).

This PR fixes this issue by selectively disabling sandboxing on those tests.

### DIFF
--- a/haskell/doctest.bzl
+++ b/haskell/doctest.bzl
@@ -183,15 +183,15 @@ def _haskell_doctest_single(target, ctx):
             hs.env,
         ),
         execution_requirements = {
-          # Prevents a race condition among concurrent doctest tests on Linux.
-          #
-          # The reason is that the doctest process uses its own PID to determine the name
-          # of its working directory. In presence of PID namespacing, this occasionally results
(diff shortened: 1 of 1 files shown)


gdeest:
CI currently read because of linter. I can't run `buildifier-fix` locally (cf. #616 ).

regnat:
Isn't there a cleaner solution for fixing these tests? I remember we had the same issue with haddock and we could fix it by explicitely specifying the temp directory. Isn't it something that we can fix the same way?

gdeest:
@regnat 

Indeed, I am not too happy about it either, but `doctest` does not support being passed an explicit directory. I have filed an issue here: https://github.com/sol/doctest/issues/219

In the mean time, disabling sandboxing for those tests seems reasonable to me, but I am open to better ideas.

mboes:
@gdeest could you include a link to the issue you created in the source code? That way if it gets fixed then we'll eventually move the code to a cleaner solution.

PR #624, opened 2019-01-23
Setup the appropriate host_platform configuration when using nix shell

NinjaTrappeur:
We currently cannot build rules_haskell using the default nix shell script alone. We either
need to setup the appropriate host_platform flags in the
bazelrc.local file, either pass the appropriate flags to the CLI.

This PR adds a small nix shell hook that'll append the appropriate flags to the
.bazelrc.local if they are missing.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -23,6 +23,22 @@ mkShell {
   ] ++ lib.optionals docTools [graphviz python36Packages.sphinx zip unzip];
 
   shellHook = ''
+    # Add nix config flags to .bazelrc.local.
+    #
+    BAZELRC_LOCAL=".bazelrc.local"
+    if [ ! -e "$BAZELRC_LOCAL" ]
+    then
+      ARCH=""
+      if [ "$(uname)" == "Darwin" ]; then
+        ARCH="darwin"
+      elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+        ARCH="linux"
+      fi
+      echo "[!] It looks like you are using a ''${ARCH} nix-based system. In order to build this project, you probably need to add the two following host_platform entries to your .bazelrc.local file."
+      echo ""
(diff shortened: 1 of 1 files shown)


NinjaTrappeur:
Fair enough.

shell.nix
```diff
@@ -23,6 +23,25 @@ mkShell {
...
+    then
+      ARCH=""
+      if [ "$(uname)" == "Darwin" ]; then
+        ARCH="darwin"
+      elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+        ARCH="linux"
+      fi
+      echo "[!] It looks like you are using a ''${ARCH} nix-based system. In order to build this project, you probably need to add the two following host_platform entries to your .bazelrc.local file."
+      echo ""
+      echo "build --host_platform=@io_tweag_rules_haskell//haskell/platforms:''${ARCH}_x86_64_nixpkgs"
```
mboes: I believe you don't need both. `build` is inherited when you say `test`.

shell.nix
```diff
@@ -23,6 +23,25 @@ mkShell {
   ] ++ lib.optionals docTools [graphviz python36Packages.sphinx zip unzip];
 
   shellHook = ''
+    # Add nix config flags to .bazelrc.local.
+    #
+    BAZELRC_LOCAL=".bazelrc.local"
+    if [ ! -e "$BAZELRC_LOCAL" ] || (\
+       ! grep -s -q "\-\-host_platform\=" "$BAZELRC_LOCAL" && \
```
mboes: We're invariably going to get false negatives here (aliases, implied flags via `--config` etc). I'd simplify the script to the point that we're only testing for the existence of `.bazelrc.local`. If it already exists, then presumably the user knows what they're doing.

shell.nix
NinjaTrappeur: Did not know that.

PR #627, opened 2019-01-24
Handle indirect dynamic CC library dependencies

### LINKED ISSUE #170: Transitive C dependencies are not linked

### LINKED ISSUE #626: Dynamically linked Haskell binaries don't load indirect cc_library dependencies
Indirect dynamic `cc_library` dependencies are not reflected in the dynamic header of a `haskell_binary` causing dynamic linker errors at runtime.

The issue is demonstrated in [this commit][demo_commit].

The example takes a `haskell_binary`, which depends on a `cc_library` (`libvalue`) in a different package, which in turn depends on yet another `cc_library` (`libadd`) in a different package. The `haskell_binary` is built with `linkstatic = False`.

The resulting binary loads the direct `cc_library` dependency `libvalue` dynamically. However, it does not reference the indirect `cc_library` dependency `libadd`.

```
$ readelf -d bazel-bin/tests/binary-with-doubly-indirect-cbits/binary-with-doubly-indirect-cbits
...
 0x0000000000000001 (NEEDED)             Shared library: [libtests_Sbinary-with-doubly-indirect-cbits_Slib-direct_Svalue_libvalue.so]
...
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN/../../../../../bazel-out/k8-fastbuild/bin/tests/binary-with-doubly-indirect-cbits:...]
...
```

Dynamic `cc_library`s generated by Bazel do not declare dynamic dependencies themselves. Instead, Bazel expects the final binary to explicitly load all transitive shared library dependencies. In this example, the code in `libvalue` will fail to execute due to a missing symbol from `libadd`.

E.g. on MacOS
```
$ bazel run //tests/binary-with-doubly-indirect-cbits
...
Executing tests from //tests/binary-with-doubly-indirect-cbits:binary-with-doubly-indirect-cbits
-----------------------------------------------------------------------------
dyld: lazy symbol binding failed: Symbol not found: _add
  Referenced from: /private/var/tmp/_bazel_builder/d1983bacedc8743311e709889ea53957/execroot/io_tweag_rules_haskell/bazel-out/darwin-fastbuild/bin/tests/binary-with-doubly-indirect-cbits/../../../../../bazel-out/darwin-fastbuild/bin/tests/binary-with-doubly-indirect-cbits/lib-direct/libvalue.so
  Expected in: flat namespace

dyld: Symbol not found: _add
  Referenced from: /private/var/tmp/_bazel_builder/d1983bacedc8743311e709889ea53957/execroot/io_tweag_rules_haskell/bazel-out/darwin-fastbuild/bin/tests/binary-with-doubly-indirect-cbits/../../../../../bazel-out/darwin-fastbuild/bin/tests/binary-with-doubly-indirect-cbits/lib-direct/libvalue.so
  Expected in: flat namespace
```

Interestingly, the same test-case does not fail on Linux. The reason is most likely that the code of `value` and `add` is inlined into the binary on Linux, but not on MacOS. At least `objdump -d` indicates this. 

[demo_commit]: https://github.com/tweag/rules_haskell/commit/a1d6b15342b3dc25bf942245ce132390d64f2488

aherrmann:
Closes #626 
Closes #170 

Bazel's dynamic `cc_library` build results do not instruct the dynamic linker to load their own dynamic library dependencies. Instead, Bazel expects the final `cc_binary` to explicitly load all its transitive dynamic library dependencies. rules_haskell did not follow this convention, and the final Haskell binary would only explicitly load its direct dynamic library dependencies. This gave rise to #170 and #626.

Previously, rules_haskell did not properly capture indirect dynamic library dependencies. The reason is that these are not present in `dep.files` as was expected by `gather_dep_info`. This PR changes `gather_dep_info` to instead consult the new C++ API's `CcInfo` provider to obtain all transitive dynamic library dependencies.

On MacOS an additional step is required to get the final Haskell binary to explicitly load all transitive dynamic library dependencies. On MacOS GHC forcibly passes the `--dead_strip_dylibs` flag to the linker. The linker will then strip any indirect or redundant dynamic library dependencies from the binary's load list. This goes directly against Bazel's dependency model. This PR passes `-u` flags to the linker to enforce keeping loads of indirect dynamic library dependencies in the binary's header. First, we obtain a list of all undefined symbols in all transitive dynamic library dependencies. Then we instruct the linker to resolve all of these undefined symbols, effectively turning indirect dynamic library dependencies into direct dynamic library dependencies.

The downside of this approach is that it causes more load commands to appear in the final binary on MacOS, which increases the risk of exceeding the MACH-O header size limit. GHC passes `--dead_strip_dylibs` specifically to decrease that risk. However, it seems that Bazel's approach to handling indirect dynamic library dependencies forces this on us.

### DIFF
--- a/haskell/cc.bzl
+++ b/haskell/cc.bzl
@@ -130,6 +130,7 @@ def cc_interop_info(ctx):
         "cc": cc_toolchain.compiler_executable(),
         "ld": cc_toolchain.ld_executable(),
         "cpp": cc_toolchain.preprocessor_executable(),
+        "nm": cc_toolchain.nm_executable(),
     }
 
     # If running on darwin but XCode is not installed (i.e., only the Command
--- a/haskell/private/actions/link.bzl
+++ b/haskell/private/actions/link.bzl
@@ -20,6 +20,95 @@ def backup_path(target):
 
     return "/".join([".."] * n)
 
+def _merge_parameter_files(hs, file1, file2):
+    """Merge two GHC parameter files into one.
+
+    Args:
+      hs: Haskell context.
+      file1: The first parameter file.
+      file2: The second parameter file.
+
+    Returns:
+      File: A new parameter file containing the parameters of both input files.
+        The file name is based on the file names of the input files. The file
+        is located next to the first input file.
+    """
+    params_file = hs.actions.declare_file(
+        file1.basename + ".and." + file2.basename,
+        sibling = file1,
+    )
+    hs.actions.run_shell(
+        inputs = [file1, file2],
+        outputs = [params_file],
+        command = """
+            cat {file1} {file2} > {out}
+        """.format(
+            file1 = file1.path,
+            file2 = file2.path,
+            out = params_file.path,
+        ),
+    )
+    return params_file
+
+def _darwin_create_extra_linker_flags_file(hs, cc, objects_dir, executable, dynamic, solibs):
+    """Write additional linker flags required on MacOS to a parameter file.
+
+    Args:
+      hs: Haskell context.
+      cc: CcInteropInfo, information about C dependencies.
+      objects_dir: Directory storing object files.
+        Used to determine output file location.
+      executable: The executable being built.
+      dynamic: Bool: Whether to link dynamically or statically.
+      solibs: List of dynamic library dependencies.
+
+    Returns:
+      File: Parameter file with additional linker flags. To be passed to GHC.
+    """
+
+    # On Darwin GHC will pass the dead_strip_dylibs flag to the linker. This
+    # flag will remove any shared library loads from the binary's header that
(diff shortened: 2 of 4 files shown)


mboes:
Nice work!

PR #630, opened 2019-01-24
Add runghc target similar to ghci

shmish111:
I needed to be able to spawn runghc from some haskell and this ghc
must have all the packages that I have built in the project. The
simplest way to do this is to copy the @repl targets.


mboes:
Could you say a little about your use case?

Profpatsch:
Yet another copy of the same bzl files? Now we have repl, doctest, runghc, … all are close copies of each other.

shmish111:
Indeed, some factoring out due. I also realised I can get rid of the wrapper script by pulling out ` -package base -package directory` from the ghci script and putting them in `{ARGS}`.

mboes:
@Profpatsch could you work with @shmish111 to refactor this to avoid the code duplication? There's some logic in here that is making me a little uneasy about code duplication.

mboes:
@shmish111 plausibly. I'll defer to @Profpatsch opinion on this.

Profpatsch:
I’d tentatively merge this for now and add a P2-priority issue that we need to refactor these files before any additional changes are done to them. Otherwise might bite quite a bit in the future.

Issue #611, opened 2019-01-22
Test GHC bindist on Linux on CI


nmattia:
The test suite should be run with the bindist-provided GHC as well

NinjaTrappeur:
Depends on #629

NinjaTrappeur:
I am locally [1] able to run the tests the azure CI is able to run. I'll send a PR as soon the previously linked issue is fixed. We'll need to pin a new rules_nixpkgs commit.

[1] In a vanilla debian Docker image.

NinjaTrappeur:
Closed by #637

PR #637, opened 2019-01-28
Test GHC bindist on Linux on CI

### LINKED ISSUE #629: Cannot load WORKSPACE is nix is not on path on linux/darwin
We are currently assuming that any linux or darwin platform will have nix in the $path. Since we started supporting GHC's bindist, It's not true anymore.

Checks like https://github.com/tweag/rules_haskell/blob/master/haskell/nixpkgs.bzl#L123 prevents us to use the binary distribution system on linux/darwin. We need to refactor these checks while keeping in mind that some linux/darwin users won't be able to run nix.

NinjaTrappeur:
We add a new CI job based on a vanilla Debian docker image (ie. without nix in `$PATH`) testing the ghc_bindist Haskell toolchain.

In order to evaluate the `WORKSPACE` file on linux without having `nix-build` in $PATH, we had to make a couple of changes in the nixpkgs-supported platforms detection.

Note: we'll need to keep the build targets of the build-linux-ghc-bindist circleci job in sync with the azure's ones.

Close #629 and #611

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,28 @@
 version: 2
 
 jobs:
+  build-linux-ghc-bindist:
+    docker:
+      - image: debian
+    working_directory: ~/rules_haskell
+    resource_class: large
+    steps:
+      - checkout
+      - run:
+          name: Setup test environment
+          command: |
+            apt-get update
+            apt-get install -y wget gnupg golang make libgmp3-dev pkg-config zip g++ zlib1g-dev unzip python bash-completion
+            wget "https://github.com/bazelbuild/bazel/releases/download/0.21.0/bazel_0.21.0-linux-x86_64.deb"
+            dpkg -i bazel_0.21.0-linux-x86_64.deb
+            echo "common:ci --build_tag_filters -requires_hackage,-requires_zlib,-requires_doctest,-requires_c2hs,-requires_threaded_rts,-requires_missing_compiler_flags,-ghcbindist_th_failure" > .bazelrc.local
+      - run:
+          name: Build tests
+          command: |
+            bazel build --config ci //tests/...
 
   # ATTN: when you change anything here, don’t forget to copy it to the build-darwin section
-  build-linux:
+  build-linux-nixpkgs:
     docker:
       - image: nixos/nix:2.1.3
     working_directory: ~/rules_haskell
@@ -73,5 +92,6 @@ workflows:
   version: 2
   build:
     jobs:
-      - build-linux
+      - build-linux-ghc-bindist
+      - build-linux-nixpkgs
       - build-darwin
--- a/haskell/nixpkgs.bzl
+++ b/haskell/nixpkgs.bzl
@@ -99,6 +99,9 @@ def haskell_nixpkgs_packages(name, base_attribute_path, packages, **kwargs):
         base_repository = name,
     )
 
+def _is_nix_platform(repository_ctx):
+    return repository_ctx.which("nix-build") != None
+
 def _gen_imports_impl(repository_ctx):
     repository_ctx.file("BUILD", "")
     extra_args_raw = ""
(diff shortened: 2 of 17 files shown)


.circleci/config.yml
```diff
@@ -1,9 +1,32 @@
...
+          name: Install Bazel
+          command: |
+            apt-get update
+            apt-get install -y curl gnupg golang make libgmp3-dev
+            echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
+            curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
+            apt-get update
+            apt-get install -y bazel
+      - run:
+          name: Build tests
```
mboes: Don't copy the test script from Windows (which is a WIP target with only few tests passing)! Copy it from the other build jobs in this very file (Linux and Darwin).

haskell/nixpkgs.bzl
```diff
@@ -99,6 +99,9 @@ def haskell_nixpkgs_packages(name, base_attribute_path, packages, **kwargs):
         base_repository = name,
     )
 
+def _is_nix_platform(repository_ctx):
+    return repository_ctx.execute(["bash", "-c", "type -p nix-build"]).return_code == 0
```
mboes: This will need a justification inline as to why it's not just `repository_ctx.which`.

.circleci/config.yml
NinjaTrappeur: I have not managed to build any other tests using the `rules_haskell` master using the bindist toolchain. I though it was an expected behaviour as we get most of the tests hackage deps using the nixpkgs.


See the error logs at:

https://circleci.com/gh/tweag/rules_haskell/3646

NinjaTrappeur:
Alright, the Bazel team just migrated the APT repository to the new stable bazel release: bazel 0.22. This change breaks this PR, `rules_haskell` does not seems to be compatible with our current codebase.

I guess we need to address these breaking changes before merging this PR. I'm closing it for now.

Note to @mboes: I addressed your remarks on a local branch.

I'll re-open this after moving `rules_haskell` to Bazel 0.22.

NinjaTrappeur:
```
root@8c05c7426914:/# apt-cache policy bazel
bazel:
  Installed: (none)
  Candidate: 0.22.0
  Version table:
     0.22.0 500
        500 http://storage.googleapis.com/bazel-apt stable/jdk1.8 amd64 Packages
```

Looks like they only put the latest version in the stable apt repo.

mboes:
We don't need to use their APT repo. They have binary distributions too.

NinjaTrappeur:
I managed to pin bazel to the 0.21 version by directly downloading the .deb package.

As I mentioned before, we cannot build/test `rules_haskell` without the nixpkgs. See #641 .

What should I do:

1. Stall this PR until #641 is resolved.
2. Only test what's tested on azure. (these tests can succeed)

mboes:
Your assessment that only the Azure tests work is incorrect. The error you are seeing regards those parts of the test suite that do depend on Hackage packages being furnished by Nixpkgs (since we don't use Hazel yet). Essentially, all tests should pass except the c2hs and protobuf ones. You can disable those via tags and the `--test_tag_filters` command-line flag. This classification will be useful for Windows too, where we likewise are not interested in testing c2hs and protobuf at this time.

NinjaTrappeur:
I added some flags and filtered the tests according them as advised. However, some test actions seems to fail: https://circleci.com/gh/tweag/rules_haskell/3702?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link .

According to blame, @gdeest recently re-wrote those. Do you have an idea about what's the problem here?

mboes:
That test depends on the toolchain being setup the way it is in Nixpkgs. See what's happening here: https://github.com/tweag/rules_haskell/blob/master/WORKSPACE#L85. `ghc_bindist` does not support `compiler_flag` yet (although it should). I'd say disable this test as well for now, and then add the required support to GHC bindist for custom flags in a separate PR, reenabling this test.

NinjaTrappeur:
This PR is ready for review.

Two small notes:

1. I disabled the tests relying on the nixpkgs ghc toolchain flags on the ghc bindist toolchain. I'll shortly create open another PR fixing these problems.
1. There's a strange TH-related issue (https://github.com/tweag/rules_haskell/pull/637/files#diff-ce4ec027cf2123f505d222d92aaab7daR17 ). I'm not really sure if this is linked to the toolchain flags. I'll open an issue and investigate more on this after we'll merge this PR.

PR #652, opened 2019-01-29
Fixing bindist CI after #642 merge.

NinjaTrappeur:
The #642 and #637 prs have been merged quickly. Basically #642 breaks
`tests:test-cc_haskell_import-output` which was not tested neither by
circleci or azure.

This commit fixes this test.


NinjaTrappeur:
Huuum, we don't use the same GHC version on bindists anymore. That's actually not that simple.

NinjaTrappeur:
Clearly, we don't want to use GHC 8.6.2 on nixpkgs.

Disabled the test with the `ghc_bindist` toolchain.

PR #656, opened 2019-01-29
Enable more passing tests on Windows CI

nmattia:
@NinjaTrappeur I think this is related to some of the work you're doing


NinjaTrappeur:
More or less, I went the other way around: blacklisting the tests that do not succeed though the `--build_tag_filters` as advised by @mboes .

The tags are now on master, you probably can also use them on azure.

mboes:
This will do for now. We can switch to the tags once more tests pass.

PR #662, opened 2019-01-29
Expose compiler flags on ghc bindist toolchain

NinjaTrappeur:
Exposing the compiler flags on the ghc bindist toolchain.
Forwarding these  flags exposed to the `haskell_toolchain` generated rules.

We can now build the tests needing some custom compiler flags on the
bindist CI. 

Removing the now useless `requires_missing_compiler_flags` tags.


mboes:
This is literally just forwarding two extra arguments to the rule. No other change. I don't think it's useful to aim for 100% test coverage. Even if you did want that, you could always add a simpler haddock test that doesn't attempt to check that Cabal packages coming from Nix are haddocked properly.

PR #663, opened 2019-01-29
 Use CcInfo provider for C library dependencies

aherrmann:
- Extract information on C library dependencies from Bazel's builtin  CcInfo provider. This allows to benefit from Bazel's central `_solib`  directory which holds all dynamic C library dependencies, and thereby  allows to reduce the amount of RUNPATH entries. This also allows to  benefit from Bazel's builtin library name mangling without having to   reimplement it in rules_haskell
- Accordingly, this PR removes the C library name mangling implementation in rules_haskell.
- Bazel only mangles names of dynamic library dependencies. If a library  dependency is available as both static and dynamic library, then each  will have a different "mangled" name. However, GHC expects both  static and dynamic libraries to be available under the same name.  The reason is that library dependencies have to be listed in the  extra-libraries section of the package configuration file. To link the  final binary statically, we only provide static archives at linking.  However, for GHC's own linker (TemplateHaskell, REPL, etc.) we require  dynamic libraries. Both have to match the entries in extra-libraries.  We achieve this by manually mangling the static archives based on  Baezl's builtin mangling if the static and dynamic library "mangled" names differ.
- On MacOS GHC's builtin linker hard-codes the expectation that any  dynamic library dependencies are `.dylib` files. However, Bazel  produces `.so` files. To work around this issue we create  symlinks to adapt any `.so` dependencies as `.dylib` dependencies.  Note, that this is only required when interacting with GHC's builtin  linker. I.e. for compilation, doctests, REPL. In particular, this is  not required for the final linking. This is good news as it allows us  to still benefit from Bazel's central `_solib` directory on MacOS.
- Adds a regression test for indirect cc_library system library dependencies. Previously these did not work due to manual name mangling.


haskell/private/providers.bzl
```diff
@@ -16,6 +16,70 @@ DefaultCompileInfo = provider(
     },
 )
 
+HaskellCcInfo = provider(
+    doc = "Haskell cc dependency information. Part of HaskellBuildInfo.",
+    fields = {
+        "static_linking": "static linking mode parameters. " +
```
mboes: Use triple quotes here.

mboes:
@guibou could you test this on your codebase before we merge?

aherrmann:
Not quite there, yet.

This seems to trigger segfaults in `happy`, when compiling with GHC > 8.2.2. Probably related to `-fPIC` behaviour.

guibou:
@mboes test in progress

aherrmann:
I've looked into the `happy` segfault. This turns out to be an issue in Hazel. Hazel always links against the threaded runtime in cbits. Now that transitive C library dependencies are tracked properly by rules_haskell, this dependency ends up being forwarded into the happy binary. `happy` ends up being linked against both the threaded and non-thread rts and that causes the segfault at runtime.

A simple workaround in Hazel is to only include the rts headers into cbits targets, not the library.

I'll address @mboes 's comments, and then merge this PR, and fix the Hazel issue separately.

Profpatsch:
I’ll review tomorrow.

haskell/private/providers.bzl
```diff
@@ -27,19 +91,22 @@ HaskellBuildInfo = provider(
...
-        "extra_libraries": "The direct non-Haskell dependencies of this target. These are added to the extra-libraries field in the package's database.",
+        "cc_dependencies": "Direct cc library dependencies. See HaskellCcInfo.",
+        "transitive_cc_dependencies": "Transitive cc library dependencies. See HaskellCcInfo.",
+        "import_dependencies": "Direct haskell_cc_import library dependencies.",
+        "transitive_import_dependencies": "Transitive haskell_cc_import library dependencies.",
     },
 )
 
-def external_libraries_get_mangled(ext_lib):
+def get_mangled_lib(ext_lib):
```
Profpatsch: Unused now

haskell/private/actions/link.bzl
```diff
@@ -259,19 +268,40 @@ def link_binary(
...
+            trans_link_ctx.libraries_to_link.to_list() + trans_import_libs,
+        )
+
+    # Transitive dynamic library dependencies to have in RUNPATH.
+    trans_cc_solibs = trans_link_ctx.dynamic_libraries_for_runtime.to_list()
+
+    # Configure RUNPATH.
+    solibs = trans_cc_solibs + trans_import_libs + set.to_list(dep_info.dynamic_libraries)
+    for rpath in set.to_list(_infer_rpaths(hs.toolchain.is_darwin, executable, solibs)):
+        args.add("-optl-Wl,-rpath," + rpath)
```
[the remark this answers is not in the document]
aherrmann: Done

haskell/private/providers.bzl
aherrmann: Done

haskell/private/providers.bzl
aherrmann: Removed.

aherrmann:
Rebased and addressed all comments.

haskell/private/actions/link.bzl
```diff
@@ -520,36 +559,14 @@ def link_library_dynamic(hs, cc, dep_info, extra_srcs, objects_dir, my_pkg_id):
...
-    cc_solibs = trans_link_ctx.dynamic_libraries_for_runtime.to_list()
-
-    # Configure RUNPATH.
-    solibs = cc_solibs + trans_import_libs + set.to_list(dep_info.dynamic_libraries)
-    for rpath in set.to_list(_infer_rpaths(hs.toolchain.is_darwin, dynamic_library, solibs)):
-        args.add("-optl-Wl,-rpath," + rpath)
+    (cc_link_libs, _cc_solibs) = _link_dependencies(
+        hs,
+        dep_info,
+        True,  # dynamic linking mode
```
Profpatsch: You can use named arguments in invocations in skylark, `dynamic = True`.

I’d strongly suggest we switch all calls to explicitly name arguments.

PR #682, opened 2019-01-31
Test indirect cbits on Windows

nmattia:


gdeest:
Will be addressed in an incoming PR along with many other tests.

PR #685, opened 2019-02-05
Fix uniplate on MacOS

aherrmann:
The Hazel build of the package `uniplate` fails on MacOS due to failure of GHC's builtin linker to load the dynamic library of the `text` package's `cbits`. This PR resolves that issue.

- Before calling `dlopen` GHC tries to determine the location of library dependencies within directories listed in `LIBRARY_PATH`. This PR defines `LIBRARY_PATH` to list all directories holding static or dynamic library dependencies. On MacOS `LIBRARY_PATH` will point to the `.dylib` versions of `.so` library dependencies, since GHC assumes that dynamic libraries end on `.dylib` on MacOS.
- In some cases GHC will generate and load intermediate dynamic libraries (e.g. `/run/user/1000/ghc2_0/libghc_6.so`). On MacOS these libraries will be loaded before rules_haskell had an opportunity to fix-up the load commands in that library. Therefore, they will reference dynamic C library dependencies by their "library name", i.e. their unmangled name. To make these libraries visible to the dynamic loader, we expose their locations in `LD_LIBRARY_PATH`. Note, that this does not influence linking of the final binary.
- Unfortunately, it is not clear to me how `uniplate` triggers this issue, while none of the tests in rules_haskell do. I haven't found a regression test for this issue.


haskell/doctest.bzl
```diff
@@ -187,6 +189,7 @@ def _haskell_doctest_single(target, ctx):
         # in this case.
         env = dicts.add(
             {
+                "LIBRARY_PATH": library_path,
```
Profpatsch: Is `LIBRARY_PATH` a magic env variable used by GHC’s internal linker? If so, it should be documented somewhere.

haskell/private/providers.bzl
```diff
@@ -124,20 +135,38 @@ def get_solibs_for_ghc_linker(hs, build_info):
...
+        dynamic loading.
     """
     trans_link_ctx = build_info.transitive_cc_dependencies.dynamic_linking
-    trans_libs = get_mangled_libs(trans_link_ctx.libraries_to_link.to_list())
-    trans_import_libs = set.to_list(build_info.transitive_import_dependencies)
 
-    _library_deps = trans_libs + trans_import_libs
+    unmangled_libs = get_unmangled_libs(trans_link_ctx.libraries_to_link.to_list())
+    mangled_libs = get_mangled_libs(trans_link_ctx.libraries_to_link.to_list())
+    import_libs = set.to_list(build_info.transitive_import_dependencies)
```
Profpatsch: In this case it might be slightly faster do do both in one manual list comprehension, this way you only do one pass.

haskell/doctest.bzl
aherrmann: It's documented in the [GHCi user-guide](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/ghci.html?highlight=library_path#extra-libraries).
Not too clearly though. In this case setting `LD_LIBRARY_PATH` alone was not sufficient.

haskell/private/providers.bzl
aherrmann: Hmm, I don't know how to create two lists out of a single list comprehension. But, I've turned it into a single traversal.

aherrmann:
@Profpatsch 
> These duplicate code changes in each PR are getting ridiculous. :(

Yes, I'm not too happy about that either. I've factored out the generation of `(LD_)LIBRARY_PATH`.

Profpatsch:
> Yes, I'm not too happy about that either

It’s not from your PRs, but the result of our ambient duplication. :)

PR #687, opened 2019-02-06
Fix module name in package conf file on Windows

gdeest:
From time to time, but not always for some unknown reason, exposed module names in packages `.conf` files would end up with `\` instead of `.` on Windows, because of the different path segment separator.

This small PR fixes this issue.

### DIFF
--- a/haskell/private/ls_modules.py
+++ b/haskell/private/ls_modules.py
@@ -69,6 +69,8 @@ interface_files = (
 )
 
 modules = (
+    # replace directory separators by . to generate module names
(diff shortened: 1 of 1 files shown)


Profpatsch:
Is that a bug in the upstream `skylib` library paths module?

gdeest:
I don't think so, as it is generated by the `ls_modules.py` Python executable. Both `/` and `\` are valid path separators on Windows it seems, with `\` being the default, so I wouldn't necessarily call that a bug anyway.

guibou:
LGTM, have a look at my comment suggestion and merge.

Issue #639, opened 2019-01-28
Fix incompatible_disable_legacy_crosstool_fields


NinjaTrappeur:
It prevents us to build `rules_haskell` using bazel 0.22.

```
ERROR: /root/.cache/bazel/_bazel_root/d00c9458fa4b9d008dbcdeefe3e0fc7c/external/local_config_cc/BUILD:47:1: in cc_toolchain_suite rule @local_config_cc//:toolchain: compilation_mode_flags is disabled by --incompatible_disable_legacy_crosstool_fields, please migrate your CROSSTOOL (see https://github.com/bazelbuild/bazel/issues/6861 for migration instructions).
ERROR: /root/.cache/bazel/_bazel_root/d00c9458fa4b9d008dbcdeefe3e0fc7c/external/local_config_cc/BUILD:47:1: in cc_toolchain_suite rule @local_config_cc//:toolchain: linking_mode_flags is disabled by --incompatible_disable_legacy_crosstool_fields, please migrate your CROSSTOOL (see https://github.com/bazelbuild/bazel/issues/6861 for migration instructions).
ERROR: /root/.cache/bazel/_bazel_root/d00c9458fa4b9d008dbcdeefe3e0fc7c/external/local_config_cc/BUILD:47:1: in cc_toolchain_suite rule @local_config_cc//:toolchain: compiler_flag is disabled by --incompatible_disable_legacy_crosstool_fields, please migrate your CROSSTOOL (see https://github.com/bazelbuild/bazel/issues/6861 for migration instructions).
ERROR: /root/.cache/bazel/_bazel_root/d00c9458fa4b9d008dbcdeefe3e0fc7c/external/local_config_cc/BUILD:47:1: in cc_toolchain_suite rule @local_config_cc//:toolchain: cxx_flag is disabled by --incompatible_disable_legacy_crosstool_fields, please migrate your CROSSTOOL (see https://github.com/bazelbuild/bazel/issues/6861 for migration instructions).
ERROR: /root/.cache/bazel/_bazel_root/d00c9458fa4b9d008dbcdeefe3e0fc7c/external/local_config_cc/BUILD:47:1: in cc_toolchain_suite rule @local_config_cc//:toolchain: linker_flag is disabled by --incompatible_disable_legacy_crosstool_fields, please migrate your CROSSTOOL (see https://github.com/bazelbuild/bazel/issues/6861 for migration instructions).
[...]

PR #691, opened 2019-02-06
Remove --all_incompatible_changes

guibou:
This comment https://github.com/bazelbuild/bazel/issues/6861#issuecomment-460145291 and this process https://docs.bazel.build/versions/master/backward-compatibility.html detail the incompatible change process.

`--all_incompatible_changes` is not the right setting for us because it will:

- enable changes which may never be set by default, hence forcing us to do useless work
- enable changes which may be set by default, but at a time in which the rest of the bazel ecosystem is not ready yet. See for example #639.

Future aware developers of rules_haskell needs to track the bazel issue tracker for possible incompatible changes and decide if they justify preliminary work. They may also occasionally enable `--all_incompatible_changes` to test, but it should not be the default.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -24,10 +24,8 @@ test:windows --experimental_enable_runfiles
 # We also need to setup an utf8 locale
 test --test_env=LANG=en_US.utf8 --test_env=LOCALE_ARCHIVE
 
-# Making incompatible changes fail.
-#
 # ToRemove: remove the three exceptions when they'll stop failing
 # on Google's protobuf_rules.
-test --all_incompatible_changes  --incompatible_disable_deprecated_attr_params=false --incompatible_new_actions_api=false --incompatible_expand_directories=false
+test --incompatible_disable_deprecated_attr_params=false --incompatible_new_actions_api=false --incompatible_expand_directories=false
 
 try-import .bazelrc.local


guibou:
close #639

.bazelrc
```diff
@@ -24,10 +24,8 @@ test:windows --experimental_enable_runfiles
 # We also need to setup an utf8 locale
 test --test_env=LANG=en_US.utf8 --test_env=LOCALE_ARCHIVE
 
-# Making incompatible changes fail.
-#
 # ToRemove: remove the three exceptions when they'll stop failing
 # on Google's protobuf_rules.
-test --all_incompatible_changes  --incompatible_disable_deprecated_attr_params=false --incompatible_new_actions_api=false --incompatible_expand_directories=false
+test --incompatible_disable_deprecated_attr_params=false --incompatible_new_actions_api=false --incompatible_expand_directories=false
```
NinjaTrappeur: I'm late to the party but I think we don't need all those `--incompatible_*` then.

NinjaTrappeur: https://github.com/tweag/rules_haskell/pull/692

guibou: You are right. My PR was half-baked. Sorry.

Profpatsch:
Huh, I’m a bit confused, because in https://github.com/tweag/rules_haskell/issues/489 @laurentlb made it look like setting `--all_incompatible_changes` was the way to go to be future-proof.
If we can’t support something, we add a flag that disables the incompatible change for the time being, then we have an explicit list of things we need to fix before the next bazel release.

See also my comment on the linked PR comment by @hlopko: https://github.com/bazelbuild/bazel/issues/6861#issuecomment-461340744

hlopko:
Answered on bazelbuild/bazel#6861.

PR #694, opened 2019-02-08
Improvements to the linking utils debugging library

Profpatsch:
These were done to make sense of more intricate errors when working on https://github.com/tweag/rules_haskell/pull/614.

- For the client codebase the naïve way of working without memoization was too slow, and the call to `objdump` could be sped up a lot as well.
- `rules_haskell` uses python2, while the client codebase used python3, so add a small amount of compatibility functionality
- Added a test which actually uncovers problems with following symlinks the way `ld.so` does (ldd has the same problem; what do you do when your debugging tool is buggy …)
- Added a motivating example and tutorial to the `README`, so hopefully other people will be able to use it, too.

The most important feedback for me is whether the tutorial makes sense.

### DIFF
--- a/debug/linking_utils/README.md
+++ b/debug/linking_utils/README.md
@@ -1,52 +1,265 @@
 # Debugging linking errors
 
-The usual utilties like `nm`, `objdump` and of course `ldd` (see
+The usual utilities, like `nm`, `objdump`, and of course `ldd` (see
 [here](https://linux-audit.com/elf-binaries-on-linux-understanding-and-analysis/#tools-for-binary-analysis)
-for a good overview of existing tools) go a long way, but when
-debugging non-trivial runtime linker failures one would often like to
-filter outputs programmatically with more than just simple `grep` and
-`sed` expressions.
+for a good overview of existing tools) go a long way. Yet, when
+debugging non-trivial runtime linker failures one would oftentimes
+like to filter outputs programmatically, with more advanced query
+logic than just simple `grep` and `sed` expressions.
 
-This library provides a small set of utility subroutines that can help
-debug complicated linker errors.
+This library provides a small set of utility subroutines. These can
+help debug complicated linker errors.
 
-The main function is `ldd(f, elf_path)`, which is in the same spirit
-as `ldd(1)`, but returns a tree of shared dependencies instead of a
-flat list. Additionally, it expects a function `f` which is applied to
-each recursion level of dependencies.
+The main function is `ldd(f, elf_path)`. It is in the same spirit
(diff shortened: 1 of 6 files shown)


debug/linking_utils/README.md
```diff
@@ -1,51 +1,265 @@
...
-as `ldd(1)`, but returns a tree of shared dependencies instead of a
-flat list. Additionally, it expects a function `f` which is applied to
-each recursion level of dependencies.
+The main function is `ldd(f, elf_path)`. It is in the same spirit
+as `ldd(1)`, but instead of a flat list of resolved libraries, it
+returns a tree of structured information.
 
-Functions that can be passed to `ldd`:
+When we use the term `ldd` in the following document, it refers
+to the `ldd` function exported from [./ldd.py][].
```
aherrmann: Incomplete link.

debug/linking_utils/README.md
```diff
@@ -1,51 +1,265 @@
 # Debugging linking errors
 
-The usual utilties like `nm`, `objdump` and of course `ldd` (see
+The usual utilties, like `nm`, `objdump`, and of course `ldd` (see
```
guibou:
```suggestion
The usual utilities, like `nm`, `objdump`, and of course `ldd` (see
```

debug/linking_utils/ldd.py
```diff
@@ -1,9 +1,18 @@
 import subprocess
 import os
+import sys
+import re
 
 
 ### helper functions
 
+if sys.version_info > (3, 0):
```
guibou: Why not just `d.items()`? Is the performance difference worth the special case?

debug/linking_utils/ldd.py
```diff
@@ -29,31 +38,75 @@ def f(y):
         return x
     return f
 
+def memoized(cache, f, arg):
```
guibou: FYI there is, in python3, `functool.lru_cache`.

debug/linking_utils/ldd.py
Profpatsch: Does `.items()` work in both python2 and python3?

debug/linking_utils/ldd.py
Profpatsch: I saw that, but I want to be compatible with 2 and 3

mboes:
CI red. Fix CI and then good to merge.

Profpatsch:
Only the `bindists` build is failing, because the upstream server from `downloads.haskell.org` is returning a 0-byte file …

Profpatsch:
Pinged the people at #haskell-infrastructure, they should have fixed the binary.

Issue #690, opened 2019-02-06
Refactor the similarities in haskell/private/actions/


Profpatsch:
Some of the code in `haskell/private/actions/` has extreme similarities, ostensibly because it was mostly copied and pasted at the time.

For example:

```
$ nix-shell -p pmd jre --run 'run.sh cpd --minimum-tokens 50 --files haskell/private/actions/* --language python'
Found a 34 line (149 tokens) duplication in the following files: 
Starting at line 53 of /home/philip/kot/bazel/rules_haskell/haskell/private/actions/repl.bzl
Starting at line 47 of /home/philip/kot/bazel/rules_haskell/haskell/private/actions/runghc.bzl

    args += expose_packages(
        build_info,
        lib_info,
        use_direct = False,
        use_my_pkg_id = None,
        custom_package_caches = package_caches,
        version = version,
    )

    if lib_info != None:
        for idir in set.to_list(lib_info.import_dirs):
            args += ["-i{0}".format(idir)]

    link_ctx = build_info.cc_dependencies.dynamic_linking
    libs_to_link = link_ctx.dynamic_libraries_for_runtime.to_list()
    import_libs_to_link = set.to_list(build_info.import_dependencies)

    # External shared libraries that we need to make available to the REPL.
    # This only includes dynamic libraries as including static libraries here
    # would cause linking errors as ghci cannot load static libraries.
    # XXX: Verify that static libraries can't be loaded by GHCi.
    seen_libs = set.empty()
    for lib in libs_to_link + import_libs_to_link:
        lib_name = get_lib_name(lib)
        if is_shared_library(lib) and not set.is_member(seen_libs, lib_name):
            set.mutable_insert(seen_libs, lib_name)
            args += ["-l{0}".format(lib_name)]

    # Transitive library dependencies to have in runfiles.
    library_deps = get_solibs_for_ghc_linker(hs, build_info)
    ld_library_path = make_path(
        library_deps,
        prefix = "$RULES_HASKELL_EXEC_ROOT",
    )
```

Though there is less some obvious duplication as well.
See the discussion in https://github.com/tweag/rules_haskell/pull/630.

Profpatsch:
There’s also some copypaste in `protobuf.bzl` and `doctest.bzl`:

```
> nix-shell -p pmd jre --run 'run.sh cpd --minimum-tokens 50 --files **/*.bzl --language python'
Found a 17 line (77 tokens) duplication in the following files: 
[...]

Issue #380, opened 2018-07-30
Multi-package repl


regnat:
Rules haskell provides `-repl` targets to open custom repls, so that if I have a target `:foo`, I can `bazel run //:foo-repl` to open a ghci session with all of `foo` available (like `stack repl foo`). However if I have two targets `:foo` and `:bar`, there's no way to open a ghci session with both `foo` and `bar` (like `stack repl foo bar` would do).

shmish111:
To be clear, what would be great is if all dependencies that are themselves workspace targets would be loaded directly from their source code rather than from compiled targets. 

I think this would mean that if I am working on `Bar.hs` in haskell_library `A` that depends on `Foo.hs` in haskell_library `B` and I change `Foo` then when I `:r`, the changes that occurred in `Foo` are available to `Bar`. At least that's the behaviour that I am looking for for a development workflow.

shmish111:
Additionally, it would be great to have the option to not load any files initially, stack has this option. This is because in an IDE situation, you want the repl to load as quickly as possible and you probably don't need it to compile everything in your project. Instead you will load one file and that will trigger all the dependencies in it's dependency tree to be compiled only.

dajmaki:
I did a quick try to see if I could use the existing repl mechanisms to load sources of multiple packages:
https://github.com/dajmaki/rules_haskell/tree/experiment-haskell-repl

With this you define a repl target:
```haskell_repl(name = "combined", deps = ["//:libraryA", "//:libraryB"])```.
This creates a ghci script that loads the sources of both libraryA and libraryB, and package dependencies from both should be loaded. I haven't tested it heavily and don't know what the shortcomings are.

Thoughts on this approach? Any obvious issues?

dajmaki:
It loads the source files (though looks like the package dbs for both are passed to ghci as well so it does need a bit more work...)

PR #695, opened 2019-02-08
Generate a json file with ghc build info for IDE tools

u-quark:
We want to be able to expose information that IDE tools need
and typically obtain from cabal files. Our current use case
for this is to be able to run ghci (and ghcid) over multiple
packages (as in #380).

What we do here is, on the `@repl` target, we write not only
the ghci runner script, but also a json file that contains
all the package-dependent information (e.g., path to ghci,
environment variables, packages, compiler flags, etc). In
order to run ghci on multiple targets, we use a script that
essentially collects all the json files for the packages 
mentioned, takes the union of all their arguments, and calls
ghci with that. See this example:
https://gist.github.com/u-quark/e58cf5e3b2d9cdeb7ea1b4e8ccc21547.

For this PR, we tried to keep the changes to a bare minimum,
but it would probably make sense to later change the ghci
and runghc wrappers to just take the required information
from the json file, which should help reducing the current
code duplication (#690).

### DIFF
--- a/haskell/lint.bzl
+++ b/haskell/lint.bzl
@@ -8,7 +8,7 @@ load(
     "HaskellLintInfo",
 )
 load(":private/context.bzl", "haskell_context")
-load(":private/packages.bzl", "expose_packages")
+load(":private/packages.bzl", "expose_packages", "pkg_info_to_ghc_args")
 load(
     ":private/path_utils.bzl",
     "target_unique_name",
@@ -54,14 +54,14 @@ def _haskell_lint_aspect_impl(target, ctx):
         "--make",
     ])
 
-    args.add_all(expose_packages(
+    args.add_all(pkg_info_to_ghc_args(expose_packages(
         build_info,
         lib_info,
         use_direct = False,
         use_my_pkg_id = None,
         custom_package_caches = None,
         version = ctx.rule.attr.version,
-    ))
+    )))
 
     sources = set.to_list(
         lib_info.source_files if lib_info != None else bin_info.source_files,
--- a/haskell/private/actions/repl.bzl
+++ b/haskell/private/actions/repl.bzl
@@ -59,9 +59,11 @@ def build_haskell_repl(
     )
     args += pkg_info_to_ghc_args(pkg_ghc_info)
 
+    lib_imports = []
     if lib_info != None:
         for idir in set.to_list(lib_info.import_dirs):
             args += ["-i{0}".format(idir)]
+            lib_imports.append(idir)
 
     link_ctx = build_info.cc_dependencies.dynamic_linking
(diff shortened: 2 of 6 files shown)


haskell/private/actions/repl.bzl
```diff
@@ -133,6 +140,37 @@ def build_haskell_repl(
...
+            "repl": repl_ghci_args,
+            "toolchain_repl": hs.toolchain.repl_ghci_args,
+        },
+    }
+    ghc_info_file = hs.actions.declare_file(
+        target_unique_name(hs, "ghc-info"),
+    )
+    hs.actions.write(
+        output = ghc_info_file,
+        content = str(ghc_info),
```
Profpatsch: Ceci n’est pas une JSON string. :smile: 
Instead of a `dict`, you need to use a `struct` (which is a `dict` with immutable fields) and use its [`to_json()` function](https://docs.bazel.build/versions/master/skylark/lib/struct.html#to_json).

haskell/lint.bzl
```diff
@@ -54,14 +54,15 @@ def _haskell_lint_aspect_impl(target, ctx):
         "--make",
     ])
 
-    args.add_all(expose_packages(
+    pkg_args, _ = expose_packages(
```
Profpatsch: Ignoring the second return value is a code smell, see comment below.

haskell/private/packages.bzl
```diff
@@ -54,10 +58,19 @@ def expose_packages(build_info, lib_info, use_direct, use_my_pkg_id, custom_pack
...
+        package_dbs.append(cache.dirname)
 
-    return args
+    ghc_info = {
+        "has_version": has_version,
+        "packages": packages,
+        "package_ids": package_ids,
+        "package_dbs": package_dbs,
+    }
+    return args, ghc_info
```
Profpatsch: From what I can see, all info that is added to `ghc_info` here is also appended to `args` in the previous lines.

So it makes more sense to actually split this function into two incremental functions: the first collects `ghc_info` and the second appends the information to args.

This way we can assign the `ghc_info` when it is needed and don’t have to ignore `ghc_info` in most calls.

haskell/private/actions/repl.bzl
u-quark: Yeap, just found about this an hour ago. Will fix!

haskell/private/packages.bzl
u-quark: Yeap, makes sense. Will fix.

u-quark:
@Profpatsch fixed the issues you mentioned.

u-quark:
@Profpatsch Thanks!

PR #697, opened 2019-02-11
Test start script

Profpatsch:
First commit is only a refactor. Second one changes start script to have a working example haskell binary, last one adds a test.

### DIFF
--- a/tests/start-script/BUILD
+++ b/tests/start-script/BUILD
@@ -1,0 +1,9 @@
+load(
+    "//:bazel/inline_tests.bzl",
+    "sh_inline_test",
+)
+
+sh_test(
+    name = "start-script",
+    srcs = ["@start_script_repository_dummy//:hello_world"],
+)
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -34,6 +34,10 @@ load(
     "haskell_package_repository_dummy",
 )
 load(
+    "@io_tweag_rules_haskell//tests/start-script:start_script_repository_dummy.bzl",
+    "start_script_repository_dummy",
+)
+load(
     "@io_tweag_rules_haskell//:constants.bzl",
     "bindists_ghc_version",
     "ghc_version",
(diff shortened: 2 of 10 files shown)


tests/start-script/BUILD
```diff
@@ -0,0 +1,9 @@
+load(
+    "//:bazel/inline_tests.bzl",
+    "sh_inline_test"
```
NinjaTrappeur: Do we really need `sh_inline_test` in scope here?

Profpatsch: We don’t, if you can find another way to make the start script build a Haskell binary without writing the source file to the user’s repository

mboes:
A lot of extra test weight here, including putting litter in the user's repository. Can you state the problem you're trying to fix?

Profpatsch:
> including putting litter in the user's repository

Huh, where?

mboes:
> Huh, where?

My bad - misintepretation of a comment your wrote above.

> The `start` script broke in the past, because the `rules_haskell` API changed.

We already test the fetch [here](https://github.com/tweag/rules_haskell/blob/master/tests/RunTests.hs#L52-L69). Can't we extend that to test the build?

Profpatsch:
> We already test the fetch [here](https://github.com/tweag/rules_haskell/blob/master/tests/RunTests.hs#L52-L69). Can't we extend that to test the build?

Unless the fetch tests some different state, this tests is a complete superset of that test, running inside of the `bazel test //...` instead of a different bazel run.

mboes:
And also adding extra test related code that we can do without.

Profpatsch:
So after talking to @mboes, the way we go forward with this is the following: Bazel sets the current repository boundary to where it finds another `WORKSPACE` file. This means you can use `local_repository` to import a subdirectory (`c2hs` already does that).

The plan forward:
1. Change the start script to be an a workspace directory.
1. test this workspace
1. Generate the `start` script from these files

We can use the same strategy to migrate https://github.com/tweag/rules_haskell_examples back into the `rules_haskell` repository.

PR #699, opened 2019-02-12
Generating coverage output for Haskell tests

SebastianKG:
This is my work towards a code coverage solution for Haskell in Bazel. Running this is done best in this form:

`bazel coverage //tests/two-libs:two-libs -s --test_output=all`

This will show you the coverage report of the chosen `haskell_test` rule. This is not the standard Bazel coverage report, but instead a custom-made solution to get around the current restrictions for `bazel coverage`'s report-collection capabilities.

However, the above won't do anything on its own. To get a coverage report for a `haskell_test`, you must establish a required base level of coverage, like this:

```
haskell_test(
    name = "two-libs",
    srcs = ["Main.hs"],
    expected_expression_coverage = 80,
    deps = [
        ":two",
        "//tests/hackage:base",
    ],
)
```

This will require expression coverage >= 80%. Only expression coverage is currently supported, for simplicity.

I have noted that not all of the test rules pass yet with this new functionality, but I don't know if immediately handling complex cases is the best approach. My proposal is that we incrementally improve the flexibility of Haskell code coverage in Bazel. After all, it has worked for 0% of the test cases until now. `two-libs` is an example of a test case which works.


haskell/private/haskell_impl.bzl
```diff
@@ -58,13 +60,42 @@ def haskell_test_impl(ctx):
...
     return hpc_outputs
 
+def _should_inspect_coverage(ctx, hs, is_test):
+    return hs.coverage_enabled and is_test and ctx.attr.expected_expression_coverage > 0
+
+def _wrap_binary_for_coverage(binary_path, hpc_path, tix_file_path, expected_expression_coverage, mix_file_paths):
+    return runfiles_boilerplate + """
+ERRORCOLOR='\033[1;31m'
+CLEARCOLOR='\033[0m'
+binary_path=$(rlocation {})
```
guibou: could you use format with explicit names:
```
""" foo bar {baz} spam eggs""".format(baz=something) """
```

I prefer that for readability instead of positional arguments.

haskell/private/haskell_impl.bzl
```diff
@@ -179,14 +217,41 @@ def _haskell_binary_common_impl(ctx, is_test):
         bin_info = bin_info,
     )
 
+    executable = binary
+
+    if inspect_coverage:
+        binary_path = ctx.workspace_name + "/" + binary.short_path
```
guibou: Use path building tools, such as `paths.join`. I don't know how this affect portability, but that's the way to do it.

haskell/private/haskell_impl.bzl
```diff
@@ -179,14 +217,41 @@ def _haskell_binary_common_impl(ctx, is_test):
...
+        hpc_path = ctx.workspace_name + "/" + hs.toolchain.tools.hpc.short_path
+        tix_file_path = hs.label.name + ".tix"
+        mix_file_paths = ""
```
guibou: why this is not a list?

SebastianKG: I did it this way since it needs to be inserted into the bash script this way. I could add a couple steps in here and make it nicer, for sure.

Profpatsch:
Am going to review tomorrow.

haskell/private/shell_utils.bzl
```diff
@@ -0,0 +1,25 @@
+"""Utilities for shell scripts."""
```
mboes: I think we can get rid of this file (see above).

haskell/private/haskell_impl.bzl
SebastianKG: I've solved this problem a different way, through expand_template as per Mathieu's suggestion below.

haskell/private/haskell_impl.bzl
SebastianKG: I gave your suggestions a shot and I actually don't feel bad about the current implementation. If I used a list comprehension, I would still have to iterate through again and concatenate the strings with a for loop, since Starlark's lists have no .join(...) method. Remember, this is to get the filenames to be of form "a.txt b.txt c.txt" for insertion into the bash template.

If I'm missing something here, please let me know. I'll leave this unresolved for the moment.

haskell/private/haskell_impl.bzl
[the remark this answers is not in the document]
SebastianKG: This is the documented use case for this function: https://github.com/bazelbuild/bazel/blob/master/tools/bash/runfiles/runfiles.bash#L63

@Profpatsch may know more about this.

haskell/private/haskell_impl.bzl
guibou: @mboes trivia about python (the CPython implementation, I don't have any idea about skylark), but string concatenation with `a + b` is `O(len(b))` with potentially 0 allocation (depending on the overallocated space in `a`). So string building with concatenation is basically `O(n)`.

All of that thanks to a check on the reference counter.

haskell/private/haskell_impl.bzl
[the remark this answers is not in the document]
Profpatsch: This is a fragment of runfiles “design”. The path given to `rlocation` is `<workspace_name>/path/to/target`.

If you get that wrong, it fails silently.

haskell/private/haskell_impl.bzl
guibou: https://docs.bazel.build/versions/master/skylark/lib/string.html#join for a join method.

haskell/private/haskell_impl.bzl
```diff
@@ -179,14 +189,46 @@ def _haskell_binary_common_impl(ctx, is_test):
...
+        wrapper = hs.actions.declare_file("coverage_wrapper.sh")
+        ctx.actions.expand_template(
+            template = ctx.file._coverage_wrapper_template,
+            output = wrapper,
+            substitutions = {
+                "{binary_path}": binary_path,
+                "{hpc_path}": hpc_path,
+                "{tix_file_path}": tix_file_path,
+                "{expected_expression_coverage}": str(expected_expression_coverage),
+                "{mix_file_paths}": mix_file_paths,
```
Profpatsch: All these should be wired through the [corresponding `skylib` `shell` module functions](https://github.com/bazelbuild/bazel-skylib/blob/master/lib/shell.bzl), `quote` and `array_literal`.

Then you can also pass a list instead of doing manual string mangling for `mix_file_paths`.

haskell/private/coverage_wrapper.sh.tpl
```diff
@@ -0,0 +1,47 @@
...
+$binary_path "$@"
+$hpc_path report $tix_file_path $hpc_dir_args > __hpc_coverage_report
+echo "Overall report"
+cat __hpc_coverage_report
+expression_coverage=$(grep "expressions used" __hpc_coverage_report | cut -c 1-3)
+if [ $expression_coverage -lt $expected_expression_coverage ]
+then
+  echo -e "\n==>$ERRORCOLOR Inadequate expression coverage.$CLEARCOLOR Expected $expected_expression_coverage%, but actual coverage was $ERRORCOLOR$(($expression_coverage))%$CLEARCOLOR.\n"
+  exit 1
+fi
```
Profpatsch: missing `\n` :p

haskell/private/haskell_impl.bzl
```diff
@@ -179,14 +189,46 @@ def _haskell_binary_common_impl(ctx, is_test):
...
             runfiles = ctx.runfiles(
-                files = solibs,
+                files =
+                    solibs +
+                    c.conditioned_mix_files +
+                    [
+                        ctx.file._bash_runfiles,
+                        hs.toolchain.tools.hpc,
+                        binary,
+                    ],
```
Profpatsch: These are unconditionally added, right? Can we only add them when `inspect_coverage` is `True`?

haskell/private/haskell_impl.bzl
SebastianKG: With Philip's advice below, I've used a list comprehension along with using shell.array_literal.

PR #701, opened 2019-02-13
Fix broken haddock links

### LINKED ISSUE #700: haddock links broken
On current master e08bb91ef379b2215808f0cc60b6dfee30c09ac2 haddock links are not working anymore.

Step to reproduce:

```
$ bazel build //tests/haddock
$ chromium bazel-bin/tests/haddock/haddock/index/index.html
# Observes the broken links
```

Bisected to this commit:

```
commit 327f2fa872bf5663cc176e39d8c4588d2bf06c60
Author: Mathieu Boespflug <m@tweag.io>
Date:   Sun Dec 9 17:34:37 2018 +0100

    Move to Bazel v0.20
    
    Fixes #492.
    
    v0.20 has breaking changes that we can't adapt to without
    simultaneously losing compatibility with earlier versions. So the
    minimum requirement is now Bazel >= v0.20. This is also true of our
    upstream dependencies, so we had to update them as part of this
    change.
    
    We're now using the new C++ toolchain API. This in principle should
    enable us to shed long standing limitations around linking. But that
    will have to be kept as future work.
    
    DON'T MERGE YET: this PR is using a fork of Nixpkgs while Bazel
    v0.20.0 gets merged upstream.
```

guibou:
This pull requests fix haddock links which were broken since 327f2fa872bf5663cc176e39d8c4588d2bf06c60. It closes #700.

It also introduces:

- A regression test, which checks that local links inside haddock are correctly resolved. This does not check links against system libraries (such as the ones stored in `/nix/store`). This can be enabled, but it will also check remote links (such as mathjax links).
- Fixs for a few broken links.

### DIFF
--- a/haskell/haddock.bzl
+++ b/haskell/haddock.bzl
@@ -221,7 +221,7 @@ def _haskell_doc_rule_impl(ctx):
             command = """
       mkdir -p "{doc_dir}"
       # Copy Haddocks of a dependency.
-      cp -R -L "{html_dir}" "{target_dir}"
+      cp -R -L "{html_dir}/." "{target_dir}"
       """.format(
                 doc_dir = doc_root_path,
                 html_dir = html_dir.path,
--- a/shell.nix
+++ b/shell.nix
@@ -20,6 +20,8 @@ mkShell {
     git
     # Needed to get correct locale for tests with encoding
     glibcLocales
+    # to check haddock outputs
+    linkchecker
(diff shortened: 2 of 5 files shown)


mboes:
LGTM, but the PR is currently failing its own regression test.

guibou:
Should be fixed now. Waiting for CI. I did not realized that CI was building in a different directory.

PR #702, opened 2019-02-13
Fix Hazel builds on Windows with GHC < 8.6

gdeest:
With GHC < 8.6, long package names would trigger failures with Hazel on Windows due to path length exceeding the 260 characters limit. While disabling this limit globally in the Windows 10 registry is possible, this would not be honored by GHC binaries (such as ghc-pkg), as a special manifest must be provided for this to work with mingw-derived executables. This would cause issues in particular with packages compiled via Hazel, as those are fetched in external repositories, which makes for even longer file names than the usual.

This commit fixes this issue by using a .bat script on Windows to find out the "short path" of the package database, and passing that to `ghc-pkg recache`.

Additionally, we also expose the rts-headers in GHC bindists, as expected by the cbits library targets generated by Hazel.

### DIFF
--- a/haskell/private/actions/package.bzl
+++ b/haskell/private/actions/package.bzl
@@ -38,7 +38,7 @@ def _get_extra_libraries(dep_info):
         set.mutable_insert(extra_lib_dirs, lib.dirname)
     return (set.to_list(extra_lib_dirs), extra_libs)
 
-def package(hs, dep_info, interfaces_dir, interfaces_dir_prof, static_library, dynamic_library, exposed_modules_file, other_modules, my_pkg_id, static_library_prof):
+def package(hs, dep_info, interfaces_dir, interfaces_dir_prof, static_library, dynamic_library, exposed_modules_file, other_modules, my_pkg_id, static_library_prof, short_path):
     """Create GHC package using ghc-pkg.
 
     Args:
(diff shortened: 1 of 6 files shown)


haskell/private/haskell_impl.bzl
```diff
@@ -308,6 +308,7 @@ def haskell_library_impl(ctx):
         other_modules,
         my_pkg_id,
         static_library_prof = static_library_prof,
+        short_path = ctx.executable._short_path if hs.toolchain.is_windows else None,
```
Profpatsch: I’d call this `executable_short_path`.

haskell/private/actions/package.bzl
```diff
@@ -38,7 +38,8 @@ def _get_extra_libraries(dep_info):
         set.mutable_insert(extra_lib_dirs, lib.dirname)
     return (set.to_list(extra_lib_dirs), extra_libs)
 
-def package(hs, dep_info, interfaces_dir, interfaces_dir_prof, static_library, dynamic_library, exposed_modules_file, other_modules, my_pkg_id, static_library_prof):
+
+def package(hs, dep_info, interfaces_dir, interfaces_dir_prof, static_library, dynamic_library, exposed_modules_file, other_modules, my_pkg_id, static_library_prof, short_path):
```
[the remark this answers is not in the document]
gdeest: This is indeed growing out of control !

haskell/private/haskell_impl.bzl
gdeest: What do you think of `shorten_path` instead ? `executable_short_path` does not carry the same meaning to me. It sounds like this is really the short path of some executable, while this is instead an executable (script) used to shorten the path of (in this case) a directory.

gdeest: I'll go with `short_path_executable`.

PR #708, opened 2019-02-15
Update to Bazel 0.22

### LINKED ISSUE #689: Remove unnecessary -rpath entries added by GHC
GHC adds a few `RUNPATH`s to binaries that are not useful to bazel at runtime (because the runtime paths differ from built time paths).

There is a [commit that fixes this](https://github.com/tweag/rules_haskell/commit/7426a03fedb915ac07152819661ea0994028676d), but according to https://github.com/tweag/rules_haskell/pull/686#issue-250443536 it “stops GHC from adding RPATH entries for core-packages or the rts, causing runtime failures on [MacOS](https://circleci.com/gh/tweag/rules_haskell/4180) and [non-Nix Linux](https://circleci.com/gh/tweag/rules_haskell/4181).”.

For now we are not hitting the Mach-O header size limits on Darwin anymore, but if that should happen again we should find a way to set `dynload` to `deploy`.

Plus this might speed up startup of dynamic binaries somewhat, because it saves on quite a few `stat`s the runtime linker has to make when it checks for library locations.

### LINKED ISSUE #696: Support for Bazel v0.22
current `master` fails with bazel 0.22 with:

```
	File "/home/guillaume/tweag/rules_haskell/tests/c-compiles/BUILD", line 9
		haskell_library(name = 'hs-lib')
	File "/home/guillaume/.cache/bazel/_bazel_guillaume/80819d925a91d7b49ed839165b536ee7/external/io_tweag_rules_haskell/haskell/private/haskell_impl.bzl", line 197, in haskell_library_impl
		gather_dep_info(ctx)
	File "/home/guillaume/.cache/bazel/_bazel_guillaume/80819d925a91d7b49ed839165b536ee7/external/io_tweag_rules_haskell/haskell/private/dependencies.bzl", line 155, in gather_dep_info
		_HaskellCcInfo_from_CcInfo(ctx, dep[CcInfo])
	File "/home/guillaume/.cache/bazel/_bazel_guillaume/80819d925a91d7b49ed839165b536ee7/external/io_tweag_rules_haskell/haskell/private/dependencies.bzl", line 23, in _HaskellCcInfo_from_CcInfo
		cc_info.linking_context.static_mode_params_for_executable
object of type 'CcLinkingContext' has no field 'static_mode_params_for_executable'
```

`cc_info.linking_context.stati_mode_params_for_executable` is not provided anymore. I'm not able to find any reference to this change online.

I've found the list of libraries using:

```
    libs = cc_info.linking_context.libraries_to_link
    
    static_linking = [lib.static_library for lib in libs if lib.static_library]
    dynamic_linking = [lib.dynamic_library for lib in libs if lib.dynamic_library]
```

Libraries are provided as `File`, so they don't have the `artifact` and `original_artifact` previously used. Note that `libs` also contains a `pic_static_library` field which may be of use to fix #429.

aherrmann:
~**depends on #705 and #706**~

- Update the nixpkgs revision to use Bazel 0.22
- Update rules_go for Bazel 0.22 compatibility
- Bazel 0.22 further modified the C++ Starlark API. Update rules_haskell accordingly. (See commit message for details)
    - Unmangled libraries are no longer available
    - On MacOS use a `cc` wrapper script (based on [the one in the Bazel source tree](https://github.com/bazelbuild/bazel/blob/6aab2719d678057c4ef8edc3dfcad56803454732/tools/cpp/osx_cc_wrapper.sh.tpl)) to patch load commands and drop redundant rpath entries.
    - No longer requires a temporary output on MacOS due to `cc` wrapper script.

Closes #696 
Closes #689


haskell/private/dependencies.bzl
```diff
@@ -19,63 +20,115 @@ load(
 )
 load(":private/set.bzl", "set")
 
+def _cc_get_static_lib(lib_info):
```
guibou: I have doubt about this strategy.  Why the order? Shouldn't the library artifact used taken based on the compilation context? If we build static, it should be a static library with the right pic status for example.

aherrmann: > Shouldn't the library artifact used taken based on the compilation context?

For static vs. dynamic linking that's what `_cc_get_static_lib` and `_cc_get_dynamic_lib` do.

Before, Bazel had the `static_mode_params_for_executable` and `dynamic_mode_params_for_executable` attributes which would distinguish between library dependencies for static or dynamic linking mode. Now, they removed that distinction and instead provide direct access to the various variants of a library (`(pic_)static`, `dynamic`, `interface`). But, these can be `None` if a target does not provide a particular variant. E.g. for a `cc_library` in our test-suite I found that `static` is `None` and `pic_static` points to a `.a`. Similarly, I found that `interface` is `None` and `dynamic` points to a `.so`. These functions pick the most appropriate available variant. E.g. if linking statically, we'll prefer a static library. But, if none is available we'll pick a dynamic one. 

This PR makes no attempt to further exploit the new C++ API to e.g. distinguish pic static libraries from regular static libraries. This PR is only aiming at making rules_haskell work with the Bazel 0.22 C++ API again.

> If we build static, it should be a static library with the right pic status for example.

Where can the pic status for static linking be found? I don't see it available in `CcInfo` or `CcToolchainInfo`. Maybe I'm missing something?

Profpatsch:
I’ll try to review tomorrow.

haskell/private/dependencies.bzl
guibou: Great answer, thank you.

haskell/private/osx_cc_wrapper.sh.tpl
```diff
@@ -14,93 +14,299 @@
...
+
+parse_rpath() {
+    # Parse the given -rpath argument and decide whether it should be
+    # forwarded to the compiler/linker.
+    local rpath="$1"
+    if [[ "$rpath" =~ ^/ || "$rpath" =~ ^@ ]]; then
+        # Absolute rpaths or rpaths relative to @loader_path or similar, are
+        # passed on to the linker. Other relative rpaths are dropped, these
+        # are auto-generated by GHC, but are useless because rules_haskell
+        # constructs dedicated rpaths to the _solib or _hssolib directory.
```
Profpatsch: Add a link to https://github.com/tweag/rules_haskell/issues/689

Profpatsch:
I assume this means we have to drop support for any bazel before 0.22?

If so, the PR should patch the corresponding README as well (and add a preliminary CHANGELOG entry so I don’t forget next release).

haskell/private/osx_cc_wrapper.sh.tpl
aherrmann: done

aherrmann:
> I assume this means we have to drop support for any bazel before 0.22?
> 
> If so, the PR should patch the corresponding README as well (and add a preliminary CHANGELOG entry so I don’t forget next release).

@Profpatsch Correct, I've updated the README and added an `[Unreleased]` section to CHANGELOG.

Profpatsch:
Awesome. LGTM.

PR #725, opened 2019-03-05
Remove CcSkylarkApiProviderHacked

aherrmann:
Since Bazel 0.22 the new C++ Starlark API allows to construct `CcInfo` providers within Starlark, removing the need for the `CcSkylarApiProviderHacked`.

- Remove `CcSkylarkApiProviderHacked`.
- Remove references to `.cc` attributes of targets.
- Use the `CcInfo` provider instead.
- Implement `haskell_cc_import` in terms of `CcInfo`.
- Remove the now redundant `import_dependencies` and `transitive_import_dependencies` from `HaskellBuildInfo`.
- Mark `haskell_cc_import` as deprecated.
- Replace uses of `haskell_cc_import` by `cc_library`.
- Update bindist and Windows CI to Bazel 0.22.

~Note, when testing on a large, proprietary code base on MacOS this causes failures due to exceeding the MACH-O header size limit on intermediate dynamic libraries generated by GHC during the `HaskellBuildLibrary` step. Resolving this will probably require further work on `osx_cc_wrapper.sh.tpl`.~

### DIFF
--- a/haskell/cc.bzl
+++ b/haskell/cc.bzl
@@ -253,12 +253,15 @@ absolute one, it's understood as a repository-relative path.
 Use this to make `.so`, `.dll`, `.dylib` files residing in external
 [external repositories][bazel-ext-repos] available to Haskell rules.
 
-*This rule is temporary replacement for [cc_import][cc_import] and
-will be deprecated in the future.*
+*This rule is temporary replacement for [cc_import][cc_import] and is
+deprecated. Use [cc_library][cc_library] instead as shown in the example.*
 
 Example:
   ```bzl
-  haskell_cc_import(name = "zlib", shared_library = "@zlib//:lib")
+  # Deprecated, use cc_library instead.
+  # haskell_cc_import(name = "zlib", shared_library = "@zlib//:lib")
+
+  cc_library(name = "zlib", srcs = ["@zlib//:lib"])
 
   haskell_import(
     name = "base_pkg",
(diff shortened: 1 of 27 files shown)


aherrmann:
> Note, when testing on a large, proprietary code base on MacOS this causes failures due to exceeding the MACH-O header size limit on intermediate dynamic libraries generated by GHC during the HaskellBuildLibrary step.

Turns out this is unrelated to this PR. When testing these changes on another repo I accidentally broke a workaround specific to that repo.

haskell/cc.bzl
```diff
@@ -45,40 +44,21 @@ def cc_interop_info(ctx):
...
+        [f for cc in ccs for f in cc.compilation_context.quote_includes],
     ))
     system_include_directories = set.to_list(set.from_list(
-        [f for cc in ccs for f in cc.system_include_directories],
+        [f for cc in ccs for f in cc.compilation_context.system_includes],
     ))
 
     cpp_flags = (
-        ["-D" + define for cc in ccs for define in cc.defines] +
+        ["-D" + define for cc in ccs for define in cc.compilation_context.defines] +
```
Profpatsch: We walk through the `ccs` list multiple times in the lines above, which is might be a hit on analysis speed.

aherrmann: Good point. I've changed it to a single iteration.

PR #727, opened 2019-03-07
Add unix-includes to GHC BUILD

brandon-leapyear:
This isn't a problem when using GHC from Nix, because the `nixpkgs_package` rule lets you override the BUILD file used ([ref](https://github.com/FormationAI/hazel/blob/561366d7aceea056ae815ca72cbc0ad46c361748/README.md#setting-up-a-new-project)).

It would be great after merging hazel into this package (#733) if the `haskell_register_ghc_bindist` command used the same BUILD file as hazel

### DIFF
--- a/haskell/ghc.BUILD
+++ b/haskell/ghc.BUILD
@@ -25,7 +25,17 @@ cc_library(
     )[0],
 )
 
-# This is needed for Hazel targets.
+# Needed for Hazel; see FormationAI/hazel/BUILD.ghc
+
+cc_library(
+    name = "unix-includes",
+    hdrs = glob(["lib/ghc-*/unix-*/include/*.h"]),
+    strip_include_prefix = glob(
+        ["lib/ghc-*/unix-*/include"],
+        exclude_directories = 0,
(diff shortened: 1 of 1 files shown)


Profpatsch:
I’m afraid the `[0]` index fails on Windows, I suspect there is no `unix` directory in the Windows release?

aherrmann:
@Profpatsch It's a bit hacky, but something like `(glob(...) + [""])[0]` might solve that issue.

brandon-leapyear:
if the `unix` directory doesn't work, what should be put into `hdrs`? should both of those attributes be set to empty strings on windows?

Profpatsch:
cc @gdeest @nmattia

Profpatsch:
I think this was included in the hazel merge: https://github.com/tweag/rules_haskell/blob/91badfe6d1a29e638a97fd6f3984de12c319966f/haskell/ghc.BUILD#L28-L36

Does that look right @brandon-leapyear?

brandon-leapyear:
Looks great! Thanks!

Issue #698, opened 2019-02-11
Migrate rules_haskell_examples back to this repository.


Profpatsch:
So after talking to @mboes, the way we go forward with this is the following: Bazel sets the current repository boundary to where it finds another `WORKSPACE` file. This means you can use `local_repository` to import a subdirectory (`c2hs` already does that).

The plan forward:
1. Change the start script to be an a workspace directory.
1. test this workspace
1. Generate the `start` script from these files

We can use the same strategy to migrate https://github.com/tweag/rules_haskell_examples back into the `rules_haskell` repository.

_Originally posted by @Profpatsch in https://github.com/tweag/rules_haskell/pull/697#issuecomment-462383094_

Profpatsch:
Once #731 is merged, we need to deprecate the merged repository.

Profpatsch:
Deprecation PR is up: https://github.com/tweag/rules_haskell_examples/pull/17

Profpatsch:
Fixed up the readmes https://github.com/tweag/rules_haskell/pull/750

Profpatsch:
Done in https://github.com/tweag/rules_haskell/pull/750

PR #731, opened 2019-03-07
Merge rules_haskell_examples into this repo

Profpatsch:
This merges https://github.com/tweag/rules_haskell_examples and applies a few fixup commits, while keeping the original history of that repo.

The last few commits by me are relevant for review.

Relevant step for https://github.com/tweag/rules_haskell/issues/698


Profpatsch:
Fixes https://github.com/tweag/rules_haskell_examples/issues/14

mboes:
@Profpatsch I removed a few of your patches, which turned out unnecessary. The workaround was to move the tutorial directory out of examples.

This PR now depends on #737.

Profpatsch:
Meh, I guess I was too deep in, trying to work around `WORKSPACE` deficiencies, to think about the obvious solution.

Should do a repro and create an upstream bug report nonetheless.

Profpatsch:
Though we have to check in complete copies of `vector` and multiple other libraries though? This makes grepping through `rules_haskell` extremely noisy.

Profpatsch:
(rebased with //:buildifier-fix)

mboes:
These are vendored dependencies. Common practice (including in Bazel itself). `grep <pattern> haskell` in any case. We could download sdists from Hackage instead, but those would be external repos, hence recursive workspace problem again.

Issue #716, opened 2019-02-21
Move hazel into rules_haskell


Profpatsch:
`rules_haskell` should always be compatible with `hazel`, but `hazel` depends on `rules_haskell`, so we cannot add tests here. Since the two are closely connected, they should be merged into one repository.

For the complete reasoning, see https://github.com/tweag/rules_haskell/issues/703

The licenses are the same so that is not a problem.

- [x] get the approval of `hazel` authors
- [x] move the files (if possible merge git histories)
- [ ] add integration tests

Update:

- [ ] migrate issues
- [ ] deprecate old repository and point to `rules_haskell`

thumphries:
Full approval from me (current Hazel "maintainer"). Will stay apprised of this ticket and all too happy to deprecate the Hazel repo on success.

@judah had a few requests about how this should be done in https://github.com/tweag/rules_haskell/issues/357#issuecomment-408224455 ; worth checking if he still feels strongly about them, particularly the `cabal2bazel` component.

judah:
Sure, I also support this change.  FYI @blackgnezdo

To copy what I said from the other thread: Hazel is at its core a fork of (a subset of) https://github.com/google/cabal2bazel.  So I would prefer you to preserve Hazel's commit history when it's merged into rules_haskell, for example, via `git merge --allow-unrelated-histories`.  We plan to push out another version of `cabal2bazel` at some point, and that would make it easier for us to share improvements in both directions.

I do still think it would be better to keep what was forked from `cabal2bazel` in a distinct directory (in Hazel, it's currently `third_party/cabal2bazel`), to simplify any future merges and/or cherry-picks, and to make the license attribution easier .  But if that doesn't fit well into `rules_haskell`, let us know and we can probably figure out a way to work around it.

nmattia:
Someone brought up that these commands should do the job:

- `cd $(mktemp -d)`: start fresh.
- `git clone git@github.com:tweag/rules_haskell.git`
- `cd rules_haskell`
- `git remote add hazel git@github.com:FormationAI/hazel.git` add hazel as a remote
- `git fetch hazel`: download current branches from hazel repo
- `git checkout hazel/master`: checks out the current master of hazel
- `git checkout -b merge-hazel`: create a new local branch
- `mkdir hazel; mv * .bazelrc .circleci .gitignore hazel/`: move everything but .git to the hazel subdir
- `git add .; git commit -m "move hazel to subdir"`: create a commit
[...]

PR #733, opened 2019-03-07
Merge hazel into rules haskell

Profpatsch:
This merges all commits from https://github.com/DACH-NY/hazel/commits/da-master into `rules_haskell`.
So far, only the mechanical work has been done (moving files and merging the branches).

There are a few unsolved questions before we can merge:
- [x] Should we keep the hazel `WORKSPACE` file, and if yes, which parts
- [x] How to manage attribution and licenses
  The fork is from Digital Asset, while the original work and “official upstream” is by FormationAI. We chose the fork because it is compatible with current `rules_haskell`.

cc @mboes @judah @thumphries @aherrmann @gdeest @nmattia


aherrmann:
> Should we keep the hazel WORKSPACE file, and if yes, which parts

I decided to remove the Hazel `WORKSPACE` file. The reasoning is as follows. The Hazel `WORKSPACE` file is only needed if we run Bazel from within the `Hazel` directory, so that Hazel is the "main workspace". However, when doing that I encountered similar issues as @Profpatsch described in https://github.com/tweag/rules_haskell/issues/740. Instead, using the rules_haskell workspace as the "main workspace" and importing Hazel as `local_repository` works fine. However, in that case the Hazel `WORKSPACE` file is redundant. So, I've removed its contents and all files that were only referenced from `WORKSPACE`. Note, this does not affect users of Hazel.

aherrmann:
I've rebased on master with the following procedure:
- Checkout the first commit before merge of the Hazel and rules_haskell repositories.
- Merge into the new rules_haskell master `git merge --allow-unrelated-histories master` as described in https://github.com/tweag/rules_haskell/issues/716#issuecomment-469781373
- Cherry-pick all commits from after the merge.

---

- rules_haskell's `run-tests` now includes tests for Hazel.
    1. It runs `bazel test @ai_formation_hazel//...`
    2. It tries to build all Hackage packages that were previously listed in `test-packages.txt`.
        (The list is now coded into `RunTests.hs`)
- I had to update the Stackage snapshot to 13.14 since the previous 10.5 did not support GHC 8.6.4.
    I also had to remove `hsndfile` as it is no longer contained in the Stackage snapshot.
- I have modified `ghc.BUILD` to be compatible with Hazel. While doing that I noticed that the glob for the `rts-headers` target only covered the headers in the Windows bindist, not in the Unix distributions. I've changed it in this PR. This should also fix issues with missing `HsFFI.h` headers and similar. cc @nmattia . This effectively includes and closes #727 . 
- The `unix-time` package in the LTS-13.14 snapshot does not specify to run its configure script which makes its build fail. This PR bumps its version to 0.4.5 which resolves that issue.
- The same `-liconv` issue from https://github.com/tweag/rules_haskell/pull/448/files#diff-724f2339e55251a76e30620025e40c39R164 resurfaced here when building `cabal2bazel`, so I've added the same fix, which is to pass `-liconv`.

hazel/.circleci/config.yml
```diff
@@ -0,0 +1,50 @@
+version: 2
```
nmattia: This should go away

hazel/.gitignore
```diff
@@ -0,0 +1 @@
+/bazel-*
```
nmattia: Should go away

hazel/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
nmattia: Do we want to keep this license?

hazel/README.md
```diff
@@ -0,0 +1,113 @@
+# Hazel: Automatic Bazel rules for third-party Haskell dependencies
+
+[![CircleCI](https://circleci.com/gh/tweag/rules_haskell/tree/master.svg?style=svg)](https://circleci.com/gh/tweag/rules_haskell/tree/master)
```
nmattia: Should go away

hazel/test/BUILD
```diff
@@ -0,0 +1,37 @@
+load(
```
nmattia: Is this being run on CI?

hazel/third_party/cabal2bazel/LICENSE
```diff
@@ -0,0 +1,203 @@
+Copyright 2018 Google LLC.  All rights reserved.
```
nmattia: Do we want to keep this license?

hazel/third_party/cabal2bazel/Readme.md
```diff
@@ -0,0 +1,33 @@
...
+#### Different semantics of `rules_haskell`
+- `hsc` files can be passed directly to build rules.
+- All source files need to be given a common `src_strip_prefix`; solve this
+  by symlinking them all into a common location.
+- Locate data files correctly in the rule's external repository.
+- Work around `haskell_library` not supporting libraries without `srcs`.
+
+#### Hazel repository structure
+- The components of each package live in an autogenerated external repository
+  (for example: `@haskell_{package}_{hash}//...`), rather than
```
nmattia: package name is outdated

hazel/tools/bazel.rc
```diff
@@ -0,0 +1,2 @@
+# Use the overrides from @local_config_cc on MacOS as well:
```
nmattia: I'm guessing this should either disappear and bubble up to top-level

hazel/tools/ghc.bzl
```diff
@@ -0,0 +1,28 @@
+load("@bazel_tools//tools/cpp:lib_cc_configure.bzl", "get_cpu_value")
+
+default_ghc_workspaces = {
```
nmattia: What is `@ghc`? Are these good defaults, are they correct?

hazel/workspace.bzl
```diff
@@ -0,0 +1,114 @@
...
+
+    hazel_custom_package_github(
+        package_name = "text-metrics",
+        github_user = "mrkkrp",
+        github_repo = "text-metrics",
+        repo_sha = "5d10b6f6ec4ff4b014e5e512f82d23e7606cc260",
+        build_file = "@ai_formation_hazel//third_party/haskell:BUILD.text-metrics",
+    )
+
+    hazel_custom_package_github(
```
nmattia: This won't work because of the new CI names. Also `embedFile` should be fixed once and for all.

hazel/tools/ghc.bzl
gdeest: These defaults expect `rules_haskell` _users_ to define a `@ghc` workspace. I think we should rather use the default names for bindists.

hazel/LICENSE
nmattia: Yeah we'll keep the licenses.

hazel/test/BUILD
nmattia: Yes

hazel/tools/bazel.rc
nmattia: Gone.

hazel/.circleci/config.yml
nmattia: Gone.

hazel/.gitignore
nmattia: Gone.

hazel/repositories.bzl
```diff
@@ -0,0 +1,20 @@
...
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+def hazel_repositories():
+    """Provide all repositories that are necessary for `hazel` to function.
+    """
+    excludes = native.existing_rules().keys()
+    if "io_tweag_rules_nixpkgs" not in excludes:
+        http_archive(
+            name = "io_tweag_rules_nixpkgs",
+            strip_prefix = "rules_nixpkgs-cd2ed701127ebf7f8f21d37feb1d678e4fdf85e5",
```
nmattia: this doesn't look right

hazel/README.md
nmattia: Gone

Profpatsch:
Can you squash all WIP commits? That makes it easier to review.

Profpatsch:
Looks good! Let’s merge once CI is green.

PR #734, opened 2019-03-07
Unify ghc command with/without params_file

brandon-leapyear:


Profpatsch:
ping @guibou and @gdeest

Profpatsch:
@gdeest I did the merge, the logic conflicted with a change you did recently, can you check whether the new diff is good?

PR #736, opened 2019-03-08
Multi target haskell_repl

aherrmann:
Closes https://github.com/tweag/rules_haskell/issues/380

- Adds a new rule `haskell_repl` which constructs a ghci wrapper that loads multiple targets by source.
    Example usage
    ```
    haskell_repl(
        name = "my-repl",
        # Collect all transitive Haskell dependencies from these targets.
        deps = [
            "//package-a:target-1",
            "//package-b:target-2",
        ],
        # Load targets by source that match these patterns.
        include = [
            "//package-a/...",
            "//packaga-b/...",
            "//common/...",
        ],
        # Don't load targets by source that match these patterns.
        exclude = [
            "//package-a/vendored/...",
        ],
    )
    ```
    - Packages which are not loaded by source, will be loaded as compiled packages using `-package-id`, `-package-db` flags.
    - Prebuilt package dependencies are always loaded as `-package`.
    - Packages from external workspaces are never loaded by source, i.e. always as compiled packages.
    - All packages loaded by source contribute their `compiler_flags` and `repl_ghci_flags`.
    - Sets the `RUNFILES_DIR` environment variable, so that code that requires runfiles can be tested in the REPL.
- Modifies the pre-existing `ghci_repl_wrapper.sh` so it can be used for the multi target repl.


SebastianKG:
Above is a problem I ran into when using this branch.

aherrmann:
@SebastianKG Thanks for testing this. That commit is out-of-date with this PR. Could you retry on the current state of this PR (branch `workspace_repl`) with the following commands?
```
$ nix-shell --pure --run 'bazel run //tests/multi_repl:c_only_repl'
$ nix-shell --pure --run 'bazel run //tests/multi_repl:c_multi_repl'
```
It's surprising that the `:add` command lists no source files. This PR [includes tests which load the REPLs](https://github.com/tweag/rules_haskell/pull/736/commits/842612ddbc8024f64a1e50a758b176ad224596b6#diff-c76605c0e23d435535ed453213261af8R72) that [pass on CI](https://circleci.com/gh/tweag/rules_haskell/4687?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link).

aherrmann:
Noticed an issue with the generated `package-db` flags in the repl wrapper. Those were relative to execution directory instead of the execroot, which could lead to missing package dbs. I've fixed it by making those paths relative to the execroot.

SebastianKG:
Awesome, I gave it another look, and your latest code seems to load the correct modules. Great work!

haskell/repl.bzl
```diff
@@ -0,0 +1,571 @@
...
+    if not is_eligible:
+        return []
+
+    target_info = _create_HaskellReplCollectInfo(target, ctx)
+    deps_infos = [
+        dep[HaskellReplCollectInfo]
+        for dep in ctx.rule.attr.deps
+        if HaskellReplCollectInfo in dep
+    ]
+    collect_info = _merge_HaskellReplCollectInfo(*([target_info] + deps_infos))
```
Profpatsch: Why not pass a list?

haskell/repl.bzl
```diff
@@ -0,0 +1,571 @@
...
+)
+"""
+Haskell REPL aspect.
+
+Used to implement the haskell_repl rule. Does not generate an executable REPL
+by itself.
+"""
+
+def _haskell_repl_impl(ctx):
+    collect_info = _merge_HaskellReplCollectInfo(*[
```
Profpatsch: As above, could just pass a list.

haskell/repl.bzl
```diff
@@ -0,0 +1,571 @@
...
+    """Whether a package should be loaded by source or as binary."""
+    for pat in exclude:
+        if _match_label(pat, lbl):
+            return False
+
+    for pat in include:
+        if _match_label(pat, lbl):
+            return True
+
+    return False
```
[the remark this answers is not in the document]
aherrmann: > If I remember correctly, there is a way to call bazel query from within bazel rules, via some builtin primitive. If its possible, we should use that at least.

I'm not aware of any primitive or Starlark library for this task. The only related feature I'm aware of is [`genquery`](https://docs.bazel.build/versions/master/be/general.html#genquery). However, we cannot use `genquery` for two reasons. First, `genquery` does not allow wildcards:

> The only difference between the queries allowed here and on the command line is that queries containing wildcard target specifications (e.g. //pkg:* or //pkg:all) are not allowed here. 

Second, `genquery` uses `bazel query` instead of `bazel cquery`, so you quickly encounter issues like the following:

```
build aborted: no such target '@nixpkgs_cc_toolchain//:bin/clang': target 'bin/clang'
```

>  If not, we should open an upstream feature request describing our use-case and mark the attributes with an experimental- prefix for now (until we know whether there is going to be an upstream feature).

If opened a feature request on the Bazel tracker: https://github.com/bazelbuild/bazel/issues/7763

haskell/repl.bzl
```diff
@@ -0,0 +1,571 @@
...
+
+    hs.actions.expand_template(
+        template = ctx.file._ghci_repl_wrapper,
+        output = output,
+        is_executable = True,
+        substitutions = {
+            "{ENV}": render_env(ghc_env + {
+                # Export RUNFILES_DIR so that targets that require runfiles
+                # can be executed in the REPL.
+                "RUNFILES_DIR": paths.join("$RULES_HASKELL_EXEC_ROOT", output.path + ".runfiles"),
```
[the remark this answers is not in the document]
aherrmann: Indeed, this will not work on Windows. I've changed it so that the repl wrapper script itself derives `RUNFILES_DIR` and `RUNFILES_MANIFEST_FILE` variables in a way that is compatible with Bazel's Bash runfiles library.

haskell/repl.bzl
aherrmann: No good reason. I've changed it.

haskell/repl.bzl
aherrmann: Done.

haskell/repl.bzl
```diff
@@ -0,0 +1,571 @@
...
+
+Example:
+  ```bzl
+  haskell_repl(
+      name = "repl",
+      deps = [
+          "//lib:some_lib",
+          "//exe:some_exe",
+      ],
+      include = [
```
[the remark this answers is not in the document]
aherrmann: I've named it `experimental_from_source`, pending on https://github.com/bazelbuild/bazel/issues/7763 as requested.

haskell/repl.bzl
```diff
@@ -0,0 +1,571 @@
...
+          "//lib/...",
+          "//exe/...",
+          "//common/...",
+      ],
+      exclude = [
```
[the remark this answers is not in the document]
aherrmann: I've named it `experimental_from_binary`, pending on https://github.com/bazelbuild/bazel/issues/7763 as requested.

haskell/private/actions/repl.bzl
```diff
@@ -1,5 +1,6 @@
 """GHCi REPL support"""
 
+load(":private/context.bzl", "render_env")
```
regnat: Wouldn't it be possible to factor all this file out with the new `haskell_repl` rule?

haskell/repl.bzl
```diff
@@ -0,0 +1,585 @@
...
+            "-package-db",
+            paths.join("$RULES_HASKELL_EXEC_ROOT", package_cache.dirname),
+        ])
+
+    # Load C library dependencies
+    link_ctx = repl_info.load_info.cc_dependencies.dynamic_linking
+    libs_to_link = link_ctx.dynamic_libraries_for_runtime.to_list()
+
+    # External C libraries that we need to make available to the REPL.
+    seen_libs = set.empty()
```
regnat: Do we have to re-implement this bit of logic once again? :thinking:

haskell/repl.bzl
```diff
@@ -0,0 +1,585 @@
...
+        ),
+        "repl_ghci_args": attr.string_list(
+            doc = "Arbitrary extra arguments to pass to GHCi. This extends `compiler_flags` and `repl_ghci_args` from the toolchain",
+            default = [],
+        ),
+        "repl_ghci_commands": attr.string_list(
+            doc = "Arbitrary extra commands to execute in GHCi.",
+            default = [],
+        ),
+        "collect_data": attr.bool(
```
regnat: What's the reason for making this configurable? (or rather: Is there a downside in having this set to `True` ?)

haskell/private/actions/repl.bzl
aherrmann: Not quite at the moment. `haskell_repl` is its own rule and can return a `DefaultInto` and thereby properly use runfiles. `haskell/private/actions/repl.bzl` creates an extra output in the regular `haskell_*` rules and cannot return a dedicated `DefaultInfo` with runfiles. Instead it has to use the `ln` trick to force it's dependencies and refer to them in `bazel-out`.

I agree that there is duplication that we don't want to keep around in the long term. One possible way forward could be to drop `haskell/private/actions/repl.bzl` completely and replace it by the `haskell/repl.bzl` aspect, similar to `haskell/haddock.bzl`. But, I'd prefer to do that in a separate PR.

haskell/repl.bzl
aherrmann: Building the `data` dependencies may take additional time that you may not want to wait if you don't intend to use those files in your REPL session. It may also prevent you from opening the REPL if one of the `data` dependencies fails to build, even if you're not interested in that `data` dependency.

Here's a concrete example from a client's project. They have a very large Haskell target which in its `data` attribute transitively depends on itself. I.e. if `collect_data` is `True`, then the target has to be built before the REPL can be built and opened. Now suppose you have a compiler error in your target and you want to use the REPL to resolve that compiler error. If `collect_data` is hard-coded to `True` then you can't.

haskell/repl.bzl
aherrmann: Good point. I've factored out `link_libraries` in the two repl files and the runghc file.

haskell/repl.bzl
regnat: > They have a very large Haskell target which in its data attribute transitively depends on itself

Ugh :anguished: 
OK, that makes sense then :+1:

haskell/repl.bzl
```diff
@@ -0,0 +1,571 @@
...
+        hs,
+        repl_info.transitive_cc_dependencies,
+        path_prefix = "$RULES_HASKELL_EXEC_ROOT",
+    )
+    library_path = [paths.dirname(lib.path) for lib in library_deps]
+    ld_library_path = [paths.dirname(lib.path) for lib in ld_library_deps]
+
+    # Load source files
+    add_sources = [
+        "*" + f.path
```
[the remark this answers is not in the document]
Profpatsch: Ah, can you add that link in a comment? The syntax is a bit elusive.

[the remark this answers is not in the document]
aherrmann: Sure, done.

Profpatsch:
Great, great work

shmish111:
Love it!

PR #739, opened 2019-03-11
README: use `git rebase --exec` for `//:buildifier-fix`

Profpatsch:
This way we prevent the sort of “formatting fixup” commits that make
`git blame` harder to use.


Profpatsch:
The tool asks you to `git commit --amend` when the formatting tool does a change, it might be worthwhile to automate that, but I haven’t tested the workflow for that.

Issue #740, opened 2019-03-12
examples: can’t build from within subdirectory


Profpatsch:
```
[nix-shell:~/kot/bazel/rules_haskell/examples]$ bazel build @io_tweag_rules_haskell_examples//vector/...
INFO: Invocation ID: 73d1dafd-1afc-4db2-ba27-93291bc342dc
INFO: Analysed 6 targets (3 packages loaded, 78 targets configured).
INFO: Found 6 targets...
INFO: Elapsed time: 26.572s, Critical Path: 25.82s
INFO: 32 processes: 32 linux-sandbox.
INFO: Build completed successfully, 59 total actions


[nix-shell:~/kot/bazel/rules_haskell/examples]$ bazel build //vector/...
INFO: Invocation ID: 2cdca67c-c1d8-4392-8aea-983e4b4f1265
INFO: Analysed 6 targets (0 packages loaded, 0 targets configured).
INFO: Found 6 targets...
ERROR: /home/philip/kot/bazel/rules_haskell/examples/vector/BUILD.bazel:18:1: error executing shell command: '/nix/store/1sa1jld0zjhdykmd8gp99bzkffba805v-bash/bin/bash -c 
        export PATH=${PATH:-} # otherwise GCC fails on Windows

        # this is equivalent to 'readarray'. We do use 'readarray' in o...' failed (Exit 1) bash failed: error executing command /nix/store/1sa1jld0zjhdykmd8gp99bzkffba805v-bash/bin/bash -c ... (remaining 1 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox

vector/Data/Vector/Fusion/Bundle.hs:110:2: error:
     fatal error: vector.h: No such file or directory
     
      ^         
    |
110 | 
    |  ^
compilation terminated.
`gcc' failed in phase `C pre-processor'. (Exit code: 1)
INFO: Elapsed time: 0.390s, Critical Path: 0.15s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
```

This might be some part of `rules_haskell` interacting weirdly with the /[-}>}?¹! workspace system.

Profpatsch:
Oh wow, this is absurd. It’s not even idemportent when you are in `/examples`!?!!

Second `bazel query` fails (after build).

```
[nix-shell:~/kot/bazel/rules_haskell/examples]$ bazel clean --expunge
INFO: Invocation ID: 883efe31-27de-4d08-bee1-ba7a73c51dc5
[...]

PR #750, opened 2019-03-13
Fix examples and tutorial

Profpatsch:
Assorted changes to make the examples repo merge more complete.

We still have the problem with `cd`ing into examples, see #740, that’s why we add a warning to both the examples and tutorial `README`s.


.gitignore
```diff
@@ -1,2 +1 @@
 /bazel-*
-.bazelrc.local
```
mboes: Why? We do need this line here. A `.bazelrc.local` should not be checked in.

README.md
```diff
@@ -88,100 +88,24 @@ haskell_register_toolchains()
...
-
-[ci-ssh]: https://circleci.com/docs/2.0/ssh-access-jobs/
 
 ## Rules
 
 See https://api.haskell.build for the reference documentation on provided
 rules. Using [./serve-docs.sh](./serve-docs.sh), you can also view
 this documentation locally.
 
+
```
mboes: No double lines at the end of sections please. The original file wasn't written that way. Leave the style the way it was. (Applies to following lines as well.)

examples/.bazelrc.local
```diff
@@ -0,0 +1 @@
+../.bazelrc.local
```
mboes: Good idea.

tutorial/README.md
```diff
@@ -3,3 +3,35 @@
 Code for the [Bazel Haskell tutorial][bazel-haskell-tutorial].
 
 [bazel-haskell-tutorial]: https://rules-haskell.readthedocs.io/en/latest/haskell.html
+
+## **Important**
+
+Run all commands from the root of `rules_haskell`.
+If you `cd examples/`, bazel *will* [break on
```
mboes: "if you `cd tutorial/`".

mboes: Does this mean that the instructions in the tutorial itself are broken? If so that's a big deal, and we'll need to find a better workaround.

mboes:
Travis CI just reared its head. Weird. @Profpatsch any idea why?

.gitignore
Profpatsch: Good catch. Was a WIP thing that slipped through (because I added the symlinks to `tutorial` and `examples`).

tutorial/README.md
Profpatsch: Hm, it looks like it does work for `tutorial/` (not for `examples/` though).

Profpatsch:
I have no idea why travis suddenly activated before, now it didn’t. Should be ready for merge.

PR #755, opened 2019-03-19
Revert "Fix Hazel builds on Windows with GHC < 8.6"

gdeest:
This reverts commit 8a6ca46b41cf01e1c10300cfa93073c7965f5cfb.

This hack is detrimental to remote caching on Windows, as it introduces absolute file paths into package databases files (the `short_path.bat` script would turn relative paths into absolute ones). Lack of sandboxing on Windows means that those paths triggered build failures in downstream builds.

A better solution is to enable long file names on Windows, which is a one-time operation. This is now done in the Azure image we use: https://github.com/Microsoft/azure-pipelines-image-generation/pull/719/files .

An unfortunate consequence is that rules_haskell won't work on Windows for any GHC version prior to 8.6 (because if inherent filesystem limitations of older MINGW binaries), nor on any Windows version but 10+ (since enabling long paths is not possible there) ; but any reasonably-sized build would already fail under these conditions, especially in conjuction with Hazel.

Likely fixes #544 .


haskell/ghc.BUILD
```diff
@@ -25,13 +25,6 @@ cc_library(
     )[0],
 )
 
-# This is needed for Hazel targets.
```
nmattia: Why remove this? I depend on this on a different project. Could you either keep it or replace it with something like

``` python
filegroup(name = "all", srcs = glob(["**"]), visibility = ["//visibility:public"])
```
so that people can still access the header files?

gdeest: Huh, just a revert failure on my part, I guess :)

mboes:
I think requiring Windows 10+ and GHC 8.6 is reasonable.

Issue #531, opened 2018-12-20
Header dependencies are not forwarded between Haskell libraries


aherrmann:
Consider the following situation:
- `cc_library` `cbits` contains a header file `a.h`.
- `haskell_library` `intermediate` depends on `cbits` and refers to `a.h` via `CPP` `#include "a.h"`.
- `haskell_library` `final` depends on `intermediate` and refers to `a.h` via `CPP` `#include "a.h"`.

Currently, `intermediate` will succeed to compile. But, `final` will fail to compile, because the header file `a.h` cannot be found.

This behaviour of `haskell_library` is in contrast to that of `cc_library`, where headers are forwarded, and also to Cabal, where headers are forwarded as well.

- https://github.com/tweag/rules_haskell/commit/01156cd7093cad3ab0b7f7b8a37316be02490c2a demonstrates the issue (`bazel build //tests/library-with-stripped-includes`)

This issue manifests itself in Hazel. For example, when a Cabal library target contains a header file in `extra-source-files` and configures `include-dirs`, and when an executable target in the same package depends on the library target and tries to include the header file. Hazel generates `cc_library` targets for headers and other cbits. The `haskell_library` will depend on the `cc_library` that captures the headers, but the `haskell_binary` target will only depend on the `haskell_library` which currently doesn't forward the headers.

Could `rules_haskell` be extended to forward headers and [includes](https://docs.bazel.build/versions/master/be/c-cpp.html#cc_library.includes)? This would resolve the issue on the Hazel side, and also align `haskell_library`'s behaviour with that of `cc_library` and Cabal.

guibou:
I wonder if your example is correct.

> * `cc_library` `cbits` contains a header file `a.h`.
> * `haskell_library` `intermediate` depends on `cbits` and refers to `a.h` via `CPP` `#include "a.h"`.
> * `haskell_library` `final` depends on `intermediate` and refers to `a.h` via `CPP` `#include "a.h"`.

If `final` directly refer to `a.h`, it *MUST* depends on `:cbits`, that's not an indirect dependency, that's a direct dependency and it should be listed as such. Adding `:cbits` as dependencies of final fix the build.

A scenario with indirect dependencies may be:

* `cc_library` `cbits-indirect` contains a header file `a.h`.
* `cc_library` `cbits-direct` contains a header file `b.h` which includes `a.h`.
[...]

PR #757, opened 2019-03-19
WIP: Add a `haskell_group` target

### LINKED ISSUE #753: Add a `filegroup`-like rule for haskell targets
While refactoring a codebase, I tried to internally split some packages while keeping the same interface, *i.e* structure them as follows:

```bzl
haskell_library(
  name = "sub-lib1",
  …,
  visibility = ["//visibility:private"],
)

haskell_library(
  name = "sub-lib2",
  …,
  visibility = ["//visibility:private"],
)

haskell_library(
  name = "sub-lib3",
  …,
  visibility = ["//visibility:private"],
)

filegroup(
  name = "mylib",
  srcs = [ ":sub-lib1", ":sub-lib2", ":sub-lib3" ],
  visibility = ["//visiblity:public"],
)
```

so that the dependencies which used to depend on `//mylib:mylib` can continue to do so and the split remains internal.

However using a `filegroup` for that doesn't work since it won't forward the `HaskellLibraryInfo` providers and thus `//mylib:mylib` won't be recognized as a set of haskell deps and will be ignored.

This led me to the idea of a `haskell_target_group` rule which would combine a set of `haskell_library`/`haskell_binary` targets (meaning properly combine the `HaskellLibrary`/`HaskellBinary` providers so that it looks like a single target from the outside)

In addition to the pattern above, this would give "for free" a multi-targets repl rule (actually this would probably borrow a lot of the implementation from #736 and make the actual repl rule just a thin layer on top of it)

regnat:
Bundles together a list of `haskell_library` targets so that they appear as a single one from the outside.

### TODO

- [X] Initial implementation
- [X] Add a `@repl` target to the group
- [ ] Add tests
- [ ] Allow bundling `haskell_binary` targets

Fix #753

### DIFF
--- a/haskell/group.bzl
+++ b/haskell/group.bzl
@@ -3,16 +3,21 @@
 load(":private/set.bzl", "set")
 load(
     ":private/providers.bzl",
-    "empty_HaskellCcInfo",
-    "merge_HaskellCcInfo",
     "HaskellBuildInfo",
     "HaskellLibraryInfo",
+    "empty_HaskellCcInfo",
+    "merge_HaskellCcInfo",
 )
 load(":private/pkg_id.bzl", "pkg_id")
 load(
     ":private/path_utils.bzl",
     "target_unique_name",
 )
+load(
+    ":private/actions/repl.bzl",
+    "build_haskell_repl",
+)
+load(":private/context.bzl", "haskell_context")
 
 def appendBuildInfos(bi1, bi2):
     return HaskellBuildInfo(
@@ -24,7 +29,6 @@ def appendBuildInfos(bi1, bi2):
         dynamic_libraries = set.union(bi1.dynamic_libraries, bi2.dynamic_libraries),
         interface_dirs = set.union(bi1.interface_dirs, bi2.interface_dirs),
         prebuilt_dependencies = set.union(bi1.prebuilt_dependencies, bi2.prebuilt_dependencies),
-        # direct_prebuilt_deps = set.union(bi1.direct_prebuilt_deps, bi2.direct_prebuilt_deps),
         cc_dependencies = merge_HaskellCcInfo(bi1.cc_dependencies, bi2.cc_dependencies),
         transitive_cc_dependencies = merge_HaskellCcInfo(bi1.transitive_cc_dependencies, bi2.transitive_cc_dependencies),
     )
@@ -54,13 +58,13 @@ def appendLibraryInfos(ctx, li1, li2, idx):
         ctx.attr.name + "-exposed-modules-" + str(idx),
     )
     ctx.actions.run_shell(
-        inputs = [ li1.exposed_modules_file, li2.exposed_modules_file ],
-        outputs = [ exposed_modules_file ],
+        inputs = [li1.exposed_modules_file, li2.exposed_modules_file],
+        outputs = [exposed_modules_file],
         command = "cat {file1} <(echo ',') {file2} > {out}".format(
             file1 = li1.exposed_modules_file,
             file2 = li2.exposed_modules_file,
-            out = exposed_modules_file
-        )
+            out = exposed_modules_file,
+        ),
     )
     return HaskellLibraryInfo(
         package_id = li1.package_id + [li2.package_id],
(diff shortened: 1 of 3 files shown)


mboes:
I would prefer following precedent established by core rules where it exists. Consider what the CC rules allow:

```python
cc_library(
    name = "foo",
    srcs = ["foo.c"],
)

cc_library(
    name = "bar",
    srcs = ["bar.c"],
)

cc_library(
    name = "baz",
    deps = [":foo", ":bar"],
)

cc_binary(
    name = "main",
    srcs = ["main.c"],
#    deps = [":baz"],
)
```

The build fails. If you uncomment the last line, then it works. So `cc_library` is a monoid under `deps` composition (you can have an empty `cc_library` as well).

regnat:
> So cc_library is a monoid under deps composition

That would mean making any indirect dependency a direct dependency, right? Wouldn't that be a (potentially quite annoying) breaking change? Because haskell doesn't impose that your modules are unique within your transitive dependency graph, but only that they are unique within your direct dependency graph.

For example, consider that `foo` depends on `hashmap` and `bar` depends and `foo` and `unordered-containers`. Both `hashmap` and `unordered-containers` expose the `Data.HashMap` module, but `bar` can import it without ambiguity because it doesn't directly depend on `hashmap` so the compiler knows that the `Data.HashMap` it imports comes from `unordered-containers`.
Now if we flatten the dependency graph, this doesn't hold anymore and unless we use the syntactically heavy `PackageImport` we can't import `Data.HashMap` anymore.


What we could have is something similar to nix's `buildInput` and `propagatedBuildInputs` where we explicitely specify which inputs are exposed to the dependent packages

aherrmann:
> What we could have is something similar to nix's buildInput and propagatedBuildInputs where we explicitely specify which inputs are exposed to the dependent packages

That's an interesting idea. There's precedent for this kind of behavior in the builtin Java rules and rules_scala with the [`exports` attribute](https://docs.bazel.build/versions/master/be/java.html#java_library.exports).

It's also worth pointing out that the Java rules do explicitly not follow the monoid model in the `deps` attribute (emphasis mine):

> Listing rules here will make them available to parent rules, as if the parents explicitly depended on these rules. **This is not true for regular (non-exported) deps.**

---

> So cc_library is a monoid under deps composition (you can have an empty cc_library as well).

If rules_haskell were to follow that model, how would this interact with `cc_library` dependencies? In the past [it was decided]( https://github.com/tweag/rules_haskell/issues/531#issuecomment-449052316) that only header files of direct C library dependencies should be visible. By the monoid logic it seems that also headers of indirect dependencies should be visible.

> If final directly refer to a.h, it MUST depends on :cbits, that's not an indirect dependency, that's a direct dependency and it should be listed as such. Adding :cbits as dependencies of final fix the build.

mboes:
We already have an `exports` attribute: http://api.haskell.build/haskell/haskell.html#haskell_library.exports.

regnat:
But the `exports` attribute that we have isn't the same as the `export` attribute from java rules because it's much more granular

mboes:
Arguably the `exports` attribute that we have was misnamed. It's about modules whereas Java's is about libraries, i.e. the Java one talks about targets whereas ours does not. I'm not sure that anyone as using the `exports` attribute yet apart from maybe @lunaris, so depending on his input we could consider a BC here (will be a lot harder down the line).

lunaris:
@mboes I don't think we're using this feature so have no objections from that point of view. I must confess though, while I see the benefits this could bring for refactoring (as an analog to `filegroup`), I have reservations about its use in supporting REPLs across multiple libraries. Naturally I'm biased, but we (Habito) ended up building a solution atop #695 that I think works quite well (and e.g. we are using to support `ghcid` etc. too) -- this is preferable to me since it doesn't require defining anything like `haskell_repl` / `haskell_group` up front. But I appreciate I'm perhaps missing the true point of `haskell_group` here/lacking understanding around limitations of REPL targets in general.

regnat:
> I have reservations about its use in supporting REPLs across multiple libraries

As far as I'm concerned, the REPL thing was just a happy side-effect, which is anyways superseded by #736 and (I think) wouldn't make sense with the "forward transitive dependencies" approach.

But if nobody has any objection against changing the `exports` attribute I think we can go on.
@mboes should we keep the old attribute under a new name or remove it altogether? (afair it had been introduced in an effort to support backpack, but I don't think there has been much advances on that front)

mboes:
I think we should expose it under a new name. It's exposing a feature of GHC that predates Backpack (and independently useful, though rarely so).

Profpatsch:
Ping: what’s the status of this PR?

regnat:
> Ping: what’s the status of this PR?

I'll take a few hours this week to update it, it has been stalled for too long

Profpatsch:
Have we decided we want this feature?

regnat:
Closing since I'm not needing this anymore and nobody else seems to want to use it atm

PR #758, opened 2019-03-20
Workaround base linking error

nmattia:
This patches the Windows bindist to work around a GHC bug where a library is missing from base's dependencies.

See https://gitlab.haskell.org/ghc/ghc/issues/16466

### DIFF
--- a/haskell/assets/ghc_8_6_4_win_base.patch
+++ b/haskell/assets/ghc_8_6_4_win_base.patch
@@ -1,0 +1,11 @@
+--- lib/package.conf.d/base-4.12.0.0.conf	2019-03-19 18:04:35.186653529 +0100
++++ lib/package.conf.d/base-4.12.0.0.conf	2019-03-19 18:04:48.958873769 +0100
+@@ -79,7 +79,7 @@
+ data-dir: $topdir\x86_64-windows-ghc-8.6.2\base-4.12.0.0
+ hs-libraries: HSbase-4.12.0.0
+ extra-libraries:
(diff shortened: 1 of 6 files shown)


haskell/ghc_bindist.bzl
```diff
@@ -96,6 +98,13 @@ def _ghc_bindist_impl(ctx):
         _execute_fail_loudly(ctx, ["./configure", "--prefix", bindist_dir.realpath])
         _execute_fail_loudly(ctx, ["make", "install"])
 
+    # We apply some patches, if needed. As the patches may touch the package DB
+    # we regenerate the cache.
+    patch(ctx)
```
aherrmann: Doesn't matter on Windows, but in general, what's the reason for patching after `./configure`, `make` instead of before?

haskell/assets/ghc_8_6_4_win_base.patch
```diff
@@ -0,0 +1,11 @@
+--- lib/package.conf.d/base-4.12.0.0.conf	2019-03-19 18:04:35.186653529 +0100
++++ lib/package.conf.d/base-4.12.0.0.conf	2019-03-19 18:04:48.958873769 +0100
+@@ -79,7 +79,7 @@
+ data-dir: $topdir\x86_64-windows-ghc-8.6.2\base-4.12.0.0
```
aherrmann: `ghc-8.6.2` doesn't match file name `ghc_8_6_4`.

mboes:
Could you link to the upstream ticket for the bug we are working around?

mboes:
Nvm the ticket link is already there. LGTM.

PR #759, opened 2019-03-20
Fix error on missing tool

nmattia:
Before this commit we checked that a key was contained in a dict by
using `my_dict[key]`, which fails directly if the key is not found.
Instead, using `my_dict.get(key)` will return `None` if the key is not
found, which is what we want.

### DIFF
--- a/haskell/toolchain.bzl
+++ b/haskell/toolchain.bzl
@@ -113,7 +113,7 @@ def _haskell_toolchain_impl(ctx):
         for file in ctx.files.tools:
             if tool == paths.split_extension(file.basename)[0]:
                 ghc_binaries[tool] = file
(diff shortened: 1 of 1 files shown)


haskell/toolchain.bzl
```diff
@@ -113,7 +113,7 @@ def _haskell_toolchain_impl(ctx):
         for file in ctx.files.tools:
             if tool == paths.split_extension(file.basename)[0]:
                 ghc_binaries[tool] = file
-        if not ghc_binaries[tool]:
+        if not ghc_binaries.get(tool):
```
aherrmann: I think `if tool not in ghc_binaryes` should also work.

nmattia:
@aherrmann nice!

PR #760, opened 2019-03-20
Pick up on versioned tools

nmattia:
Some tools have the version prepended to the executable name, e.g.:

```
    haddock-8.6.4.exe
```

Most tools are copied (`ghc.exe`, `ghc-8.6.4.exe`) but not all.

### DIFF
--- a/haskell/toolchain.bzl
+++ b/haskell/toolchain.bzl
@@ -111,7 +111,6 @@ def _haskell_toolchain_impl(ctx):
     ghc_binaries = {}
     for tool in _GHC_BINARIES:
         for file in ctx.files.tools:
-
             if tool in ghc_binaries:
                 continue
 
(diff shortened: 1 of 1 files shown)


haskell/toolchain.bzl
```diff
@@ -111,9 +111,12 @@ def _haskell_toolchain_impl(ctx):
     ghc_binaries = {}
     for tool in _GHC_BINARIES:
         for file in ctx.files.tools:
-            if tool == paths.split_extension(file.basename)[0]:
```
regnat: Shouldn't we try the one with the version set first to ensure that we pick the most specific one in case of conflict?

nmattia: Your call. In practice they are exactly the same.

regnat: I'd invert them just in case they happen not to be the same then

nmattia: @regnat I realized the order of the checks doesn't matter. What matters is the order in which the `ctx.files.tools` are read.

Issue #765, opened 2019-03-20
Coverage should be run as part of test suite


SebastianKG:
The new coverage functionality is not protected by testing, and should therefore be set up to run as part of the CI build and checked every time there's a pull request. It may not make sense to run `bazel coverage //...`, but coverage should be run on at least some subset of the tests.

SebastianKG:
Issue cleanup! I should have closed this when the above PR was merged.

PR #768, opened 2019-03-25
Add coverage tests to CI

SebastianKG:
Just a few coverage tests added to CI, to avoid this functionality being broken. To be extended in the future.


SebastianKG:
Merging this should resolve https://github.com/tweag/rules_haskell/issues/765

SebastianKG:
Something is messing with the build system in this PR, so I'll be changing the PR to test that. Don't merge it while I do that -- I'll take the tag off when I've fixed that issue.

Issue #767, opened 2019-03-21
Wrong rpath is computed for dependencies of haskell_test


facundominguez:
**Describe the bug**

When trying to build inline-java with bazel, I can build a test. However, when I try to run the test with `bazel run //:spec`, I get an error like
```
/home/centos/.cache/bazel/_bazel_centos/589e581f43ed24e379c7bd0e2ec2d636/execroot/io_tweag_inline_java/bazel-out/k8-fastbuild/bin/spec.runfiles/io_tweag_inline_java/spec: error while loading shared libraries: libjvm.so: cannot open shared object file: No such file or directory
```

When I check the the runpath of the binary I find that it doesn't contain `libjvm.so` because the path is wrong. It is
```
  RUNPATH              $ORIGIN/../_solib_k8/_U_S_S_Copenjdk___Uexternal_Sopenjdk_Slib_Sopenjdk_Sjre_Slib_Samd64_Sserver:$ORIGIN/../_solib_k8/_U_S_Sjni_Copenjdk___Uexternal_Sopenjdk_Slib_Sopenjdk_Sjre_Slib_Samd64_Sserver:/nix/store/q8rg4mca5zqv98arpxd7164pqa72hf1n-ncurses-6.1/lib:/nix/store/hhzvw77vrmn5f506kfyb1yh1sdinkbwf-gmp-6.1.2/lib:/nix/store/83lrbvbmxrgv7iz49mgd42yvhi473xp6-glibc-2.27/lib:/nix/store/zikcg4s36w5k6fvlba2h4p4riw1qc7fp-gcc-7.3.0-lib/lib
```
while it should be 
```
[...]

PR #769, opened 2019-03-25
link.bzl: fix create_rpath_entry() for binaries in project root

Profpatsch:
We miscounted the number of parent directories by just taking the
length of the path returned by `get_parent_dir()`. Since it would
return `"."` if a file had no parent, we’d count 1 and put one `".."`
instead of zero.

Now we correctly return “no parent dir” in that case.


facundominguez:
Fixes the problem in https://github.com/tweag/rules_haskell/issues/767. :+1

PR #770, opened 2019-03-25
Update readme to add explicit http_archive import

chris-bacon:
The native http_archive import is depreciated, so have added it as an explicit import for the readme manual instruction


mboes:
Thanks!

PR #777, opened 2019-03-26
Uncovered expression count as a coverage metric

SebastianKG:
An alternate way of evaluating expression coverage: the maximum allowed count of expressions not covered.

### DIFF
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -130,12 +130,16 @@ def _mk_binary_rule(**kwargs):
         ),
     )
 
-    # Tests have an extra fields regarding expected code coverage percentages.
+    # Tests have an extra fields regarding code coverage.
     if is_test:
         attrs.update({
-            "expected_expression_coverage": attr.int(
+            "expected_covered_expressions_percentage": attr.int(
                 default = 0,
                 doc = "The expected percentage of expressions covered by testing.",
+            ),
+            "expected_uncovered_expression_count": attr.int(
+                default = 0,
+                doc = "The expected number of expressions which are not covered by testing.",
             ),
             "_coverage_wrapper_template": attr.label(
                 allow_single_file = True,
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -141,6 +141,10 @@ def _mk_binary_rule(**kwargs):
                 default = -1,
                 doc = "The expected number of expressions which are not covered by testing.",
             ),
+            "strict_coverage_analysis": attr.bool(
+                default = False,
+                doc = "Requires that the coverage metric is matched exactly, even doing better than expected is not allowed.",
+            ),
             "_coverage_wrapper_template": attr.label(
                 allow_single_file = True,
                 default = Label("@io_tweag_rules_haskell//haskell:private/coverage_wrapper.sh.tpl"),
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -134,11 +134,11 @@ def _mk_binary_rule(**kwargs):
     if is_test:
         attrs.update({
             "expected_covered_expressions_percentage": attr.int(
-                default = 0,
+                default = -1,
                 doc = "The expected percentage of expressions covered by testing.",
             ),
             "expected_uncovered_expression_count": attr.int(
-                default = 0,
+                default = -1,
                 doc = "The expected number of expressions which are not covered by testing.",
             ),
             "_coverage_wrapper_template": attr.label(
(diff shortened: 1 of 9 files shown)


haskell/private/coverage_wrapper.sh.tpl
```diff
@@ -42,9 +43,15 @@ $binary_path "$@"
...
-if [ $expression_coverage -lt $expected_expression_coverage ]
+covered_expression_percentage=$(grep "expressions used" __hpc_coverage_report | cut -c 1-3)
+if [ $covered_expression_percentage -lt $expected_covered_expressions_percentage ]
 then
-  echo -e "\n==>$ERRORCOLOR Inadequate expression coverage.$CLEARCOLOR Expected $expected_expression_coverage%, but actual coverage was $ERRORCOLOR$(($expression_coverage))%$CLEARCOLOR.\n"
+  echo -e "\n==>$ERRORCOLOR Inadequate expression coverage percentage.$CLEARCOLOR Expected $expected_covered_expressions_percentage%, but actual coverage was $ERRORCOLOR$(($covered_expression_percentage))%$CLEARCOLOR.\n"
+  exit 1
+fi
+uncovered_expression_count=$(grep "expressions used" __hpc_coverage_report | sed s/.*\(//g | cut -f1 -d "/")
+if [ $uncovered_expression_count -gt $expected_uncovered_expression_count ]
```
ghorn: Could you implement a ratcheting mechanism? If you only check greater than then noone will tighten the bounds. What about having both `expected_uncovered_expression_count` and `max_expected_uncovered_expression_count` and then doing something like:
```
# check max allowed
if [ $uncovered_expression_count -gt $max_expected_uncovered_expression_count ]
then
  echo "Too many uncovered expressions..."
  exit 1
fi

# check exact number
if [ $uncovered_expression_count -ne $expected_uncovered_expression_count ]
then
  echo "You have xxx uncovered expressions but expected YYY."

  if [ $uncovered_expression_count -lt $expected_uncovered_expression_count ]
  then
    echo "Nice job! You got rid of uncovered expressions, please reduce the expected number..."
  elif [ $uncovered_expression_count -gt $expected_uncovered_expression_count ]
    echo "Uh oh please don't make test coverage worse...."
  fi

  exit 1
fi
```

haskell/haskell.bzl
```diff
@@ -130,13 +130,21 @@ def _mk_binary_rule(**kwargs):
...
     )
 
-    # Tests have an extra fields regarding expected code coverage percentages.
+    # Tests have an extra fields regarding code coverage.
     if is_test:
         attrs.update({
-            "expected_expression_coverage": attr.int(
-                default = 0,
+            "expected_covered_expressions_percentage": attr.int(
+                default = -1,
```
guibou: Is this `-1` documented?

guibou: Actually I don't like sentinel values, but the bazel API does not allows us to use something similar to a `Maybe`. However any sentinel value must be documented as much as possible, OR using something more common such as `None` if possible.

haskell/private/haskell_impl.bzl
```diff
@@ -62,7 +62,8 @@ def haskell_binary_impl(ctx):
     return _haskell_binary_common_impl(ctx, is_test = False)
 
 def _should_inspect_coverage(ctx, hs, is_test):
-    return hs.coverage_enabled and is_test and ctx.attr.expected_expression_coverage > 0
+    return hs.coverage_enabled and is_test and (ctx.attr.expected_covered_expressions_percentage != None or
```
guibou: `None` or `-1`?

haskell/haskell.bzl
```diff
@@ -130,13 +130,21 @@ def _mk_binary_rule(**kwargs):
...
                 doc = "The expected percentage of expressions covered by testing.",
             ),
+            "expected_uncovered_expression_count": attr.int(
+                default = -1,
+                doc = "The expected number of expressions which are not covered by testing.",
+            ),
+            "strict_coverage_analysis": attr.bool(
```
guibou: Shouldn't this be `True` by default?

SebastianKG: No, I don't think the behaviour described by strict code coverage is what people expect by default. People expect to be setting minimums/maximums, not exact values. There is value, however, in setting exact values if your team is willing to constantly re-set them for every change (namely, it forces awareness and constant improvement of test coverage).

haskell/haskell.bzl
SebastianKG: Added a comment internally, but I don't think this should be documented to the user. This is the Nothing value, and if the user provides anything it should be a Just.

PR #789, opened 2019-03-27
WIP: Invoke bazel fewer times in tests

nmattia:
Marked as WIP because we first need to compare CI times. If there's no improvement there's no point in merging.


nmattia:
No improvement there unfortunately.

PR #790, opened 2019-03-28
Reenable cachix

nmattia:
Shaves of ~ 7 minutes in the "Build tests" job (19mn21s -> 11mn54s). The "Run tests" step took about 2 minutes longer but I think that's only due to performance flakiness.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,6 +40,13 @@ jobs:
             # CircleCI and Nix sandboxing don't play nice. See
             # https://discourse.nixos.org/t/nixos-on-ovh-kimsufi-cloning-builder-process-operation-not-permitted/1494/5
             mkdir -p /etc/nix && echo "sandbox = false" > /etc/nix/nix.conf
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -40,6 +40,13 @@ jobs:
             # CircleCI and Nix sandboxing don't play nice. See
             # https://discourse.nixos.org/t/nixos-on-ovh-kimsufi-cloning-builder-process-operation-not-permitted/1494/5
             mkdir -p /etc/nix && echo "sandbox = false" > /etc/nix/nix.conf
+            nix-env -iA cachix -f https://github.com/NixOS/nixpkgs/tarball/db557aab7b690f5e0e3348459f2e4dc8fd0d9298
+      - run:
+          name: Run cachix
+          command: |
+            cachix use tweag
+            cachix push tweag --watch-store
```
Profpatsch: Does this also set up `cachix` as a binary substitutor that nix uses?

nmattia: `cachix use` does, it writes to `nix.conf`.

Profpatsch: Ah, that’s … interesting.

PR #791, opened 2019-03-28
Don't waste time updating channels

nmattia:


Profpatsch:
Why don’t we need to update channels? Because we pin?

nmattia:
Yes, as far as I can tell everything is pinned. From a `grep -r '<nixpkgs>'` the only occurrences are those from nixpkgs_package which get `NIX_PATH=nixpkgs=...` set by the `repositories`.

PR #794, opened 2019-03-28
Make CI more robust

nmattia:
* Print a dot every minute to keep the CI alive
* ~Disable cachix for the nixpkgs build~


gdeest:
That's a strong :+1: from me on the dot part.
Why the need to disable cachix on `nixpkgs` builds ?

nmattia:
It appears to be causing the CI to OOM and is mostly needed for Darwin
anyway.

PR #795, opened 2019-03-28
Add disk cache to CI (darwin)

nmattia:
We don't want to keep old artifacts around so we always build from
scratch on master builds and upload the new cache afterwards. Because
Circle doesn't allow skipping a "restore_cache" we create a dummy
"empty" cache that's only ever pulled on master. Alternatively we could
ask Bazel to clean up old items (LRU style) but the documentation is
very terse and I could not figure how to do it:
    https://docs.bazel.build/versions/master/remote-caching.html
It also appears that there's ongoing work but the feature is not ready:
    https://github.com/bazelbuild/bazel/issues/5139

Currently the disk cache is only implemented for the Darwin builds,
which were the slowest ones. There is no reason why a disk cache
couldn't be used for the other jobs: I just haven't gotten around to
doing it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,22 @@
 version: 2
 
+# NOTE:
+#   Disk cache:
+#       We don't want to keep old artifacts around so we always build from
+#       scratch on master builds and upload the new cache afterwards. Because
+#       Circle doesn't allow skipping a "restore_cache" we create a dummy
+#       "empty" cache that's only ever pulled on master. Alternatively we could
+#       ask Bazel to clean up old items (LRU style) but the documentation is
+#       very terse and I could not figure how to do it:
+#           https://docs.bazel.build/versions/master/remote-caching.html
+#       It also appears that there's ongoing work but the feature is not ready:
+#           https://github.com/bazelbuild/bazel/issues/5139
+#
+#       Currently the disk cache is only implemented for the Darwin builds,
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -117,6 +144,25 @@ jobs:
...
+      - save_cache:
+          key: v1-rules_haskell-cache-{{ .Branch }}-{{ .BuildNum }}
+          paths:
+              - ~/.cache/bazel/
+
+      - run:
+          name: Clean up cache
+          shell: /bin/bash -eilo pipefail
+          command: |
+            rm -rf ~/.cache/bazel/
```
Profpatsch: If you remove it each time, you don’t need to use the Circle cache, right?

nmattia: Clarified the reasoning somewhere else.

.circleci/config.yml
```diff
@@ -117,6 +144,25 @@ jobs:
...
+            mkdir -p ~/.cache/bazel/
+
+      - save_cache:
+          key: v1-rules_haskell-empty-master-{{ .BuildNum }}
```
gdeest-da: Isn't this being run for every branch ?

nmattia: It is, there unfortunately is no way around it. But it's not a problem.

gdeest-da: Got it.

PR #799, opened 2019-03-31
Add support for GHC plugins

mboes:
This introduces new `ghc_plugin` rule type, and an extra attribute to
`haskell_*` rules to add plugins to the compilation process. Plugins
only affect the compile action, not any other action. We take care to
ensure that plugin dependencies do not induce extra runtime
dependencies, by using `-plugin-package` for plugin dependencies
instead of `-package`.

### DIFF
--- a/haskell/private/haskell_impl.bzl
+++ b/haskell/private/haskell_impl.bzl
@@ -30,7 +30,7 @@ load(
 )
 load(":private/pkg_id.bzl", "pkg_id")
 load(":private/set.bzl", "set")
-load(":private/providers.bzl", "HaskellCoverageInfo")
+load(":private/providers.bzl", "GhcPluginInfo", "HaskellCoverageInfo")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@bazel_skylib//lib:shell.bzl", "shell")
 
@@ -66,7 +66,11 @@ def _should_inspect_coverage(ctx, hs, is_test):
 
 def _haskell_binary_common_impl(ctx, is_test):
     hs = haskell_context(ctx)
-    dep_info = gather_dep_info(ctx)
+    dep_info = gather_dep_info(ctx, ctx.attr.deps)
+    plugin_dep_info = gather_dep_info(
+        ctx,
+        [dep for plugin in ctx.attr.plugins for dep in plugin[GhcPluginInfo].deps],
+    )
 
     # Add any interop info for other languages.
     cc = cc_interop_info(ctx)
@@ -84,6 +88,7 @@ def _haskell_binary_common_impl(ctx, is_test):
         cc,
         java,
         dep_info,
+        plugin_dep_info,
         srcs = srcs_files,
         ls_modules = ctx.executable._ls_modules,
         import_dir_map = import_dir_map,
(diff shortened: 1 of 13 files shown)


haskell/private/haskell_impl.bzl
```diff
@@ -236,7 +244,11 @@ def _haskell_binary_common_impl(ctx, is_test):
 
 def haskell_library_impl(ctx):
     hs = haskell_context(ctx)
-    dep_info = gather_dep_info(ctx)
+    dep_info = gather_dep_info(ctx, ctx.attr.deps)
```
guibou: There is a code duplication here.

haskell/private/haskell_impl.bzl
```diff
@@ -66,7 +66,11 @@ def _should_inspect_coverage(ctx, hs, is_test):
 
 def _haskell_binary_common_impl(ctx, is_test):
     hs = haskell_context(ctx)
-    dep_info = gather_dep_info(ctx)
+    dep_info = gather_dep_info(ctx, ctx.attr.deps)
```
guibou: Code duplication here.

haskell/private/packages.bzl
```diff
@@ -2,18 +2,20 @@
 
 load(":private/set.bzl", "set")
 
-def pkg_info_to_ghc_args(pkg_info):
+def pkg_info_to_ghc_args(pkg_info, for_plugin = False):
     """
     Takes the package info collected by `ghc_info()` and returns the actual
     list of command line arguments that should be passed to GHC.
+    Last argument indicates whether the package is a plugin dependency.
```
guibou: Please be specific in the name of the argument, instead of "Last argument". This will prevent lost of sync between comment and code when a new argument will be added to the function.

haskell/toolchain.bzl
```diff
@@ -17,7 +17,7 @@ load(":private/actions/package.bzl", "package")
 
 _GHC_BINARIES = ["ghc", "ghc-pkg", "hsc2hs", "haddock", "ghci", "runghc", "hpc"]
 
-def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, params_file = None, env = None, progress_message = None):
+def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, params_file = None, env = None, progress_message = None, input_manifests = None):
```
guibou: What is this?

haskell/toolchain.bzl
```diff
@@ -96,6 +96,7 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, params_file = None, e
 
     hs.actions.run_shell(
         inputs = inputs,
+        input_manifests = input_manifests,
```
guibou: And this?

mboes: It's an argument to `run_shell()`, so it's documented in the [API documentation](https://docs.bazel.build/versions/master/skylark/lib/actions.html#run_shell) for `run_shell()`.

guibou: Nevermind. I was just surprised to discover new arguments to `run_shell` and `_run_ghc` which were not used elsewhere in this PR.

haskell/private/haskell_impl.bzl
mboes: @guibou What is duplicated here is one list comprehension in two places. See what our [coding guidelines have to say about that](https://github.com/tweag/guides/blob/master/style/Haskell.md#general-style) (that part originally from the GHC coding style document).

haskell/toolchain.bzl
mboes: Ah, you're right. They're used in a followup PR (not submitted yet). Though that extra argument does no harm at this point.

PR #803, opened 2019-04-01
Ci disallow local nix builds

Profpatsch:
Experiment to guarantee all (nixpkgs) dependencies are binary cached, at least for `nixpkgs-linux`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,12 +51,20 @@ jobs:
       - run:
           name: System dependencies
           command: |
+            set -e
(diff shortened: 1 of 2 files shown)


Profpatsch:
Update: adding an uncached dependency (that is one that has to be built) stops the build as expected, but only *after* downloading all binary substitutions.

Here’s the upstream issue tracking that: https://github.com/NixOS/nix/issues/2753

I still think it’s a net win for `rules_haskell`.

PR #804, opened 2019-04-01
Move start script to bindists & document

Profpatsch:
This PR improves the start script and the bindists.

- make the start script compile a runnable hello world module
- generate bindist file from official upstream hash files
- move from nixpkgs to bindists

### DIFF
--- a/haskell/gen_ghc_bindist.py
+++ b/haskell/gen_ghc_bindist.py
@@ -1,0 +1,151 @@
+#!/usr/bin/env python
+
+# This is a happy-path tool to download the bindist
+# download paths and hashes, for maintainers.
+# It uses the hashes provided by download.haskell.org.
+
+from __future__ import print_function
+
+import pprint
+import sys
+import urllib2
+
+# All GHC versions we generate.
+# `version` is the version number
+# `distribution_version` is a corrected name
+# (sometimes bindists have errors and are updated by new bindists)
+# `ignore_prefixes` is the prefix of files to ignore
+# `ignore_suffixes` is the suffix of files to ignore
+VERSIONS = [
+    { "version": "8.6.4" },
+    { "version": "8.6.3" },
+    { "version": "8.6.2" },
+    { "version": "8.4.4" },
+    { "version": "8.4.3" },
+    { "version": "8.4.2" },
+    { "version": "8.4.1" },
+    { "version": "8.2.2" },
+    { "version": "8.0.2",
+      "ignore_suffixes": [".patch"] },
+    { "version": "7.10.3",
+      "distribution_version": "7.10.3b",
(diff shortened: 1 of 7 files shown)


haskell/ghc_bindist_generated.bzl
```diff
@@ -0,0 +1,145 @@
+# Generated with `bazel run @io_tweag_rules_haskell//haskell:gen-ghc-bindist | sponge haskell/ghc_bindist_generated.bzl`
```
mboes: This file should be in `private/`.

haskell/gen_ghc_bindist.py
```diff
@@ -0,0 +1,132 @@
...
+    return "https://downloads.haskell.org/~ghc/{ver}/SHA256SUMS".format(
+        ver = version
+    )
+
+# Parses the tarball hashsum file for a version.
+def parse_sha256_file(content, version):
+    res = {}
+    for line in content:
+        # f5763983a26dedd88b65a0b17267359a3981b83a642569b26334423f684f8b8c  ./ghc-8.4.3-i386-deb8-linux.tar.xz
+        hash_file = line.strip().split("  ./")
```
guibou: Nitpick, but replace the following 4 lines by:

```
(hash, file) = line.strip().split("  ./")
```

You'll get the assert behavior, more compact code and no indices access.

haskell/gen_ghc_bindist.py
```diff
@@ -0,0 +1,132 @@
...
+        file_ = hash_file[1]
+        prefix = "ghc-{ver}-".format(ver = version)
+        suffix = ".tar.xz"
+        # ghc-8.4.3-i386-deb8-linux.tar.xz
+        if file_.startswith(prefix) and file_.endswith(suffix):
+            # i386-deb8-linux
+            name = file_[len(prefix):-len(suffix)]
+            res[name] = hash
+        else:
+            eprint("Info: Can't parse the sha256 field for {ver}: {entry}".format(ver = version, entry = line.strip()))
```
guibou: Is this supposed to fail? `eprint` does not.

Profpatsch: No, just an info field, because some of the releases have special files. Though you are right, it might be more correct to fail in these cases.

Profpatsch: Okay, I changed it so that it fails for everything that was not explicitely filtered out.

Profpatsch:
So, importing the `GHC_BINDIST` dict from a different module blows up with our (deprecated and very old) version of Skydoc. For now I’m copying the generated code into the file.

mboes:
Our version of Skydoc is deprecated but unfortunately there is no good replacement for it as of yet. The new Stardoc is a lot less feature complete.

Issue #764, opened 2019-03-20
Code coverage with scoping of which deps are included


SebastianKG:
Currently, code coverage functionality (activated using `bazel coverage`) builds all included libraries (srcs *and* deps) with coverage enabled for a `haskell_test` rule. This means the resulting values of expression coverage are based on coverage of every single Haskell library built from source by your local Bazel (not, however, dependencies pulled from Nix, etc.).

It would be nice to have some capacity to restrict the scope of the coverage test. To only the most directly included `haskell_library`s? To libraries in the same package? How configurable should this be? Still TBD.

Profpatsch:
See also the recent Multi Target REPL PR: https://github.com/tweag/rules_haskell/pull/736. It includes a way to filter targets (in order to define which ones should be loaded from source and which as binaries).

PR #813, opened 2019-04-05
Reused path patterns from REPL and implemented coverage scoping.

SebastianKG:
This allows a haskell_test to scope which targets in its dependencies are included in coverage analysis. By default, they all are, even transitive dependencies.

Documentation update to come in a later PR.

This resolves: https://github.com/tweag/rules_haskell/issues/764

### DIFF
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -145,7 +145,7 @@ def _mk_binary_rule(**kwargs):
                 default = False,
                 doc = "Requires that the coverage metric is matched exactly, even doing better than expected is not allowed.",
             ),
-            "coverage_source_patterns": attr.string_list(
+            "experimental_coverage_source_patterns": attr.string_list(
                 default = ["//..."],
                 doc = "The path patterns specifying which targets to analyze for test coverage metrics. Wildcarding is allowed.",
             ),
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -145,6 +145,10 @@ def _mk_binary_rule(**kwargs):
                 default = False,
                 doc = "Requires that the coverage metric is matched exactly, even doing better than expected is not allowed.",
             ),
+            "coverage_source_patterns": attr.string_list(
+                default = ["//..."],
+                doc = "The path patterns specifying which targets to analyze for test coverage metrics. Wildcarding is allowed.",
+            ),
             "_coverage_wrapper_template": attr.label(
                 allow_single_file = True,
                 default = Label("@io_tweag_rules_haskell//haskell:private/coverage_wrapper.sh.tpl"),
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -147,7 +147,12 @@ def _mk_binary_rule(**kwargs):
             ),
             "experimental_coverage_source_patterns": attr.string_list(
                 default = ["//..."],
-                doc = "The path patterns specifying which targets to analyze for test coverage metrics. Wildcarding is allowed.",
+                doc = """The path patterns specifying which targets to analyze for test coverage metrics.
+
+                Wild-card targets such as //... or //:all are allowed. The paths must be relative to the workspace, which means they must start with "//".
+
+                Note, this attribute may leave experimental status depending on the outcome of https://github.com/bazelbuild/bazel/issues/7763.
+                """,
             ),
             "_coverage_wrapper_template": attr.label(
                 allow_single_file = True,
--- a/haskell/private/haskell_impl.bzl
+++ b/haskell/private/haskell_impl.bzl
@@ -25,7 +25,9 @@ load(":private/mode.bzl", "is_profiling_enabled")
 load(
     ":private/path_utils.bzl",
     "ln",
+    "match_label",
     "module_name",
+    "parse_pattern",
     "target_unique_name",
 )
 load(":private/pkg_id.bzl", "pkg_id")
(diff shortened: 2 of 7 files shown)


haskell/haskell.bzl
```diff
@@ -145,6 +145,10 @@ def _mk_binary_rule(**kwargs):
                 default = False,
                 doc = "Requires that the coverage metric is matched exactly, even doing better than expected is not allowed.",
             ),
+            "coverage_source_patterns": attr.string_list(
+                default = ["//..."],
```
Profpatsch: We don’t want to expose the pattern matching implementation as a stable feature yet, because we are hoping for upstream bazel to provide it. You should prefix anything that depends on it with `experimental_`.

haskell/private/path_utils.bzl
```diff
@@ -318,3 +318,103 @@ def copy_all(ctx, srcs, dest):
             command = "mkdir -p {dest} && cp -L -R \"$@\" {dest}".format(dest = dest.path),
             arguments = [args],
         )
+
+def parse_pattern(pattern_str):
+    """Parses a string label pattern.
```
Profpatsch: If we move this to the general utilities we should link to https://github.com/bazelbuild/bazel/issues/7763 and require that every piece of code using this has to be an experimental API.

haskell/haskell.bzl
```diff
@@ -145,6 +145,10 @@ def _mk_binary_rule(**kwargs):
...
+                doc = "The path patterns specifying which targets to analyze for test coverage metrics. Wildcarding is allowed.",
```
Profpatsch: “Wildcarding” should point to a description of the syntax and semantics that is allowed.

haskell/private/haskell_impl.bzl
```diff
@@ -65,6 +67,19 @@ def _should_inspect_coverage(ctx, hs, is_test):
...
                                                 ctx.attr.expected_uncovered_expression_count != -1)
 
+def _make_coverage_data(mix_file, target_label):
+    return struct(
+        mix_file = mix_file,
+        target_label = target_label,
+    )
+
+def _coverage_enabled_for_target(coverage_source_patterns, label):
+    for pat in coverage_source_patterns:
```
guibou:
```
return any([match_label(pat, label) for pat in coverage_source_patterns])
```

Feels more readable for me, but it is discutable.

haskell/private/path_utils.bzl
```diff
@@ -318,3 +318,103 @@ def copy_all(ctx, srcs, dest):
...
+        `...`, `:all`, and `:*`. Also the `//some/package` shortcut is allowed.
+
+    Returns:
+      A struct of
+        package: A list of package path components. May end on the wildcard `...`.
+        target: The target name. None if the package ends on `...`. May be one
+          of the wildcards `all` or `*`.
+
+    """
+
```
guibou: Should you start by trimming leading and trailing whitespaces from the pattern and the needle? I'm not sure, but this may be discussed.

haskell/private/path_utils.bzl
```diff
@@ -318,3 +318,103 @@ def copy_all(ctx, srcs, dest):
...
+    # necessary to specify a workspace. Therefore, we don't allow it.
+    if pattern_str.startswith("@"):
+        fail("Invalid haskell_repl pattern. Patterns may not specify a workspace. They only apply to the current workspace")
+
+    # To keep things simple, all patterns have to be absolute.
+    if not pattern_str.startswith("//"):
+        fail("Invalid haskell_repl pattern. Patterns must start with //.")
+
+    # Separate package and target (if present).
+    package_target = pattern_str[2:].split(":", maxsplit = 2)
```
guibou:
```suggestion
    package_target = pattern_str[2:].split(":", maxsplit = 1)
```

I don't understand the 2, is that a typo or is there a specific reason (if, please add a comment). In python, `maxsplit` gives the number of time the separator is looked for. So with `maxsplit = 2`, you'll get a maximum of two splits, so three tokens. I suppose that the bazel API followed this.

haskell/private/path_utils.bzl
```diff
@@ -318,3 +318,103 @@ def copy_all(ctx, srcs, dest):
...
+        package = package,
+        target = target_str,
+    )
+
+def match_label(patterns, label):
+    """Whether the given local workspace label matches any of the patterns.
+
+    Args:
+      patterns: A list of parsed patterns to match the label against.
+        Apply `_parse_pattern` before passing patterns into this function.
```
guibou:
```suggestion
        Apply `parse_pattern` before passing patterns into this function.
```

Typo?

haskell/private/path_utils.bzl
```diff
@@ -318,3 +318,103 @@ def copy_all(ctx, srcs, dest):
...
+      label: Match this label against the patterns.
+
+    Returns:
+      A boolean. True if the label is in the local workspace and matches any of
+      the given patterns. False otherwise.
+
+    """
+
+    # Only local workspace labels can match.
+    # Despite the docs saying otherwise, labels don't have a workspace_name
```
guibou: Please add a link to the documentation and bazel version here. This kind of comment looks weird when outdated.

haskell/private/path_utils.bzl
```diff
@@ -318,3 +318,103 @@ def copy_all(ctx, srcs, dest):
...
+        elif patterns.package[i] != package[i]:
+            return False
+
+    # If no wild-card or mismatch was encountered, the lengths must match.
+    # Otherwise, the label's package is not covered.
+    if len(patterns.package) != len(package):
+        return False
+
+    # Match target.
+    if patterns.target == "all" or patterns.target == "*":
```
guibou:
```
return (patterns.target == "all" or pattern.target == "*") or (pattern.target == target)
```

Also, I'll be tempted to "normalize" `all` and `*` in `parse_pattern`

haskell/repl.bzl
```diff
@@ -174,114 +173,14 @@ def _merge_HaskellReplCollectInfo(args):
         transitive_cc_dependencies = transitive_cc_dependencies,
     )
 
-def _parse_pattern(pattern_str):
```
guibou: Ignore most of my previous comments, this PR is not supposed to clean `parse_pattern` at the same time it moves it.

Profpatsch: If you want to do that, you can do so in a different commit. Might be best to split the move from the other stuff anyway.

haskell/haskell.bzl
SebastianKG: Matched what is found in haskell_repl

haskell/repl.bzl
SebastianKG: @guibou I resolved your comments re: pattern parsing, since they're out of scope for this PR. I think there will be some work on that code in another PR, so I'll keep them in mind.

PR #814, opened 2019-04-06
Move to Bazel 0.24

mboes:


Profpatsch:
`_GHC_DEFAULT_VERSION` is still set to `"8.4.4"` in `ghc_bindist.bzl`, do we want to bump that? Or use `ghc_version`?

mboes:
Let's.

PR #822, opened 2019-04-11
Don't clean in between steps in CI script

mboes:
Repl uses runfiles now. So there is no need to keep this. It slows
down CI significantly.


Profpatsch:
\o/

Issue #809, opened 2019-04-04
Code coverage HTML report support


SebastianKG:
It is useful to inspect which files are actually being tested in code coverage analysis.

The best solution is to have an HTML report that is generated as a file.

However, tests in Bazel can't generate files, only a text output. It isn't nice, but it may only be possible to emit the HTML contents to the test output.

Additionally, this will require forwarding of all source files which have been instrumented for code coverage along in the dependency graph, because `hpc markup` needs to know where to find them.

SebastianKG:
This has been completed per the two PRs merged above.

PR #830, opened 2019-04-15
Html report extraction tool

SebastianKG:
Since `bazel coverage` can't output any files, this is the second step for generating an HTML coverage report. For a `haskell_test` rule with attr `coverage_report_format = "html"`, first run it with `bazel coverage`, then, with this new functionality, run:

```
kazel run //tools/coverage-reports:coverage-report-renderer -- \
  --testlog=<absolute path to testlog test.xml file>           \
  --destdir=<absolute path to destination folder>
```

This will generate all of the relevant `hpc markup` report files in the destination folder, under `coverage-reports`.

Generally, the testlog test.xml is found at `{WORKSPACE_ROOT}/bazel-testlogs/tests/{PACKAGE_NAME}/{HASKELL_TEST_TARGET_NAME}/test.xml`.

One thing I haven't figured out is how to pass relative paths in as arguments, since the coverage-report-renderer isn't run in the directory the user is in. If there's a way of getting that information from an environment variable or something like that, that would be great.

Merging this will complete: https://github.com/tweag/rules_haskell/issues/809. Documentation to come later, after seeing how users use this in the wild.

### DIFF
--- a/haskell/private/coverage_wrapper.sh.tpl
+++ b/haskell/private/coverage_wrapper.sh.tpl
@@ -55,7 +55,7 @@ do
 done
 
 # run the test binary, and then generate the report
-$binary_path "$@" > /dev/null
+$binary_path "$@" > /dev/null 2>&1
 $hpc_path report "$tix_file_path" $hpc_dir_args $hpc_exclude_args > __hpc_coverage_report
 
 # if we want a text report, just output the file generated in the previous step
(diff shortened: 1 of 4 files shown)


Profpatsch:
A high-level question: Are those XML reports in a standardized format? Because then we could integrate with some tooling that certainly has to exist instead of building our own HTML output.

SebastianKG:
@Profpatsch

The outer report is a JUnit XML output. In Python, there’s a library for parsing it: https://pypi.org/project/junitparser/. In Haskell (not that we have to use Haskell), there isn’t an equivalent. However, all we do with the outer XML is parse the top-level list of test suites, and take the name for each testcase & its system output, so I’m also not sure if it matters.

The inner reports are just HTML files, in a format defined by hpc. Those, we don’t even parse. And I’ve wrapped those with my own tags, so I can use XML parsing instead of splitting on a separator or something.

haskell/private/coverage_wrapper.sh.tpl
```diff
@@ -109,12 +109,14 @@ fi
...
   cd hpc_out
   for file in *.html **/*.hs.html; do
     [ -e "$file" ] || continue
-    echo "<coverage-result-file name=\"$file\">"
+    echo "<coverage-report-part name=\"$file\">"
+    echo "<![CDATA["
     cat $file
-    echo "</coverage-result-file>"
+    echo "]]>"
+    echo "</coverage-report-part>"
```
Profpatsch: Of course this breaks if the concatenated file uses `CDATA`.

PR #831, opened 2019-04-16
Full CC/Haskell/CC sandwich using CC Skylark API

mboes:
We had the full sandwich previously, but it relied on two wrapper
rules, `cc_haskell_import` and `haskell_cc_import` along with various
hacks in order to do the job. Both of those rules are now deprecated.
Haskell rules now return a `CcInfo` provider the way any `cc_library`
does.


mboes:
I have pushed a new commit excluding haskell libraries from the interop logic, out of precaution. The interop logic should really just go away at this point. We should be treating Haskell and CC libraries uniformly as far as linking is concerned. But that will be a pretty massive refactor - for later PR's.

Issue #834, opened 2019-04-17
Handling of system libraries [Hazel]


nmattia:
Currently rules_haskell's support for dependencies on system libraries (libraries _not_ provided through Bazel but present on the system like `pthread`, `stdc++`, etc) is very hand wavy.

Say a cabal package specifies a dependency on a system library:

```
extra-libraries:
  - mysyslib
```

This library won't be found in the bazel dependency set of known libraries, because the library was _not_ provided through Bazel. However we shouldn't fail the build either since this library _is_ present on the system and can be linked against. Our [current solution](https://github.com/tweag/rules_haskell/blob/2c25c4d9e49efb893b8bcd8a108fcf2622ee8f72/hazel/third_party/cabal2bazel/bzl/cabal_package.bzl#L45-L54) is to check against a set of library names that we know are system libraries. If `mysyslib` is included, then we let the build go through, do tell GHC to link against `mysyslib` but let GHC find the library in the system library paths.

The current solution is not ideal because it requires us to keep an exact list of system libraries, which is not feasible since this may vary from system to system. Instead we should allow the user to override the set of system libraries when initializing hazel.

Another related issue is that of system library dependencies that are _not_ specified in cabal files. Both cabal and stack link against some libraries by defaut (e.g. `stdc++`) and packages generally assume they do not have to specify a dependency on those packages. We should probably do the same.

aherrmann:
Hazel support has been removed and replaced by `stack_snapshot`. The `extra_deps` attribute to `stack_snapshot` can be used to define per package C library dependencies.

PR #835, opened 2019-04-18
Hazel: Make excluded C libraries configurable

aherrmann:
Some packages list libraries in the extra-libraries section of their cabal file, which are always provided by the system and do not need to be provided as Bazel targets. E.g. libstdc++, or libpthread.

Previously, Hazel held a hard-coded list of such libraries for which it would not expect a user-provided Bazel target.

This change allows users to configure this list of system libraries themselves by passing empty string entries in the extra_libs argument to hazel_repositories. A default selection is provided.

Addresses the user configurability aspect of #834.

### DIFF
--- a/hazel/hazel.bzl
+++ b/hazel/hazel.bzl
@@ -15,6 +15,7 @@ load(
     "@bazel_tools//tools/build_defs/repo:utils.bzl",
     "patch",
 )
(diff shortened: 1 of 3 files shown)


nmattia:
Sidenote: on windows, it might be `winpthread`

PR #836, opened 2019-04-18
Write the -optP flags to a response because they can be very long

ndmitchell:
Before I failed to build grpc-haskell because I had 42Kb of command line. After, I can build it because 27K of that can go in a response file passed to the preprocessor. While the fix is only necessary on Windows, I suspect it's good practice to apply it universally, for consistency.

We are using this patch in the digital-asset/daml repo.

### DIFF
--- a/haskell/private/actions/compile.bzl
+++ b/haskell/private/actions/compile.bzl
@@ -214,7 +214,7 @@ def _compilation_defaults(hs, cc, java, dep_info, plugin_dep_info, srcs, import_
 
     ghc_args += ["-i{0}".format(d) for d in set.to_list(import_dirs)]
 
-    # Write the -optP flags to a response because they can be very long on Windows
+    # Write the -optP flags to a parameter file because they can be very long on Windows
     # e.g. 27Kb for grpc-haskell
     # Equivalent to: ghc_args += ["-optP" + f for f in cc.cpp_flags]
     optp_args_file = hs.actions.declare_file("optp_args_%s" % hs.name)
--- a/haskell/private/actions/compile.bzl
+++ b/haskell/private/actions/compile.bzl
@@ -213,7 +213,18 @@ def _compilation_defaults(hs, cc, java, dep_info, plugin_dep_info, srcs, import_
             set.mutable_insert(import_dirs, idir)
 
     ghc_args += ["-i{0}".format(d) for d in set.to_list(import_dirs)]
-    ghc_args += ["-optP" + f for f in cc.cpp_flags]
+
+    # Write the -optP flags to a response because they can be very long on Windows
+    # e.g. 27Kb for grpc-haskell
+    # Equivalent to: ghc_args += ["-optP" + f for f in cc.cpp_flags]
+    optp_args_file = hs.actions.declare_file("optp_args_%s" % hs.name)
+    optp_args = hs.actions.args()
+    optp_args.add_all(cc.cpp_flags)
+    optp_args.set_param_file_format("multiline")
+    print(cc.cpp_flags)
+    hs.actions.write(optp_args_file, optp_args)
+    ghc_args += ["-optP@" + optp_args_file.path]
+
     ghc_args += cc.include_args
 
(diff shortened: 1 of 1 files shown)


ndmitchell:
I am unable to see the CircleCI failure - it gives me a 404 (suggesting i should perhaps log in - and only lets me do that by granting CircleCI access to all my repos...)

mboes:
I don't see a place in CircleCI to avoid having to login. We're considering standardizing on Azure Pipelines for all platforms (we already use it for Windows). Perhaps this would be another reason to do so.

haskell/private/actions/compile.bzl
```diff
@@ -213,7 +213,18 @@ def _compilation_defaults(hs, cc, java, dep_info, plugin_dep_info, srcs, import_
...
-    ghc_args += ["-optP" + f for f in cc.cpp_flags]
+
+    # Write the -optP flags to a response because they can be very long on Windows
+    # e.g. 27Kb for grpc-haskell
+    # Equivalent to: ghc_args += ["-optP" + f for f in cc.cpp_flags]
+    optp_args_file = hs.actions.declare_file("optp_args_%s" % hs.name)
+    optp_args = hs.actions.args()
+    optp_args.add_all(cc.cpp_flags)
+    optp_args.set_param_file_format("multiline")
+    print(cc.cpp_flags)
```
aherrmann:
```suggestion
```
Looks like debugging left-overs?

ndmitchell: Oops - thanks!

haskell/private/actions/compile.bzl
```diff
@@ -213,7 +213,17 @@ def _compilation_defaults(hs, cc, java, dep_info, plugin_dep_info, srcs, import_
             set.mutable_insert(import_dirs, idir)
 
     ghc_args += ["-i{0}".format(d) for d in set.to_list(import_dirs)]
-    ghc_args += ["-optP" + f for f in cc.cpp_flags]
+
+    # Write the -optP flags to a response because they can be very long on Windows
```
guibou: What do you mean by a "response"?

mboes: [response file](https://github.com/bazelbuild/bazel/issues/3750).

guibou: If you like this kind of dialog: https://github.com/bazelbuild/bazel/issues/3750#issuecomment-330201248

tl;dr; For me a response file is a file which contains response to questions, for an automatic installation processes for example. Perhaps the term in not precise enough and may surprise future developers.

ndmitchell: I called it a parameter file in the comment, to match the Bazel terminology

aherrmann:
I had a look at the CI failure on this one. The failure can be reproduced (at least on nixpkgs Linux) with `bazel build //tests/haddock:haddock`. The trouble is that the haddock rule [copies over all ghc_args for a target and forwards them to haddock as `--optghc`](https://github.com/tweag/rules_haskell/blob/040174836072dd4622d5ee0c470fd4acd45668f7/haskell/haddock.bzl#L81-L82). This includes the `-optP@...` flag. However, the parameter file itself is not forwarded to the action inputs in the haddock rule and the action fails with
```
gcc: error: bazel-out/k8-fastbuild/bin/tests/haddock/optp_args_haddock-lib-a: No such file or directory
```

I haven't checked, but the same issue might also appear in runghc/repl/doctest rules. 

To demonstrate, the issue can be worked around by adding the optp parameter file to the `extra_source_files` entry in the returned struct [in `_compilation_defaults`](https://github.com/tweag/rules_haskell/pull/836/files#diff-33a8c46955e2a9917090b5258346aee2L337):
```
        extra_source_files = depset(transitive = [extra_srcs, depset([optp_args_file])]),
```

aherrmann:
netlify still fails with `InvalidAccessKeyId`.

mboes:
It deployed this time around.

PR #837, opened 2019-04-18
Fix early issues with HTML coverage reports and coverage scoping

SebastianKG:
This is a little silly: genfiles lose their path qualification (by `bazel-out/k8-fastbuild/genfiles/`) when being passed as runfiles, but .mix files remember their paths as being qualified. So, to avoid losing that information, we have to copy all genfiles to double-qualified names, pass them as runfiles, and then tell `hpc` where to find them. We can't tell `hpc` how to find them when they're not double-qualified, because then their paths don't match the .mix file expectation regardless.

### DIFF
--- a/haskell/private/coverage_wrapper.sh.tpl
+++ b/haskell/private/coverage_wrapper.sh.tpl
@@ -114,12 +114,15 @@ then
   $hpc_path markup "$tix_file_path" $hpc_dir_args $hpc_exclude_args \
     --srcdir "." --srcdir "$package_path" --destdir=hpc_out > /dev/null 2>&1
   cd hpc_out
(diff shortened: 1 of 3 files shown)


SebastianKG:
^ additional work: I added in a separator to the testlog output which denotes where HTML reports can be found, because there was noise coming into the output in some environments.

SebastianKG:
^ another change: removed duplicates in the coverage wrapper scripts parameters which are lists. I noticed that in a larger codebase, the test was timing out because the quoted list took up over 300,000 characters (!!). In the case I saw, this was almost entirely because there were many duplicate entries (there were diamond patterns in the dependency graph).

This is an initial fix, but in fact some codebases may hit this limit legitimately. Something to think about down the line.

PR #840, opened 2019-04-23
ls_modules is more robust with duplicate packages

### LINKED ISSUE #839: Failure: Multiple versions of the following packages installed
**Describe the bug**

My environment contains many version of the same package, e.g.:

```
$ bazel run @ghc-toolchain//:bin/ghc-pkg list | grep haskell-gi-
    haskell-gi-overloading-0.0
    haskell-gi-overloading-1.0
```

A bazel build of a target (which does not make any reference to `haskell-gi-overloading` fails with the following:

```
$ bazel build //atarget
...
Multiple versions of the following packages installed: 
haskell-gi-overloading, haskell-src-meta

This is not currently supported.
```

**To Reproduce**

Use commit 5796e5d9c3b28fd63aabd07e9a0e6f95715b472d

then

```
cd multiple_version_repro
nix-shell
bazel build //:foo
```

guibou:
ls_modules was crashing with an error if an haskell package was
duplicated in the package database. I.e if it exists a least two
package with the same package name and a different package id.

This commit changes the error behavior, it now only crash when the
package is explicitly used by the reexported module phase. So as long
as the duplicate is not explicitly used, it does not crash.

Closes #839

### DIFF
--- a/haskell/private/ls_modules.py
+++ b/haskell/private/ls_modules.py
@@ -32,19 +32,16 @@ with io.open(global_pkg_db_dump, "r", encoding='utf8') as f:
 
     # A few sanity checks.
     assert len(names) == len(ids)
+
+    # compute duplicate, i.e. package name associated with multiples ids
+    duplicates = set()
     if len(names) != len(set(names)):
-        duplicates = [
+        duplicates = set([
             name for name, count in collections.Counter(names).items()
             if count > 1
-        ]
-        sys.exit(
-            "\n".join([
-                "Multiple versions of the following packages installed: ",
-                ", ".join(duplicates),
-                "\nThis is not currently supported.",
-            ])
-        )
+        ])
 
+    # This associate pkg name to pkg id
     pkg_ids_map = dict(zip(names, ids))
 
 with io.open(hidden_modules_file, "r", encoding='utf8') as f:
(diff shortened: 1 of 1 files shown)


haskell/private/ls_modules.py
```diff
@@ -57,10 +54,36 @@
...
+    # Any duplicate used during the replacement phase will trigger a
+    # script failure later
+    duplicate_used = set()
+
+    def replace_pkg_by_pkgid(match):
+        pkgname = match.group(1)
+
+        if pkgname in duplicates:
+            duplicate_used.add(pkgname)
+            return "" # arbitrary return value, the script will crash later
```
mboes: Why not fail right then and here?

guibou: Bette user experience. If there are many *used* duplicates, it will fail and report all used duplicates, instead of the first one.

That's a tradeoff, but I can simplify the source code and fail on the first error.

mboes: I think it's better to simplify the logic. It is unlikely that there will be many duplicate modules to deal with, so it's not onerous for the user to print only the first one. I think you can even get rid of the inner function definition this way.

guibou: @mboes I fixed the logic. However I don't understand when you say:

> I think you can even get rid of the inner function definition this way

If you are talking about the newly introduced `replace_pkg_by_pkgid`, you are right that everything can be put inside a `lambda`, but from my point of view this will lead to an unreadable chunk of code. In case you were meaning something else, could you detail?

mboes:
Making it a function is fine.

PR #842, opened 2019-04-25
Add GHC 8.6.5 bindist

cocreature:


haskell/gen_ghc_bindist.py
```diff
@@ -18,6 +18,7 @@
 # `ignore_suffixes` is the suffix of files to ignore
 VERSIONS = [
     { "version": "8.6.4" },
+    { "version": "8.6.5" },
```
hurryabit: My OCD says this should move one line further up. 😃

mboes:
The CI failure is spurious: CircleCI has apparently decided all of a sudden that external contributors are "not authorized" to run macOS builds (and then marking the whole build as failed). Or maybe that's true of all of us since we overran our macOS build minutes budget this month...

haskell/gen_ghc_bindist.py
cocreature: Fixed :smile:

Issue #838, opened 2019-04-23
Support static linking of Haskell libraries in CC binaries


mboes:
**Is your feature request related to a problem? Please describe.**

CC binaries can either be "mostly static" or "dynamic". To get the latter we need to set `linkstatic = False`. Without this attribute, CC binaries that depend on Haskell binaries fail with a link error, because prebuilt libraries do not advertise to the CC rules where to find their respective static library file.

**Describe the solution you'd like**

Linking against Haskell binaries should be supported even with `linkstatic = False` in `cc_binary`, just like it is already supported in `haskell_binary`. For this to be possible, *all* Haskell libraries (including the RTS) should return a `CcInfo` provider.

PR #843, opened 2019-04-25
Rename haskell_import to haskell_toolchain_library

mboes:
This is a substantial breaking change. But a simple one to handle: it
can be done with

```
sed -i s/^haskell_import/haskell_toolchain_library/ **/BUILD{,.bazel}
```

everywhere. I considered marking this as deprecated instead, but that
won't help much: the deprecation spam will make things unbearable
without switching to the new name anyways.

The motivation for this change can be found in #440 and #838.
`haskell_import` was always poorly named, because it did not
correspond to precedent in the CC and Java rules. Whereas those rules
"import" concrete artifacts with some metadata into the dependency
graph, our old `haskell_import` was merely giving a name in the
dependency graph to stuff that is magically available from somewhere,
namely the GHC boot libraries.

I will be submitting a followup PR to implement a new `haskell_import`
that is similar in spirit to `cc_import` and `java_import`, once
I manage to solve/workaround the two following bugs:

* https://groups.google.com/forum/?nomobile=true#!topic/bazel-discuss/gQRd7CGmVu8
* https://github.com/bazelbuild/bazel/issues/8129

Meanwhile this PR just prepares the ground, by moving the existing
`haskell_import` out of the way and renaming it to
`haskell_toolchain_library`.


mboes:
Yes. TBH I don't quite understand `haskell_import_new`. Maybe @regnat can comment on whether this can replace `haskell_import_new`, which AFAIK was an experiment that at this point is only used in the rules_haskell test suite. `haskell_import_new` enables an alternative modular way to interoperate with Nix. I don't know whether we'll be able to recover that (though it's not very usable at the moment because it's very slow).

PR #828, opened 2019-04-15
hazel: windows: fix unix-compat

aherrmann:
- Allow to apply patches to Hazel packages fetched from Hackage or custom URLs.
- Apply patch to `unix-compat` fixing issue #796 due to missing version macros in `hsc` files.

Closes #796.

PR #847, opened 2019-04-26
Generate version macros for preprocessors

### LINKED ISSUE #821: Missing version macros on hsc files.
**Describe the bug**
rules_haskell tells ghc to generate version macros if a target has a version attribute. However, with hsc files rules_haskell runs them through the hsc2hs preprocessor first, which doesn't seem to generate version macros in that case.

**To Reproduce**
Checkout https://github.com/tweag/rules_haskell/commit/25fc107002d9b8bb260da15ea058e163a28f65b3 and execute `nix-shell --pure --run 'bazel test //tests/package-name:bin-c'`

**Expected behavior**
Version macros should also be generated for `.hsc` files, so that the example above passes.

**Environment**
 - NixOS + 18.03:
 - Bazel version: `rules_haskell` Nix shell.
 - Version of the rules: Revision 25fc107002d9b8bb260da15ea058e163a28f65b3

**Additional context**
https://github.com/tweag/rules_haskell/commit/25fc107002d9b8bb260da15ea058e163a28f65b3 adds a regression test which currently fails with the following error due to missing version macros:
```
Main.hsc:12:24: error: missing binary operator before token "("
Main.hsc:18:24: error: missing binary operator before token "("
Main.hsc: In function 'main':
Main.hsc:12:24: error: missing binary operator before token "("
Main.hsc:18:24: error: missing binary operator before token "("
Main.hsc:12:24: error: missing binary operator before token "("
Main.hsc:18:24: error: missing binary operator before token "("
```

This also seems to be the cause for https://github.com/tweag/rules_haskell/issues/796.

aherrmann:
GHC is capable of defining version macros itself. However, this does not affect the preprocessors `c2hs` and `hsc2hs`. Cabal, for example, generates its own version macros header file for these preprocessors. This PR adds this capability to rules_haskell.

Closes #821 

- Generate version macros for dependencies that define the `version` attribute.
- Add `version_macros` field to `HaskellBuildInfo`.
- Pass `-include <version_macros_header>` through `cpp_flags` as defined by `cc_interop_info`.
- Add regression test for `c2hs`, `hsc2hs`, and regular Haskell files with `-XCPP`.
- Remove patch to `unix-compat` inlining version macros. See #828


haskell/private/haskell_impl.bzl
```diff
@@ -401,10 +401,29 @@ def haskell_library_impl(ctx):
             set.singleton(c_p.interfaces_dir),
         )
 
+    version_macros = set.empty()
```
guibou: Can this be factored with the code in `haskell_import.bzl`?

haskell/private/version_macros.py
```diff
@@ -0,0 +1,101 @@
...
+    return "#define {macro}{params} {val}".format(
+        macro = macro,
+        params = "({})".format(",".join(params)) if params else "",
+        val = val,
+    )
+
+
+def cpp_ifndef(macro, body):
+    """CPP ifndef block."""
+    return "#ifndef {macro}\n{body}\n#endif /* {macro} */".format(
```
guibou:
```
return f"#ifndef {macro}\n{body}\n#endif /* {macro} */"
```

Is enough (provided that python version in >= 3.6

haskell/private/version_macros.py
```diff
@@ -0,0 +1,101 @@
...
+    [1]: http://hackage.haskell.org/package/Cabal-2.4.1.0/docs/src/Distribution.Simple.Build.Macros.html#generatePackageVersionMacros
+    """
+    (major1, major2, minor) = version_components(version)
+    escaped_name = cpp_escape_name(name)
+    return "\n".join([
+        # #define VERSION_pkg "1.2.3"
+        cpp_ifndef_define(
+            "VERSION_" + escaped_name,
+            [],
+            "\"{}\"".format(version),
```
guibou: Nit. I don't like escaping.

```
'"{}"'
```

haskell/private/version_macros.py
```diff
@@ -0,0 +1,101 @@
...
+        ),
+        # #define MIN_VERSION_pkg(major1, major2, minor) ...
+        cpp_ifndef_define(
+            "MIN_VERSION_" + escaped_name,
+            ["major1", "major2", "minor"],
+            " \\\n".join([
+                "(",
+                "  (major1) < {} ||".format(major1),
```
guibou: This is also a good place for f strisgs

```
f"  (major1) < {major1} ||"
```

...

haskell/private/version_macros.py
aherrmann: If I try that I'm getting 
```
    return f"#ifndef {macro}\n{body}\n#endif /* {macro} */"
                                                          ^
SyntaxError: invalid syntax
```
despite
```
$ nix-shell --pure --run 'python3 --version'
Python 3.6.8
```
Is there some escape I'm missing?

haskell/private/haskell_impl.bzl
aherrmann: I've factored it out into `haskell/private/version_macros.bzl`.

haskell/private/version_macros.py
aherrmann: done

haskell/private/version_macros.py
aherrmann: Same issue here
```
    f"  (major1) < {major1} ||",
                              ^
SyntaxError: invalid syntax
```

haskell/private/version_macros.py
guibou:
```
Python 3.6.8 (default, Dec 24 2018, 03:01:30) 
[GCC 7.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> macro = "hello"
>>> body = "world"
>>> f"#ifndef {macro}\n{body}\n#endif /* {macro} */"
'#ifndef hello\nworld\n#endif /* hello */'
```

You should be right.

Perhaps the python used by your bazel install is not a python 3.6, despite what your shell claim.

Don't waste time with this nitpick ;)

aherrmann:
rebased on master

PR #852, opened 2019-04-27
Simplify and rationalize provider model

mboes:
Depends on #844.

We used to have the following main providers:

* `HaskellBuildInfo` exposing information common to both binaries and
  libraries,
* `HaskellBinaryInfo` exposing information specific to binaries,
* `HaskellLibraryInfo` exposing information specific to libraries.

We now have:

* `HaskellInfo`, a generic provider returned by all core Haskell
  rules.
* `HaskellLibraryInfo`, returned only by `haskell_library`.

Most of the provider attributes are now in `HaskellInfo` (though
several have been removed entirely). A library is something that
additionally has an identifier and a version number, that's it.

It's best to review this commit by commit, but for each commit, first
look at the changes in `haskell/providers.bzl`. All other changes in
the commit are a direct consequence of any changes in that first file.


haskell/haddock.bzl
```diff
@@ -121,10 +121,7 @@ def _haskell_doc_aspect_impl(target, ctx):
             depset(trans_libs),
             depset(transitive_haddocks.values()),
             depset(transitive_html.values()),
-            # Need to give source files this way because the source_files field of
-            # HaskellLibraryInfo provider contains files that are already
-            # pre-processed by hsc2hs and these should be visible to Haddock.
-            set.to_depset(target[HaskellLibraryInfo].header_files),
```
Profpatsch: Hm, doesn’t that remove the `hsc2hs` preprocessing?

haskell/cc.bzl
```diff
@@ -312,8 +305,8 @@ cc_haskell_import = rule(
     attrs = {
         "dep": attr.label(
             doc = """
-Target providing a `HaskellLibraryInfo` or `HaskellBinaryInfo`, such as
-`haskell_library` or `haskell_binary`.
+Target providing a `HaskellLiInfo` such as `haskell_library` or
```
Profpatsch: `HaskellLiInfo`

haskell/haddock.bzl
```diff
@@ -121,10 +121,7 @@ def _haskell_doc_aspect_impl(target, ctx):
...
+            target[CcInfo].compilation_context.headers,
```
Profpatsch: Hm, doesn’t that remove the `hsc2hs` preprocessing?

Profpatsch: We found out that the comment was originally talking about two lines down.

mboes:
@aherrmann this branch interacts badly with #847 in ways that I don't understand. Could you have a look?

aherrmann:
@mboes Sure, I'll have a look.

aherrmann:
I've rebased this PR on master. I had to change the version macros generation because it was relying on the headers attribute of the HaskellLibrary provider. See [commit message](https://github.com/tweag/rules_haskell/pull/852/commits/a5826f1ba905cef2a7f99fa8ebbf38b7c428caa3) for details.

PR #853, opened 2019-04-29
Install GHC bindist in relocatable mode on Unix

aherrmann:
On Unix the GHC bindist requires to execute `./configure && make && make install`. In its default configuration, it hard-codes the absolute path to the bindist installation dir into various places in the final bindist. This affects the hashes and thereby cache keys, as the absolute install path will likely differ on different machines. This patch builds the bindist in relocatable mode on Unix and patches the generated wrapper scripts to no longer hard-code the absolute installation path.

- Pin the `--libdir` to `EPREFIX/lib`. On some platforms autotools defaults to `EPREFIX/lib64`, instead.
- Patch the configure script to build in relocatable mode.
- Patch the generated executable wrapper scripts to no longer hard-code the absolute path to the installation directory.

### DIFF
--- a/haskell/ghc_bindist.bzl
+++ b/haskell/ghc_bindist.bzl
@@ -212,8 +212,18 @@ def _ghc_bindist_impl(ctx):
 
     # On Windows the bindist already contains the built executables
     if os != "windows":
+        _execute_fail_loudly(ctx, ["sed", "-i", "s/RelocatableBuild = NO/RelocatableBuild = YES/", "mk/config.mk.in"])
         _execute_fail_loudly(ctx, ["./configure", "--prefix", bindist_dir.realpath, "--libdir", "${exec_prefix}/lib"])
         _execute_fail_loudly(ctx, ["make", "install"])
+        ctx.file("patch_bins", executable = True, content = """#!/usr/bin/env bash
+grep -lZ {bindist_dir} bin/* | xargs -0 --verbose \\
+    sed -i \\
+        -e '2iDISTDIR="$( dirname "$(readlink -f "$0")" )/.."' \\
+        -e 's:{bindist_dir}:$DISTDIR:'
+""".format(
+            bindist_dir = bindist_dir.realpath,
+        ))
+        _execute_fail_loudly(ctx, ["./patch_bins"])
 
     ctx.template(
         "BUILD",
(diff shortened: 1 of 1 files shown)


haskell/ghc_bindist.bzl
```diff
@@ -212,8 +212,20 @@ def _ghc_bindist_impl(ctx):
 
     # On Windows the bindist already contains the built executables
     if os != "windows":
-        _execute_fail_loudly(ctx, ["./configure", "--prefix", bindist_dir.realpath])
+        _execute_fail_loudly(ctx, ["sed", "-i", "s/RelocatableBuild = NO/RelocatableBuild = YES/", "mk/config.mk.in"])
+        _execute_fail_loudly(ctx, ["./configure", "--prefix", bindist_dir.realpath, "--libdir", "${exec_prefix}/lib"])
         _execute_fail_loudly(ctx, ["make", "install"])
+        ctx.file("patch_bins", executable=True, content="""#!/usr/bin/env bash
+set -x
```
mboes: The tracing stuff should probably go.

haskell/ghc_bindist.bzl
```diff
@@ -212,8 +212,20 @@ def _ghc_bindist_impl(ctx):
 
     # On Windows the bindist already contains the built executables
     if os != "windows":
-        _execute_fail_loudly(ctx, ["./configure", "--prefix", bindist_dir.realpath])
+        _execute_fail_loudly(ctx, ["sed", "-i", "s/RelocatableBuild = NO/RelocatableBuild = YES/", "mk/config.mk.in"])
+        _execute_fail_loudly(ctx, ["./configure", "--prefix", bindist_dir.realpath, "--libdir", "${exec_prefix}/lib"])
```
mboes: Where does `${exec_prefix}` come from?

haskell/ghc_bindist.bzl
aherrmann: Removed. Sorry, overlooked that.

haskell/ghc_bindist.bzl
aherrmann: `${exec_prefix}` is defined within `mk/install.mk`. However, it turns out that `${exec_prefix}/lib` is the default for `RelocatableBuild = YES`. So the `--libdir` flag is redundant. I've removed it.

PR #854, opened 2019-04-30
haskell_toolchain tools as a list

### LINKED ISSUE #848: Custom haddock executable

I want to change the `haddock` executable. For now I'm providing a `tool` argument to `haskell_toolchain` which point to a directory in which I manually overrode the `haddock` executable. This is not convenient.

I want a way to explicitly set the `haddock` executable.

I see two alternatives:

- The current implementation looks for binaries (`haddock`, but also `ghc`, `hpc`, ...) in the label provided by the `tools` attribute of `haskell_toolchain`. I propose to change this attribute to a list of label. Binaries will be searched from this list in order.
- Implement an `haddock` toolchain similar to the `c2hs_toolchain` recently introduced by @m.

I'm in favor of the first solution that I consider simpler and more generic. @m what were the reasons for the introduction of `c2hs_toolchain`?

guibou:
`haskell_toolchain` `tools` attribute is now a list of
labels. Binaries (such as `ghc`, `haddock`, ...) are now located in
this list, first entries take precedence.

This is a backward incompatible change, but the conversion of
repository is trivial (i.e. add list bracket on your `tools` attribute).

Closes #848

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ genrule(
 # Make it visible to rules_haskell rules:
 haskell_toolchain(
     name = "ghc",
-    tools = "@ghc//:bin",
+    tools = ["@ghc//:bin"],
     version = "8.4.1",
     extra_binaries = [":toolchain_as"], # <----
 )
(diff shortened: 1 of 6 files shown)


guibou:
Thanks. The build failure is trivial (a new instance of `tools` to update which appears when I rebased on top of `master`). I'll merge once CI pass.

PR #859, opened 2019-05-04
Bazelify compiler's global package database

### LINKED ISSUE #838: Support static linking of Haskell libraries in CC binaries

mboes:
Exposing GHC's global package database to Bazel enables us to stop
treating prebuilt libraries specially. They now export
a `HaskellLibraryInfo` like any other library. Even better, they now
also export a `CcInfo` provider like any other native code library,
which means that CC rules have a full view of all static archives
needed for linking a binary statically (see the `cc_haskell_import`
test, which no longer requires `linkstatic = False`).

`HaskellPrebuiltPackageInfo` is still there but now has no content.
It's only there to prevent aspects from reaching too far. We'll get
rid of it in future commits.

To get here, we had to make the `haskell_toolchain` macro stop
defining a `toolchain` target. This design is less like rules_go, but
more like the built-in CC rules, so we're still covered by precedent.

Fixes #838


haskell/private/pkgdb_to_bzl.py
```diff
@@ -0,0 +1,170 @@
...
+                        for include_dir in pkg.include_dirs
+                        if path_to_label(include_dir, pkg.pkgroot)
+                    ]),
+                    includes = [
+                        "/".join([repo_dir, path_to_label(include_dir, pkg.pkgroot)])
+                        for include_dir in pkg.include_dirs
+                        if path_to_label(include_dir, pkg.pkgroot)
+                    ],
+                    static_library = "glob({})".format([
+                        path_to_label("{}/lib{}.a".format(library_dir, hs_library), pkg.pkgroot)
```
aherrmann: Doesn't this also catch profiling libraries `libHSxyz_p.a`?

haskell/private/pkgdb_to_bzl.py
```diff
@@ -0,0 +1,170 @@
...
+                        if path_to_label(dynamic_library_dir, pkg.pkgroot)
+                        for ext in ["dll", "dylib", "so"]
+                    ]),
+                    deps = pkg.depends,
+                    linkopts = pkg.ld_options + [
+                        "-L{}".format(library_dir)
+                        for library_dir in pkg.dynamic_library_dirs + pkg.library_dirs
+                        if not path_to_label(library_dir, pkg.pkgroot)
+                    ] + [
+                        "-l{}".format(extra_library)
```
aherrmann: If `extra_library` is a static library, do these `linkopts` end up in the right position in the linker flags? I.e. behind the object that depends on the `extra_library`.

haskell/toolchain.bzl
```diff
@@ -206,6 +212,10 @@ _haskell_toolchain = rule(
             doc = "GHC and executables that come with it. First item take precedance.",
             mandatory = True,
         ),
+        "libraries": attr.label_list(
+            doc = "The set of libraries that come with GHC.",
+            #mandatory = True,
```
aherrmann:
```suggestion
```

haskell/doctest.bzl
```diff
@@ -80,7 +81,7 @@ def _haskell_doctest_single(target, ctx):
       File: the doctest log.
     """
 
-    if HaskellInfo not in target:
+    if HaskellInfo not in target or HaskellPrebuiltPackageInfo in target:
```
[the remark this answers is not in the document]
mboes: This test will be removed in a future PR (that actually and finally removes `HaskellPrebuiltPackageInfo`). In this PR I'm just doing the simplest thing.

haskell/private/pkgdb_to_bzl.py
```diff
@@ -0,0 +1,170 @@
...
+def path_to_label(path, pkgroot):
+    """Substitute one pkgroot for another relative one to obtain a label."""
+    topdir_relative_path = path.replace(pkgroot, "$topdir")
+    if topdir_relative_path.find("$topdir") != -1:
+        return topdir_relative_path.replace("$topdir", topdir).replace('\\', '/')
+
+pkg_confs = {}
+for conf in glob.glob(os.path.join(topdir, "package.conf.d", "*.conf")):
+    with open(conf, 'r') as f:
+        pkg_confs[conf] = f.read()
```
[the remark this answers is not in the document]
mboes: Good point. I will fix this to be more lazy and reduce peak memory usage.

haskell/private/pkgdb_to_bzl.py
mboes: In theory no. There is no wildcard in the glob here. The only reason we use `glob` is so that if the library file is missing (some packages don't have *both* a shared and a static library), then it won't error out.

haskell/private/pkgdb_to_bzl.py
mboes: Needs a comment.

haskell/private/pkgdb_to_bzl.py
aherrmann: Oh, I see. Sorry, assumed there was a wild-card.

hazel/third_party/cabal2bazel/bzl/cabal_package.bzl
```diff
@@ -62,13 +63,13 @@ def _get_extra_libs(libnames, extra_libs):
             result.append(lookup)
     return result
 
-def _get_core_dependency_includes(ghc_workspace):
+def _get_core_dependency_includes(name):
```
aherrmann: IIRC this is only necessary, because `haskell_import` used to not include headers. Now that it does, I suspect that we can completely drop this part.

hazel/third_party/cabal2bazel/bzl/cabal_package.bzl
```diff
@@ -371,8 +370,13 @@ def _get_build_attrs(
             # Collect direct cbits dependencies.
             cdeps[condition] += [hazel_cbits(p)]
             if p in _CORE_DEPENDENCY_INCLUDES:
-                cdeps[condition] += [_CORE_DEPENDENCY_INCLUDES[p]]
-                deps[condition] += [_CORE_DEPENDENCY_INCLUDES[p]]
+                # Create per-package unique import of core dependency.
+                haskell_toolchain_library(
+                    name = _CORE_DEPENDENCY_INCLUDES[p],
```
aherrmann: Hazel generates `haskell_toolchain_library` and `cc_library` targets for all items of `core_packages`. This includes `unix`, and `rts`. So, they should be available as `@haskell_unix//:unix(-cbits)` and `@haskell_rts//:rts(-cbits)`. However, the `-cbits` targets are empty dummy targets. Now that `haskell_toolchain_library` contains a `CcInfo`, we can probably drop much of the `cbits` logic.

hazel/third_party/cabal2bazel/bzl/cabal_package.bzl
```diff
@@ -425,6 +429,12 @@ def _get_build_attrs(
 
     elibs_targets = _get_extra_libs(build_info.extraLibs, extra_libs)
 
+    # Create per-package unique import of core dependency.
+    rts_import_name = "__hazel_{}_rts".format(name)
+    haskell_toolchain_library(
```
aherrmann: Same here. That should already exist in `@haskell_rts//:rts`.

hazel/third_party/cabal2bazel/bzl/cabal_package.bzl
```diff
@@ -200,7 +201,6 @@ def _get_build_attrs(
         generated_srcs_dir,
         extra_modules,
         ghc_version,
-        ghc_workspace,
```
aherrmann: :tada: Good riddance!

hazel/third_party/cabal2bazel/bzl/cabal_package.bzl
aherrmann: Could creating these per-package targets be the reason for the large `(LD_)LIBRARY_PATH`. IIUC each transitive Hackage dependency will add another instance of these libraries to the dependencies.

hazel/third_party/cabal2bazel/bzl/cabal_package.bzl
```diff
@@ -436,7 +446,7 @@ def _get_build_attrs(
                  ]),
         defines = [o[2:] for o in build_info.ccOptions if o.startswith("-D")],
         textual_hdrs = list(headers),
-        deps = ["{}//:rts-headers".format(ghc_workspace)] + select(cdeps) + cc_deps + elibs_targets,
+        deps = [rts_import_name] + select(cdeps) + cc_deps + elibs_targets,
```
aherrmann: I'm not sure if this is related to the MacOS segfault. But, a side effect of this is that `HSrts-ghc8.6.4` appears in the `extra-libraries` field of Hazel haskell_library targets.

Generally, in the following setup
```
haskell_library(name = "hs-lib-a", ...)
cc_library(name = "c-lib", deps = [":hs-lib-a"], ...)
haskell_library(name = "hs-lib-b", deps = [":c-lib", "hs-lib-a"], ...)
```
`hs-lib-b`'s package configuration file would list `hs-lib-a` both in `depends` and in `extra-libraries`.

mboes:
> This logic will then cause those entries to be stripped away silently, causing compiler errors due to missing headers and the like later on.

Best to start with at least failing early for now.

PR #860, opened 2019-05-05
docs: added missing dangling comma

ghuntley:


mboes:
I don't know why CircleCI now thinks it's fair to refuse to build forks (on Darwin only) but nonetheless fail the build. Will file a support ticket.

PR #861, opened 2019-05-06
Hazel: Fix zlib on Windows

aherrmann:
- Add `includes = ["cbits"]` attribute to the `zlib` build definition.
    Using `strip_include_prefix` alone caused the Windows build to fail with the following error
    ```
    this rule is missing dependency declarations for the following files included by 
    'external/haskell_zlib/cbits/adler32.c':
      'external/haskell_zlib/cbits/zutil.h'
      'external/haskell_zlib/cbits/zlib.h'
      'external/haskell_zlib/cbits/zconf.h'
    ```
    Adding `includes` resolves that issue. It seems that Bazel on Windows does not take `strip_include_prefix` into account when checking for undeclared dependencies. 
- Add `zlib` to Windows CI.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -33,13 +33,7 @@ jobs:
       /c/bazel/bazel.exe build --config windows "@haskell_cryptonite//..."
       /c/bazel/bazel.exe build --config windows "@haskell_unix__compat//..."
       /c/bazel/bazel.exe build --config windows "@haskell_hostname//..."
-
-      # FIXME:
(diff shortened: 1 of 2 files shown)


mboes:
>  It seems that Bazel on Windows does not take strip_include_prefix into account when checking for undeclared dependencies.

@aherrmann is there an upstream issue about this?

aherrmann:
@mboes Not yet, I still need to create a minimal repro.

PR #876, opened 2019-05-13
Packaged the coverage report renderer for Hackage

SebastianKG:
Since it's not feasible to establish a WORKSPACE function for clients to build the coverage report renderer using a BUILD file inside of Rules_Haskell, the next best thing is to push it up to Hackage and let them pull it down themselves. This is the work necessary to make that possible.


tools/coverage-reports/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+If we have a target defined by the Rules_Haskell `haskell_test` rule, such as:
+
+```
+haskell_test(
+    name = "two-libs",
+    srcs = [
+        ...
+    ],
+    coverage_report_format = "html", # this must be set
```
Profpatsch: `# this must be set and can be one of …`

tools/coverage-reports/package.yaml
```diff
@@ -0,0 +1,36 @@
...
+maintainer:          "m@tweag.io"
+copyright:           "2019 Tweag I/O Limited"
+
+extra-source-files:
+- LICENSE
+- package.yaml
+- README.md
+- stack.yaml
+
+synopsis:            Generate coverage report from Rules_Haskell bazel coverage output. See README at <https://github.com/tweag/rules_haskell/blob/master/tools/coverage-reports/README.md>
```
Profpatsch: You’re distributing the `README.md` with the source, so this link should not be needed.

tools/coverage-reports/package.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - cmdargs
+    - directory
+    - filepath
+    - hxt
+    - hxt-xpath
+    - listsafe
+    ghc-options:
+    - -rtsopts
+    - -threaded
+    - -with-rtsopts=-N
```
Profpatsch: Is this actually something which benefits from parallelizing?

SebastianKG: probably not. Removed

Issue #874, opened 2019-05-13
Direct Cabal support


mboes:
**Is your feature request related to a problem? Please describe.**

We have @formationai's Hazel checked into this repository. Using Hazel, we can emulate Cabal on top of basic rules_haskell primitives, like `haskell_library`, `haskell_binary` etc. The problem is that Hazel is not simple. More specifically,

* Cabal is a beast with by now very complicated semantics that don't map very well to assumptions made by rules_haskell. For example, we end up needing to push versions through `haskell_library` and also reimplement our own CPP version macros for those tools that don't generate them themselves. This adds some complexity and extra code to the core rules. But there are also cases where rules_haskell makes Hazel's life hard, e.g. because rules_haskell insists on compiling cbits separately from Haskell code, leading to hacky workarounds in Hazel.
* Getting Hazel support to the level where all Cabal packages work even on Windows has proved challenging. Between .buildinfo files, configure scripts and the like, the surface area offered by Cabal is quite large, and getting everything right on all platforms is tough.

**Describe the solution you'd like**

There is another way to build Cabal libraries: call Cabal itself instead of emulating it. This is not a fantastic idea for new libraries, but for existing ones that already have a .cabal file (+ configure scripts and custom hooks potentially), better to reuse the existing build system as a black box rather than attempt translating it. We can still get the benefit of sandboxing, remote caching and distributed builds, if at a less granular level.

We would need two components.

One is, provided an unpacked source distribution, a way to call `Setup.hs`, to populate `HaskellInfo`, `HaskellLibraryInfo` and `CcInfo` providers in particular including a package database for the package and a (static, dynamic) library pair.

```python
haskell_cabal_library(
    name = "transformers-0.5.2.0",
    srcs = glob(["**"]),
    deps = ...,
)
```

If the above exposes the exact same providers as a `haskell_library`, then they become interchangeable.

The other component is a way to generate a dependency graph from a package set, so that the above doesn't need to be written manually. We can introduce a workspace rule that:

* given a list of Cabal dependencies, computes the transitive closure of all Cabal packages we'll need,
* downloads all source distributions for all transitive Cabal dependencies and unpacks them.

We could design it along the same lines as rules_jvm_external. Key differences with Hazel:

* There is a single external repository hosting all sources, rather than one per package.
* This makes it possible to have multiple versions of a Cabal package available, although defined in separate external repositories.
[...]

PR #879, opened 2019-05-19
Add haskell_cabal_library

mboes:
(This PR does not include Windows support, which should be easy to add later.)

This is an alternative solution for handling Cabal libraries published
on Hackage to both Hazel and Nixpkgs. The idea in this one is to use
Cabal to build Cabal packages, and Stack to download source
distributions from a Stackage snapshot and generate a dependency
graph. See #874 for the rationale. The Stack part will be submitted as
the next PR in the series.

This rule exposes all the same providers as `haskell_library`, so one
can be substituted for the other.

I haven't tried building all of Stackage, but as a data point, this
rule is able to build the close to 200 dependencies of Stack. The vast
majority of Cabal packages in Stackage build out-of-the-box, including
those that have system dependencies like `zlib`. Only packages that do
very funky things like attempt creating their own files in the CWD
fail (like Stack itself), because Bazel sensibly marks the source
directory as read-only.

cc @ndmitchell @judah

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -1,0 +1,245 @@
+"""Cabal packages"""
+
+load("@bazel_skylib//lib:dicts.bzl", "dicts")
+load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
+load(":cc.bzl", "cc_interop_info")
+load(":private/context.bzl", "haskell_context")
+load(":private/dependencies.bzl", "gather_dep_info")
+load(":private/mode.bzl", "is_profiling_enabled")
+load(":private/set.bzl", "set")
+load(
+    ":providers.bzl",
+    "HaskellInfo",
+    "HaskellLibraryInfo",
+    "empty_HaskellCcInfo",
+    "get_libs_for_ghc_linker",
+    "merge_HaskellCcInfo",
+)
+
+def _execute_or_fail_loudly(repository_ctx, arguments):
+    exec_result = repository_ctx.execute(arguments)
+    if exec_result.return_code != 0:
+        fail("\n".join(["Command failed: " + " ".join(arguments), exec_result.stderr]))
+    return exec_result
+
+def _so_extension(hs):
+    return "dylib" if hs.toolchain.is_darwin else "so"
+
+def _dirname(file):
+    return file.dirname
+
+def _haskell_cabal_library_impl(ctx):
+    hs = haskell_context(ctx)
+    dep_info = gather_dep_info(ctx, ctx.attr.deps)
+    cc = cc_interop_info(ctx)
+    cc_info = cc_common.merge_cc_infos(
+        cc_infos = [dep[CcInfo] for dep in ctx.attr.deps if CcInfo in dep],
+    )
+    name = hs.label.name
+    with_profiling = is_profiling_enabled(hs)
+
+    # Check that a .cabal file exists. Choose the root one.
+    cabal = None
+    for f in ctx.files.srcs:
+        if f.extension == "cabal":
+            if not cabal or f.dirname < cabal.dirname:
+                cabal = f
+    if not cabal:
+        fail("A .cabal file was not found in the srcs attribute.")
+
+    # Check that a Setup script exists. If not, create a default one.
+    setup = None
+    for f in ctx.files.srcs:
+        if f.basename in ["Setup.hs", "Setup.lhs"]:
+            if not setup or f.dirname < setup.dirname:
+                setup = f
+    if not setup:
+        setup = hs.actions.declare_file("Setup.hs", sibling = cabal)
+        hs.actions.write(
+            output = setup,
+            content = """
+module Main where
+import Distribution.Simple
+
+main :: IO ()
+main = defaultMain
+""",
+        )
+
+    package_database = hs.actions.declare_file(
+        "_install/package.conf.d/package.cache",
+        sibling = cabal,
+    )
+    interfaces_dir = hs.actions.declare_directory(
+        "_install/iface",
+        sibling = cabal,
+    )
+    static_library = hs.actions.declare_file(
+        "_install/lib/libHS{}.a".format(name),
+        sibling = cabal,
+    )
+    static_library_prof = None
+    if with_profiling:
+        static_library_prof = hs.actions.declare_file(
+            "_install/lib/libHS{}_p.a".format(name),
+            sibling = cabal,
+        )
+    dynamic_library = hs.actions.declare_file(
+        "_install/lib/libHS{}-ghc{}.{}".format(name, hs.toolchain.version, _so_extension(hs)),
+        sibling = cabal,
+    )
+
+    args = hs.actions.args()
+    package_databases = set.to_depset(dep_info.package_databases)
+    extra_headers = cc_info.compilation_context.headers
+    extra_include_dirs = cc_info.compilation_context.includes
+    (library_deps, ld_library_deps, ghc_env) = get_libs_for_ghc_linker(
+        hs,
+        dep_info.transitive_cc_dependencies,
+    )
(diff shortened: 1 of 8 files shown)


haskell/private/cabal_wrapper.sh.tpl
```diff
@@ -0,0 +1,68 @@
...
+#
+# cabal_wrapper.sh <PKG_NAME> <SETUP_PATH> <PKG_DIR> <PACKAGE_DB_PATH> [SETUP_ARGS...]
+
+set -euo pipefail
+execroot=$(pwd)
+
+# Set these variables if unset.
+: ${LD_LIBRARY_PATH:=}
+: ${LIBRARY_PATH:=}
+export LD_LIBRARY_PATH=$execroot/$LD_LIBRARY_PATH
```
aherrmann: `LD_LIBRARY_PATH` and `LIBRARY_PATH` are colon separated lists of paths. This only prefixes the `$execroot` to the first entry in that list.

haskell/private/cabal_wrapper.sh.tpl
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
+#
+# cabal_wrapper.sh <PKG_NAME> <SETUP_PATH> <PKG_DIR> <PACKAGE_DB_PATH> [SETUP_ARGS...]
+
+set -euo pipefail
+execroot=$(pwd)
```
aherrmann:
```suggestion
execroot="$(pwd)"
```
In principle the path to the execroot could contain whitespace. Granted, a lot of bazel rules would fail in that case. (Same for all other variables below).

haskell/cabal.bzl
```diff
@@ -0,0 +1,247 @@
...
+        },
+    )
+
+    # Make the Cabal configure/build/install steps one big action so
+    # that we don't have to track all inputs explicitly between steps.
+    ctx.actions.run(
+        executable = cabal_wrapper,
+        arguments = [args],
+        inputs = depset([setup], transitive = [
+            depset([hs.tools.ghc, hs.tools.ghc_pkg, hs.tools.runghc]),
```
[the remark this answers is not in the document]
mboes: No reason. I'll apply your suggestion.

haskell/cabal.bzl
```diff
@@ -0,0 +1,247 @@
...
+        ] + ([static_library_prof] if with_profiling else []),
+        env = dicts.add(ghc_env, hs.env),
+        mnemonic = "HaskellCabalLibrary",
+        progress_message = "HaskellCabalLibrary {}".format(hs.label),
+        use_default_shell_env = True,
+    )
+
+    default_info = DefaultInfo(files = depset([static_library, dynamic_library]))
+    hs_info = HaskellInfo(
+        package_ids = set.empty(),
```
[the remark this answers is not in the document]
mboes: This field only gets overwritten during dependency info gathering. I'm not sure why it's part of the provider TBH.

haskell/private/cabal_wrapper.sh.tpl
mboes: This line should include the variable substitution that is happening a few lines below. Will fix.

haskell/cabal.bzl
```diff
@@ -0,0 +1,247 @@
...
+    with_profiling = is_profiling_enabled(hs)
+
+    # Check that a .cabal file exists. Choose the root one.
+    cabal = None
+    for f in ctx.files.srcs:
+        if f.extension == "cabal":
+            if not cabal or f.dirname < cabal.dirname:
+                cabal = f
+    if not cabal:
+        fail("A .cabal file was not found in the srcs attribute.")
```
Profpatsch: The `cabal` file is always named `projectname.cabal` and has to be toplevel afaik.

The new support for cabal projects is going to be an important TODO for later.

haskell/cabal.bzl
```diff
@@ -0,0 +1,247 @@
...
+)
+
+def _execute_or_fail_loudly(repository_ctx, arguments):
+    exec_result = repository_ctx.execute(arguments)
+    if exec_result.return_code != 0:
+        fail("\n".join(["Command failed: " + " ".join(arguments), exec_result.stderr]))
+    return exec_result
+
+def _so_extension(hs):
+    return "dylib" if hs.toolchain.is_darwin else "so"
```
Profpatsch: I think we have this logic in multiple places by now.

haskell/cabal.bzl
```diff
@@ -0,0 +1,247 @@
...
+                cabal = f
+    if not cabal:
+        fail("A .cabal file was not found in the srcs attribute.")
+
+    # Check that a Setup script exists. If not, create a default one.
+    setup = None
+    for f in ctx.files.srcs:
+        if f.basename in ["Setup.hs", "Setup.lhs"]:
+            if not setup or f.dirname < setup.dirname:
+                setup = f
```
Profpatsch: Same for the setup script, it should always be on the toplevel if it exists. Otherwise we might run into packages that have multiple setup files because they use the cabal project stuff.

haskell/cabal.bzl
```diff
@@ -0,0 +1,247 @@
...
+    )
+    cc_info = cc_common.merge_cc_infos(
+        cc_infos = [
+            CcInfo(
+                compilation_context = compilation_context,
+                linking_context = linking_context,
+            ),
+            cc_info,
+        ],
+    )
```
Profpatsch: The whole block constructing the `CCInfo` looks extremely boilerplatey, can we factor out the boring parts?

haskell/cabal.bzl
```diff
@@ -0,0 +1,247 @@
...
+A `haskell_cabal_library` can be substituted for any
+`haskell_library`. The two are interchangeable in most contexts.
+However, using a plain `haskell_library` sometimes leads to better
+build times, and does not require drafting a `.cabal` file.
+
+"""
+
+def _chop_version(name):
+    """Remove any version component from the given package name."""
+    return name.rpartition("-")[0]
```
Profpatsch: unused?

haskell/private/cabal_wrapper.sh.tpl
```diff
@@ -0,0 +1,68 @@
...
+libdir=$pkgroot/iface
+dynlibdir=$pkgroot/lib
+package_database=$pkgroot/package.conf.d
+
+%{ghc_pkg} recache --package-db=$package_database
+
+# Cabal really wants the current working directory to be directory
+# where the .cabal file is located. So we have no choice but to chance
+# cd into it, but then we have to rewrite all relative references into
+# absolute ones before doing so (using $execroot).
```
Profpatsch: We should open an upstream issue for that, that’s clearly bad behaviour. :(

haskell/private/cabal_wrapper.sh.tpl
```diff
@@ -0,0 +1,68 @@
...
+    --enable-relocatable \
+    --builddir=$distdir \
+    --prefix=$pkgroot \
+    --libdir=$libdir \
+    --dynlibdir=$dynlibdir \
+    --libsubdir= \
+    --package-db=clear \
+    --package-db=global \
+    "${@/=/=$execroot/}" \
+    --package-db=$package_database # This arg must come last.
```
Profpatsch: I think all of these should be quoted as well.

Profpatsch: I normally quote things nonetheless (because you cannot distinguish between bugs and known implicit guarantees and explicit quoting is the way to work around the biggest `sh` misfeature)

haskell/cabal.bzl
Profpatsch: And there’s no breakage because of cabal projects with multiple `Setup.hs`?

mboes: Cabal projects are not in source distributions. This deals with source distributions, i.e. single Cabal packages.

PR #880, opened 2019-05-20
Make ghc-paths work on Windows

ndmitchell:
Before building ghc-paths lead to the error:

```
bazel-out\x64_windows-fastbuild\bin\external\haskell_ghc__paths\GHC\Paths.hs:7:15: error:
    lexical error in string/character literal at character 'U'
  |
7 | libdir  = "C:\Users\ndmit_000\_bazel_Neil\cqtxsy6d\external\io_tweag_rules_haskell_ghc_windows_amd64\lib"
  |               ^
```

I fix that by using `sed` to replace `\` with `/`.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -39,6 +39,7 @@ jobs:
       /c/bazel/bazel.exe build --config windows "@haskell_unix__compat//..."
       /c/bazel/bazel.exe build --config windows "@haskell_hostname//..."
       /c/bazel/bazel.exe build --config windows "@haskell_zlib//..."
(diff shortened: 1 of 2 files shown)


Profpatsch:
Really sorry about the darwin stuff, CircleCI decided to not trust out-of-repo branches on Darwin a few days back and only @mboes has access to the necessary buttons afaik.

mboes:
Perhaps so, but I don't know what buttons to click. Unless you mean the "force the merge anyways button" (which I just pressed).

Profpatsch:
Ah, I can’t do that.

PR #886, opened 2019-05-22
Add arguments to haskell_register_ghc_nixpkgs

### LINKED ISSUE #883: haskell_register_ghc_nixpkgs missing arguments

`haskell_register_ghc_nixpkgs` does not exposes the following arguments from `haskell_toolchain`:

- `tools`
- `locale`

And does not exposes the following argument from the associated `nixpkgs_package`:

- `nix_file_content`.

I'd like theses arguments to be exposed so the user can set / override them.

Still using `haskell_toolchain`, but it does not work anymore.

guibou:
`haskell_register_ghc_nixpkgs` wraps both `haskell_toolchain` and
`nixpkgs_package` but was not forwarding all the arguments of both
rules.

- `haskell_toolchain`: `tools` and `locale`
- `nixpkgs_package`: `nix_file_content

This closes #883

### DIFF
--- a/haskell/nixpkgs.bzl
+++ b/haskell/nixpkgs.bzl
@@ -62,6 +62,7 @@ haskell_toolchain(
     # On Darwin we don't need a locale archive. It's a Linux-specific
     # hack in Nixpkgs.
     locale_archive = {locale_archive},
+    locale = {locale},
 )
         """.format(
             toolchain_libraries = toolchain_libraries,
@@ -72,6 +73,7 @@ haskell_toolchain(
             haddock_flags = repository_ctx.attr.haddock_flags,
             repl_ghci_args = repository_ctx.attr.repl_ghci_args,
             locale_archive = locale_archive,
+            locale = repr(repository_ctx.attr.locale),
         ),
     )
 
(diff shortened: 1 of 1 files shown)


guibou:
Note: this does not includes the `verbose` option which was added recently in `rules_nixpkgs` because `rules_haskell` uses an old version of `rules_nixpkgs`.

guibou:
> Do you need this because you want to override haddock?

Yes.

mboes:
If you want a hacked toolchain, then you should provide the hacked toolchain wholesale to the `haskell_register_ghc_nixpkgs()` macro. In fact the DAML repo (cc @aherrmann) does that as well. What I mean is, write a Nix file that exposes a GHC nix derivation that includes the Haddock you want (the Nixpkgs `symlinkJoin` or `buildEnv` functions might come in handy).

aherrmann:
For reference [here is the DAML repos' call of `haskell_register_ghc_nixpkgs`](https://github.com/digital-asset/daml/blob/b09cbd037bb2f5440330d0a33981926e205ca4a1/WORKSPACE#L239-L267).

guibou:
@mboes comment adressed. `tools` was removed.

haskell/nixpkgs.bzl
```diff
@@ -297,6 +298,7 @@ haskell_toolchain(
             haddock_flags = repository_ctx.attr.haddock_flags,
             repl_ghci_args = repository_ctx.attr.repl_ghci_args,
             locale_archive = locale_archive,
+            locale = repr(repository_ctx.attr.locale) or [],
```
mboes: This looks suspicious. Isn't `locale` an `attr.string()`? In fact, if it has a default, why do you need the `or` clause?

mboes:
```suggestion
            locale = repr(repository_ctx.attr.locale),
```

guibou: Good point!

PR #887, opened 2019-05-22
Add stack_install

### LINKED ISSUE #874: Direct Cabal support

mboes:
With `haskell_cabal_library` (added in #882), we have the ability to make
one-shot calls to Cabal, assuming,

1. that the sdist for the Cabal package has been fetched and unpacked, and 
2. that dependencies have been declared correctly.

Declaring these dependencies by hand for all packages on Hackage is a
fool's errand. This commit uses Stack to,

1. resolve package names to package versions using the given snapshot,
2. fetch the sdist and unpack it,
3. find out the dependency graph and generate a `BUILD` file encoding it.

Stack only outputs dependency information in GraphViz format. So have to
parse that.

This assumes Stack is in the `PATH` (hence build is non-hermetic). But it's
pretty unreasonable to build our own Stack inside a workspace rule (Bazel's
build engine doesn't work there). The best we can do is download prebuilt
binaries, which we can do in the future (after figuring out how to
seamlessly work on NixOS).

BTW @snoyberg: I noticed the upcoming `--global-hints` feature of
Stack is marked as "experimental" and "could be removed". Just FTR
that will be a very useful feature for us (we'll be able to generate
a dependency graph without first downloading a GHC toolchain).

@regnat this PR removes the "new-style" `haskell_import` rule. We now
no longer need that rule in the tests and I believe no one else is
using it either.

Closes #874.


mboes:
We can remove dead code separately (and later if downstream depends on this experimental feature). @regnat I removed 2984010732dae38ce456e8c70ccc7967f8cc8089 from this PR and will push it in a followup PR.

haskell/cabal.bzl
```diff
@@ -243,3 +243,231 @@ However, using a plain `haskell_library` sometimes leads to better
...
+        fail("Cannot find stack command in your PATH.")
+    stack = [stack_cmd, "--no-nix", "--skip-ghc-check", "--system-ghc"]
+    if versioned_packages:
+        _execute_or_fail_loudly(repository_ctx, stack + ["unpack"] + versioned_packages)
+    stack = [stack_cmd, "--no-nix", "--skip-ghc-check", "--system-ghc", "--resolver", repository_ctx.attr.snapshot]
+    if unversioned_packages:
+        _execute_or_fail_loudly(repository_ctx, stack + ["unpack"] + unversioned_packages)
+    exec_result = _execute_or_fail_loudly(repository_ctx, ["ls"])
+    unpacked_sdists = exec_result.stdout.splitlines()
+    stack_yaml_content = "\n".join(["resolver: none", "packages:"]) + "\n" + "\n".join([
```
regnat: Wouldn't it be cleaner and easier to read to use [to_json](https://docs.bazel.build/versions/master/skylark/lib/struct.html#to_json) rather than writing the yaml file by hand

mboes: Great idea. Will try that.

tests/haddock/LibB.hs
```diff
@@ -5,17 +5,12 @@ module LibB where
...
 import LibA.A (a)
 import LibA (f)
-import LibC (mytype, LibCType)
 import TH (foo)
 
 -- | Doc for 'x' using 'f' and 'a' and 'Int'.
 x :: Int
 x = const f a
 
--- | This uses a type from an undocumented package
```
regnat: Don't we remove a potentially useful test here?

I assume that's because we can't easily extend the stackage snapshot whith some lib which isn't on hackage but maybe we could use an existing library with no documentation instead

WORKSPACE
```diff
@@ -33,6 +33,32 @@ load("@os_info//:os_info.bzl", "is_linux", "is_windows")
...
+        "hspec",
+        "hspec-core",
+        "lens-family-core",
+        "data-default-class",
+        "lens-labels",
+        "proto-lens",
+        "lens-family",
+    ],
+    snapshot = "lts-13.15",
+    deps = ["@zlib.dev//:zlib"],
```
aherrmann: How do we know which package to forward these to? Or are they generally visible to all packages? In the latter case that means that adding a `deps` entry here would rebuild all imported stackage packages, right?

WORKSPACE
```diff
@@ -33,6 +33,32 @@ load("@os_info//:os_info.bzl", "is_linux", "is_windows")
 # bazel dependencies
 haskell_repositories()
 
+load("@io_tweag_rules_haskell//haskell:cabal.bzl", "stack_install")
+
+stack_install(
+    name = "stackage",
+    packages = [
```
aherrmann: Is there any way to override a package with a Bazel target? E.g. inject a Bazel built version of, say, `streaming`.

WORKSPACE
```diff
@@ -159,12 +170,29 @@ cc_library(
...
     srcs = [":lib"],
-    testonly = 1,
 )
 """,
     repository = "@nixpkgs",
 )
 
+nixpkgs_package(
+    name = "c2hs",
+    attribute_path = "haskellPackages.c2hs",
```
aherrmann: Why pull these from nixpkgs rather than stackage?

haskell/cabal.bzl
```diff
@@ -243,3 +243,225 @@ However, using a plain `haskell_library` sometimes leads to better
...
+    "process",
+    "rts",
+    "stm",
+    "template-haskell",
+    "terminfo",
+    "text",
+    "time",
+    "transformers",
+    "unix",
+    "xhtml",
```
aherrmann: Interestingly `hazel/packages.bzl` does not list `xhtml` as a core-package. Did this change between stackage releases?

haskell/cabal.bzl
```diff
@@ -243,3 +243,225 @@ However, using a plain `haskell_library` sometimes leads to better
...
+
+    # Unpack all given packages, then compute the transitive closure
+    # and unpack anything in the transitive closure as well.
+    stack_cmd = repository_ctx.which("stack")
+    if not stack_cmd:
+        fail("Cannot find stack command in your PATH.")
+    stack = [stack_cmd, "--no-nix", "--skip-ghc-check", "--system-ghc"]
+    if versioned_packages:
+        _execute_or_fail_loudly(repository_ctx, stack + ["unpack"] + versioned_packages)
+    stack = [stack_cmd, "--no-nix", "--skip-ghc-check", "--system-ghc", "--resolver", repository_ctx.attr.snapshot]
```
aherrmann: As this is a repository rule this just picks whatever `ghc` is in `$PATH`, correct? Does this have any impact on the generated dependency graph? E.g. conditional dependencies based on the `ghc` version, or similar? If so then this seems like another iteration of https://github.com/FormationAI/hazel/issues/80.

haskell/cabal.bzl
```diff
@@ -243,3 +243,225 @@ However, using a plain `haskell_library` sometimes leads to better
...
+    exec_result = _execute_or_fail_loudly(
+        repository_ctx,
+        stack + ["dot", "--external"],
+    )
+    dependencies = {k: [] for k in all_packages}
+    for line in exec_result.stdout.splitlines():
+        tokens = [w.strip('";') for w in line.split(" ")]
+
+        # All lines of the form `"foo" -> "bar";` declare edges of the
+        # dependency graph in the Graphviz format.
```
aherrmann: It would be great if stack could spit this out in a more robust and easier parseable format.

haskell/cabal.bzl
```diff
@@ -243,3 +243,225 @@ However, using a plain `haskell_library` sometimes leads to better
...
+    "text",
+    "time",
+    "transformers",
+    "unix",
+    "xhtml",
+]
+
+def _compute_dependency_graph(repository_ctx, versioned_packages, unversioned_packages):
+    """Given a list of root packages, compute a dependency graph.
+
```
aherrmann: As I understand it this does not just return the dependency graph of the transitive closure of the requested packages, but of all of Stackage, correct? That's probably fine, but not very clear from the docstring.

shell.nix
```diff
@@ -16,6 +16,9 @@ mkShell {
     perl
     python3
     bazel
+    # For stack_install. TODO Remove ghc when move to Stack 2.
```
aherrmann: What does Stack 2 improve in this respect?

haskell/cabal.bzl
```diff
@@ -243,3 +243,225 @@ However, using a plain `haskell_library` sometimes leads to better
...
+            """alias(name = "{name}", actual = ":{actual}", visibility = {visibility})""".format(
+                name = unversioned_package,
+                actual = package,
+                visibility = visibility,
+            ),
+        )
+    build_file_content = "\n".join(build_file_builder)
+    repository_ctx.file("BUILD.bazel", build_file_content, executable = False)
+
+stack_install = repository_rule(
```
aherrmann: Comparing to Hazel, here are a few things that come to mind which `stack_install` and `haskell_cabal_library` don't seem to offer, yet. Those might be good future extensions.
- Override individual packages with custom Bazel builds. Probably easiest by injecting a mapping from package name to Bazel targets.
- Stack's `extra-deps` feature to override a package version or just it's source repository. Also to add additional Hackage packages, or packages from other sources.
- Apply source patches. This is very useful if a package just requires minor adjustment to build within Bazel. E.g. file paths to embedded data files. In such cases a full fork imposes a lot of overhead.
- Overriding cabal flags. E.g. `integer-simpl` vs. `integer-gmp`.
- Binary targets. As I understand `haskell_cabal_library` does currently not expose any binary targets. I suspect this has two consequences. A) Users cannot use e.g. a Stackage provided `c2hs`. B) Stackage packages that depend on other Stackage packages' binaries, e.g. `happy`, will probably not work.

WORKSPACE
aherrmann: I can answer this myself now. Yes, they are forward to all packages.

WORKSPACE
mboes: What you can do is override the version found in the snapshot. So in this sense `packages` is a merge between the `packages` and `extra-deps` field in the `stack.yaml`. This is implied in the rule documentation but the documentation could be more explicit.

WORKSPACE
mboes: We can,

* ask the user to specify exactly which package we should expose the system lib to,
* blanket forward the dependency to all packages.

The latter means more rebuilds as you say. I like minimal incremental rebuilds. But OTOH with the latter option we no longer have a conventional `deps` attribute with a conventional type (label list). Moreover, as Hazel taught me, Cabal seems to expose a lot of C stuff to all transitive downstream dependencies. So if we change the set of system libs, we change the inputs for most of the dependency graph and therefore have to rebuild most of it anyways. It's perhaps why in both Stack and Cabal, this is a global option that affects the inputs to all packages.

WORKSPACE
```diff
@@ -52,29 +78,10 @@ load(
...
-    "haskell_package_repository_dummy",
-)
-load(
-    "@io_tweag_rules_haskell//:constants.bzl",
-    "test_ghc_version",
-)
 
-haskell_nixpkgs_package(
+nixpkgs_package(
     name = "ghc",
```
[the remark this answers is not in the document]
mboes: This is hard to achive in practice: Nixpkgs is pretty aggressive about prunning old minor releases when a new one comes out. Either way we're still just as reproducible.

WORKSPACE
mboes: Because we don't yet have `haskell_cabal_binary`.

haskell/cabal.bzl
mboes: xhtml has been a core library for a long time. Probably an omission in Hazel. Or maybe an omission in Stackage.

haskell/cabal.bzl
mboes: No not all of Stackage. It computes only the transitive closure.

haskell/cabal.bzl
mboes: The nice thing about using Stackage is that in principle they deal with this for us, provided we stick to using the same GHC version as Stackage does for any given snapshot.

haskell/cabal.bzl
mboes: Agreed. I'll file a ticket upstream.

haskell/cabal.bzl
mboes: > Override individual packages with custom Bazel builds.

Not currently possible, but it's possible to override the version at least.

> Stack's extra-deps feature to override a package version or just it's source repository.

See above. Not possible to override with one's own source repository. Good future feature.

> Also to add additional Hackage packages, or packages from other sources.

Outside of the cases listed above, these wouldn't be part of the transitive closure, right? So don't need to be part of the `stackage` external repository?

> Apply source patches ... Overriding cabal flags ... Binary targets

Agreed. Let's create tickets about those.

shell.nix
mboes: It generates the dependency graph without the spurious requirement that GHC be in the environment.

tests/haddock/LibB.hs
mboes: We are. But the value of the test needs to be measured against the cost of migrating it. It's quite difficult.

> maybe we could use an existing library with no documentation instead

Good idea. If you can find one then we could revive this test.

mboes:
@regnat 

> With this approach, we still have to explicitely list all our dependencies in the WORKSPACE file (though I assume any attempt at lifting this would result in a need of a "load a generated file" thing in the workspace which is always painful)

Yes, either what you mention (à la Hazel), or we could just download all packages in the snapshot. In fact I'm pondering making it so that if no `packages` are specified, or if using some boolean flag, then all of Stackage is downloaded, and all of it becomes available. For a small project that's a comparatively big download. But for e.g. DAML, which is a large repo with lots of dependencies anyways, downloading all of Stackage isn't all that crazy.

> The external dependencies are shared amongst all the packages, which means that adding a new one will require rebuilding the whole dependency set

That's true. See discussion with @aherrmann above. This seems inevitable to me given a design flaw in Cabal. But we could consider trying to do better. I'm just not sure it'll pan out all that well.

haskell/cabal.bzl
symbiont-joseph-kachmar: > Stack's extra-deps feature [...]

I haven't tested any of these new features yet, however in practice one can make a custom snapshot file, host it remotely, and point `stack` to it (e.g. `stack build --resolver=<link_to_snapshot.>`).

Depending on how `rules_haskell` and `stack` interact here, this could be a neat workaround for `extra-deps` at the moment.

mboes: Hm, so you're saying #910 could be unnecessary? It hasn't been merged yet. Not too late to block it. :)

jkachmar: Unfortunately I can't say _for sure_, since I've been unable to get the quickstart instructions for `rules_haskell` working on macOS; I'm basing this purely off of my knowledge of how `stack` treats resolvers.

I do think that having something like #910 would probably be useful, regardless of this, though. I just wanted to suggest it just in case it actually did work, and could help unblock someone in the meantime 😄

Profpatsch: @jkachmar You can use `rules_haskell` master for now, on MacOS there was a minor bugfix, see https://github.com/tweag/rules_haskell/pull/893

I’m gonna push a point-release later today.

mboes: There's a singular point missing in our CI matrix currently, which I guess is the point you're in: macOS + ghc bindists (as opposed to GHC-from-Nixpkgs).

Profpatsch: https://github.com/tweag/rules_haskell/releases/tag/v0.9.1

Issue #438, opened 2018-09-10
Profiling is unecessary slow


guibou:
*This is an updated description which summarize the discussion*

- Building with `-c dbg` is really slow when template haskell is used because we use `-fexternal-interpreter` and this GHC bug: https://ghc.haskell.org/trac/ghc/ticket/16256#ticket . In summary, profiling build is 15x times slower than normal build.
- There is two solutions to build haskell with profiling if template haskell is involved. `-fexternal-interpreter`, which is now used in rules_haskell, and another solution which build two times, once without profiling (so template haskell can execute non profiled code) and one with profiling.
- rules_haskell was initially using the two build solution, but the switch to `-fexternal-interpreter` was accidentally done in be5cf67#diff-33a8c46955e2a9917090b5258346aee2R205. Since #892 the dead code of the two build solution was removed.

We have two solution here:

- Keep the approach with `-fexternal-interpreter` and wait for an upstream fix for GHC
- Get back to the approach without `-fexternal-interpreter` and reimplement the two steps build.

guibou:
More informations:

- the `haskell_library` have `deps` with `cc_library`.
- explicitly adding `-lanHaskellLibrary` to the `compiler_flags` of the `haskell_library` fix the build.

guibou:
Related to #439?

guibou:
I'm still not able to generate a small reproducible example, but I now have a lot of question related to the implementation of `-prof` in rules_haskell.

So, rules_haskell fails building some targets in the following context:

- `TemplateHaskell` is activated
- More than one haskell file in the `src` package
- Dependencies to other haskell libraries which have cbit

When experimenting, I found that removing the `-fexternal-interpreter` flag from https://github.com/tweag/rules_haskell/blob/f7b355a9c5da4e0b81fc60bc963cd78f0dc4fe0b/haskell/private/actions/compile.bzl#L202 leads to progress, i.e. targets which cannot be built previously can now be built.

I've tried the same change in rules_haskell codebase and the result is surprising. `bazel test -c dbg //...` fails all tests, but `bazel test -c dbg //a_specific_test` works for all tests. I'll try to understand that later.

There are two ways to build with profiling using ghc and TemplateHaskell, `-fexternal-interpreter` and a two phases build where you first build without `-prof` and then with. This is detailed in https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#using-template-haskell-with-profiling

[...]

PR #892, opened 2019-05-23
Drop non-profiling inputs to profiling build

aherrmann:
The profiling builds took the non-profiling object file outputs as an additional input with a comment stating that these are required for template Haskell. However, it seems these inputs are unnecessary as removing them does not seem to negatively affect `bazel test //tests/... -c dbg`. As I understand, this is because we are [passing `-fexternal-interpreter`](https://github.com/tweag/rules_haskell/blob/a7a054c67016b0e3ee867abdfd5a1fa749f805b1/haskell/private/actions/compile.bzl#L260) on profiling builds. Removing that flag would indeed cause errors due to missing object files.

### DIFF
--- a/haskell/private/haskell_impl.bzl
+++ b/haskell/private/haskell_impl.bzl
@@ -161,13 +161,7 @@ def _haskell_binary_common_impl(ctx, is_test):
             srcs = srcs_files,
             ls_modules = ctx.executable._ls_modules,
             import_dir_map = import_dir_map,
-            # NOTE We must make the object files compiled without profiling
-            # available to this step for TH to work, presumably because GHC is
-            # linked against RTS without profiling.
-            extra_srcs = depset(transitive = [
-                depset(ctx.files.extra_srcs),
-                depset([c.objects_dir]),
-            ]),
+            extra_srcs = depset(ctx.files.extra_srcs),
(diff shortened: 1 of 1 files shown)


guibou:
I fully agree with theses changes, however have a look at #438 

There is two options for profiling:

- building twice (with and without profiling)
- building with `-fexternal-interpreter`

The current situation is that, starting from the first solution, the flag for the second solution was introduced in a totally unrelated commit, see #438 for all the details, and progressively the code evolved such that the first solution was not working anymore and that's what your are removing now (which is a good thing).

But considering that `-fexternal-interpreter` is ~ 15x times slower than the other approach when it involves template haskell, see https://gitlab.haskell.org/ghc/ghc/issues/16256 I have concerns.

That's a tradeoff between simplifying rules_haskell by fully switching to `-fexternal-interpreter` or removing the flag and fixing the previous approach. Pro and cons of approach:

- Fixing the first will be more work than the second
- The first generates more code in rules_haskell
- The second is slower
- The second will be faster if the bug if fixed upstream.

aherrmann:
@guibou Thanks for the pointer. It's a tricky tradeoff. As you point out this PR doesn't change the current state of affairs. 

> This can be extended by removing the objets generation code at all.

I'm working on this for a separate PR. I'll continue the discussion there.

Issue #884, opened 2019-05-22
Can not build projects on MacOS Mojave


alonsodomin:
**Describe the bug**
When trying to build a project using these rules, Bazel will attempt to download the necessary archives and packages to end up failing saying that it can't find the `@io_tweag_rules_haskell_ghc_darwin_amd64//` package.

Here is the standard out:

```bash
sed: 1: "mk/config.mk.in": invalid command code m
INFO: An error occurred during the fetch of repository 'io_tweag_rules_haskell_ghc_darwin_amd64'
INFO: Call stack for the definition of repository 'io_tweag_rules_haskell_ghc_darwin_amd64':
 - /private/var/tmp/_bazel_domingueza/cf40c2340e1460b987f62a06aab24e21/external/io_tweag_rules_haskell/haskell/ghc_bindist.bzl:364:5
 - /private/var/tmp/_bazel_domingueza/cf40c2340e1460b987f62a06aab24e21/external/io_tweag_rules_haskell/haskell/ghc_bindist.bzl:400:9
 - /Users/domingueza/Workspace/Projects/herd/WORKSPACE:35:1
INFO: Repository 'io_tweag_rules_haskell_ghc_darwin_amd64' used the following cache hits instead of downloading the corresponding file.
 * Hash 'cccb58f142fe41b601d73690809f6089f7715b6a50a09aa3d0104176ab4db09e' for https://downloads.haskell.org/~ghc/8.6.4/ghc-8.6.4-x86_64-apple-darwin.tar.xz
If the definition of 'io_tweag_rules_haskell_ghc_darwin_amd64' was updated, verify that the hashes were also updated.
ERROR: /private/var/tmp/_bazel_domingueza/cf40c2340e1460b987f62a06aab24e21/external/io_tweag_rules_haskell_ghc_darwin_amd64-toolchain/BUILD:4:1: no such package '@io_tweag_rules_haskell_ghc_darwin_amd64//': Traceback (most recent call last):
	File "/private/var/tmp/_bazel_domingueza/cf40c2340e1460b987f62a06aab24e21/external/io_tweag_rules_haskell/haskell/ghc_bindist.bzl", line 215
		_execute_fail_loudly(ctx, ["sed", "-i", "s/RelocatableB..."])
	File "/private/var/tmp/_bazel_domingueza/cf40c2340e1460b987f62a06aab24e21/external/io_tweag_rules_haskell/haskell/ghc_bindist.bzl", line 177, in _execute_fail_loudly
[...]

PR #893, opened 2019-05-23
Fix darwin bindist by touching WORKSPACE file

Profpatsch:
Closes: https://github.com/tweag/rules_haskell/issues/884

### DIFF
--- a/haskell/ghc_bindist.bzl
+++ b/haskell/ghc_bindist.bzl
@@ -266,6 +266,7 @@ haskell_toolchain(
         },
         executable = False,
     )
+    ctx.file("WORKSPACE")
 
 _ghc_bindist = repository_rule(
     _ghc_bindist_impl,


mboes:
How does this work?

Profpatsch:
> How does this work?

Some rule tried to access `io_tweag_rules_haskel_ghc_darwin_amd64` as a bazel repository, but without a `WORKSPACE` file things are not counted as repository. If we `touch` it, it’s recognized as a repository. I am not 100% certain it *should* be counted as a repository, so please stop me if that’s wrong.

mboes:
This looks suspicious to me. Are you sure it works with this patch but fails without? Also, Bazel tends to automatically generate a `WORKSPACE` file for external repositories.

Profpatsch:
I’ll try what happens if I remove this patch (once I can access the test Macbook again)

mboes:
Easiest is to ssh into circleci workers

PR #896, opened 2019-05-24
Exclusive profiling mode

### LINKED ISSUE #895: CcInfo does not take profiling mode into account
**Describe the bug**
The `CcInfo` provider [returned by `haskell_library`](https://github.com/tweag/rules_haskell/blob/ad7e2af7d763b4dc021bc03eaa4f4070845dc976/haskell/private/haskell_impl.bzl#L545) does not take the profiling mode into account. Similarly, [`haskell_toolchain_library`](https://github.com/tweag/rules_haskell/blob/ad7e2af7d763b4dc021bc03eaa4f4070845dc976/haskell/private/haskell_impl.bzl#L642) ignores the profiling mode artifacts of toolchain libraries. (Relatedly, it also ignores the `_thr`, `_debug` versions of the `rts`).

**Expected behavior**
If the `CcInfo` provider is to faithfully describe the compilation results of a Haskell target, then it should reflect profiling mode. I.e. in regular compilation mode `CcInfo` would carry the regular shared and static libraries. In profiling mode it would only carry the static profiling library, as GHC does not produce dynamic libraries in profiling mode.

**Additional context**
Ideally, `HaskellInfo` would not need to carry any `.so`/`.a` files itself, instead those would all be carried by the `CcInfo` provider. However, this may not be fully achievable.

Notably, dropping all non-profiling libraries in profiling mode seems to enforce compilation with `-fexternal-interpreter`, which imposes a significant performance penalty. See https://github.com/tweag/rules_haskell/issues/438.

Also, the static archives of GHC's core packages shipped with nixpkgs or the GHC bindist on Unix are compiled without `-fPIC`. Which may be an issue with GHCi/Template Haskell.

Nonetheless, it seems reasonable to let `CcInfo` handle the *regular* compilation path. I.e. not worrying about GHCi/Template Haskell. Where necessary, the `HaskellInfo` provider (or another dedicated provider if deemed appropriate) could carry any libraries/objects required for GHCi/Template Haskell.

aherrmann:
Closes #895 

- When profiling mode is active (`compilation_mode == "dbg"`) then we compile only profiling libraries and binaries instead of compiling both profiling and non-profiling. In profiling mode the toolchain libraries only carry their static archives, as profiling mode doesn't provide any shared libraries.
- Some test cases strictly require shared objects. Those are disabled in profiling mode.

Wrt. https://github.com/tweag/rules_haskell/issues/438 this PR goes further down the road of `-fexternal-interpreter`. Given the large performance penalty that that implies we may yet want to compile non-profiling objects for GHCi/Template Haskell. However, as mentioned in #895 it seems better to not expose these in the `CcInfo` provider, but instead handle them separately in `HaskellInfo`.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -155,16 +155,13 @@ def _haskell_cabal_library_impl(ctx):
         "_install/iface",
         sibling = cabal,
     )
+    static_library_filename = "_install/lib/libHS{}.a".format(name)
+    if with_profiling:
+        static_library_filename = "_install/lib/libHS{}_p.a".format(name)
     static_library = hs.actions.declare_file(
-        "_install/lib/libHS{}.a".format(name),
+        static_library_filename,
         sibling = cabal,
     )
-    static_library_prof = None
-    if with_profiling:
-        static_library_prof = hs.actions.declare_file(
-            "_install/lib/libHS{}_p.a".format(name),
-            sibling = cabal,
-        )
     dynamic_library = hs.actions.declare_file(
         "_install/lib/libHS{}-ghc{}.{}".format(name, hs.toolchain.version, _so_extension(hs)),
         sibling = cabal,
@@ -189,7 +186,7 @@ def _haskell_cabal_library_impl(ctx):
             interfaces_dir,
             static_library,
             dynamic_library,
-        ] + ([static_library_prof] if with_profiling else []),
+        ],
         env = c.env,
         mnemonic = "HaskellCabalLibrary",
         progress_message = "HaskellCabalLibrary {}".format(hs.label),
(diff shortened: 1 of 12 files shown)


aherrmann:
CI is green now. The handling of `//debug/linking_utils:test-ldd` on Darwin in `bazel test prof` is not very satisfactory. In `.circleci/config.yml` we're setting `echo "common:ci --test_tag_filters -dont_test_on_darwin" >> .bazelrc.local`. However, it seems that these filters don't compose and passing an additional `--test_tag_filters` argument within `RunTests.hs` overrides the original filter. So, https://github.com/tweag/rules_haskell/pull/896/commits/798ec961506b96c1af883f877f6a650bf6cd6ee7 simply replicates the original filter within `RunTests.hs` on Darwin.

PR #899, opened 2019-05-26
Remove special handling of prebuilt libraries

### LINKED ISSUE #440: Get rid of `HaskellPrebuiltPackageInfo`

mboes:
Now that prebuilt libraries look exactly like any other Haskell
libraries except for the fact that we have no source code for them, we
can get rid of `HaskellPrebuiltPackageInfo`. This in turn simplifies
a lot of internals:

* we no longer need the `prebuilt_deps` and `direct_prebuilt`
  attributes to `HaskellInfo` anymore.
* the Haddock wrapper is vastly simplified.

Closes #440

### DIFF
--- a/haskell/lint.bzl
+++ b/haskell/lint.bzl
@@ -5,7 +5,6 @@ load(
     "HaskellInfo",
     "HaskellLibraryInfo",
     "HaskellLintInfo",
-    "HaskellPrebuiltPackageInfo",
 )
 load(":private/context.bzl", "haskell_context", "render_env")
 load(":private/packages.bzl", "expose_packages", "pkg_info_to_compile_flags")
@@ -31,7 +30,7 @@ def _haskell_lint_rule_impl(ctx):
 def _haskell_lint_aspect_impl(target, ctx):
     hs = haskell_context(ctx, ctx.rule.attr)
 
-    if HaskellInfo not in target or HaskellPrebuiltPackageInfo in target:
+    if HaskellInfo not in target:
         return []
 
     hs_info = target[HaskellInfo]
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -166,8 +166,6 @@ main = defaultMain
         dynamic_libraries = set.insert(dep_info.dynamic_libraries, dynamic_library),
         interface_dirs = set.insert(dep_info.interface_dirs, interfaces_dir),
         compile_flags = [],
-        prebuilt_dependencies = set.empty(),
-        direct_prebuilt_deps = set.empty(),
         cc_dependencies = dep_info.cc_dependencies,
         transitive_cc_dependencies = dep_info.transitive_cc_dependencies,
     )
--- a/haskell/doctest.bzl
+++ b/haskell/doctest.bzl
@@ -13,7 +13,6 @@ load(
     "@io_tweag_rules_haskell//haskell:providers.bzl",
     "HaskellInfo",
     "HaskellLibraryInfo",
-    "HaskellPrebuiltPackageInfo",
 )
 
 def _doctest_toolchain_impl(ctx):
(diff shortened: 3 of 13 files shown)


haskell/lint.bzl
```diff
@@ -31,7 +30,7 @@ def _haskell_lint_rule_impl(ctx):
 def _haskell_lint_aspect_impl(target, ctx):
     hs = haskell_context(ctx, ctx.rule.attr)
 
-    if HaskellInfo not in target or HaskellPrebuiltPackageInfo in target:
```
regnat: What will happen for the prebuilt dependencies here? Are they still ignored by the lint?

PR #900, opened 2019-05-26
Remove overuse of sets

mboes:
Sets were introduced earlier on in the development of rules_haskell,
for reasons I do not remember (I believe @fuuzetsu made the decision).
In the end however, using sets means more copying during the analysis
phase than would otherwise happen with depsets, which is a data
structure that supports O(1) union. What's more, we ended up having to
use conversion functions in many places, be it `set.to_depset()` or
`set.to_list()`. We now use far fewer conversion functions, making the
code easier to read, and hopefully helping with performance too.


mboes:
Depends on #899.

Fuuzetsu:
Hmm, I'm sure there was some reason why we didn't use depset at the time but I can't remember... Maybe because iterations were expensive or something... If the current code works then great.

haskell/private/dependencies.bzl
```diff
@@ -140,32 +140,55 @@ def gather_dep_info(ctx, deps):
...
-            package_ids = acc.package_ids
             if HaskellLibraryInfo not in dep:
                 fail("Target {0} cannot depend on binary".format(ctx.attr.name))
-            if HaskellLibraryInfo in dep:
-                set.mutable_insert(package_ids, dep[HaskellLibraryInfo].package_id)
             acc = HaskellInfo(
-                package_ids = package_ids,
-                package_databases = set.mutable_union(acc.package_databases, binfo.package_databases),
+                package_ids = acc.package_ids,
+                package_databases = acc.package_databases,
```
guibou: Why no union between `acc.package_databases` and `binfo.package_databases`?

mboes: Because we already computed the set of all package databases a few lines earlier. Look at the other hunks as well.

PR #904, opened 2019-05-27
Add haskell_cabal_binary

mboes:
This rule reuses much of the code from `haskell_cabal_library`. It
builds binaries from a Cabal file. It exposes the same providers as
`haskell_binary`.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -289,6 +289,105 @@ build times, and does not require drafting a `.cabal` file.
 
 """
 
+def _haskell_cabal_binary_impl(ctx):
+    hs = haskell_context(ctx)
+    dep_info = gather_dep_info(ctx, ctx.attr.deps)
+    cc = cc_interop_info(ctx)
+    cc_info = cc_common.merge_cc_infos(
+        cc_infos = [dep[CcInfo] for dep in ctx.attr.deps if CcInfo in dep],
+    )
+
+    cabal = _find_cabal(hs, ctx.files.srcs)
+    setup = _find_setup(hs, cabal, ctx.files.srcs)
+    package_database = hs.actions.declare_file(
+        "_install/package.conf.d/package.cache",
+        sibling = cabal,
+    )
+    binary = hs.actions.declare_file(
+        "_install/bin/{}".format(hs.label.name),
+        sibling = cabal,
+    )
+    c = _prepare_cabal_inputs(
+        hs,
+        cc,
+        dep_info,
+        cc_info,
+        cabal = cabal,
+        setup = setup,
+        srcs = ctx.files.srcs,
+        cabal_wrapper_tpl = ctx.file._cabal_wrapper_tpl,
+        package_database = package_database,
+    )
+    ctx.actions.run(
+        executable = c.cabal_wrapper,
+        arguments = [c.args],
+        inputs = c.inputs,
+        outputs = [
+            package_database,
+            binary,
+        ],
+        env = c.env,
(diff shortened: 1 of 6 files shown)


haskell/cabal.bzl
```diff
@@ -289,6 +289,108 @@ build times, and does not require drafting a `.cabal` file.
...
+    default_info = DefaultInfo(files = depset([binary]), executable = binary)
+
+    return [hs_info, cc_info, default_info]
+
+haskell_cabal_binary = rule(
+    _haskell_cabal_binary_impl,
+    executable = True,
+    attrs = {
+        "srcs": attr.label_list(allow_files = True),
+        "deps": attr.label_list(),
```
aherrmann: If a Cabal package has a library and an executable and the executable depends on the library. Should a user add the corresponding `haskell_cabal_library` to `deps`?

haskell/cabal.bzl
```diff
@@ -289,6 +289,108 @@ build times, and does not require drafting a `.cabal` file.
...
+Example:
+  ```bzl
+  haskell_cabal_binary(
+      name = "happy",
+      srcs = glob(["**"]),
+  )
+  ```
+
+This rule assumes that the .cabal file defines a single executable
+with the same name as the package.
```
aherrmann: Can that restriction be lifted in the future? E.g. allow multi executable Cabal packages.

haskell/cabal.bzl
```diff
@@ -289,6 +289,108 @@ build times, and does not require drafting a `.cabal` file.
...
+        interface_dirs = dep_info.interface_dirs,
+        compile_flags = [],
+        prebuilt_dependencies = dep_info.prebuilt_dependencies,
+        cc_dependencies = dep_info.cc_dependencies,
+        transitive_cc_dependencies = dep_info.transitive_cc_dependencies,
+    )
+    cc_info = cc_common.merge_cc_infos(
+        cc_infos = [dep[CcInfo] for dep in ctx.attr.deps if CcInfo in dep],
+    )
+    default_info = DefaultInfo(files = depset([binary]), executable = binary)
```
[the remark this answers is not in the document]
mboes: Yes. That's part of the next PR, which will fix that for libraries also.

haskell/cabal.bzl
mboes: Not necessary.

haskell/cabal.bzl
mboes: It very well could, by asking the user to name the binaries. Otherwise we'd have to parse the .cabal files to find out the binary names at workspace rule evaluation time. Good future task.

haskell/cabal.bzl
```diff
@@ -289,6 +289,108 @@ build times, and does not require drafting a `.cabal` file.
...
+        static_libraries = dep_info.static_libraries,
+        static_libraries_prof = dep_info.static_libraries_prof,
+        dynamic_libraries = dep_info.dynamic_libraries,
+        interface_dirs = dep_info.interface_dirs,
+        compile_flags = [],
+        prebuilt_dependencies = dep_info.prebuilt_dependencies,
+        cc_dependencies = dep_info.cc_dependencies,
+        transitive_cc_dependencies = dep_info.transitive_cc_dependencies,
+    )
+    cc_info = cc_common.merge_cc_infos(
```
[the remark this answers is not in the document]
mboes: True. It was a copy/paste from the `haskell_cabal_library` rule, but it can be removed entirely in the binary case. I will do so before merging.

PR #907, opened 2019-05-28
Add a tools argument to Stack and Cabal rules

mboes:
Tools are built using the host configuration. Currently, they are not
very useful unless they are tools part of a whitelist defined by
Cabal. For example, Cabal lets you customize which happy or c2hs you
want to use. We support that.

Packages that correspond to the whitelisted tools are detected as such
and filtered from the dependency graph in `stackage_snapshot`. That
way only packages that produce libraries remain.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -33,6 +33,17 @@ load("@os_info//:os_info.bzl", "is_linux", "is_windows")
 # bazel dependencies
 haskell_repositories()
 
+http_archive(
+    name = "happy",
+    build_file_content = """
+load("@io_tweag_rules_haskell//haskell:cabal.bzl", "haskell_cabal_binary")
+haskell_cabal_binary(name = "happy", srcs = glob(["**"]), visibility = ["//visibility:public"])
+    """,
+    sha256 = "22eb606c97105b396e1c7dc27e120ca02025a87f3e44d2ea52be6a653a52caed",
+    strip_prefix = "happy-1.19.10",
+    urls = ["http://hackage.haskell.org/package/happy-1.19.10/happy-1.19.10.tar.gz"],
+)
+
 load("@io_tweag_rules_haskell//haskell:cabal.bzl", "stack_snapshot")
 
 stack_snapshot(
@@ -56,6 +67,7 @@ stack_snapshot(
         "lens-family",
     ],
     snapshot = "lts-13.15",
+    tools = ["@happy"],
     deps = ["@zlib.dev//:zlib"],
 )
 
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -72,7 +72,14 @@ main = defaultMain
         )
     return setup
 
-def _prepare_cabal_inputs(hs, cc, dep_info, cc_info, cabal, setup, srcs, cabal_wrapper_tpl, package_database):
+_CABAL_TOOLS = ["alex", "c2hs", "cpphs", "doctest", "happy"]
+
+def _cabal_tool_flag(tool):
+    """Return a --with-PROG=PATH flag if input is a recognized Cabal tool. None otherwise."""
+    if tool.basename in _CABAL_TOOLS:
+        return "--with-{}={}".format(tool.basename, tool.path)
+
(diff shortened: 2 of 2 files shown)


haskell/cabal.bzl
```diff
@@ -581,6 +615,10 @@ stack_snapshot = repository_rule(
         "deps": attr.label_list(
             doc = "Dependencies of the package set, e.g. system libraries or C/C++ libraries.",
         ),
+        "tools": attr.label_list(
+            doc = """Tool dependencies. They are built using the host configuration, since
```
aherrmann: I don't see the host configuration enforced anywhere. [`label_list`](https://docs.bazel.build/versions/master/skylark/lib/attr.html#label_list) seems to suppor the `cfg = "host"` attribute, so, maybe that should be set?

haskell/cabal.bzl
```diff
@@ -166,11 +177,14 @@ def _haskell_cabal_library_impl(ctx):
         "_install/lib/libHS{}-ghc{}.{}".format(name, hs.toolchain.version, _so_extension(hs)),
         sibling = cabal,
     )
+    (tool_inputs, tool_input_manifests) = ctx.resolve_tools(tools = ctx.attr.tools)
```
aherrmann: Fascinating, I didn't know about [`resolve_tools`](https://docs.bazel.build/versions/master/skylark/lib/ctx.html#resolve_tools). Looks like [it was introduced in 0.23.0](https://github.com/bazelbuild/bazel/commit/52446a12b7435e850347dda144566280625cf498), there's a [design doc](https://docs.google.com/document/d/1xPsvTY-vWav9zX--7ieXjUilcl7M46m-88_oNV0QhEU/edit#) for anyone interested.

WORKSPACE
```diff
@@ -33,6 +33,17 @@ load("@os_info//:os_info.bzl", "is_linux", "is_windows")
 # bazel dependencies
 haskell_repositories()
 
+http_archive(
```
aherrmann: Having to provision tools like that is quite a lot of boilerplate. Especially, if you want to keep snapshot versions in sync. Hazel can use the tools straight from the Stackage snapshot. Is that something that `stack_snapshot` could be taught as well in the future?

mboes: Well, there's a lot that you can do if you parse `.cabal` files. We don't, and Stack doesn't tell us which dependency is a tool dependency and which one is not.

mboes:
@aherrmann Since we whitelist build tools anyways (like Cabal does), come to think of it we could teach the whitelist to `stack_snapshot` and have it emit a `haskell_cabal_binary` instead of `haskell_cabal_library` when it sees something in the whitelist. That would solve all the boilerplate issues you speak of. We'd get getter build parallelization as a bonus because Bazel would be able to build other libraries in parallel to the tools, rather than building the tools upfront. However, there are plenty of other binaries in a snapshot, many of which are not tools. We would still generate bad target definitions for those. We would be in a situation where sometimes binaries listed in the `packages` attribute are handled correctly (when they are build tools), sometimes not. Thoughts?

In principle we could detect binary targets and absence of library sections by grepping the .cabal files for `^library` or `^executable [a-Z0-9-]+`, but that's venturing into pretty hacky territory.

haskell/cabal.bzl
```diff
@@ -72,7 +72,14 @@ main = defaultMain
         )
     return setup
 
-def _prepare_cabal_inputs(hs, cc, dep_info, cc_info, cabal, setup, srcs, cabal_wrapper_tpl, package_database):
+_CABAL_TOOLS = ["alex", "c2hs", "cpphs", "doctest", "happy", "tar"]
```
aherrmann: Which `tar` is that. Is it the binary `tar`, or the Hackage package [`tar`](http://hackage.haskell.org/package/tar)?

WORKSPACE
aherrmann: Maybe ultimately a short-coming of stack. Apart from scanning Cabal files, there's also Cabal's [`plan.json`](http://hackage.haskell.org/package/cabal-plan), but last I checked the format wasn't documented and it was unclear if it's stable, yet.

haskell/cabal.bzl
mboes: Removed.

haskell/cabal.bzl
mboes: Fixed.

PR #910, opened 2019-05-30
[WIP] Implement stack_snapshot.vendored_packages

### LINKED ISSUE #902: stack_install: let user apply source patches
With Hazel some Hackage packages require small source patches to build under Bazel. E.g. to adjust file paths to embedded files or runtime data dependencies. See `wai-app-static` for example. The new `stack_install` does not allow users to apply source patches.

`stack_install` should allow users to specify source patches. Ideally in a way similar to [`http_archive`](https://github.com/bazelbuild/bazel/blob/62e3df71ca74d493a48351fa36760fa5bd889322/tools/build_defs/repo/http.bzl#L56).

Users could fork the package repository and if `stack_install` allowed to override a package's source repository users could inject patches that way. For small patches this is usually too much overhead.

**Additional context**
Opening issue as suggested in https://github.com/tweag/rules_haskell/pull/887#discussion_r287280815.

mboes:
This feature allows overriding packages in the snapshot with targets
defined manually. Useful when the source distribution needs to be
patched, or when the `BUILD` file needs to be customized.

Closes #902.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -548,7 +548,15 @@ Specify a fully qualified package name of the form <package>-<version>.
                 dependencies[src].append(dest)
     return (dependencies, transitive_unpacked_sdists)
 
+def _invert(d):
+    """Invert a dictionary."""
+    return dict(zip(d.values(), d.keys()))
+
+def _label_to_string(label):
+    return "@{}//{}:{}".format(label.workspace_name, label.package, label.name)
+
 def _stack_snapshot_impl(repository_ctx):
+    vendored_dependencies = _invert(repository_ctx.attr.vendored_packages)
     packages = repository_ctx.attr.packages
     non_core_packages = [
         package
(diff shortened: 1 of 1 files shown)


haskell/cabal.bzl
```diff
@@ -628,15 +635,19 @@ haskell_cabal_library(
...
     _stack_snapshot_impl,
     attrs = {
         "snapshot": attr.string(
             doc = "The name of a Stackage snapshot.",
+            mandatory = True,
         ),
         "packages": attr.string_list(
             doc = "A set of package identifiers. For packages in the snapshot, version numbers can be omitted.",
         ),
+        "vendored_packages": attr.label_keyed_string_dict(
```
aherrmann: This doesn't seem to be covered by the test-suite.

mboes:
> I.e. we have to write the BUILD file ourselves, and in particular we have to list the dependencies for network ourselves.

Yes, this is correct.

mboes:
I'm parking this PR as WIP for now. The reason is, while the change is fairly lightweight, it's not super convenient to test, but much more importantly, the work is a little ahead of our actual use cases at this point, especially given that I'd expect many use cases to be covered by Stack custom snapshots instead. We should merge this once downstream starts feeling the pinch of not having this feature.

PR #911, opened 2019-05-31
Add a tools attribute to core haskell rules

mboes:
This attribute works in much the same way the tools attribute for
plugins. This attribute can be used to expose GHC preprocessors like
`hspec-discover`.


tests/binary-with-tool/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
...
+
+package(default_testonly = 1)
+
+haskell_test(
+    name = "binary-with-tool",
+    srcs = ["Main.hs"],
+    args = ["$(location //tests/binary-simple)"],
+    expected_covered_expressions_percentage = 100,
+    tags = ["coverage-compatible"],
+    tools = ["//tests/binary-simple"],
```
aherrmann: This test case confuses me. I understand the `tools` attribute as compile time dependencies, plugins and preprocessors. But, then using it as a `$(location ...)` arg sounds more like a `data` attribute dependency. The test-case also doesn't actually use the tool at compile time, so I'm not sure what it really tests.

mboes: It passes the tool as an argument to the binary. The binary checks that there is an argument. If there is none, it fails. Not that we would get that far if `tools` didn't exist: because then `$(location ...)` would fail.

aherrmann: My point is it doesn't test anything about the handling of `tools` in rules_haskell. To illustrate this:
- if the attribute is changed to `data = ["//tests/binary-simple"]`, then the test-case still passes.
- also if `_resolve_preprocessors` is replaced by the following dummy, then the test-case still passes. 
    ```
    def _resolve_preprocessors(ctx, preprocessors):
        return struct(inputs = depset(), input_manifests = [])
    ```

haskell/toolchain.bzl
```diff
@@ -72,6 +72,7 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, params_file = None, e
         params_file_src = "<(:)"  # a temporary file with no contents
 
     script = """
+find
```
aherrmann: Is that a leftover?

tests/binary-with-tool/Main.hs
```diff
@@ -0,0 +1,6 @@
+{-# OPTIONS_GHC -F -pgmF bazel-out/host/bin/tests/binary-with-tool/cat #-}
```
aherrmann: Perfectly fine for this test-case, though I wonder if there's a way to avoid hard-coding that path. Maybe if we'd expand `$(location ...)` in `compiler_flags`?

mboes: Well, now you're asking for another feature. ;) Which happens to be the subject of the next PR. Though I'm not sure it will help this particular case. This path is at least stable across architectures, and seems to arise because the build is happening in one configuration while the tool was built in another. It appears that none of it should be considered Bazel private, but rather a stable Bazel convention for how to name things.

PR #912, opened 2019-05-31
Compatibility with bazel 0.25

guibou:
With theses changes, a subset of rules_haskell is now compatible with
bazel 0.25.

- `load` arguments must be placed first in bazel files
- `default_python_version` is now `python_version`
- Some rule path where fixed. Bazel is now more strict about the
  syntax package:rule. Previously accepted //:package/path/rulename
  must be converted to //package/path:rulename

*NOTE*

I tested these changes using a bazel 0.25 hacked from nixpkgs. It does not fully build rules_haskell (i.e. java issues, which are related to nixpkgs packaging), but everything else works. I'm successfully using bazel 0.25 with rules_haskell on a client codebase with these changes.

### DIFF
--- a/debug/linking_utils/ldd_test.bzl
+++ b/debug/linking_utils/ldd_test.bzl
@@ -1,5 +1,5 @@
 load(
-    "//:tests/inline_tests.bzl",
+    "//tests:inline_tests.bzl",
     "py_inline_test",
 )
 
--- a/haskell/ghc_bindist.bzl
+++ b/haskell/ghc_bindist.bzl
@@ -1,5 +1,7 @@
 """Workspace rules (GHC binary distributions)"""
 
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "patch")
+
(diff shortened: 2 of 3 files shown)


guibou:
@aherrmann 

```
  # bazel 0.25 not yet in nixpkgs
  # it builds, but fails the integrated tests
  # This override build it, and do not run the tests ;)
  # See for the discussion. https://github.com/NixOS/nixpkgs/pull/61097
  bazel = pkgs.bazel.overrideAttrs(oldAttrs: rec {
    version = "0.25.2";
    name = "bazel-${version}";

    src = pkgs.fetchurl {
      url = "https://github.com/bazelbuild/bazel/releases/download/${version}/${name}-dist.zip";
      sha256 = "7456032199852c043e6c5b3e4c71dd8089c1158f72ec554e6ec1c77007f0ab51";
    };

    doInstallCheck = false;
});
```

Works if you don't use any java stuff. See the `doInstallCheck` which disable tests for java.

PR #913, opened 2019-05-31
[WIP] No global pkgdb

aherrmann:
**WIP do not merge**

- Compile with `-no-global-package-db`
- Rewrite the toolchain library's package configuration, dropping any library references
    - Allows us to control when we use static or dynamic Haskell or C libraries
    - Avoids issues due to mismatch names between dynamic and static libraries
- Propagate all libraries (Haskell and C) through `CcInfo` providers
    - Resolves the Haskell/Cc/Haskell sandwich issue
- Use static libraries wherever possible, including Template Haskell (PIC static libraries)
    - This reduces overall build times if dynamic libraries never need to be produced
    - This avoids any MACH-O header size limit issues on MacOS, as those only occur with dynamic libraries.
    - At least for linking it also resolves the static C library name clash issue, as we can pass the `.a` files with their full path in `ld-options`.

---

- Already works for most of the Haskell targets under `//tests/...`.
- The following are missing
    - [ ] REPL - rules need to be adjusted
    - [ ] doctest - depends on Cabal/stack
    - [ ] Cabal/stack - we can't use the Cabal generated package configuration
    - [ ] In some cases the gcc command-lines become too long.
        The reason is that the `user_link_flags` in the `haskell_import` libraries repeat flags many times. In particular the `rts` package has a large number of flags. Probably the best solution would be to have `haskell_import` generate a proper target providing `CcInfo` containing only its own `ld-options`. To that end we need to work around the limitation that all toolchain dependencies are built in the `host` configuration.


aherrmann:
Changing package configuration files to list haskell libraries under `extra-libs` instead of `hs-libraries` conflicts with GHC plugins. When GHC loads a plugin it requires the module to be available in an `hs-libraries` item in the package providing the plugin. This is a blocker for this PR.

rules_haskell does already carry both Haskell and C libraries in `CcInfo` providers. See https://github.com/tweag/rules_haskell/pull/921

rules_haskell already supports GHC with a static RTS even on Unix, which allows to use only static Haskell libraries. See https://github.com/tweag/rules_haskell/pull/970

In light of this I'm closing this PR.

PR #915, opened 2019-05-31
Expand location and make variables in compiler flags

mboes:
The documentation for both the C/C++ rules and the Java rules
carefully notes in any attribute that is a list of flags to something
that this attribute is subject to "Make variable substitution" (and
bourne shell tokenization). I don't know how to do the latter, and
I don't know that it matters. But for consistency with the Java and
C/C++ rules, we now do the same too.

This can occasionally come in handy. For example to pass in the
location of a tool that belongs to the java toolchain.

### DIFF
--- a/haskell/private/haskell_impl.bzl
+++ b/haskell/private/haskell_impl.bzl
@@ -136,6 +136,21 @@ def _resolve_preprocessors(ctx, preprocessors):
         input_manifests = input_manifests,
     )
 
+def _expand_make_variables(name, ctx, strings):
+    # All labels in all attributes should be location-expandable.
+    label_attrs = [
+        ctx.attr.srcs,
+        ctx.attr.extra_srcs,
+        ctx.attr.data,
+        ctx.attr.deps,
+        ctx.attr.plugins,
+        ctx.attr.tools,
+    ]
+    targets = [target for attr in label_attrs for target in attr]
+    strings = [ctx.expand_location(str, targets) for str in strings]
+    strings = [ctx.expand_make_variables(name, str, {}) for str in strings]
+    return strings
+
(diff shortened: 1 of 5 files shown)


haskell/private/haskell_impl.bzl
```diff
@@ -136,6 +136,21 @@ def _resolve_preprocessors(ctx, preprocessors):
...
+        ctx.attr.srcs,
+        ctx.attr.extra_srcs,
+        ctx.attr.data,
+        ctx.attr.deps,
+        ctx.attr.plugins,
+        ctx.attr.tools,
+    ]
+    targets = [target for attr in label_attrs for target in attr]
+    strings = [ctx.expand_location(str, targets) for str in strings]
+    strings = [ctx.expand_make_variables(name, str, {}) for str in strings]
```
aherrmann:
```suggestion
    # ctx.expand_make_variables is deprecated. Replace when #5859 is resolved:
    # https://github.com/bazelbuild/bazel/issues/5859
    strings = [ctx.expand_make_variables(name, str, {}) for str in strings]
```

tests/binary-with-tool/Main.hs
```diff
@@ -1,4 +1,5 @@
-{-# OPTIONS_GHC -F -pgmF bazel-out/host/bin/tests/binary-with-tool/cat #-}
+{-# LANGUAGE CPP #-}
+{-# OPTIONS_GHC -F -pgmF CAT #-}
```
aherrmann: That's a nice trick.

Profpatsch:
Has the merge bot failed us? oO

mboes:
Yeah, weird. Merging my hand.

PR #916, opened 2019-06-02
Nixpkgs config is explicitly set to {}

guibou:
Else, nixpkgs will try to load it from ~/.config/nixpkgs/config.nix
which is not reproducible at all.

Future version of rules_nixpkgs won't accept it anymore.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -292,7 +292,7 @@ nixpkgs_package(
     # a workaround for
     # https://github.com/bazelbuild/bazel/issues/2927.
     nix_file_content = """
-    with import <nixpkgs> {};
+    with import <nixpkgs> { config = {}; };
     runCommand "nodejs-rules_haskell" { buildInputs = [ nodejs ]; } ''
       mkdir -p $out/nixpkgs_nodejs
       cd $out/nixpkgs_nodejs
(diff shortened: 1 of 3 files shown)


WORKSPACE
```diff
@@ -292,7 +292,7 @@ nixpkgs_package(
     # a workaround for
     # https://github.com/bazelbuild/bazel/issues/2927.
     nix_file_content = """
-    with import <nixpkgs> {};
+    with import <nixpkgs> { config = {}; };
```
aherrmann:
```suggestion
    with import <nixpkgs> { config = {}; overlays = []; };
```
`nixpkgs` also defaults the `overlays` attribute to files in `~/.config/nixpkgs`.

docs/haskell-use-cases.rst
```diff
@@ -144,7 +144,7 @@ This definition assumes a ``ghc.nix`` file at the root of the
 repository. In this file, you can use the Nix expression language to
 construct a compiler with all the packages you depend on in scope::
 
-  with (import <nixpkgs> {});
+  with (import <nixpkgs> { config = {}; });
```
aherrmann:
```suggestion
  with (import <nixpkgs> { config = {}; overlays = []; });
```

nixpkgs/cc-toolchain.nix
```diff
@@ -1,4 +1,4 @@
-with import ./. {};
+with import ./. { config = {}; };
```
aherrmann:
```suggestion
with import ./. { config = {}; overlays = []; };
```

PR #921, opened 2019-06-04
Remove HaskellCcInfo in favor of CcInfo

aherrmann:
Removes the `HaskellCcInfo` provider and instead makes full use of the `CcInfo` provider. This change allowed to factor out and simplify a couple of places related to handling library dependencies. And remove a couple now unused functions.

The main functions to use `CcInfo` are `get_extra_libs` (for linking) and `get_ghci_extra_libs` (for loading in GHCi). Both these functions internally take care of matching static and dynamic library names, adjust dynamic library file extensions, and collecting dynamic libraries into one directory to avoid issues with the MACH-O header size limit.

Constructing linker flags has also been factored out to reduce duplication.

This PR is a step towards #913.

### DIFF
--- a/haskell/providers.bzl
+++ b/haskell/providers.bzl
@@ -2,76 +2,12 @@
 
 load(
     ":private/path_utils.bzl",
-    "darwin_convert_to_dylibs",
+    "get_lib_name",
     "make_path",
-    "windows_convert_to_dlls",
+    "mangle_static_library",
+    "symlink_dynamic_library",
+    "target_unique_name",
 )
-
-HaskellCcInfo = provider(
-    doc = "Haskell cc dependency information. Part of HaskellInfo.",
-    fields = {
-        "static_linking": """static linking mode parameters.
-            A struct of
-            (libraries_to_link, dynamic_libraries_for_runtime, user_link_flags).
-            Libraries in libraries_to_link are struct(lib, mangled_lib)
-            because the Darwin linker needs the original library path,
-            while the Linux linker needs the mangled path.
-            """,
-        "dynamic_linking": """static linking mode parameters.
-            A struct of
-            (libraries_to_link, dynamic_libraries_for_runtime, user_link_flags).
-            Libraries in libraries_to_link are struct(lib, mangled_lib)
-            because the Darwin linker needs the original library path,
-            while the Linux linker needs the mangled path.
-            """,
-    },
-)
-
-def empty_HaskellCcInfo():
-    return HaskellCcInfo(
-        static_linking = struct(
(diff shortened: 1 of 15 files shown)


aherrmann:
I've tested this on the daml repository. I think that this requires the custom package configuration files for compile/link flags to be fully applicable. It still makes sense to keep this in a separate PR to make it more digestable (It's already pretty big). But it might be best to hold this one until the package config files are ready.

haskell/providers.bzl
```diff
@@ -218,3 +79,203 @@ GhcPluginInfo = provider(
...
+        if lib_to_link.pic_static_library:
+            static_lib = lib_to_link.pic_static_library
+        elif lib_to_link.static_library:
+            static_lib = lib_to_link.static_library
+
+        if dynamic_lib:
+            dynamic_lib = symlink_dynamic_library(hs, dynamic_lib, fixed_lib_dir)
+        static_lib = mangle_static_library(hs, dynamic_lib, static_lib, fixed_lib_dir)
+
+        if dynamic and dynamic_lib:
```
mboes: What does it mean to be `not (dynamic and dynamic_lib)`? In that case it gets added to `static_libs`, like when `static_lib != None`. Is that right?

haskell/providers.bzl
```diff
@@ -218,3 +79,203 @@ GhcPluginInfo = provider(
...
+        static_lib = mangle_static_library(hs, dynamic_lib, static_lib, fixed_lib_dir)
+
+        lib = static_lib if static_lib else dynamic_lib
+        if dynamic and dynamic_lib:
+            lib = dynamic_lib
+
+        if lib:
+            libs.append(lib)
+
+    # NOTE: We can avoid constructing these in the future by instead generating
```
mboes: Let's turn this into a ticket and link to the ticket in the comment. It's a fairly important note, because some corner cases essentially require us to resolve it.

haskell/providers.bzl
aherrmann: Yes, `dynamic` is a bool, `True` means prefer dynamic lib, `False` means prefer `static_lib`. We can only use `dynamic_lib` if it exists. That explains the first branch. Second branch, if there is no `static_lib` then we have to use the `dynamic_lib`. Third, default to `static_lib`. First and second could be merged into `if (dynamic and dynamic_lib) or not static_lib`, though I'm not sure if thats more readable.

mboes:
Looks fine

haskell/providers.bzl
aherrmann: Done #926

PR #925, opened 2019-06-06
Deduplicate toolchain libraries

### LINKED ISSUE #917: linkopts in haskell_import grow exponentially
**Describe the bug**
The `haskell_import` rule constructs the `HaskellImportHack` provider with a field to carry the `linkopts` field meant for `CcInfo`. It does so by combining the `HaskellImportHack.linkopts` of all its dependencies. This causes `linkopts` to grow exponentially. This is generally wasteful, and quickly exceeds the maximum command line length (tested on Linux) if those `linkopts` are passed to a tool like `gcc` (e.g. in #913).

**To Reproduce**
Build a Haskell target, e.g. `//tests/binary-simpl`, and observe the toolchain libararies' `linkopts`. E.g. adding the following to `haskell_import_impl`
```
    print(id, "len(linkopts)", len(import_info.linkopts))
```
produces the following output (formatted for readability)
```
ghc-prim-0.5.3            len(linkopts) 83
rts                       len(linkopts) 83
integer-gmp-1.0.2.0       len(linkopts) 86
array-0.5.3.0             len(linkopts) 252
base-4.12.0.0             len(linkopts) 252
filepath-1.4.2.1          len(linkopts) 252
ghc-boot-th-8.6.5         len(linkopts) 252
transformers-0.5.6.2      len(linkopts) 252
xhtml-3000.2.2.1          len(linkopts) 252
terminfo-0.4.1.2          len(linkopts) 255
ghc-heap-8.6.5            len(linkopts) 418
deepseq-1.4.4.0           len(linkopts) 504
mtl-2.2.2                 len(linkopts) 504
stm-2.5.0.0               len(linkopts) 504
time-1.8.0.2              len(linkopts) 756
pretty-1.1.3.6            len(linkopts) 839
bytestring-0.10.8.2       len(linkopts) 925
containers-0.6.0.1        len(linkopts) 1091
ghc-compact-0.1.0.0       len(linkopts) 1260
template-haskell-2.14.0.0 len(linkopts) 1343
unix-2.7.2.2              len(linkopts) 1937
binary-0.8.6.0            len(linkopts) 2520
directory-1.3.3.0         len(linkopts) 3197
text-1.2.3.1              len(linkopts) 4622
hpc-0.6.0.3               len(linkopts) 5548
process-1.6.5.0           len(linkopts) 6142
parsec-3.1.13.0           len(linkopts) 6303
ghc-boot-8.6.5            len(linkopts) 7398
haskeline-0.7.4.3         len(linkopts) 14807
ghci-8.6.5                len(linkopts) 17396
libiserv-8.6.3            len(linkopts) 24625
Cabal-2.4.0.1             len(linkopts) 30348
ghc-8.6.5                 len(linkopts) 50690
```

**Expected behavior**
`ld-options` should not grow more than linearly. Ideally, `linkopts` should only hold the `ld-options` of the current package being imported. The `linkopts` should only be combined later on using `merge_cc_infos`, so that Bazel can avoid unnecessarily duplicating `linkopts`.

**Environment**
 - OS name + version: openSUSE Tumbleweed 20190529
 - Bazel version: 0.24.0 provided by `shell.nix`
 - Version of the rules: 1672e2bff44ce7ef66cf638e7819215ffd9936b5

**Additional context**
As detailed in https://github.com/bazelbuild/bazel/issues/8129 we cannot directly produce a `CcInfo` in `haskell_import` as any toolchain dependencies would be built in the `host` configuration. As further detailed in https://github.com/bazelbuild/bazel/pull/5650 there is work in progress to allow to control the configuration of toolchain dependencies, but it is not available, yet. The recommended work around seems to be to define such targets separately from the toolchain and have all rules depend on them directly.

`rules_haskell` currently implements this with the separation of `haskell_import` and `haskell_toolchain_library`. However, this is still not fully satisfactory. Consider the following example:
```
haskell_toolchain_library(name = "base")
haskell_toolchain_library(name = "transformers")
haskell_library(name = "foo", deps = [":base", ":transformers"], ...)
```
Here, `base` and `transformers` will share common toolchain library dependencies, e.g. `rts`, however, as there is no functioning target providing `CcInfo` for `rts`, both `:base` and `:transformers` will still have to duplicate the corresponding `linkopts` (and other fields) internally.

It seems difficult to come up with an alternative implementation while maintaining the user interface. Here's a first attempt (which doesn't work):
- `haskell_import` will continue to return a provider like `HaskellImportHacked`.
- We add a (private) repository rule, say `make_toolchain_libraries`, that is called automatically by `haskell_repositories`, which defines unique global targets for all toolchain libraries. It does so by calling what is now `haskell_toolchain_library`. This would produce targets like `@io_tweag_rules_haskell_toolchain_libraries//:base`. These targets would produce the proper `CcInfo` and `HaskellInfo` providers, which don't manually duplicate `linkopts` and others, but instead use regular dependencies and `merge_cc_info`.
- We replace the user facing `haskell_toolchain_library` by a macro that just generates an `alias` to the global toolchain library targets.

Unfortunately, this doesn't work because `make_toolchain_libraries` wouldn't know which toolchain libraries are actually provided by the toolchain (especially in case of nixpkgs). Also, in order to know the dependencies of a toolchain library we need access to the toolchain's package db.

aherrmann:
Closes #917 
Workaround for https://github.com/bazelbuild/bazel/issues/8129 and https://github.com/bazelbuild/bazel/pull/5650

- Deduplicate the `CcInfo` providers for toolchain libraries.
- Remove unused `id_file` in `haskell_toolchain_library_impl`.

**What rules_haskell did before**
- `haskell_import` loaded libraries included in the GHC distribution into Bazel.
    This collected the files and flags of all transitive dependencies into a single target. The consequence was duplication. First, the way transitive dependencies were collected would duplicate contents yielding exponential growth. Second, if a regular target depended on two toolchain libraries, e.g. `base` and `transformers`, this would yield additional duplication.
- `haskell_toolchain` depended on all `haskell_import` targets and collected them into a `dict`.
    This is the source of the *host* configuration issue, because `haskell_import` targets are a dependency of the toolchain, they are always built in the *host* configuration. Therefore, we cannot use `CcInfo` providers generated in `haskell_import` to build `haskell_...` targets in the *target* configuration.
- `haskell_toolchain_library` depended on the toolchain, plucked out the requested library from the `dict` and returned its providers.

**What this PR does**
- `haskell_import` still loads libraries included in the GHC distribution into Bazel.
    However, this does no longer collect files or flags of transitive dependencies, it only collects the files and flags of that particular library. Dependencies are handled later on.
- `haskell_toolchain` *unchanged*
- `haskell_toolchain_libraries` (*plural*) Depends on the toolchain and generates `CcInfo` for all toolchain libraries.
    This introduces an additional layer of indirection, which allows to work around the host vs. target configuration issue. We define a single global instance of this rule within rules_haskell. The user is not meant to interact with it directly. Since this is a regular target, it will be built in the *target* configuration. This rule reads the libraries `dict` from the toolchain and generates a `CcInfo` provider (in the *target* configuration) for each library. This is where we handle toolchain library dependencies. We use the standard `cc_common.merge_cc_infos` at this point to merge the `CcInfo`s of all transitive dependencies of a library. These `CcInfo` providers are unique per library and Bazel can properly take care to not duplicate any libraries or flags. This rule exposes a `dict` from library name to a `struct` of providers. This is easily extensible to handle transitive dependencies on other providers, e.g. combining `HaskellInfo` providers.
- `haskell_toolchain_library` (*singular*) now implicitly depends on `haskell_toolchain_libraries`. All it does is pluck out the providers of the requested library and return them.

Note, the API is unchanged. The user does not need to care about `haskell_import` or `haskell_toolchain_libraries`. They call `haskell_toolchain_library` as before to get a handle on a toolchain library. However, now these do not cause any duplication. Even in the following silly example
```
haskell_toolchain_library(name = "base")
haskell_toolchain_library(name = "base_dup", package = "base")
haskell_toolchain_library(name = "transformers")
haskell_library(
    name = "mylib",
    deps = [":base", ":base_dup", ":transformers"],
    ...
)
```

### DIFF
--- a/haskell/BUILD.bazel
+++ b/haskell/BUILD.bazel
@@ -58,6 +58,10 @@ toolchain_type(
     visibility = ["//visibility:public"],
 )
 
+# XXX: This target is a work aroud for the lack toolchain transitions not being
+# implemented, yet. See
+# https://github.com/bazelbuild/proposals/blob/master/designs/2019-02-12-toolchain-transitions.md
+# This will need to be revisited once that proposal is implemented.
 haskell_toolchain_libraries(
     name = "toolchain-libraries",
     visibility = ["//visibility:public"],
--- a/haskell/BUILD.bazel
+++ b/haskell/BUILD.bazel
@@ -1,3 +1,8 @@
+load(
+    "@io_tweag_rules_haskell//haskell:haskell.bzl",
+    "haskell_toolchain_libraries",
+)
+
 exports_files(
     glob(["*.bzl"]) + [
         "assets/ghci_script",
@@ -50,5 +55,10 @@ toolchain_type(
 
 toolchain_type(
     name = "doctest-toolchain",
+    visibility = ["//visibility:public"],
+)
+
+haskell_toolchain_libraries(
+    name = "toolchain-libraries",
     visibility = ["//visibility:public"],
 )
--- a/haskell/haskell.bzl
+++ b/haskell/haskell.bzl
@@ -31,6 +31,7 @@ load(
     _haskell_import_impl = "haskell_import_impl",
     _haskell_library_impl = "haskell_library_impl",
     _haskell_test_impl = "haskell_test_impl",
+    _haskell_toolchain_libraries_impl = "haskell_toolchain_libraries_impl",
     _haskell_toolchain_library_impl = "haskell_toolchain_library_impl",
 )
 load(
@@ -314,25 +315,28 @@ haskell_import = rule(
     },
 )
 
+haskell_toolchain_libraries = rule(
+    _haskell_toolchain_libraries_impl,
+    attrs = {
+        "_cc_toolchain": attr.label(
+            default = Label("@bazel_tools//tools/cpp:current_cc_toolchain"),
+        ),
+    },
+    toolchains = [
+        "@io_tweag_rules_haskell//haskell:toolchain",
+    ],
+)
+
 haskell_toolchain_library = rule(
     _haskell_toolchain_library_impl,
     attrs = dict(
         package = attr.string(
             doc = "The name of a GHC package not built by Bazel. Defaults to the name of the rule.",
         ),
-        _version_macros = attr.label(
-            executable = True,
-            cfg = "host",
-            default = Label("@io_tweag_rules_haskell//haskell:version_macros"),
-        ),
-        # XXX We'll no longer need this once HaskellImportHack is removed.
-        _cc_toolchain = attr.label(
-            default = Label("@bazel_tools//tools/cpp:current_cc_toolchain"),
+        _toolchain_libraries = attr.label(
+            default = Label("@io_tweag_rules_haskell//haskell:toolchain-libraries"),
         ),
     ),
-    toolchains = [
-        "@io_tweag_rules_haskell//haskell:toolchain",
-    ],
 )
 """Import packages that are prebuilt outside of Bazel.
 
--- a/haskell/private/haskell_impl.bzl
+++ b/haskell/private/haskell_impl.bzl
@@ -609,16 +609,17 @@ def haskell_library_impl(ctx):
 # TODO Get rid of this by computing a CcInfo in haskell_import
 # instead. Currently blocked on upstream.
 HaskellImportHack = provider()
+HaskellToolchainLibraries = provider()
 
 def haskell_toolchain_library_impl(ctx):
-    hs = haskell_context(ctx)
-
     if ctx.attr.package:
         package = ctx.attr.package
     else:
         package = ctx.label.name
 
-    target = hs.toolchain.libraries.get(package)
+    libraries = ctx.attr._toolchain_libraries[HaskellToolchainLibraries].libraries
+    target = libraries.get(package)
+
     if not target:
         fail(
             """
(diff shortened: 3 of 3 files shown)


haskell/BUILD.bazel
```diff
@@ -52,3 +57,8 @@ toolchain_type(
     name = "doctest-toolchain",
     visibility = ["//visibility:public"],
 )
+
```
mboes: It should be documented in the code that this target is a workaround for https://github.com/bazelbuild/proposals/blob/master/designs/2019-02-12-toolchain-transitions.md not existing yet. Because this design is no longer going to work as-is once Bazel gets "configurations" (see [here](https://blog.bazel.build/2019/02/11/configurable-builds-part-1.html)). When that happens, there might be multiple configurations during the build, each of which require their own toolchain libraries.

aherrmann: :+1: I've added a comment to that effect.

Profpatsch:
Can we tell mergify include the PR description in the merge commit? I think we are losing lots of valuable history information if we don’t include that.

mboes:
Can you upstream that feature request?

PR #927, opened 2019-06-06
TemplateHaskell with cbits dependencies

aherrmann:
Before we never linked (`-l`) C library dependencies of the current package during compilation (already before #921). This made TemplateHaskell depending on such C libraries fail with
```
ghc: panic! (the 'impossible' happened)
  (GHC version 8.6.5 for x86_64-unknown-linux):
        Loading temp shared object failed: /run/user/1000/ghc4_0/libghc_6.so: undefined symbol: c_add_one
```
This fixes this issue. This also highlights that we need to use dynamic libraries during compilation so long as we're using the dynamic rts.

### DIFF
--- a/haskell/private/actions/compile.bzl
+++ b/haskell/private/actions/compile.bzl
@@ -10,6 +10,7 @@ load("@bazel_skylib//lib:paths.bzl", "paths")
 load(
     ":private/path_utils.bzl",
     "declare_compiled",
+    "link_libraries",
     "module_name",
     "target_unique_name",
 )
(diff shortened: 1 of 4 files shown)


aherrmann:
Rebasing this PR on #1039 resolves the CI failure with `@haskell_wai__app__static//:lib`.

Profpatsch:
Huh, @aherrmann, did we forget about this?

aherrmann:
@Profpatsch Yes, forgot about this one. Added to merge queue.

PR #928, opened 2019-06-06
Toolchain haddock is back

### LINKED ISSUE #901: haddock can't link to prebuild packages
**Describe the bug**

When building the docs for a package, haddock can't link to any identifier in prebuilt packages:

> Warning: LibB: could not find link destinations for:
    Int String

and the links indeed don't show up on the html doc:

image

**To Reproduce**

```sh
bazel build //tests/haddock:haddock-transitive
firefox bazel-bin/tests/haddock/haddock-transitive/testsZShaddockZShaddock-lib-b/LibB.html
```

Check that the `Int` or `String` types in the signatures aren't links.

**Expected behavior**

The generated documentation should contain links to `Int` or `String` from the standard library

**Environment**
- NixOS 19.09.git.c713d03 
- bazel 0.24.0 (from the project's shell.nix)
- rules_haskell at 9ba36c36c856cc0a196c9bb4573b3aaa4e8cf5b0 (current master)

**Additional context**

Might be a regression introduced by https://github.com/tweag/rules_haskell/pull/859 since this changes the way we handle prebuilt dependencies

guibou:
This closes #901 

This PR introduces the following:

- `haskell_import` have new fields to mimic the GHC package database haddoc informations
- `haddock_interfaces` field of `HaddockInfo` provider is now a list, to be coherent with what can be stored in the GHC package database `haddock-interfaces` field https://downloads.haskell.org/~ghc/8.6.4/docs/html/users_guide/packages.html
- the `pkgdb_to_bzl` script can now fill the haddock fields of `haskell_import`.
- a special case was introduced for nixpkgs (and others ?). If `haddock-interfaces` or `haddock-html` point to files which are outside of the `WORKSPACE` containing the ghc toolchain, `pkgdb_to_bzl` generates local bazel label and the bazel process generates the associated symbolic links. The information is passed by the script using a comment in the generated `BUILD` file.


guibou:
This work in a limited context and is full of hacks. More information will come.

haskell/haddock.bzl
```diff
@@ -232,30 +238,36 @@ def _haskell_doc_rule_impl(ctx):
...
+                        package_id,
+                        interface.path,
+                    ))
 
     args.add_all(all_caches, map_each = _dirname, format_each = "--optghc=-package-db=%s")
     locale_archive_depset = (
         depset([hs.toolchain.locale_archive]) if hs.toolchain.locale_archive != None else depset()
     )
 
+    haddocks = []
```
aherrmann: You could also use
```
haddocks = depset(transitive = [depset(i) for i in transitive_haddocks.values()])
```
If `transitive_haddocks` contained `depset`s then you could also drop the list comprehension.

Same in `_haskell_doc_aspect_impl`.

haskell/nixpkgs.bzl
```diff
@@ -32,6 +32,14 @@ def _ghc_nixpkgs_haskell_toolchain_impl(repository_ctx):
         fail("Error executing pkgdb_to_bzl.py: {stderr}".format(stderr = result.stderr))
     toolchain_libraries = result.stdout
 
+    # Haddock files on nixpkgs are stored outside of the ghc package
+    # The pkgdb_to_bzl.py program generate bazel label for theses files
```
aherrmann:
```suggestion
    # The pkgdb_to_bzl.py program generates bazel labels for theses files
```

haskell/nixpkgs.bzl
```diff
@@ -32,6 +32,14 @@ def _ghc_nixpkgs_haskell_toolchain_impl(repository_ctx):
...
+    # and ask the parent process to generate the associated bazel symlink
```
aherrmann:
```suggestion
    # and asks the parent process to generate the associated bazel symlink
```

haskell/private/pkgdb_to_bzl.py
```diff
@@ -108,13 +110,60 @@ def hs_library_pattern(name, mode = "static", config = ""):
...
 
     pkg_id_map.append((pkg.name, pkg.id))
+
+    # Haddock handling
+    # Haddock files may be packaged inside your ghc distribution (as
+    # it is the case for ghc_bindist) or elsewhere if you pull GHC
+    # from something such as nixpkgs which split ghc from its
+    # documentation
+
+    # For each paths in haddock_html and haddock_interfaces, we will
```
aherrmann:
```suggestion
    # For each path in haddock_html and haddock_interfaces, we will
```

haskell/haddock.bzl
guibou: Thank you!

guibou:
Yeah, I fixed the windows CI. Just need to understand the errors from darwin.

guibou:
It fails on darwing because the link checker do not find some paths. However they are broken as well on linux, but the link checker do not find them because it explicitly black list them ;)

The issue is that packages from the toolchain does not know their haddock dependencies. i.e. `base` depends on `ghc-prim` but that's not know for now.

This may also explain the weird behavior I described in #901 where some transitive dependencies were not previously working.

I'll try to fix that.

guibou:
Note: this works by explictly adding the transitive dependency in the `haskell_doc` call.

aherrmann:
> The issue is that packages from the toolchain does not know their haddock dependencies

@guibou https://github.com/tweag/rules_haskell/pull/925 addresses a part of the toolchain dependency handling, in that case regarding `CcInfo`. The changes in that PR might be relevant here.

mboes:
@guibou thanks to this PR, most Haddock warnings have disappeared. Though we still have Haddock warnings during the two protobuf tests. Maybe something you're doing in this PR needs to be replicated in the protobuf rules?

guibou:
@mboes yes. I observed something similar in another codebase even after the regression of #901. Actually I wrote #928 as a preliminary work to fix the issue you're discussing ;)

I'll open an issue with a description as soon as I understand this issue a bit more.

PR #932, opened 2019-06-09
Remove spurious use_default_shell_env in a compile actions

mboes:
If the `PATH` changed (which happens *very* frequently with
`nix-shell`), pretty much the entire cache had to be trashed because
all object files needed to be rebuilt. This commit fixes the problem.


Profpatsch:
Great catch. The (very sparse) docs of `use_default_shell_env` read like enabling it should actually *improve* the cache life, but apparently it is not so.

PR #934, opened 2019-06-10
Propagate haddock provider through protobuff toolchain

guibou:
Similar to #928, `HaddockInfo` is now correctly propagated through protobuff rules.


guibou:
Note: it may be tempting to share this with the code in #928 but each line have subtles differences.

mboes:
Great stuff! Haddock woes seem to be finally behind us.

PR #937, opened 2019-06-11
Fix ghc bindist scripts on macOS

Profpatsch:
We have to be compatible with BSD utils in repository_rules if we want to be compatible with MacOS for bindists.


haskell/ghc_bindist.bzl
```diff
@@ -170,6 +170,9 @@ GHC_BINDIST = \
 def _execute_fail_loudly(ctx, args):
     """Execute a command and fail loudly if it fails.
 
+    ATTN: All commands have to be cross-compatible between BSD tools and GNU tools,
+    because we want to support MacOS. Please cross-reference the MacOS man-pages.
```
mboes: s/MacOS/macOS/.

PR #941, opened 2019-06-12
Fix various pitfalls for new users

Profpatsch:
There was some confusion about how to best setup `rules_haskell` for development.

cc @ulysses4ever


shell.nix
```diff
@@ -41,9 +41,10 @@ mkShell {
       elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
         ARCH="linux"
       fi
-      echo "[!] It looks like you are using a ''${ARCH} nix-based system. In order to build this project, you probably need to add the two following host_platform entries to your .bazelrc.local file."
+      echo "[!] It looks like you are using a ''${ARCH} nix-based system. In order to build this project, you need to add the two following host_platform entries to your .bazelrc.local file."
       echo ""
-      echo "test --host_platform=@io_tweag_rules_haskell//haskell/platforms:''${ARCH}_x86_64_nixpkgs"
```
ulysses4ever: But now `test` is not covered, is it?

Profpatsch: Yes, see commit message

ulysses4ever: Oh, good to know about build covering both, build and rest, thanks!

PR #942, opened 2019-06-13
Add haskell_hlint rule

aherrmann:
- Adds `hlint_toolchain` to specify the `hlint` executable and optionally an `.hlint.yaml` configuration file.
- Adds the `haskell_hlint_aspect` which runs `hlint` on the sources of the given target and its dependencies. Similar to the `doctest` aspect.
- Adds the `haskell_hlint` rule which uses the `haskell_hlint_aspect` similar to the `doctest` rule.
- Adds test cases for both, a case were `hlint` passes and a case where `hlint` fails.

Based on the `hlint` aspect and rule from the [`daml` repository](https://github.com/digital-asset/daml/blob/106e90e8a626b99dc43d8a74ebbc91e30c2d8112/bazel_tools/hlint.bzl). cc @ndmitchell


aherrmann:
@mboes I've renamed the rule from `haskell_hlint` to `haskell_lint`.

haskell/lint.bzl
```diff
@@ -0,0 +1,170 @@
...
+                # seems to involve ignoring these generated paths
+                # (the f.is_source almost always returns True)
+                if all([
+                    f.path.endswith(".hs"),
+                    f.path.startswith("external/") == False,
+                    f.path.startswith("bazel-out/") == False,
+                    f.path.startswith("nix/") == False,
+                ]):
+                    inputFiles.append(f)
+                    inputPaths.append(f.path)
```
Profpatsch: What happens to generated Haskell files? Normally I’d expect they should not be linted by default.

Shouldn’t we restrict to a specified subset of rules instead of trying to heuristically filter sources?

haskell/lint.bzl
```diff
@@ -0,0 +1,170 @@
...
+
+    inputFiles = []
+    inputPaths = []
+    if hasattr(ctx.rule.attr, "srcs"):
+        for src in ctx.rule.attr.srcs:
+            for f in src.files:
+                # We want to only do native Haskell source files, which
+                # seems to involve ignoring these generated paths
+                # (the f.is_source almost always returns True)
+                if all([
```
regnat: Wouldn't [f.is_source](https://docs.bazel.build/versions/master/skylark/lib/File.html#is_source) filter the files as we want here?

aherrmann: Good point. When this was first implemented `f.is_source` was insufficient, and these patterns also filter out external repositories. Though, it seems an approach similar to [`coverage_source_patterns`](https://github.com/tweag/rules_haskell/blob/af0644b9ab8e2bbf6b6921ffe275181cf0fe8ec2/haskell/haskell.bzl#L163-L171) or [`from_source`](https://github.com/tweag/rules_haskell/blob/af0644b9ab8e2bbf6b6921ffe275181cf0fe8ec2/haskell/repl.bzl#L369-L380) would be the cleaner solution here. I'll give it a shot.

haskell/lint.bzl
aherrmann: Generated files are filtered out. Now using `f.is_source`. I've now also restricted the aspect to targets providing `HaskellInfo`.

haskell/lint.bzl
aherrmann: I've changed the code to use `f.is_source` and to allow users to control which targets to apply `hlint` to. PTAL

aherrmann:
@mboes I've pushed a commit that turns `haskell_lint` into `haskell_lint_test`, a test rule that will fail the test if `hlint` produced a warning. `hlint` is still executed during the build stage by the aspect. However, the results are stored and the rule generates a script that will print any `hlint` warnings and fail if `hlint` failed on any target.

haskell/lint.bzl
```diff
@@ -0,0 +1,294 @@
...
+        return []
+
+    hlint_toolchain = ctx.toolchains["@io_tweag_rules_haskell//haskell/lint:toolchain"]
+    hint = hlint_toolchain.hint
+
+    inputFiles = [
+        f
+        for src in getattr(ctx.rule.attr, "srcs", [])
+        for f in src.files
+        if f.is_source and f.extension in ["chs", "hs", "hs-boot", "hsc", "lhs", "lhs-boot"]
```
mboes: HLint seems to croak on `.chs` files that actually use C2hs syntax. I'm not sure we can lint those.

aherrmann: Good point, thanks fixed.

Profpatsch:
Do we have any way forward on this?

Profpatsch:
Closing in favor of the issue: https://github.com/tweag/rules_haskell/issues/1140

PR #949, opened 2019-06-14
WIP: try to run the RunTests.hs test suite for bindists

Profpatsch:
Let’s see what CI thinks about this.


mboes:
PR contains only trivial WIP change and seems dormant. Closing for now.

Issue #958, opened 2019-06-20
symlink_dynamic_library takes an unreasonable amount of time


regnat:
**Describe the bug**

Building an haskell project from a clean action cache but a fully populated remote cache (or disk_cache) is painfully slow, most of the time being spent in `Symlink` actions due to `symlink_dynamic_library`.

For example, building `rules_haskell` with a pre-populated `disk_cache` and after a `bazel clean` takes ~70s while it takes ~10s if I comment https://github.com/tweag/rules_haskell/blob/1d2083299f5bbbf24408595f9b95a73a918bb662/haskell/providers.bzl#L269-L270

**To Reproduce**

On `rules_haskell`

```bash
# Ensure that we have a fully populated disk-cache
bazel build --disk_cache=/tmp/disk_cache //...
# Clean the action cache
bazel clean
# Rebuild the project
bazel build --disk_cache=/tmp/disk_cache //...
# Observe that it is unreasonably long for something which should just be pulling from a remote cache

# Repeat the above steps with haskell/providers.bzl#L269-L270 commented and admire how fast this is
```

**Expected behavior**

A fresh build with a remote cache should be fast

**Environment**
 - OS name + version: NixOS 19.09.git.b29d4de
 - Bazel version: 0.24.0 (provided from the nix-shell), also observed on 0.26.0
 - Version of the rules: 1d2083299f5bbbf24408595f9b95a73a918bb662 (current master)

regnat:
(just to give some extra numbers, on the CI of a mid-sized haskell project I'm working on, commenting these two lines makes the build time drop from 4min to 50s)

Profpatsch:
The solution here is probably to change the definition of `ln()` (which shells out to `ln(1)` for each call) to something that can symlink multiple files in the same action.

guibou:
Are we sure that it is only `ln` which is problematic, or do we have any idea on the number of them?

```
λ paddle crevette → time bash -c "for i in \$(seq 10000); do ln -sf rules_nodejs-0.16.2 toto-\$i; done;"
bash -c   4.75s user 8.55s system 102% cpu 12.967 total
```

10K `ln` calls takes 12s. That's a lot, but @regnat is talking about 3minutes, which means roughly 150K symlinks, which feels too high.

aherrmann:
@guibou With `ln()` in Bazel each symlink is created in its own action. Creating the sandbox for each these actions surely adds some overhead.

[...]

PR #960, opened 2019-06-24
Don't sandbox symlink creation

aherrmann:
As noted in https://github.com/tweag/rules_haskell/issues/958 symlink creation takes a large amount of time in rules_haskell. This PR aims to reduce the time spent on creating symlinks by reducing the overhead due to sandboxing and caching symlinks.

A fully disk cached build of rules_haskell (`bazel clean && bazel fetch //... && bazel build //...`) takes
- Before this PR:
  - Elapsed time: 37.621s - 46.778s
  - Critical Path: 1.80s - 12.29s (varies strongly)
- After this PR:
  - Elapsed time: 22.375s - 26.950s
  - Critical Path: 1.37s - 1.76s

---

@regnat Could you test this on your use-case to verify that this PR helps?

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -390,3 +390,13 @@ buildifier_dependencies()
 load("@ai_formation_hazel//:workspace.bzl", "hazel_setup")
 
 hazel_setup()
+
+# For profiling
+# Required to make use of `bazel build --profile`.
+
+# Dummy target //external:python_headers.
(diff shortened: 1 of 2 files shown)


mboes:
@regnat is offline for at least the next week. Merging for now since this is low risk and we already have data that this improves the situation. Though it will be useful to have @regnat's numbers post facto.

PR #964, opened 2019-06-24
Allow @ in Haskell binary names

### LINKED ISSUE #962: @ in rule name
**Describe the bug**

A rule name with an `@` was used to be accepted, and it is not anymore.

For example, in current master:

```
haskell_binary(
    name = "foo@bar",
    srcs = ["Foo.hs"],
)
```

Leads to the following build failure:

```
[nix-shell:~/tweag/rules_haskell]$ bazel build //tests:foo@bar
INFO: Build option --test_env has changed, discarding analysis cache.
INFO: Analyzed target //tests:foo@bar (1 packages loaded, 1082 targets configured).
INFO: Found 1 target...
ERROR: /home/guillaume/tweag/rules_haskell/tests/BUILD.bazel:347:1: HaskellRegisterPackage tests/link-config-foo@bar/link-config-foo@bar.conf failed (Exit 1) ghc-pkg failed: error executing command external/io_tweag_rules_haskell_ghc_nixpkgs/bin/ghc-pkg recache '--package-db=bazel-out/k8-fastbuild/bin/tests/link-config-foo@bar' -v0 --no-expand-pkgroot

ghc-pkg: 0: Parse of field '
unexpected '@'
expecting "-", white space or end of input

link-config-foo@bar
' failed.
Target //tests:foo@bar failed to build
INFO: Elapsed time: 0.567s, Critical Path: 0.16s
INFO: 15 processes: 15 linux-sandbox.
```

Workaround: rename the rules.

Is this a bug or a requirement of rule name?

aherrmann:
Closes #962.
Fixes a regression introduced in #930.

Note, the Bazel documentation [specifies](https://docs.bazel.build/versions/master/build-ref.html#name)

> Target names must be composed entirely of characters drawn from the set `a–z`, `A–Z`, `0–9`, and the punctuation symbols `_/.+-=,@~`.

Currently, most of these are not supported by rules_haskell.

A possible solution would be to implement full [z-encoding](https://hackage.haskell.org/package/ghc-8.4.3/docs/src/Encoding.html#zEncodeString) in rules_haskell and add a case for `@`. Also, `ghc-pkg` does not allow more than one `-` in a row, so simply replacing characters by `-` can generate invalid package names.

### DIFF
--- a/haskell/providers.bzl
+++ b/haskell/providers.bzl
@@ -331,7 +331,7 @@ def create_link_config(hs, cc_info, binary, args, dynamic = None, pic = None):
         if not get_lib_name(lib).startswith("HS")
     ])
 
-    package_name = target_unique_name(hs, "link-config").replace("_", "-")
+    package_name = target_unique_name(hs, "link-config").replace("_", "-").replace("@", "-")
     conf_path = paths.join(package_name, package_name + ".conf")
     conf_file = hs.actions.declare_file(conf_path)
     write_package_conf(hs, conf_file, {
--- a/tests/target-name/BUILD.bazel
+++ b/tests/target-name/BUILD.bazel
@@ -1,0 +1,30 @@
+load(
+    "@io_tweag_rules_haskell//haskell:haskell.bzl",
+    "haskell_library",
+    "haskell_test",
+)
+
+package(default_testonly = 1)
+
+# Test that rules_haskell handles allowed characters in target names.
+#
+# The documentation specifies:
+# Target names must be composed entirely of characters drawn from
+# the set a–z, A–Z, 0–9, and the punctuation symbols _/.+-=,@~
+#
+# Currently, rules_haskell does not support the full set.
+
+haskell_library(
+    name = "lib0-a_Aa",
+    srcs = ["Lib.hs"],
+    deps = ["//tests/hackage:base"],
+)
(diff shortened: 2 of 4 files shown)


guibou:
> Do you have a use case where you really want @ in the binary name?

On a client codebase we use `@` for many extended targets. We can easily replace it by something else (note: as @aherrmann stated, we need to be careful about avoiding repetition of these chars, because they are replaced by `-`).

We can avoid this char, we just need to refactor on our side.

mboes:
I'd vote for banning `@`, owing to that character being used for companion targets. @aherrmann what do you think?

guibou:
@mboes 

That's ok for me, that's not a huge refactoring. I just need to find a better symbol.

aherrmann:
Generally I'm in favor of banning special characters to avoid too much complexity in the implementation. 

Though, the `@` character is convenient to have available as a user. E.g. in the `daml` repo we generate an implicit `@ghci` `haskell_repl` target for every Haskell target, which nicely matches the builtin `@xyz` targets. But it should be easy to replace that with another character when necessary, and reserving `@` for `rules_haskell` makes sense.

mboes:
OK I guess that's convincing. No need for rules_haskell to hog that namespace exclusively.

Issue #965, opened 2019-06-24
Link issue: haskell_doc + TemplateHaskell + haskell_import


guibou:
**Describe the bug**

I recently converted my codebase from the now deprecated `haskell_cc_import` to `cc_library`. `haskell_doc` fails at runtime with link issues.

**To Reproduce**

In 376123c0a3b376a9b4cb68f35b5b659d995f0f40, I'm able to reproduce the problem:

- `bazel build //tests:toto` fails with linking error related to `lz4`.
- lz4 is loaded in the WORKSPACE in the `@lz4//:lz4` `cc_library`.
- This cc_library is used as a `deps` an *empty* `haskell_library`, `//tests:lz4`
- `//tests:lz4` is used by an `haskell_library`: `//tests:utils`, which uses `TemplateHaskell`. That's important, without `TemplateHaskell`, the issue does not appears.
- All of theses libraries build
- `//tests:toto` is an `haskell_doc` which make reference to `utils` and fails to build.

Steps to reproduce the behavior.

**Expected behavior**

**Environment**
 - OS name + version:
 - Bazel version:
 - Version of the rules:

**Additional context**

guibou:
I've tried to adapt the work of #927 which seems relevant, but without success.

PR #967, opened 2019-06-24
Set LD_LIBRARY_PATH for haddock

aherrmann:
Closes: #965 

Haddock failed on the example given by @guibou in https://github.com/tweag/rules_haskell/commit/376123c0a3b376a9b4cb68f35b5b659d995f0f40 due to `liblz4.so.1` not being found. This PR resolves this issue by exporting `LD_LIBRARY_PATH` in the same way as is done for the REPL and doctests.

@guibou I've included your example as a regression test.

### DIFF
--- a/tests/Foo.hs
+++ b/tests/Foo.hs
@@ -1,0 +1,5 @@
+{-# LANGUAGE TemplateHaskell #-}
+module Foo where
+
+-- This is here just to trigger template haskell
+blork = $([| 1 + 1 |])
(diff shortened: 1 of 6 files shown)


tests/Foo.hs
```diff
@@ -0,0 +1,5 @@
+{-# LANGUAGE TemplateHaskell #-}
+module Foo where
+
+-- This is here just to trigger template haskell
+blork = $([| 1 + 1 |])
```
guibou: ``
```suggestion
blork = $([| 1 + 1 |])

```

aherrmann: Thanks, fixed.

PR #970, opened 2019-06-28
Support GHC with static RTS

aherrmann:
GHC can be compiled with a static RTS. This allows to use only static Haskell librarires including for GHCi and Template Haskell. This is the default on Windows.

This PR adds support for a static GHC on Unix as well. The mode of GHC (static or dynamic) is detected automatically during toolchain setup.

### DIFF
--- a/haskell/private/workspace_utils.bzl
+++ b/haskell/private/workspace_utils.bzl
@@ -1,0 +1,67 @@
+load("@bazel_tools//tools/cpp:lib_cc_configure.bzl", "get_cpu_value")
+
+def execute_or_fail_loudly(repository_ctx, arguments):
+    """Execute the given command
+
+    Fails if the command does not exit with exit-code 0.
+
+    Args:
+      arguments: List, the command line to execute.
+
+    Returns:
+      exec_result: The output of the command.
+
+    """
+    exec_result = repository_ctx.execute(arguments)
+    if exec_result.return_code != 0:
+        arguments = [_as_string(x) for x in arguments]
+        fail("\n".join(["Command failed: " + " ".join(arguments), exec_result.stderr]))
+    return exec_result
+
+def _as_string(v):
+    if type(v) == "string":
+        return v
+    else:
+        return repr(v)
+
+def _find_ghc(repository_ctx):
+    """Find the GHC executable in the current workspace.
+
+    Returns:
+      path, The path to the GHC executable.
+
+    """
+    if get_cpu_value(repository_ctx) == "x64_windows":
+        ghc = repository_ctx.path("bin/ghc.exe")
+    else:
+        ghc = repository_ctx.path("bin/ghc")
+
+    if not ghc.exists:
+        fail("Cannot find GHC executable in {}.".format(ghc))
+
+    return ghc
+
+def ghc_is_static(repository_ctx):
+    """Query GHC for whether the RTS is static or dynamic
+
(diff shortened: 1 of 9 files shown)


aherrmann:
To test rules_haskell with a static GHC on Unix you can apply the following patch to define the corresponding Nix override:
``` diff
diff --git a/nixpkgs/default.nix b/nixpkgs/default.nix
index f4303b0..c2e01ca 100644
--- a/nixpkgs/default.nix
+++ b/nixpkgs/default.nix
@@ -1,5 +1,31 @@
+{ ... }@args:
 import (fetchTarball {
    # Checkout from 2019-05-06.
    url = https://github.com/mboes/nixpkgs/archive/361e9185c83166b44419e05e75fc8473875df6ea.tar.gz;
    sha256 = "0c6izp9i1ln4yy2h9jlm48k94f018h8pw60jfbhq49p076aaafrb";
+}) (args // {
+  overlays = (args.overlays or []) ++ [
+    (_: pkgs: {
+      haskell = pkgs.haskell // {
+        compiler = pkgs.haskell.compiler // {
+          ghc865 = (pkgs.haskell.compiler.ghc865.override {
+            enableRelocatedStaticLibs = true;
+            enableShared = false;
+          }).overrideAttrs (attrs: {
+            # enableRelocatedStaticLibs = true only sets -fPIC. However, with
+            # enableShared = false we also require -fexternal-dynamic-refs,
+            # otherwise GHC still generates R_X86_64_PC32 relocations which
+            # prevents loading these static libraries as PIC.
+            preConfigure = ''
+              ${attrs.preConfigure or ""}
+              cat <<EOF >>mk/build.mk
+              GhcLibHcOpts += -fPIC -fexternal-dynamic-refs
+              GhcRtsHcOpts += -fPIC -fexternal-dynamic-refs
+              EOF
+            '';
+          });
+        };
+      };
+    })
+  ];
 })
```

aherrmann:
Rebased on master and awaiting review.

haskell/private/workspace_utils.bzl
```diff
@@ -0,0 +1,67 @@
...
+        ghc = repository_ctx.path("bin/ghc.exe")
+    else:
+        ghc = repository_ctx.path("bin/ghc")
+
+    if not ghc.exists:
+        fail("Cannot find GHC executable in {}.".format(ghc))
+
+    return ghc
+
+def ghc_is_static(repository_ctx):
```
mboes: This won't play nicely with RBE, or any cross-compilation for that matter.

mboes:
Rather than autodetection, it may be better to ask the user to say whether the toolchain is static or not, since this is going to complicate the cross compilation case. It's like compiler versions. In the case of Nixpkgs, compiler versions in the WORKSPACE file are redundant, since we can auto detect them. But we can only reliably detect them in narrow cases: when the build architecture matches the host architecture. Here, if the target architecture is different from host or build, `ghc --interactive` fails.

Profpatsch:
I agree that we shouldn’t do autodetection because of RBE (haven’t reviewed yet though).

aherrmann:
I was hoping we could go the opposite direction and also autodetect the version, since a wrong version setting can lead to pretty confusing error messages. However, it seems that wouldn't work RBE either then.

Static GHC seems like a niche use-case, so I'm fine with it requiring some manual setup.

PR #974, opened 2019-06-29
The start script requires exact version of Bazel now (fix #923)

### LINKED ISSUE #923: Add upper bazel version warning
At any given time we only test against a given bazel version, while upstream releases new versions every month (currently we test `0.24` and upstream is at `0.26`).
Because of breaking changes in every bazel version, we cannot really support any version we don’t test (yet).

So we should add a warning to the `start` script (and the haskell toolchain?) if the bazel version is higher than what we checked before.

ulysses4ever:


ulysses4ever:
I did exactly what was suggested in @aspiwack's [comment](https://github.com/tweag/rules_haskell/issues/923#issuecomment-506781651) because, in my experience, a too new version can crush on you as likely. 

But we certainly can!

ulysses4ever:
I'll abandon this PR and think about a new one — in case this idea of pinning an exact version will conquer the hearts at some point.

aspiwack:
I'd say the Right™ behaviour, in case the installed Bazel version is too new, is to fail by default and let folks override this if they really want an install (a warning is too easily missed or dismissed). Either by relaunching the script with a flag, or by dropping in an interactive prompt where the use has to type « yes », or something, to install with an untested version.

mboes:
Things aren't that simple: Bazel has a backwards compatibility guarantee now. We should use that.

aspiwack:
> Things aren't that simple: Bazel has a backwards compatibility guarantee now. We should use that.

I've got to shamefully confess: I don't understand how this point relates to my proposal.

What I was trying to say was that, whatever the range of Bazel versions that we accept, we want to refuse to automatically run the start script if the installed Bazel version is out of the range. And instead ask for some explicit confirmation. But it's fine to have a range, if we are confident enough.

PR #979, opened 2019-06-30
Use Stack global hints

mboes:
This requires Stack 2.1. The `--global-hints` flag is not available in 
prior versions. This flag allows us to query the dependency graph without
any GHC installed locally.


haskell/cabal.bzl
```diff
@@ -478,15 +485,20 @@ def _compute_dependency_graph(repository_ctx, versioned_packages, unversioned_pa
...
+    if stack_major_version < 2:
+        fail("Stack 2.1 or above required.")
+
+    # Unpack all given packages, then compute the transitive closure
+    # and unpack anything in the transitive closure as well.
+    stack = [stack_cmd]
     if versioned_packages:
         _execute_or_fail_loudly(repository_ctx, stack + ["unpack"] + versioned_packages)
-    stack = [stack_cmd, "--no-nix", "--skip-ghc-check", "--system-ghc", "--resolver", repository_ctx.attr.snapshot]
+    stack = [stack_cmd, "--resolver", repository_ctx.attr.snapshot]
```
aherrmann: Why are those flags no longer necessary? Or have they been removed?

shell.nix
```diff
@@ -17,7 +17,6 @@ mkShell {
     bazel
     # For stack_install. TODO Remove ghc when move to Stack 2.
```
aherrmann:
```suggestion
    # For stack_install.
```

haskell/cabal.bzl
mboes: These flags were meant to prevent Stack from trying to futz around with GHC. Newer Stack is now more lazy about pulling in GHC, so these flags are no longer required,  in particular with `--global-hints`.

PR #982, opened 2019-07-02
Fix incompatible_string_join_requires_strings

aherrmann:
Addresses [issue on CI](https://circleci.com/gh/tweag/rules_haskell/6979?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link).

It seems that stack failed at the fetch step and the Bazel error is from the code-path that constructs the error message. Since [Bazel 0.27](https://github.com/bazelbuild/bazel/issues/7802) it is no longer legal to apply string join to a list that contains non-string elements.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -17,7 +17,7 @@ load(
 def _execute_or_fail_loudly(repository_ctx, arguments):
     exec_result = repository_ctx.execute(arguments)
     if exec_result.return_code != 0:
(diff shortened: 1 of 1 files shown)


mboes:
This fix should be subsumed by #979, which I thought had been merged long ago, but flaky network conditions were preventing CI from passing. I have retried the build of that one and as soon the build goes green mergebot should be merging #979.

PR #987, opened 2019-07-06
Allow arbitrary Cabal tools

mboes:
Cabal has a whitelist of tools, for which we can supply a path
explicitly using `--with-PROG`, where `PROG` is the name of the tool.
But in practice `Setup.hs` scripts and the build process in general
sometimes use other tools that are not part of the whitelist. For
example, `postgresql-libpq` calls `pg_config` during configuration.
Unlike the whitelist, these tools can only be supplied by ensuring
they are reachable from the `PATH`. So we extend the `PATH` to make
this true, hoping that not too much else is exposed as a side effect.

We could avoid undeclared inputs leaking, but

1. we can't do this without a hefty performance cost on Windows (no
   symlinks),
2. in practice this should be the responsibility of the workspace
   rules for each dependency to expose just what we want,
3. Cabal rules are not hermetic to begin with, because they need all
   manner of standard utilities.

Adding a test for this would be awkward, and the feature is at a low
risk of regression (since it's already partially tested), so I didn't write one. Testing would require
making all of `@stackage`, and hence most tests, depend on Nixpkgs. We
don't want that, because of bindist builds.

### DIFF
--- a/haskell/private/cabal_wrapper.sh.tpl
+++ b/haskell/private/cabal_wrapper.sh.tpl
@@ -12,15 +12,33 @@
 
 set -euo pipefail
 shopt -s nullglob
-execroot="$(pwd)"
 
-# Set these variables if unset.
-: ${LD_LIBRARY_PATH:=}
-: ${LIBRARY_PATH:=}
-export LD_LIBRARY_PATH=$execroot/${LD_LIBRARY_PATH//:/:$execroot/}
-export LIBRARY_PATH=$execroot/${LIBRARY_PATH//:/:$execroot/}
+# Poor man's realpath, because realpath is not available on macOS.
+function realpath()
+{
+    [[ $1 = /* ]] && echo "$1" || echo "$PWD/${1#./}"
+}
+
+function canonicalize_path()
+{
+    new_path=""
+    while IFS=: read -r -d: entry
+    do
+	if [[ -n "$entry" ]]
+	then
+	    new_path="$new_path${new_path:+:}$(realpath "$entry")"
+	fi
+    done <<< "${1:-}:"
+    echo $new_path
+}
+
+# Remove any relative entries, because we'll be changing CWD shortly.
+LD_LIBRARY_PATH=$(canonicalize_path $LD_LIBRARY_PATH)
+LIBRARY_PATH=$(canonicalize_path $LIBRARY_PATH)
+PATH=$(canonicalize_path $PATH)
 
 name=$1
+execroot="$(pwd)"
 setup=$execroot/$2
 srcdir=$execroot/$3
 pkgroot="$(realpath $execroot/$4/..)" # By definition (see ghc-pkg source code).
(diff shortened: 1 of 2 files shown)


haskell/private/cabal_wrapper.sh.tpl
```diff
@@ -12,15 +12,27 @@
...
-export LD_LIBRARY_PATH=$execroot/${LD_LIBRARY_PATH//:/:$execroot/}
-export LIBRARY_PATH=$execroot/${LIBRARY_PATH//:/:$execroot/}
+function canonicalize_path()
+{
+    new_path=""
+    while IFS=: read -r -d: entry
+    do
+	if [[ -n "$entry" ]]
+	then
+	    new_path="$new_path${new_path:+:}$(realpath -m "$entry")"
```
aherrmann: As far as I can tell `realpath` doesn't exist on macos, so this will probably fail on non nixpkgs macos. This can also be done in bash, similar to [here](https://github.com/tweag/rules_haskell/blob/1f7a51b278f27b0cea8c9aef9b529e39ea735b75/hazel/third_party/cabal2bazel/bzl/cabal_package.bzl#L91)

mboes: macOS+Nix shouldn't be a problem, but bare macOS, yes. A pure bash solution is better. I found this on the Internet:

```
realpath() {
    [[ $1 = /* ]] && echo "$1" || echo "$PWD/${1#./}"
}
```

which I understand better than the code you link to.

mboes:
@aherrmann PTAL

haskell/private/cabal_wrapper.sh.tpl
aherrmann: The linked code does essentially the same thing, with the difference that it keeps relative paths relative. In that case it matters more, as the outputs of `./configure` are outputs of the configure step and inputs to later build steps, so absolute paths would certainly leak into cache keys. Here it should matter less, as `PATH` should typically not end up in the built library/binary.

haskell/private/cabal_wrapper.sh.tpl
```diff
@@ -12,15 +12,33 @@
...
+	    new_path="$new_path${new_path:+:}$(realpath "$entry")"
+	fi
+    done <<< "${1:-}:"
+    echo $new_path
+}
+
+# Remove any relative entries, because we'll be changing CWD shortly.
+LD_LIBRARY_PATH=$(canonicalize_path $LD_LIBRARY_PATH)
+LIBRARY_PATH=$(canonicalize_path $LIBRARY_PATH)
+PATH=$(canonicalize_path $PATH)
```
aspiwack: I'm a bit surprised that we have to do the absolute name resolution in Bash. I would have thought Bazel would give us tools to do so.

mboes: Any Bazel supplied absolute paths would wreak havoc on the action cache. Nix gets away with this because the store is always in the same location. But the Bazel cache and the action execution sandbox could be anywhere.

aspiwack: Oh! I get it! It could be quite terrible for shared cache too.

That being said, these global names should not be a dependency nor an output of the target, so I'm not sure it would matter for this particular case.

mboes: Right, I meant for any shared cache. In general there will be paths to inputs going in one or more of the above 3 environment variables.

PR #990, opened 2019-07-11
Fix $(location) references to data attribute

aherrmann:
The following kind of `$(location)` expansion failed with an "*expression is not a declared prerequisite of this rule*" error.

```
haskell_library(
    ...
    data = ["//some/genrule:target"],
    compiler_flags = ["-DSOME_DEF=$(rootpath //some/genrule:target)"],
)
```

The reason is that since #959 `expand_location` was called multiple times, iterating over all relevant attributes. However, `expand_location` fails immediately if any reference cannot be resolved. This issue didn't come up so far, because certain attributes (e.g. `tools`) are always included implicitly. However, `data` is not.

This PR changes the code to instead deduplicate the list of targets and call `expand_location` only once. This PR also modifies `//tests/binary-with-data` to serve as a regression test for this issue.

### DIFF
--- a/haskell/private/haskell_impl.bzl
+++ b/haskell/private/haskell_impl.bzl
@@ -146,8 +146,19 @@ def _expand_make_variables(name, ctx, strings):
         ctx.attr.plugins,
         ctx.attr.tools,
     ]
-    for attr in label_attrs:
-        strings = [ctx.expand_location(str, attr) for str in strings]
+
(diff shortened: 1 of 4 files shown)


aherrmann:
CI was failing on Windows and Linux bindist because `//tools/runfiles` depended on `@stackage`. We only need core-packages, so pulling those in via `haskell_toolchain_library` instead should fix the issue.

PR #991, opened 2019-07-11
Suppress noisy "Loaded package environment ..." messages

mboes:
This is a workaround for
https://gitlab.haskell.org/ghc/ghc/merge_requests/1308.

### DIFF
--- a/haskell/toolchain.bzl
+++ b/haskell/toolchain.bzl
@@ -80,7 +80,10 @@ while IFS= read -r line; do compile_flags+=("$line"); done < %s
 while IFS= read -r line; do extra_args+=("$line"); done < %s
 while IFS= read -r line; do param_file_args+=("$line"); done < %s
 
-"${compile_flags[@]}" "${extra_args[@]}" ${param_file_args+"${param_file_args[@]}"}
+# XXX Workaround https://gitlab.haskell.org/ghc/ghc/merge_requests/1308.
+set -o pipefail
+"${compile_flags[@]}" "${extra_args[@]}" ${param_file_args+"${param_file_args[@]}"} 2>&1 \
+  | while IFS= read -r line; do [[ $line =~ ^Loaded ]] || echo "$line"; done >&2
 """ % (compile_flags_file.path, extra_args_file.path, params_file_src)
 
     ghc_wrapper_name = "ghc_wrapper_%s_%s" % (hs.name, mnemonic)


haskell/toolchain.bzl
```diff
@@ -80,7 +80,11 @@ while IFS= read -r line; do compile_flags+=("$line"); done < %s
 while IFS= read -r line; do extra_args+=("$line"); done < %s
 while IFS= read -r line; do param_file_args+=("$line"); done < %s
 
-"${compile_flags[@]}" "${extra_args[@]}" ${param_file_args+"${param_file_args[@]}"}
+# XXX Workaround https://gitlab.haskell.org/ghc/ghc/merge_requests/1308.
+output=$("${compile_flags[@]}" "${extra_args[@]}" ${param_file_args+"${param_file_args[@]}"} 2>&1)
+rc=$?
+echo "$output" | while IFS= read -r line; do [[ $line =~ ^Loaded ]] || echo "$line"; done
```
aherrmann:
```suggestion
echo "$output" | while IFS= read -r line; do [[ $line =~ ^Loaded ]] || echo "$line" >&2; done
```
To keep `bazel build`'s stdout reasonably parseable.

haskell/toolchain.bzl
```diff
@@ -80,7 +80,11 @@ while IFS= read -r line; do compile_flags+=("$line"); done < %s
 while IFS= read -r line; do extra_args+=("$line"); done < %s
 while IFS= read -r line; do param_file_args+=("$line"); done < %s
 
-"${compile_flags[@]}" "${extra_args[@]}" ${param_file_args+"${param_file_args[@]}"}
+# XXX Workaround https://gitlab.haskell.org/ghc/ghc/merge_requests/1308.
+output=$("${compile_flags[@]}" "${extra_args[@]}" ${param_file_args+"${param_file_args[@]}"} 2>&1)
```
aherrmann: I think we can avoid storing the output in a variable and instead stream it right away.

```suggestion
set -o pipefail
"${compile_flags[@]}" "${extra_args[@]}" ${param_file_args+"${param_file_args[@]}"} 2>&1 \
  | while IFS= read -r line; do [[ $line =~ ^Loaded ]] || echo "$line"; done >&2
```

`set -o pipefail` forwards the exit code if GHC fails, so the `rc=$?` below becomes unnecessary.

mboes:
Great suggestions, thanks.

haskell/toolchain.bzl
```diff
@@ -80,7 +80,13 @@ while IFS= read -r line; do compile_flags+=("$line"); done < %s
...
 while IFS= read -r line; do param_file_args+=("$line"); done < %s
 
-"${compile_flags[@]}" "${extra_args[@]}" ${param_file_args+"${param_file_args[@]}"}
+# XXX Workaround https://gitlab.haskell.org/ghc/ghc/merge_requests/1308.
+set -o pipefail
+"${compile_flags[@]}" "${extra_args[@]}" ${param_file_args+"${param_file_args[@]}"} 2>&1 \
+  | while IFS= read -r line; do [[ $line =~ ^Loaded ]] || echo "$line"; done >&2
+rc=$?
+echo "$output" | while IFS= read -r line; do [[ $line =~ ^Loaded ]] || echo "$line" >&2; done
+exit $rc
```
aherrmann: Those three lines then become redundant. (I don't know how to write multi-line suggestions on GH UI).

mboes: Yeah I just noticed this was causing extraneous whitespace in the output. Fixed in the latest push.

PR #1000, opened 2019-07-21
Bazel rule deployment guidelines conformance

mboes:
This PR changes the workspace name, from `@io_tweag_rules_haskell` to
`@rules_haskell`. The new name is to be consistent with the Bazel rule
deployment guidelines. The guidelines about naming have changed. See
https://docs.bazel.build/versions/master/skylark/deploying.html. This
is a breaking change, because macros often generate BUILD files that
assume the repository has a certain name in the WORKSPACE file. But
it's also going to make importing haskell rules BUILD file after BUILD
file a fair bit more pleasant.

Other changes, also for compliance with the guidelines:

* Users should import from `haskell/defs.bzl` rather than
  `haskell/haskell.bzl` (the latter is still possible but deprecated).
* Rules that require third-party code that does not ship with GHC are
  not re-exported from `haskell/defs.bzl`, unlike what was the case
  with `haskell/haskell.bzl`. This concerns doctest and protobuf
  rules.
* workspace rules are exposed in `haskell/repositories.bzl`, with new
  names.

These other changes are backwards compatible.


CHANGELOG.md
```diff
@@ -10,11 +10,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
...
-
   0.27 is a LTS release, which means upstream guarantees all new
   releases are backwards-compatible to it for 3 months. See the [Bazel
   Stability](https://blog.bazel.build/2019/06/06/Bazel-Semantic-Versioning.html)
   blog post for more information.
+* The repository name has changed, to follow the
+  new [Bazel rule guidelines][bazel-rule-guidelines]. It was
+  previously called `@io_tweag_rules_haskell`. It is now called
+  `@rules_haskell`. You should adapt your `WORKSPACE` file to match
+  this.
```
aherrmann: I think this should also mention `--host_platform` in `.bazelrc`/`.bazelrc.local` explicitly. Many `grep` tools ignore dotfiles by default so it's easy to miss this and get confusing error messages.

PR #1001, opened 2019-07-22
Pass -hide-all-packages to ghci in wrapper script

jcpetruzza:
If ghc comes with a rich package-db (e.g., if provisioned using using nixpkgs's `ghcWithPackages`), then we need to ensure they are not in scope, or we may get package import conflicts, etc.


aherrmann:
@jcpetruzza Can you rebase on latest rules_haskell master? I think that should fix CI.

jcpetruzza:
Sorry, didn't notice you guys had rebased the branch already!

PR #1002, opened 2019-07-23
 Combined cc_wrapper for all OSes

aherrmann:
Replaces the macOS specific CC wrapper by one applicable to all OSes

- Shortens library search paths to stay below maximum path length on Windows.
    GHC generates library search paths that contain redundant up-level references (`..`). This can exceed the maximum path length on Windows, which will cause linking failures. This wrapper shortens library search paths to avoid that issue.
- Shortens rpaths and load commands on macOS.
    The rpaths and load commands generated by GHC and Bazel can quickly exceed the MACH-O header size limit on macOS. This wrapper shortens and combines rpaths and load commands to avoid exceeding that limit.
- Finds .so files if only .dylib are searched on macOS.
    Bazel's cc_library will generate .so files for dynamic libraries even on macOS. GHC strictly expects .dylib files on macOS. This wrapper hooks into gcc's --print-file-name feature to work around this mismatch in file extension.

The CC wrapper is also used for REPL and runghc targets so that these also don't require symlinks for dynamic libraries on macOS.

Closes https://github.com/tweag/rules_haskell/issues/958


aherrmann:
The CI failure on Windows is due to timeout.

Tracing the build of `@haskell_vector//:lib` on a local Windows machine using [procmon](https://docs.microsoft.com/en-us/sysinternals/downloads/procmon) (Windows defender disabled) shows the following kind of process tree (simplified for readability).

```
cc_wrapper.exe
├── Conhost.exe
└── python.exe        (command: python.exe .../cc_wrapper.zip ...)          (time: 10:05:04 - 10:05:13)
    └── python.exe    (command: python.exe .../cc_wrapper-source.py ...)    (time: 10:05:11 - 10:05:11)
        └── gcc.exe
            ...
```

The first call to `python.exe` is given a zip archive of the `cc_wrapper` `py_binary` target and unpacks it into a temporary directory. The second call to `python.exe` calls the unpacked main python module. The timings show that the unzipping and clean-up of the temporary files take up the vast majority of the ~9s execution time. This overhead is incurred on every single call to `cc_wrapper` that GHC performs. On CI it adds up to exceed the timeout.

aherrmann:
A quick test shows that a `cc_wrapper` written in C++ and built as `cc_binary` does not incur this overhead. Rewriting the `cc_wrapper` in C++ would be a possible work around.

Profpatsch:
Do you plan on doing the rewrite and opening a new PR once that is done? Or do you want to push the rewrite to this PR?

aherrmann:
Happy to do it in a separate PR. I'll close this for now. I'll be gone the next two weeks, so I won't be able to resolve this before I'm back.

Issue #704, opened 2019-02-13
Cannot run tests with bindist using shell.nix missing lib gmp


nmattia:
Build fails when using the ghc bindist if libgmp is not installed on the system. Even adding it to the nix-shell does not help:

```shell
[nix-shell:~/projects/tweag/rules_haskell]$ echo $LD_LIBRARY_PATH 
/nix/store/rd7n0v6mymvyqbw3d307caqc23c8kc71-gmp-6.1.2/lib
```

The link cannot find libgmp during build: 

``` shell
[nix-shell:~/projects/tweag/rules_haskell]$ bazel test //tests/binary-with-lib-dynamic/...
...
ERROR: /home/nicolas/projects/tweag/rules_haskell/tests/binary-with-lib-dynamic/BUILD:9:1: error executing shell command: '/nix/store/pawkx6zfnj59zvpa0hkrrqin9f9iwika-bash/bin/bash -c
        export PATH=${PATH:-} # otherwise GCC fails on Windows

        # this is equivalent to 'readarray'. We do not use 'readarray' ...' failed (Exit 1) bash failed: error executing command /nix/store/pawkx6zfnj59zvpa0hkrrqin9f9iwika-bash/bin/bash -c ... (remaining 1 argument(s) skipped)

/nix/store/3xwc1ip20b0p68sxqbjjll0va4pv5hbv-binutils-2.30/bin/ld.gold: error: cannot find -lgmp
collect2: error: ld returned 1 exit status
`gcc' failed in phase `Linker'. (Exit code: 1)
FAILED: Build did NOT complete successfully
```

Passing `action_env` and `test_env` does not help:

``` shell
[nix-shell:~/projects/tweag/rules_haskell]$ bazel test //tests/binary-with-lib-dynamic/... --action_env=LD_LIBRARY_PATH= --test_env=LD_LIBRARY_PATH
<same error>
FAILED: Build did NOT complete successfully
[nix-shell:~/projects/tweag/rules_haskell]$ bazel test //tests/binary-with-lib-dynamic/... --action_env=LD_LIBRARY_PATH=$LD_LIBRARY_PATH --test_env=LD_LIBRARY_PATH=$LD_LIBRARY_PATH
<same error>
FAILED: Build did NOT complete successfully
```

Specifying `--linkopt=-L/path/to/lib` fixes some builds:

```
[...]

PR #1003, opened 2019-07-24
BuildKite support

mboes:
Migrate Nixpkgs Linux job, which is flaky on CircleCI due to
underpowered resource classes, to BuildKite.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,14 @@
+steps:
+  - command: |
+      . /var/lib/buildkite-agent/.nix-profile/etc/profile.d/nix.sh
+      echo "build:ci --host_platform=@rules_haskell//haskell/platforms:linux_x86_64_nixpkgs" > .bazelrc.local
+      nix-shell --arg docTools false --pure --run \
+          'bazel build --config ci //tests:run-tests'
+      # TODO(Profpatsch) re-add a nixpkgs startup script
+      # and enable this test again
+      nix-shell --arg docTools false --pure --run \
+          './bazel-ci-bin/tests/run-tests --skip "/startup script/"'
+      nix-shell --arg docTools false --pure --run \
+          'bazel coverage //tests/... --config ci --build_tag_filters "coverage-compatible" --test_tag_filters "coverage-compatible" --test_output=all'
+    label: "Run tests"
+    timeout: 30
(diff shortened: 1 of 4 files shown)


.buildkite/pipeline.yml
```diff
@@ -0,0 +1,22 @@
+steps:
+  - command: |
+      . /var/lib/buildkite-agent/.nix-profile/etc/profile.d/nix.sh
```
flokli: This is very specific to how our agent is set up - I'd rather see us add a `nix=true` tag to all nix-enabled agents (which set this locally in an environment hook), and then [require that tag to be present for that specific step](https://buildkite.com/docs/agent/v3/cli-start#agent-targeting).

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,22 @@
...
+      nix-shell --arg docTools false --pure --run \
+          'bazel coverage //tests/... --config ci --build_tag_filters "coverage-compatible" --test_tag_filters "coverage-compatible" --test_output=all'
+    label: "Run tests (Nixpkgs)"
+    timeout: 30
+  - command: |
+      echo "common:ci --build_tag_filters -requires_hackage,-requires_lz4,-requires_zlib,-requires_doctest,-requires_c2hs,-requires_threaded_rts,-dont_test_with_bindist" > .bazelrc.local
+      # XXX: See .bazelrc [backward compatible options] for the the rational behind this flag
+      echo "build --incompatible_use_python_toolchains=false" >> .bazelrc.local
+      bazel build --config ci //tests/...
+      ./tests/run-start-script.sh
```
flokli: While tinkering with sourcing the nix profile inside the agent hooks, I noticed this step cannot run when the nix profile is sourced before, due to `ld.gold` not finding `gmp`. Is this a known issue / expected?

https://buildkite.com/tweag-1/rules-haskell/builds/29#b7b15f5e-178c-4bc4-b891-6e2bc2332cb7/21-174

.buildkite/pipeline.yml
mboes: I don't see this as a problem for now: the agent such that it is setup currently is temporary. We'll want to reprovision it as completely immutable infrastructure (not in scope for this PR).

.buildkite/pipeline.yml
mboes: Yes, I noticed this too. I'm not sure why this is. But it seems fair to keep Nix and non-nix environments separate.

flokli: Can you open an issue about this, so we can follow-up on it later?

flokli: so there's already https://github.com/tweag/rules_haskell/issues/704 about that, which we closed as WONTFIX. We might want to revisit that, assuming the CI agents run on NixOS, and we want to test the bindist feature.

flokli:
@mboes this would need a force-merge, as CircleCI fails (because we remove its pipeline configuration)

PR #1004, opened 2019-07-24
More robust Stack version detection

### LINKED ISSUE #998: Stack version checking seems to be incorrect in general
I've got a stack version (installed by Ubuntu) where

```
$ stack --version
Version 1.9.3
```

And this fails the version-parsing command in `cabal.bzl`:

```
line 494, in _compute_dependency_graph
		int(exec_result.stdout.split(" ")[0]...])
invalid literal for int() with base 10: "Version": For input string: "Version"
```

mboes:
Stack version strings are sometimes prefixed by the word "Version",
sometimes not.

Fixes #998

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -490,7 +490,14 @@ def _compute_dependency_graph(repository_ctx, snapshot, versioned_packages, unve
     if not stack_cmd:
         fail("Cannot find stack command in your PATH.")
     exec_result = _execute_or_fail_loudly(repository_ctx, [stack_cmd, "--version"])
-    stack_major_version = int(exec_result.stdout.split(" ")[0].split(".")[0])
+    stack_version_words = exec_result.stdout.split(" ")
+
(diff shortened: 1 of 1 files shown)


aspiwack:
I tried to test this, but I got an error message:

```
ERROR: Failed to load Starlark extension '@rules_haskell//haskell:providers.bzl'.
Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
The following chain of repository dependencies lead to the missing definition.
 - @rules_haskell
This could either mean you have to add the '@rules_haskell' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
ERROR: cycles detected during target parsing
```

(it otherwise looks like it should work)

mboes:
This looks like a bug in your WORKSPACE file. My guess: you haven't adjusted your WORKSPACE file like the changelog says.

PR #1008, opened 2019-07-25
Allow reexporting a package from a `haskell_library`

regnat:
This is the re-opening of #757 (because github doesn't allow me to directly reopen it)

I eventually found the time to finish this.

I didn't reuse the `exports` field since I find it a bit disturbing to suddenly change the meaning of an existing attribute, so the field is called `package_reexports` instead.

So the semantics now are that if a rule `foo` has `package_reexports = [ "bar ]`, then

- `foo` will depend on `bar` (as if it was in its `deps` field)
- Any rule that (directly)  depends on `foo` will see `bar` as if it was in its dependencies
- Running `foo@repl` will load both the modules of `foo` and `bar`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 * The `haskell_register_toolchains()` macro is deprecated. Use
   `rules_haskell_toolchains()` from `haskell/repositories.bzl`
   instead.
-- The `exports` attribute has been renamed to `module_exports`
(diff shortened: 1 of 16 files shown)


mboes:
> I didn't reuse the exports field since I find it a bit disturbing to suddenly change the meaning of an existing attribute, so the field is called package_reexports instead.

As we discussed, no one is using this field, so go ahead and reuse it.

regnat:
I've renamed the old `exports` attribute to `module_exports` and that one to `exports`

PR #1011, opened 2019-07-25
Download Stack if possible

mboes:
If we don't find a recent enough Stack on the PATH, then just download
one locally and use that. Since this is happening in a workspace rule,
we have no choice but to use prebuilt binaries. This behaviour brings
us closer to what rules_jvm_external does, which likewise downloads
a tool used to compute the dependency graph.

NixOS users will of course need to continue providing Stack in the
shell, because the Stack bindists don't work there.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -474,6 +474,60 @@ _CORE_PACKAGES = [
     "xhtml",
 ]
 
+_STACK_DEFAULT_VERSION = "2.1.3"
+
+# Only ever need one version, but use same structure as for GHC bindists.
+_STACK_BINDISTS = \
+    {
+        "2.1.3": {
+            "freebsd-x86_64": (
+                "https://github.com/commercialhaskell/stack/releases/download/v2.1.3/stack-2.1.3-freebsd-x86_64.tar.gz",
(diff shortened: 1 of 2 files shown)


haskell/cabal.bzl
```diff
@@ -671,53 +698,118 @@ stack_snapshot = repository_rule(
...
+        else:
+            stack_major_version = int(stack_version_words[0].split(".")[0])
+        if stack_major_version < 2:
+            print("Stack version not recent enough. Downloading a newer version...")
+        else:
+            repository_ctx.symlink(stack_cmd, "stack")
+            return
+
+    # If we can't find Stack, download it.
+    (os, arch) = _get_platform(repository_ctx)
```
aherrmann: Should we detect NixOS and fail gracefully in that case? The bindist will probably fail with the usual and quite misleading "no such file or directory" error.

Profpatsch: Yeah, we should put this behind a flag and only enable it for bindists

mboes: I don't know how to reliably detect NixOS. Maybe we could improve the error message in this case with hints? Being on NixOS will be the common case for why the binary failed...

mboes:
I improved the error message for NixOS users, but also made it possible for Nix users to supply Stack via `nixpkgs_package`, so that adding Stack to the shell is no longer necessary even there.

PR #1012, opened 2019-07-26
Use a disk cache during CI

mboes:


mboes:
This setting doesn't seem to have any effect on BuildKite build times. @flokli @aherrmann any ideas?

flokli:
I can confirm the bazel cache was created on the agent:

```
root@ubuntu-32gb-nbg1-1:~# du -sh /tmp/bazel-*
312M	/tmp/bazel-disk-cache
```

However, it doesn't seem to be used for example to avoid having to redownload `rules_haskell`:
```
Fetching @rules_haskell_ghc_linux_amd64; fetching 158s
```

aherrmann:
I ran a test on my machine which may well explain this behavior. Within the same working directory I ran `bazel build //tests/binary-with-lib` in the following configurations:

1. bindist, i.e. no Nix in PATH and no `--host-platform` set.
2. nixpkgs, i.e. Nix in PATH and `--host-platform` set to nixpkgs
3. bindist again.
4. nixpkgs again.

The build 1. took a long time because it had to fetch the GHC bindist. The build 2. took a while as well. But, builds 3. and 4. were both much faster even without `--disk_cache`.

If the worker state is persisted between builds then this may be the same phenomenon.

flokli:
But this would mean, fetched external things are not persisted in the disk cache? Is that a upstream bazel bug/featue?

mboes:
The disk cache is an action cache, not a repository cache, AFAIU. Hm... which indeed, would explain why we're not seeing a speedup.

I have a vague recollection that an external repository cache was added recently to Bazel. Rings a bell?

aherrmann:
@mboes There was a discussion on [bazel-discuss](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/bazel-discuss/V0YFIrZAK7Q/9Je8QUmtCwAJ) and a [proposal](https://docs.google.com/document/d/1aagKnM8kRmMvLqSXasShsI-NYOBZYtgNrbYWLPwBoBg/edit). I don't know if it's implemented, yet. Are you referring to that?

mboes:
Yes. Now, we understand why the bindist isn't cached, but I still don't understand why that in turn causes everything (including protobuf) to be rebuilt despite the disk cache. The two different outcomes of the CC autoconfiguration do lead to two different absolute paths being passed as inputs to GHC. But I would expect both alternatives to be cached, so that in a three build sequence such as bindist/nixpkgs/bindist, the second bindist build runs faster than the first one.

mboes:
Build times on Linux are down to ~4 minutes (both nixpkgs and bindist jobs combined). And we solve the runaway cache size problem to boot.

PR #1013, opened 2019-07-27
Run more tests when using bindists

### LINKED ISSUE #1005: Haddock output differs when using ghc bindist
**Describe the bug**

The output of our Haddock test is different when using GHC binary distributions, for reasons that escape me: https://buildkite.com/tweag-1/rules-haskell/builds/36#a5128231-3be9-43d7-b7bd-7a10f5af4be8

**Expected behavior**

The output shoud be the same, irrespective of where the compiler was obtained.

mboes:
Our test suite is only run in full for the Nixpkgs job. This PR makes
the bindist job run more tests. This uncovered three bugs:

* one bug in stack_install, where certain combinations of packages
  would cause the unpack phase to fail,
* two Haddock related bugs (one in `pkgdb_to_bzl.py` and the other in
  the HaskellHaddock action).

These bugs are fixed as part of this PR.

Fixes #1005

### DIFF
--- a/haskell/nixpkgs.bzl
+++ b/haskell/nixpkgs.bzl
@@ -37,7 +37,7 @@ def _ghc_nixpkgs_haskell_toolchain_impl(repository_ctx):
     # and asks the parent process to generate the associated bazel symlink
     for line in result.stdout.split("\n"):
         if line.startswith("#SYMLINK:"):
-            _, name, path = line.split(" ")
+            _, path, name = line.split(" ")
             repository_ctx.symlink(path, name)
 
     repository_ctx.file(
--- a/haskell/private/pkgdb_to_bzl.py
+++ b/haskell/private/pkgdb_to_bzl.py
@@ -133,39 +133,30 @@ for conf in glob.glob(os.path.join(topdir, "package.conf.d", "*.conf")):
     # this does not work, we generate a path in the workspace and
     # output a SYMLINK information for the parent process
 
-    symlinks = []
-
     # first, try to get a path within the package
     # We check if the file exists because cabal will unconditionally
     # generate the database entry even if no haddock was generated.
-    if pkg.haddock_html and os.path.exists(pkg.haddock_html):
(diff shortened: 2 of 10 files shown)


mboes:
Requesting a review from @guibou concerning the two Haddock fixes.

haskell/nixpkgs.bzl
```diff
@@ -37,7 +37,7 @@ def _ghc_nixpkgs_haskell_toolchain_impl(repository_ctx):
     # and asks the parent process to generate the associated bazel symlink
     for line in result.stdout.split("\n"):
         if line.startswith("#SYMLINK:"):
-            _, name, path = line.split(" ")
+            _, path, name = line.split(" ")
```
Profpatsch: Was this the bug?

mboes: No just a code simplification. The previous code swapped the order in one place, only to unswap here.

haskell/private/pkgdb_to_bzl.py
```diff
@@ -133,39 +133,30 @@ def hs_library_pattern(name, mode = "static", config = ""):
...
-
-    haddock_interfaces = []
+            output.append("#SYMLINK: {} {}".format(pkg.haddock_html, haddock_html))
 
     # If there is many interfaces, we give them a number
     interface_id = 0
-
+    haddock_interfaces = []
     for interface_path in pkg.haddock_interfaces:
-        if os.path.exists(interface_path):
```
guibou: You mostly removed the `os.path.exists` call. Could you elaborate why?

mboes: Because I believe that is an incorrect test, which does not match the semantics of `path_to_label()`. The latter returns `None` if `$topdir` was not found in the input. In the case of bindists, we do have a `$topdir` to expand, and the given path does not exist unless we expand the path. The path always exists because it's somewhere underneath the `$topdir`.

Issue #971, opened 2019-06-28
Startup script: have both bindist and nixpkgs in the workspace


aspiwack:
The onboarding process is harder than it need be. In particular, it is finicky to make the nixpkgs version of the toolchain works. As the startup script only builds a workspace with the ghc_bindist option.

It would presumably be better, if the generated workspace had both the ghc_bindist toolchain and the nixpkgs toolchain registered. Together with the instruction to switch between these using a platform flag (which is currently necessary for the nixpkgs toolchain anyway, so it wouldn't be much of a change for the end user, it simply saves setup time).

Profpatsch:
Duplicate of https://github.com/tweag/rules_haskell/issues/784 I think

aspiwack:
#784  seems to be suggesting a different solution to the same problem. Feel free to merge the two issues as you feel like.

Profpatsch:
@aspiwack Do you think autodiscovery is a good idea?

aspiwack:
No idea. What does it entail?

PR #1018, opened 2019-07-30
Start script nixpkgs support

Profpatsch:
Best reviewed comment-by-comment.

* Make the `start` script support nixpkgs again
* Test both modes on CI
* Improve CI a bit

Closes https://github.com/tweag/rules_haskell/issues/971

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ If you are on NixOS, skip to the [Nixpkgs](#Nixpkgs) section.
 In a fresh directory, run:
 
 ```console
-$ sh <(curl https://haskell.build/start) --use-bindists
+$ curl https://haskell.build/start | sh
 ```
 
 This will generate initial `WORKSPACE` and `BUILD` files for you. See the
@@ -51,8 +51,7 @@ syntax [here][bazel-cli]. Common [commands][bazel-cli-commands] are
 
 This rule set supports using [Nixpkgs][nixpkgs] to provision your GHC
 toolchain and to fetch hackage packages from there. To create your
-project, instead of passing `--use-bindists`, pass `--use-nix`, like
-so:
+project, pass `--use-nix`, like so:
 
 ```console
 $ sh <(curl https://haskell.build/start) --use-nix
@@ -62,8 +61,8 @@ This generates the same files as above, but uses `nixpkgs` to
 provision GHC.
 
 If you are on NixOS, this is the only way to set up your project,
-because the GHC toolchain provisioned through `--use-bindists` cannot
-be executed on NixOS.
+because the GHC toolchain provisioned through binary distributions
+cannot be executed on NixOS.
 
 [bazel-cli-commands]: https://docs.bazel.build/versions/master/command-line-reference.html#commands
 [nixpkgs]: https://nixos.org/nixpkgs/
--- a/README.md
+++ b/README.md
@@ -24,16 +24,18 @@ The full reference documentation for rules is at https://haskell.build.
 
 You'll need [Bazel >= 0.27][bazel-getting-started] installed.
 
+If you are on NixOS, skip to the [Nixpkgs](#Nixpkgs) section.
+
 ### The easy way
 
 In a fresh directory, run:
 
 ```console
-$ curl https://haskell.build/start | sh
+$ sh <(curl https://haskell.build/start) --use-bindists
 ```
 
 This will generate initial `WORKSPACE` and `BUILD` files for you. See the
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -29,11 +29,11 @@ You'll need [Bazel >= 0.27][bazel-getting-started] installed.
 In a fresh directory, run:
 
 ```console
-$ curl https://haskell.build/start | sh
+$ sh <(curl https://haskell.build/start) --use-bindists
```
mboes: The common case should remain as simple and as memorable as possible. The common is bindists, so that should be the default, if no flag is passed. That would allow keeping the instructions on this line identical to what they were.

We can mention in text displayed to the user that there is also a `--use-nixpkgs` flag that they might want to use, and how to use it.

README.md
```diff
@@ -47,45 +47,24 @@ syntax [here][bazel-cli]. Common [commands][bazel-cli-commands] are
 
 ### Nixpkgs
 
-This rule set supports [Nixpkgs][nixpkgs]. If you are on NixOS, or if
-you are using Nixpkgs on your project, consider passing the following
-argument on the command-line to select a Nixpkgs-based toolchain for
-the build:
+This rule set supports using [Nixpkgs][nixpkgs] to provision your GHC
+toolchain and to fetch hackage packages from there. To create your
+project, instead of passing `--use-bindists`, pass `--use-nix`, like
```
mboes: See above: `--use-bindists` should be allowed but not necessary.

README.md
```diff
@@ -47,45 +47,24 @@ syntax [here][bazel-cli]. Common [commands][bazel-cli-commands] are
...
 
-```
-$ bazel build --host_platform=@rules_haskell//haskell/platforms:linux_x86_64_nixpkgs //...
+```console
+$ sh <(curl https://haskell.build/start) --use-nix
 ```
 
-See [below](#saving-common-command-line-flags-to-a-file) to
-permanently set that flag.
+This generates the same files as above, but uses a `nixpkgs` backend.
```
mboes: What is a "backend"? It should at the very least be defined, but it's not a term we use anywhere else, so best avoid it.

start
```diff
@@ -11,43 +11,127 @@ MAX_BAZEL_MINOR=28
...
+    "@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl",
+    "nixpkgs_git_repository",
+)
+
+# Fetch a version of nixpkgs from GitHub.
+# For more information see the documentation of rules_nixpkgs at
+# https://github.com/tweag/rules_nixpkgs/blob/master/README.md
+nixpkgs_git_repository(
+    name = "nixpkgs",
+    # TODO: fix this to a git commit hash and insert sha256
```
mboes: Having a sha256: yes. But for the revision, using a tag works just as well: tags are assigned immutably by upstream, right?

Profpatsch: Tags are not immutable by default, and I wouldn’t be sure they are never modified (even if by accident). Using the rev hash eliminates a source of bugs.

README.md
Profpatsch: https://github.com/tweag/rules_haskell/pull/1018/commits/3a64a062d2cd87bdf551dbff2c6e5d5864bb8f2c

Profpatsch:
Implemented suggestions in the last two commits, PTAL.

README.md
```diff
@@ -47,45 +47,24 @@ syntax [here][bazel-cli]. Common [commands][bazel-cli-commands] are
...
+```console
+$ sh <(curl https://haskell.build/start) --use-nix
 ```
 
-See [below](#saving-common-command-line-flags-to-a-file) to
-permanently set that flag.
+This generates the same files as above, but uses `nixpkgs` to
+provision GHC.
+
+If you are on NixOS, this is the only way to set up your project,
```
mboes: See @bethaitman's Haskell documentation talk: don't put important information (like something being required for NixOS) at the end of the section. Start with that.

start
mboes: Tags are good enough in this case and are better documentation that won't go out-of-date. Especially when coming from  a trusted source like the Nixpkgs release managers.

mboes: Also, this is putting a `TODO` in every user's `WORKSPACE` file when the currently hardcoded value is a fine default.

README.md
Profpatsch: Added a sentence at the top of `Setup`.

Profpatsch:
Once you approve, I’ll squash the fixup commits and add the merge label.

PR #1022, opened 2019-07-30
Hardcode location of ar and strip on Darwin

### LINKED ISSUE #1019: macOS bindist project cannot find 'ar' when using 'stack_snapshot'
**Describe the bug**
`rules_haskell` fails to find the `ar` program, even when it is available on the `PATH`, when building projects using the `stack_snapshot` rule with a toolchain using GHC bindists.

This is using 8eee207eb01a9d2c369bc666fc50dfb37d95934a to get a version of `rules_haskell` that works with macOS bindists.

**To Reproduce**
[See this repository for a reproducible example](https://github.com/jkachmar/bazel_haskell_bindist_sandbox/tree/15ad1575df753abffbf7576f7cfb910c456d8217)

**Expected behavior**
`rules_haskell` should successfully build libraries using the `stack_snapshot` rules on macOS with a toolchain that uses GHC bindists.

**Environment**
 - OS name + version: macOS 10.14.5
 - Bazel version: 0.28.1
 - Version of the rules: 8eee207eb01a9d2c369bc666fc50dfb37d95934a

mboes:
The reason we do this is because `CcToolchainInfo.ar_executable()`
gives us a relative path, but a relative path we cannot include in the
sandbox, because the value is a string, not a path that we can add to
the inputs of an action. This function is deprecated anyways. The
current fix isn't the proper fix, which requires the use of some
different API that may or may not already be available. So for now we
hardcode the absolute path, which in fact we already did previously in
certain cases when on Darwin.

Fixes #1019

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -84,6 +84,8 @@ def _prepare_cabal_inputs(hs, cc, dep_info, cc_info, tool_inputs, tool_input_man
 
     (ghci_extra_libs, env) = get_ghci_extra_libs(hs, cc_info)
     env["PATH"] = _make_path(hs, tool_inputs)
+    if hs.toolchain.is_darwin:
+        env["SDKROOT"] = "macosx"  # See haskell/private/actions/link.bzl
 
     # TODO Instantiating this template could be done just once in the
     # toolchain rule.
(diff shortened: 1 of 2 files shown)


mboes:
I just force pushed a different fix, which is, I believe, the proper fix.

PR #1024, opened 2019-07-31
Add support for Bazel persistent workers

ulysses4ever:
This is the first (working) step with several shortcomings:

1. The `worker` binary is assumed to be in `~/bin`.
2. No incremental builds yet.
3. No tests or documentation.

First should be fixed ASAP, I imagine. I'd appreciate a bit of advice on this particular matter.

Third should be fixed in the due course.

On the positive side: it should support both standard (non-worker) mode as well as the worker one.

### DIFF
--- a/haskell/nixpkgs.bzl
+++ b/haskell/nixpkgs.bzl
@@ -83,6 +83,7 @@ haskell_toolchain(
     # hack in Nixpkgs.
     {locale_archive_arg}
     locale = {locale},
+    use_worker = {use_worker},
 )
         """.format(
             toolchain_libraries = toolchain_libraries,
@@ -95,6 +96,7 @@ haskell_toolchain(
             repl_ghci_args = repository_ctx.attr.repl_ghci_args,
             locale_archive_arg = "locale_archive = {},".format(repr(locale_archive)) if locale_archive else "",
             locale = repr(repository_ctx.attr.locale),
+            use_worker = repository_ctx.attr.use_worker,
         ),
     )
 
@@ -120,6 +122,7 @@ _ghc_nixpkgs_haskell_toolchain = repository_rule(
         "locale": attr.string(
             default = "en_US.UTF-8",
         ),
+        "use_worker": attr.bool(),
     },
 )
 
@@ -173,10 +176,15 @@ def haskell_register_ghc_nixpkgs(
         locale = None,
         repositories = {},
         repository = None,
-        nix_file_content = None):
+        nix_file_content = None,
+        use_worker = False):
     """Register a package from Nixpkgs as a toolchain.
 
-    Toolchains can be used to compile Haskell code. To have this
+    Args:
+      use_worker: This is a part of experimental support for the persistent worker mode.
+        It is not intended for production usage, yet.
+
+      Toolchains can be used to compile Haskell code. To have this
     toolchain selected during [toolchain
     resolution][toolchain-resolution], set a host platform that
     includes the `@rules_haskell//haskell/platforms:nixpkgs`
@@ -231,6 +239,7 @@ def haskell_register_ghc_nixpkgs(
         repl_ghci_args = repl_ghci_args,
         locale_archive = locale_archive,
         locale = locale,
+        use_worker = use_worker,
     )
 
     # toolchain definition.
--- a/haskell/BUILD.bazel
+++ b/haskell/BUILD.bazel
@@ -33,6 +33,12 @@ py_binary(
     visibility = ["//visibility:public"],
 )
 
+sh_binary(
+    name = "worker",
+    srcs = ["private/worker.sh"],
+    visibility = ["//visibility:public"],
+)
+
 py_binary(
     name = "version_macros",
     srcs = ["private/version_macros.py"],
--- a/haskell/defs.bzl
+++ b/haskell/defs.bzl
@@ -87,6 +87,11 @@ _haskell_common_attrs = {
     "_cc_toolchain": attr.label(
         default = Label("@bazel_tools//tools/cpp:current_cc_toolchain"),
     ),
+    "_worker": attr.label(
+        executable = True,
+        cfg = "host",
+        default = Label("@rules_haskell//haskell:worker"),
+    ),
 }
 
 def _mk_binary_rule(**kwargs):
--- a/haskell/private/actions/link.bzl
+++ b/haskell/private/actions/link.bzl
@@ -4,7 +4,7 @@ load(":private/packages.bzl", "expose_packages", "pkg_info_to_compile_flags")
 load(":private/pkg_id.bzl", "pkg_id")
 load(":providers.bzl", "create_link_config")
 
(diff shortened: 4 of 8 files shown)


ulysses4ever:
How come CI tries to run tests in the worker mode? @regnat It's doom to fail because the worker is not packaged with rules_haskell yet.

regnat:
@ulysses4ever according to https://blog.bazel.build/2019/06/19/list-strategy.html

> If none of the strategy flags was used, bazel will generate a default list of strategies `remote,worker,sandboxed,local` and, for every action it wants to execute, will pick up the first strategy that can execute it.

So if it is available, `worker` is privileged over `sandboxed` or `local`. That kinda sucks for us because we can specify a custom `--spawn_strategy` in `rules_haskell`'s `.bazelrc`, but all our dependents won't have this and thus will have a weird failure.

Do you think it's possible to have the worker be opt-in in the toolchain declaration? Something like adding a `use_worker :: Bool` flag to it

regnat:
Actually running it explicitely without the worker also fails for a different reason:

```
▶ bazel test //... --spawn_strategy sandboxed
INFO: Invocation ID: 724cf41e-887b-4ee0-8a0f-e7a5e1a77f78
WARNING: /home/regnat/.cache/bazel/_bazel_regnat/9b6df921bb6e07dc34fcea21f9031196/external/zlib.dev/BUILD:9:1: in linkstatic attribute of cc_library rule @zlib.dev//:zlib: setting 'linkstatic=1' is re
commended if there are no object files
INFO: Analyzed 290 targets (0 packages loaded, 0 targets configured).
INFO: Found 196 targets and 94 test targets...
ERROR: /home/regnat/Documents/taff/tweag/bazel/rules_haskell/tests/binary-custom-main/BUILD.bazel:8:1: HaskellBuildBinary //tests/binary-custom-main:binary-custom-main failed (Exit 1) worker failed: e
rror executing command bazel-out/host/bin/external/rules_haskell/haskell/worker bazel-out/k8-fastbuild/bin/tests/binary-custom-main/compile_flags_binary-custom-main_HaskellBuildBinary ... (remaining 1
 argument(s) skipped)

src/main/tools/linux-sandbox-pid1.cc:427: "execvp(bazel-out/host/bin/external/rules_haskell/haskell/worker, 0x18dc950)": No such file or directory
```

ulysses4ever:
As discussed on today's meeting with @regnat, the flag should be doable.

Strange that I don't get the same error in `binary-custom-main` as @regnat does. What I get with `bazel test //... --spawn_strategy sandboxed -k` is:

> Executed 35 out of 94 tests: 35 tests pass and 59 fail to build.

And `//tests:test-binary-custom-main PASSED`.

ulysses4ever:
The latest update is not intended to pass through CI yet, but it should be possible to run it locally as shown by @regnat :
```
bazel test //... --spawn_strategy sandboxed -k
```
It passes for me.

ulysses4ever:
I added a toolchain flag to indicate whether we want to use worker. CI should pass now.

ulysses4ever:
Could someone help me decipher CI failures, please? Why CirecleCI fails when downloading nix (I guess?)? And what is `bazel lint`? I cannot run `bazel lint` locally (`Command 'lint' not found.`)

regnat:
> Could someone help me decipher CI failures, please? Why CirecleCI fails when downloading nix (I guess?)? And what is `bazel lint`? I cannot run `bazel lint` locally (`Command 'lint' not found.`)

`bazel lint` is just the name of the test which doesn't map to an existing bazel command.
The actual command run is `bazel run //:buildifier` which checks the formatting of the starlark files.
You can automatically fix it with `bazel run //:buildifier-fix` (all that is in the README with a bit more details).

The Circle failure is… interesting. It's definitely not related to your branch and is probably transient. Hopefully that will solve itself on the next CI run :crossed_fingers: . Or maybe @Profpatsch has an idea

ulysses4ever:
CI is happy :tada: Could someone review, please?

Profpatsch:
Since CI now passes it must have been transient.

haskell/nixpkgs.bzl
```diff
@@ -173,7 +176,8 @@ def haskell_register_ghc_nixpkgs(
         locale = None,
         repositories = {},
         repository = None,
-        nix_file_content = None):
+        nix_file_content = None,
+        use_worker = False):
```
aherrmann: This flag will show up in the [API docs for master](https://api.haskell.build/haskell/nixpkgs.html#haskell_register_ghc_nixpkgs). Best document that this is not ready for use by regular users, yet.

ulysses4ever:
@aherrmann thanks for the review! I addressed your comments. Could you check it out again?

haskell/nixpkgs.bzl
```diff
@@ -173,10 +176,15 @@ def haskell_register_ghc_nixpkgs(
...
     """Register a package from Nixpkgs as a toolchain.
 
-    Toolchains can be used to compile Haskell code. To have this
+    Args:
+      use_worker: This is a part of experimental support for the persistent worker mode.
+        It is not intended for production usage, yet.
+
+      Toolchains can be used to compile Haskell code. To have this
```
aherrmann:
```suggestion
    Toolchains can be used to compile Haskell code. To have this
```

ulysses4ever:
Hmm, I wasn't able to use worker when changed the call to `haskell_register_ghc_nixpkgs` with `use_worker=True` -- the build proceeded as usual in the "non-worker" mode. Let's probably not merge until I figure it out?

ulysses4ever:
Ok, I figured it and now can run one test with worker successfully. But if I do `bazel test //...` I get 
```
ERROR: infinite symlink expansion detected
[start of symlink chain]
/home/artem/Dev/rules_haskell
/home/artem/Dev/rules_haskell/examples/bazel-examples/external/rules_haskell
/home/artem/.cache/bazel/_bazel_artem/9f09c111e94c2a1957e003be5e5101a4/execroot/rules_haskell_examples/external/rules_haskell
/home/artem/.cache/bazel/_bazel_artem/9f09c111e94c2a1957e003be5e5101a4/external/rules_haskell
[end of symlink chain]
ERROR: Failed to get information about path, for examples/bazel-examples/external/rules_haskell, skipping: Infinite symlink expansion
```
Why is that?

ulysses4ever:
My problem looks [like this](https://github.com/bazelbuild/bazel/issues/5763): I reference worker binary with `~` (`$HOME`). But I don't see why 1) I didn't hit it before; 2) it doesn't manifest in singleton tests.

ulysses4ever:
SOmething wrong was with my local copy. Clean checkout works (and tests) as expected: both in standard mode and in the persistent worker mode.

haskell/nixpkgs.bzl
ulysses4ever: Wow, @aherrmann, I missed your comment somehow, I'm terribly sorry! I opened #1044 to fix this.

PR #1028, opened 2019-08-02
Fetch stack_snapshot dependencies from snapshot, not from Hackage

### LINKED ISSUE #1027: `stack_snapshot` dependencies are always fetched from Hackage
In `stack_snapshot`, using a `local_snapshot`, dependencies are always fetched from Hackage. If the dependency has been overriden to a git repository in the custom snapshot, then this will most fail with an error: stack knows that the snapshot contains `foo-x.y.z`, and will attempt to fetch `foo-x.y.z` (inferred from the `.cabal` file in the git repository). In the best case, `foo-x.y.z` doesn't exist in Hackage, and an error is produced, in the worst case, it does and it's a different version which will be fetched!

It appears that `stack unpack foo` will fetch `foo` from the snapshot, whereas `stack unpack foo-x.y.z` will fetch `foo` from Hackage.

What should happen is that `foo` is fetched from the snapshot, as we have specified. The only way I know how which would do so, currently, is to pass `foo` without the version number. 

Currently, we pass all of our dependencies with their version numbers. An alternative is to pass none of the dependency with version numbers (of course, the packages with overridden version numbers would still be passed that way, but the dependencies computed by stack, would have no version number). That _should_ work fine, as all of these dependencies are in the snapshot. I think. But because I'm worried, I'm making this issue rather than a simple pull request.

Workaround: also specify the faulty dependency in the `stack_snapshot`.

**Environment**
 - OS name + version: Ubuntu 19.04
 - Bazel version: 0.27.0- (@non-git) (built with nixpkgs)
 - Version of the rules: 1d8c5956108729e40f488fe77e9af439b167052e

aspiwack:
**This design should really be triple check, I may have misunderstood something*

`stack unpack foo-x.y.z` fetches that very version from Hackage,
whereas `stack unpack foo` fetches the version specified in the
snapshot, including if it's a git location.

So, when computing dependencies in the snapshot, we don't pass a
version number to `stack unpack`. Otherwise git override will fail to
download.

Fixes #1027

### DIFF
--- a/.netlify/install.sh
+++ b/.netlify/install.sh
@@ -13,22 +13,3 @@ V=0.26.0
 curl -LO https://github.com/bazelbuild/bazel/releases/download/$V/bazel-$V-installer-linux-x86_64.sh
 chmod +x bazel-$V-installer-linux-x86_64.sh
 ./bazel-$V-installer-linux-x86_64.sh --user
-
-# XXX: Hack to prevent the `haskell_nixpkgs_package_list` rule from crashing:
-# This rule expects a `nix-build` executable which is used to generate a
-# store-path containing an `all-haskell-packages.bzl` file which defines the
-# `package` list. Since actually installing `nix-build` on the netlify image
-# seems difficult, we provide a dummy shell script which does exactly that.
-packages_list=$(mktemp -d)
-cat <<EOF > $packages_list/all-haskell-packages.bzl
-packages = []
-EOF
-
-mkdir -p $HOME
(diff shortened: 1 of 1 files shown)


mboes:
CI fails due to a linting error. You'll want to run `bazel run //:buildifier-fix` to autofix.

Profpatsch:
Whoops, sorry, push fail. Github decided to close this PR because there were no new commits after my push; opening another one.

Issue #1025, opened 2019-08-02
Stack has a --numeric-version flag


aspiwack:
I don't know since when it exists, but all the versions installed on my computer have it. It prints the version number without any decoration.

```
$ stack --numeric-version
1.9.3
```

This provides a simpler and more robust solution to version checking than #1004 .

mboes:
This is an easy one. @Profpatsch can you take it on?

Profpatsch:
Sure.

PR #1031, opened 2019-08-09
cabal: simplify stack_version_check

Profpatsch:
stack has a `--numeric-version` option, which prints only the semver
string, so we can simplify the logic.

Co-Authored-By: Arnaud Spiwack <arnaud.spiwack@tweag.io>
Fixes: https://github.com/tweag/rules_haskell/issues/1025


aspiwack:
Co-authored by me :stuck_out_tongue: . That's a tad of an overstatement :slightly_smiling_face: .

PR #1033, opened 2019-08-11
Make HaskellInfo.source_files a depset, not a set

mboes:
Like for other instances of this code transformation, it avoids many
conversions during the analysis phase.


haskell/private/actions/compile.bzl
```diff
@@ -322,9 +329,7 @@ def _compilation_defaults(hs, cc, java, dep_info, plugin_dep_info, cc_info, srcs
         args = args,
         compile_flags = compile_flags,
         inputs = depset(transitive = [
-            depset(header_files),
-            depset(boot_files),
-            set.to_depset(source_files),
+            source_files,
```
aherrmann: Why are `boot_files` and `header_files` no longer included here?

haskell/private/dependencies.bzl
```diff
@@ -37,15 +37,16 @@ def gather_dep_info(ctx, deps):
...
+        dep[HaskellInfo].source_files
+        for dep in deps
+        if HaskellInfo in dep
+    ])
 
     import_dirs = set.empty()
     for dep in deps:
         if HaskellInfo in dep:
-            import_dirs = set.mutable_union(source_files, dep[HaskellInfo].source_files)
+            import_dirs = set.mutable_union(import_dirs, dep[HaskellInfo].import_dirs)
```
aherrmann: Looks like a copy-paste issue. Interesting that this didn't cause any problems.

haskell/private/actions/compile.bzl
mboes: Because they're part of extra_source_files.

haskell/private/dependencies.bzl
mboes: Yes. The only place where `import_dirs` are taken from a provider is in the repl code. I don't understand `import_dirs` well enough yet to know whether the repl code really needs them. My suspicion is that `import_dirs` is unnecessary altogether.

PR #1035, opened 2019-08-11
Allow haskell_library to be empty

mboes:
An empty `haskell_library` is one where no source files are given.
This leads to no library being produced, but this is now nevertheless
legal. There are two reasons for doing this:

* Match the semantics of `cc_library`, where empty libraries are
  legal.
* Emulate empty Cabal libraries (also legal) using `haskell_library`.
  This can be useful to e.g. workaround `bytestring-builder` or
  `nats`, which are packages on Hackage that expose empty libraries.

### DIFF
--- a/haskell/private/actions/compile.bzl
+++ b/haskell/private/actions/compile.bzl
@@ -480,17 +480,24 @@ def compile_library(
             mix_file = hs.actions.declare_file(".hpc/{pkg}/{module}.mix".format(pkg = pkg_id_string, module = module))
             coverage_data.append(_coverage_datum(mix_file, src_file, hs.label))
 
-    hs.toolchain.actions.run_ghc(
-        hs,
-        cc,
-        inputs = c.inputs,
-        input_manifests = c.input_manifests,
-        outputs = c.outputs + [datum.mix_file for datum in coverage_data],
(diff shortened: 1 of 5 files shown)


haskell/private/actions/compile.bzl
```diff
@@ -480,17 +480,24 @@ def compile_library(
...
+            inputs = c.inputs,
+            input_manifests = c.input_manifests,
+            outputs = c.outputs + [datum.mix_file for datum in coverage_data],
+            mnemonic = "HaskellBuildLibrary" + ("Prof" if with_profiling else ""),
+            progress_message = "HaskellBuildLibrary {}".format(hs.label),
+            env = c.env,
+            arguments = c.args,
+        )
+    else:
+        hs.actions.run_shell(
```
aherrmann: Why is this dummy action necessary? Presumably, `c.outputs` is empty.

mboes: In this case I believe `c.outputs = [iface_dir, objects_dir]`. An apparent side effect of this dummy action is that Bazel creates these directories. Does that make sense to you?

aherrmann: Interesting, I didn't know about that behavior. Yes, that makes sense to me.

PR #1037, opened 2019-08-12
Handle empty libraries on Hackage

### LINKED ISSUE #999: haskell_cabal_library fails on packages with no modules
I'm not entirely sure of my diagnosis, but I'm seeing an error when building the `bytestring-builder` package

```
ERROR: /home/aspiwack/.cache/bazel/_bazel_aspiwack/bd7cdf0b0195cd3d7f8d5aa71bc1df4e/external/stackage/BUILD.bazel:345:1: output 'external/stackage/bytestring-builder-0.10.8.2.0/_install/lib/libHSbytestring-builder-0.10.8.2.0.a' was not created
ERROR: /home/aspiwack/.cache/bazel/_bazel_aspiwack/bd7cdf0b0195cd3d7f8d5aa71bc1df4e/external/stackage/BUILD.bazel:345:1: output 'external/stackage/bytestring-builder-0.10.8.2.0/_install/lib/libHSbytestring-builder-0.10.8.2.0-ghc8.6.4.so' was not created
ERROR: /home/aspiwack/.cache/bazel/_bazel_aspiwack/bd7cdf0b0195cd3d7f8d5aa71bc1df4e/external/stackage/BUILD.bazel:345:1: not all outputs were created or valid
```

I ould check that `external/stackage/bytestring-builder-0.10.8.2.0/_install/lib/` exists but is empty.

My guess is the following: bytestring-builder is a compability package, on recent versions of ghc, it is an empty package. This seems to cause cabal to generate no file, but the rule for cabal package seems to promise a number of outputs, which are not created, and Bazel's action validation step fails.

I haven't made a reproducer yet. But it should be sufficient to generate a dummy Haskell project which depends on bytestring-builder.

**Environment**
 - OS name + version: Ubuntu 19.03
 - Bazel version: 0.27.0 (from a recent nixpkgs)
 - Version of the rules: f1cb56a4e05ebc3a5ce0560117ecddb2e0a3d673 (from a master today)

mboes:
Empty libraries are libraries that don't expose any modules. It is
legal for a Cabal package to have an empty library. We now handle this
case properly. However, it is hard to do so in full generality. Since
these packages are rare and old, we use a blacklist for now of known
empty packages and do something special when we encounter a package in
that list.

Fixes #999

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -69,6 +69,14 @@ main = defaultMain
 
 _CABAL_TOOLS = ["alex", "c2hs", "cpphs", "doctest", "happy"]
 
+# Some old packages are empty compatibility shims. Empty packages
+# cause Cabal to not produce the outputs it normally produces. Instead
+# of detecting that, we blacklist the offending packages, on the
+# assumption that such packages are old and rare.
+#
+# TODO: replace this with a more general solution.
+_EMPTY_PACKAGES_BLACKLIST = ["bytestring-builder", "nats"]
+
 def _cabal_tool_flag(tool):
     """Return a --with-PROG=PATH flag if input is a recognized Cabal tool. None otherwise."""
     if tool.basename in _CABAL_TOOLS:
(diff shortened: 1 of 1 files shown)


aspiwack:
I can test on my use-case tomorrow.

aspiwack:
It worked for me, indeed.

PR #1039, opened 2019-08-14
cc_wrapper for all OSes

### LINKED ISSUE #958: symlink_dynamic_library takes an unreasonable amount of time

aherrmann:
Introduces `cc_wrapper` for all operating systems.

- Shortens library search paths to stay below maximum path length on Windows.
    GHC generates library search paths that contain redundant up-level references (`..`). This can exceed the maximum path length on Windows, which will cause linking failures.
- Shortens rpaths and load commands on macOS.
    The rpaths and load commands generated by GHC and Bazel can quickly exceed the MACH-O header size limit on macOS. This wrapper shortens and combines rpaths and load commands to avoid exceeding that limit.
- Finds .so files if only .dylib are searched on macOS.
    Bazel's cc_library will generate .so files for dynamic libraries even on macOS. GHC strictly expects .dylib files on macOS. This wrapper hooks into gcc's --print-file-name feature to work around this mismatch in file extension.
- Removes the need to generate dynamic library symbolic links addressing #958.
- The CC wrapper is also used for REPL and runghc targets so that these also don't require symlinks for dynamic libraries on macOS.
- Uses a separate cc_wrapper written in bash on Windows.
    This works around https://github.com/bazelbuild/bazel/issues/8981, execution of py_binary targets is too slow on Windows to be practical for cc_wrapper. On Windows the cc_wrapper only needs to shorten library search paths and not worry about rpaths or dynamic library file extensions. This allows for a much simpler cc_wrapper.
- Tested on https://github.com/digital-asset/daml on Linux, macOS, and Windows.

Second attempt at #1002, which failed due to py_binary being too slow on Windows, see  https://github.com/bazelbuild/bazel/issues/8981.

Closes #958


aherrmann:
Rebased on master to fix merge conflict.

haskell/private/actions/link.bzl
```diff
@@ -352,15 +352,10 @@ def link_library_dynamic(hs, cc, dep_info, cc_info, extra_srcs, objects_dir, my_
     )
     args.add_all(pkg_info_args)
 
-    # When linking a dynamic library we still collect static libraries for
-    # dependencies where possible. This is so that a final binary that depends
-    # on this dynamic library, is linked statically itself, will not fail at
-    # link time due to missing transitive dynamic library dependencies. In this
-    # case transitive dependencies will still be linked in statically.
```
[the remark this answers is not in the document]
aherrmann: GHC loads shared libraries with `RTLD_LAZY`. This can cause missing symbols with transitive Haskell library dependencies in dynamic linking mode. This issue also came up in https://github.com/tweag/rules_haskell/pull/927. It's not strictly related to the `cc_wrapper`. However, I tested the `cc_wrapper` on `damlc` in https://github.com/digital-asset/daml with a dynamic RTS, and that required this change to pass.

haskell/private/cc_wrapper.bzl
```diff
@@ -0,0 +1,56 @@
...
+    feature_configuration = cc_common.configure_features(
+        ctx = ctx,
+        cc_toolchain = cc_toolchain,
+        requested_features = ctx.features,
+        unsupported_features = ctx.disabled_features,
+    )
+    cc = cc_common.get_tool_for_action(
+        feature_configuration = feature_configuration,
+        action_name = ACTION_NAMES.c_compile,
+    )
```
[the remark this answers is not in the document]
aherrmann: [`CcToolchainInfo.compiler_executable`](https://docs.bazel.build/versions/0.28.0/skylark/lib/CcToolchainInfo.html#compiler_executable) has been deprecated. `get_tool_for_action` is used instead to get access to the compiler executable. See also https://github.com/bazelbuild/bazel/issues/8802. This will return the C compiler of the active cc_toolchain.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+
+        """
+        args = iter(args)
+        for arg in args:
+            out = []
+            # Poor man's pattern matching: Each handler function takes the
+            # current argument, the stream of up-coming arguments, and a
+            # reference to the list of arguments to forward. The handler must
+            # return True if it consumes the argument, and return False if
+            # another handler should consume the argument.
```
Profpatsch: This is beautifully bonkers, but I got no suggestions.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+          args: Iterable over command-line arguments.
+
+        """
+        self.action = Args.LINK
+        self.print_file_name = None
+        self.libraries = []
+        self.library_paths = []
+        self.rpaths = []
+        self.output = None
+        self._prev_ld_arg = None
```
Profpatsch: These members could do with some docstrings

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+                yield arg
+
+            for out_arg in out:
+                yield out_arg
+
+    def _handle_output(self, arg, args, out):
+        if arg == "-o":
+            # Remember the output filename.
+            self.output = next(args)
+            out.extend(["-o", self.output])
```
Profpatsch: I’d go for a more structured approach here and add `"-o"`, `"-l"` and such not here, but in the final “pretty-printing phase” of converting `Args` to a GHC command line.

That is, just save the data (`self.output` in this match) and add the lower-level structure of the args later.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+        else:
+            return False
+
+    def _handle_library(self, arg, args, out):
+        if arg == "-l" or arg == "--library":
+            library = next(args)
+        elif arg.startswith("-l"):
+            library = arg[2:]
+        elif arg.startswith("--library="):
+            library = arg[len("--library="):]
```
Profpatsch: Can we factor the CLI-logic out? in the best case there’d be just something like

```python
def _handle_library(…):
  (consumed, el) = argument(args, short = "-l", long = "--library")
  if consumed:
    self.libraries.append(el)
  return consumed
```

The API is kind of ugly, but I don’t think we can do better with the mutable approach.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+        elif arg.startswith("-Wl,"):
+            ld_args = arg.split(",")[1:]
+            if len(ld_args) == 2 and ld_args[0] == "-rpath":
+                self._handle_rpath(ld_args[1], out)
+                return True
+            else:
+                out.append(arg)
+                return True
+        else:
+            return False
```
Profpatsch: According to https://stackoverflow.com/questions/7221141/any-difference-between-wl-option-and-xlinker-option-syntax-for-gcc both `-Xlinker` and `-Wl` essentially do the same thing.

We should add a comment describing that.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+                    self._prev_ld_arg = ld_arg
+                else:
+                    out.extend(["-Xlinker", ld_arg])
+            elif self._prev_ld_arg == "-rpath":
+                self._prev_ld_arg = None
+                self._handle_rpath(ld_arg, out)
+            else:
+                # This indicates a programmer error and should not happen.
+                raise RuntimeError("Unhandled _prev_ld_arg '{}'.".format(self._prev_ld_arg))
+            return True
```
Profpatsch: From the [gcc manpage](http://man7.org/linux/man-pages/man1/gcc.1.html#OPTIONS), `-Xlinker` description: 

```
 When using the GNU linker, it is usually more convenient to pass
           arguments to linker options using the option=value syntax than as
           separate arguments.  For example, you can specify -Xlinker
           -Map=output.map rather than -Xlinker -Map -Xlinker output.map.
           Other linkers may not support this syntax for command-line
           options.
```

Do we need to handle that case?

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
+#!/usr/bin/env python3
+"""CC toolchain wrapper
+
+Usage: cc_wrapper [ARG]...
+
+Wraps the C compiler of the Bazel CC toolchain. Transforms arguments to work
+around limitations of Bazel and GHC and passes those via response file to the C
```
Profpatsch: You never define what a “response file” is. This is a good place to do so.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+      - Resolve symbolic links.
+
+    Args:
+      input_path: The path to shorten.
+
+    Returns:
+      The shortened path.
+
+    """
+    exists = os.path.exists(input_path)
```
Profpatsch: What does it mean that a given file does not exist? Should we accept that in the first place?

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+        shortened = rel
+
+    # Try normalizing the path if possible.
+    norm = os.path.normpath(shortened)
+    if len(norm) < len(shortened):
+        # Ensure that the path is still correct. Reducing up-level references
+        # may change the meaning of the path in the presence of symbolic links.
+        try:
+            if not exists or os.path.samefile(norm, shortened):
+                shortened = norm
```
Profpatsch: This is where allowing non-existent files becomes questionable, because what if we screw up the non-existent path here and don’t notice. Could lead to nasty bugs.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+            # stat may fail if the path became invalid or does not exist.
+            pass
+
+    # Try resolving symlinks.
+    try:
+        real = os.path.relpath(os.path.realpath(shortened))
+        if len(real) < len(shortened):
+            shortened = real
+    except IOError:
+        # realpath may fail if the path does not exist.
```
Profpatsch: In that case, shouldn’t we use `exists` to check before `realpath`? That is if we want to allow non-existent paths in the first place.

haskell/private/cc_wrapper.py.tpl
aherrmann: They are defined in the `Attrs` section of the class docstring [here](https://github.com/tweag/rules_haskell/pull/1039/files/71c2a6d5286a0e0802c78be3a1a8cf6e6ae8061d..a10b82374bc7b6f86844835b1656e06b449a0b0a#diff-19cb794875a66149f66f34681dd62136R66).

haskell/private/cc_wrapper.py.tpl
aherrmann: Done.

haskell/private/cc_wrapper.py.tpl
aherrmann: The less structured approach is intentional. This wrapper tries hard to maintain the relative order of the command-line arguments, structuring all flags and only then pretty-printing them could change that order. The only exception are arguments that we intend to modify and where we need to have seen all command-line flags before we can modify them, namely `--print-file-name` and `-rpath`.

Generally, it is not safe to change the order of command-line arguments. E.g. with static library linking the relative order of the arguments matters. Consider `-l foo libbar.a -l baz`, where `libfoo` and `libbaz` are also static libraries. If we would structure and then pretty print all arguments, then we'd end up with something like `-l foo -l baz libbar.a`, which changes meaning.

We could structure and also remember the position of each argument and then puzzle them back together in the right order. But, I don't see that the benefit outweighs the additional complexity.

haskell/private/cc_wrapper.py.tpl
aherrmann: Agreed, that looks better.

haskell/private/cc_wrapper.py.tpl
aherrmann: Done. AFAIK e.g. BSD ld does not accept that syntax and GHC doesn't emit it, but for completeness it makes sense to handle it.

haskell/private/cc_wrapper.py.tpl
aherrmann: Done

haskell/private/cc_wrapper.py.tpl
aherrmann: This could happen with `--library-path`. E.g. if a Haskell library has a dynamic and static output, then we'll have the static library dir (`bazel-out/.../<package>`) and the dynamic library dir (`.../_solib_XYZ/...`) listed in the package-db. But, if we're compiling in static mode, then the dynamic library won't be in the sandbox and the directory won't exist. This could also happen with Cabal libraries that have bogus `library-path` contents. GHC and GCC are lenient about this, so we should be, too. However, it's probably safer to capture this case before we even call `shorten_path`. I've changed it so that `shorten_path` requires the path to exist and is simply not called on paths that don't.

haskell/private/cc_wrapper.py.tpl
aherrmann: This seems to depend on OS and Python version and I can't trigger the issue on my machine. But, IIRC this happened on CI also in case of bad permissions or dangling links, so checking `exists` is not enough.

haskell/private/cc_wrapper.py.tpl
aherrmann: See reply in https://github.com/tweag/rules_haskell/pull/1039/files/71c2a6d5286a0e0802c78be3a1a8cf6e6ae8061d..a10b82374bc7b6f86844835b1656e06b449a0b0a#r318499294

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+def generate_response_line(arg):
+    # Gcc expects one argument per line, surrounded by double quotes, with
+    # inner double quotes escaped with backslash, and backslashes themselves
+    # escaped. shlex.quote conflicts with this format.
+    return '"{}"\n'.format(arg.replace("\\", "\\\\").replace('"', '\\"'))
+
+
+def is_darwin():
+    """Whether the execution platform is Darwin."""
+    return platform.system() == "Darwin"
```
Profpatsch: nit: should be passed in from bazel

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+    run_cc(args, exit_on_error=True)
+
+    if is_darwin():
+        darwin_rewrite_load_commands(darwin_rewrites, output)
+
+
+def shorten_rpaths(rpaths, libraries, output):
+    """Avoid redundant rpaths.
+
+    Filters out rpaths that are not required to load any library dependency.
```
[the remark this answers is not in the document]
aherrmann: No, I don't think so. This wrapper is trying to adapt between GHC and Bazel in an unobtrusive way. Redundant rpaths are not an error, but something that can be optimized away, and in the case of macOS something that we have to remove to avoid issues with the MACH-O header size limit.

E.g. GHC will generate lots of rpaths of the form `bazel-out/.../packageXYZ/`. These are redundant in normal linking mode, because rules_haskell passes "better" `$ORIGIN/.../_solib_*/...` rpaths. However, sometimes (e.g. template Haskell) GHC links temporary outputs and doesn't forward rules_haskell's rpaths. In that case we need to use the `bazel-out/...` type rpaths. Also, GHC might pass absolute rpaths, e.g. `/nix/store/...`. These may be redundant if the same library was also imported as a `cc_library` and a corresponding `$ORIGIN/.../_solib_*/...` rpath exists. That's not an error though.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+    Args:
+      rpaths: List of given rpaths.
+      libraries: List of library dependencies.
+      output: The output binary, used to resolve rpaths.
+
+    Returns:
+      List of required rpaths.
+
+    """
+    input_rpaths = sort_rpaths(rpaths)
```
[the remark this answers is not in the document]
aherrmann: We classify three types of rpaths:
- relative to output, i.e. `$ORIGIN/...`/`@loader_path/...`
- absolute, e.g. `/nix/store/...`
- relative, e.g. `bazel-out/...`.

We prefer those relative to the output. They tend to be shorter, and they typically involve Bazel's `_solib_*` directory which bundles lots of libraries (meaning less rpaths required). They're also less likely to leak information about the local installation into the Bazel cache.

Next, we prefer absolute paths. They function regardless of execution directory, and they are still likely to play well with the cache, e.g. `/nix/store/...` or `/usr/lib/...`.

Finally, we fall back to relative rpaths.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+    missing = set(libraries)
+
+    rpaths = []
+
+    for rpath in input_rpaths:
+        if not missing:
+            break
```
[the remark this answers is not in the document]
aherrmann: It is, I've rephrased things to be a bit easier to understand.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+    rewrites = []
+
+    # References to core libs take up much space. Consider detecting the GHC
+    # libdir and adding an rpath for that and making load commands relative to
+    # that. Alternatively, https://github.com/bazelbuild/bazel/pull/8888 would
+    # also avoid this issue.
+
+    # Determine solib dir and rewrite load commands relative to solib dir.
+    # This allows to replace potentially many rpaths by one.
+    solib_rpath = find_solib_rpath(input_rpaths, output)
```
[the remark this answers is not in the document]
aherrmann: The mechanism that we use these for only exists on Darwin. Namely, Darwin allows to explicitly refer to the rpath in load commands. E.g.
```
LOAD @rpath/somedir/libsomelib.dylib
```
With that we can avoid multiple rpath entries of the form
```
RPATH @loader_path/.../_solib_*/mangled_a
RPATH @loader_path/.../_solib_*/mangled_b
RPATH @loader_path/.../_solib_*/mangled_c
```
And instead have a single rpath and load commands as follows
```
RPATH @loader_path/.../_solib_*
LOAD @rpath/mangled_a/lib_a
LOAD @rpath/mangled_b/lib_b
LOAD @rpath/mangled_c/lib_c
```

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+    return sorted(rpaths, key=rpath_priority)
+
+
+def find_solib_rpath(rpaths, output):
+    """Find the solib directory rpath entry.
+
+    The solib directory is the directory under which Bazel places dynamic
+    library symbolic links on Unix. It has the form `_solib_<cpu>`.
+
+    """
```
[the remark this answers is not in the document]
aherrmann: Good idea, I've added one.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+
+    libname, ext = os.path.splitext(libname)
+    if ext in [".dll", ".dylib", ".so"]:
+        return libname
+
+    return None
+
+
+def resolve_rpath(rpath, output):
+    """Resolve the given rpath, replacing references to the binary."""
```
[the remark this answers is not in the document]
aherrmann: We have the opposite already. This here resolves the `$ORIGIN`/`@loader_path` in an rpath to determine if the rpath is useful. Note, GHC can also generate rpaths with `$ORIGIN`/`@loader_path` in them. So, this is not only about rpaths generated by rules_haskell.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+
+    if is_temporary_output(output):
+        # GHC generates temporary libraries outside the execroot. The regular
+        # relative rpaths don't work in that case and have to be converted to
+        # absolute paths.
+        if has_origin(rpath):
+            # We don't know what $ORIGIN/@loader_path was meant to refer to.
+            # Try to find an existing, matching rpath by globbing.
+            stripped = replace_origin(rpath, "")
+            candidates = glob.glob(os.path.join("**", stripped), recursive=True)
```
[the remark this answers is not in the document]
aherrmann: Yeah, we're in a tricky situation here. In some cases, e.g. template Haskell, GHC links temporary dylibs in a temporary directory outside the execroot, e.g. `/tmp/ghcXYZ`. `$ORIGIN` in the rpath was meant to refer to the final output, e.g. `bazel-out/.../mypackage/libHSmylib.so`. But, in this case we don't know that path, it's not on the command-line.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+    else:
+        _, res = run_cc_print_file_name(filename, args)
+
+    sys.stdout.write(res.stdout.decode())
+    sys.stderr.write(res.stderr.decode())
+    sys.exit(res.returncode)
+
+
+def run_cc_print_file_name(filename, args):
+    """Run cc --print-file-name on the given file name.
```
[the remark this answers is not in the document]
aherrmann: Done

haskell/private/cc_wrapper.sh.tpl
```diff
@@ -0,0 +1,162 @@
...
+    local arg="$1"
+    if [[ $IN_RESPONSE_FILE = 1 ]]; then
+        arg="$(unquote_arg "$arg")"
+    fi
+    if [[ $LIB_DIR_COMING = 1 ]]; then
+        LIB_DIR_COMING=
+        handle_lib_dir "$arg"
+    elif [[ "$arg" =~ ^@(.*)$ ]]; then
+        IN_RESPONSE_FILE=1
+        while read line; do
```
Profpatsch:
```
In - line 140:
        while read line; do
              ^--^ SC2162: read without -r will mangle backslashes.
```

haskell/private/cc_wrapper.py.tpl
aherrmann: Okay, done.

haskell/private/cc_wrapper.sh.tpl
```diff
@@ -19,18 +19,16 @@ set -euo pipefail
 # Find compiler
 
 find_exe() {
-    local exe="$1"
-    local location
+    local -n location="$1"
```
Profpatsch: whoa

haskell/private/cc_wrapper.py.tpl
Profpatsch: True

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+        if not found:
+            # Retry with .so extension.
+            found, so_res = run_cc_print_file_name("%s.so" % basename, args)
+            if found:
+                res = so_res
+    else:
+        _, res = run_cc_print_file_name(filename, args)
+
+    sys.stdout.write(res.stdout.decode())
+    sys.stderr.write(res.stderr.decode())
```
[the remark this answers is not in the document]
aherrmann: Done

haskell/private/cc_wrapper.py.tpl
```diff
@@ -0,0 +1,812 @@
...
+        new_kwargs.update(kwargs)
+        kwargs = new_kwargs
+
+    with response_file(args) as rsp:
+        res = subprocess.run([CC, "@" + rsp], **kwargs)
+
+    if exit_on_error and res.returncode != 0:
+        if capture_output:
+            sys.stdout.write(res.stdout.decode())
+            sys.stderr.write(res.stderr.decode())
```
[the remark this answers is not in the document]
aherrmann: Done

haskell/private/cc_wrapper.py.tpl
Profpatsch: Awesome, I like the improved clarity this brings to the handlers.

haskell/private/cc_wrapper.py.tpl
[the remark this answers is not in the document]
Profpatsch: So many edge cases :dizzy_face:. But I see how we can’t be strict in that case.

haskell/private/cc_wrapper.py.tpl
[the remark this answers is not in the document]
Profpatsch: Thanks, this and the added comments makes the logic a lot clearer.

haskell/private/cc_wrapper.py.tpl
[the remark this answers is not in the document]
Profpatsch: Awesome, I think you can drop that description into a comment as-is.

haskell/private/cc_wrapper.py.tpl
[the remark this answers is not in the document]
Profpatsch: Same, great description, please copy & paste into a source code comment :)

haskell/private/cc_wrapper.py.tpl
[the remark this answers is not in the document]
Profpatsch: Probably

haskell/private/cc_wrapper.py.tpl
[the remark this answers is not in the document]
Profpatsch: Ah, I think you describe that somewhere else in the PR as well.

haskell/private/cc_wrapper.bzl
```diff
@@ -45,18 +44,43 @@ _cc_wrapper = rule(
...
+def cc_wrapper(name, **kwargs):
     _cc_wrapper(
-        name = name + "-source",
-        template = template,
+        name = name + ".py",
+        template = "@rules_haskell//haskell:private/cc_wrapper.py.tpl",
+    )
+    _cc_wrapper(
+        name = name + ".sh",
+        template = "@rules_haskell//haskell:private/cc_wrapper.sh.tpl",
```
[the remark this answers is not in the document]
aherrmann: Done

haskell/private/cc_wrapper.py.tpl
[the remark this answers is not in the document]
aherrmann: Done, I've added it to the docstring of `sort_rpaths`.

haskell/private/cc_wrapper.py.tpl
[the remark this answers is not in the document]
aherrmann: Done.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -731,8 +732,24 @@ def run_cc(args, capture_output=False, exit_on_error=False, **kwargs):
...
+        # script using Bazel runfiles.
+        r = bazel_runfiles.Create()
+        cc = r.Rlocation("/".join([WORKSPACE, CC]))
+        if cc is None and platform.system() == "Windows":
+            # We must use "/" instead of os.path.join on Windows, because the
+            # Bazel runfiles_manifest file uses "/" separators.
+            cc = r.Rlocation("/".join([WORKSPACE, CC + ".exe"]))
+        if cc is None:
+            print("CC not found '{}'.".format(CC), file=sys.stderr)
+            sys.exit(1)
```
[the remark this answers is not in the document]
aherrmann: Sounds good, done.

haskell/private/cc_wrapper.sh.tpl
aherrmann: Good point.

haskell/private/cc_wrapper.sh.tpl
```diff
@@ -0,0 +1,162 @@
...
+    # escaped.
+    local arg="$1"
+    arg="${arg//\\/\\\\}"
+    arg="${arg//\"/\\\"}"
+    printf '"%s"\n' "$arg"
+}
+
+unquote_arg() {
+    local arg="$1"
+    if [[ "$arg" =~ ^\"(.*)\"[[:space:]]*$ ]]; then
```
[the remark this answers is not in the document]
aherrmann: You mean why wouldn't the argument be quoted? If it doesn't contain spaces it doesn't have to be quoted. And I've observed cases where the argument wasn't quoted.

aherrmann:
The CI failure is a timeout. I'm testing locally if there's a performance regression. I also started a rerun. The reason for the timeout may not be that anything in particular got slower, but rather that everything had to be rebuilt.

haskell/ghc_bindist.bzl
```diff
@@ -266,6 +266,14 @@ haskell_toolchain(
         haddock_flags = ctx.attr.haddock_flags,
         repl_ghci_args = ctx.attr.repl_ghci_args,
     )
+
+    if os == "windows":
+        # These libraries cause linking errors on Windows when linking
```
mboes: Is there an upstream ticket for this?

aherrmann: Not that I know. I'll try to come up with a good repro and open a ticket.

PR #1041, opened 2019-08-15
Deprecate Hazel

mboes:
Hazel was previously marked as experimental, and never shipped as part
of any release. So deprecating it is a matter of moving any references
from the documentation, and adding a note to the README. We'll still
carry the code for now, since there are still users of Hazel and we
need to continue adapting the Hazel code as rules_haskell changes.


mboes:
cc @thumphries @nhibberd. We recommend switch to the [Cabal rules](https://api.haskell.build/haskell/cabal.html), which use Cabal to build Hackage packages.

PR #1046, opened 2019-08-16
Fix c2hs on Windows

cocreature:
This fixes a few issues with c2hs_library that we encountered in daml,
see https://github.com/digital-asset/daml/pull/2574 for the upstream
PR.

1. c2hs did not have the runfiles setup properly. This works fine on
this repo since c2hs comes from nixpkgs but we get it from Hazel.

2. cpp invoked via c2hs crashes without any error if the mingw bin dir
is not in PATH. For now, we fix this in a rather hacky way by
calculating it relative to cc.

3. language-c has issues when it comes to parsing some headers shipped
with mingw, see https://github.com/haskell/c2hs/issues/199. You can
avoid this by setting certain flags in cpp so I added an `extra_args`
option to `c2hs_library` that allows passing additional arguments.

cc @aherrmann

### DIFF
--- a/haskell/c2hs.bzl
+++ b/haskell/c2hs.bzl
@@ -22,6 +22,7 @@ def _c2hs_library_impl(ctx):
     cc = cc_interop_info(ctx)
     args = hs.actions.args()
     c2hs = ctx.toolchains["@rules_haskell//haskell/c2hs:toolchain"].c2hs
+    c2hs_exe = ctx.toolchains["@rules_haskell//haskell/c2hs:toolchain"].c2hs_exe
 
     if len(ctx.files.srcs) != 1:
         fail("srcs field should contain exactly one file.")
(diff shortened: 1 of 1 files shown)


haskell/c2hs.bzl
```diff
@@ -59,23 +61,36 @@ def _c2hs_library_impl(ctx):
...
+        tools = [hs.tools.ghc, c2hs_exe],
         outputs = [hs_file, chi_file],
-        command = """
+        command =
+            # cpp (called via c2hs) gets very unhappy if the mingw bin dir is
+            # not in PATH so we add it to PATH explicitely.
+            (
+                """
+        export PATH=$PATH:{mingw-bin}
+        """.format(cc = paths.dirname(cc.tools.cc)) if hs.toolchain.is_windows else ""
```
aherrmann: `cc` doesn't match the place holder `{mingw-bin}`.

cocreature: Good catch, fixed!

PR #1055, opened 2019-08-24
Completes the prototype implementation of Bazel persistent worker mode

ulysses4ever:
Started in 132d0af6. This commit imports worker's code and sets everything to make
the persistent worker mode available to a regular `rules_haskell` user.

The worker code lives in `tools/worker`.

The mode is supported in ghc_nixpkgs-toolchain only, at this point.

### DIFF
--- a/haskell/defs.bzl
+++ b/haskell/defs.bzl
@@ -91,6 +91,12 @@ _haskell_common_attrs = {
         cfg = "host",
         default = Label("@rules_haskell//haskell:ghc_wrapper"),
     ),
+    "worker": attr.label(
+        default = None,
+        executable = True,
+        cfg = "host",
+        doc = "Experimental. Worker binary employed by Bazel's persistent worker mode. See docs/haskell-use-cases.rst",
+    ),
 }
 
 def _mk_binary_rule(**kwargs):
@@ -180,42 +186,11 @@ def _mk_binary_rule(**kwargs):
         **kwargs
     )
 
-haskell_test = _mk_binary_rule(test = True)
-"""Build a test suite.
+_haskell_test = _mk_binary_rule(test = True)
 
-Additionally, it accepts [all common bazel test rule
-fields][bazel-test-attrs]. This allows you to influence things like
-timeout and resource allocation for the test.
+_haskell_binary = _mk_binary_rule()
 
-[bazel-test-attrs]: https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes-tests
-"""
-
-haskell_binary = _mk_binary_rule()
-"""Build an executable from Haskell source.
-
-Example:
-  ```bzl
-  haskell_binary(
-      name = "hello",
-      srcs = ["Main.hs", "Other.hs"],
-      deps = ["//lib:some_lib"]
-  )
-  ```
-
-Every `haskell_binary` target also defines an optional REPL target that is
-not built by default, but can be built on request. The name of the REPL
-target is the same as the name of binary with `"@repl"` added at the end.
-For example, the target above also defines `main@repl`.
-
-You can call the REPL like this (requires Bazel 0.15 or later):
-
-```
-$ bazel run //:hello@repl
-```
-
-"""
-
-haskell_library = rule(
+_haskell_library = rule(
     _haskell_library_impl,
     attrs = dict(
         _haskell_common_attrs,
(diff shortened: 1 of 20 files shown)


haskell/private/context.bzl
```diff
@@ -40,13 +40,18 @@ def haskell_context(ctx, attr = None):
     if hasattr(ctx.executable, "_ghc_wrapper"):
         ghc_wrapper = ctx.executable._ghc_wrapper
 
+    worker = None
+    if hasattr(ctx.executable, "_worker"):
+        worker = ctx.executable._worker
```
aherrmann: Why not `worker = getattr(ctx.executable, "_worker", None)`?

tests/haskell_proto_library/BUILD.bazel
```diff
@@ -76,6 +76,7 @@ haskell_library(
         ":person_haskell_proto",
         ":stripped_address_haskell_proto",
         "//tests/hackage:base",
+        "@stackage//:proto-lens",
```
aherrmann:
```suggestion
```
I'm not sure this is still needed, I think this was an artifact of updating the proto toolchain, which is no longer necessary.

tests/haskell_proto_library/Bar.hs
```diff
@@ -4,5 +4,10 @@ import Proto.StrippedAddress
 import Proto.Tests.HaskellProtoLibrary.Person
 import Proto.Tests.HaskellProtoLibrary.Person_Fields
 
+import Data.ProtoLens.Message
+
+defPerson :: Person
+defPerson = defMessage
```
aherrmann: How is this related to the worker?

tools/worker/Compile.hs
```diff
@@ -0,0 +1,98 @@
...
+        (dflags2, fileish_args, _warns) <-
+          parseDynamicFlags dflags (map noLoc flags)
+
+        -- Normilize paths
+        let
+          normalise_hyp fp
+            | strt_dot_sl && "-" `isPrefixOf` nfp = cur_dir ++ nfp
+            | otherwise                           = nfp
+            where
+-- # if defined(mingw32_HOST_OS) -- TODO: do we need Win support at this point?
```
aherrmann: Might just as well uncomment this and enable `{-# LANGUAGE CPP #-}`.

tools/worker/Compile.hs
```diff
@@ -0,0 +1,98 @@
...
+    let dflags' = dflags { ldInputs = map (FileOption "") o_files
+                                      ++ ldInputs dflags }
+    _ <- GHC.setSessionDynFlags dflags'
+
+    targets <- mapM (uncurry GHC.guessTarget) hs_srcs
+    GHC.setTargets targets
+    ok_flag <- GHC.load LoadAllTargets
+
+    when (failed ok_flag) (liftIO $ exitWith (ExitFailure 1))
+    return ()
```
aherrmann:
```suggestion
```
This seems redundant.

tools/worker/Main.hs
```diff
@@ -0,0 +1,28 @@
...
+
+--import GHC.IO.Handle (hDuplicate, hDuplicateTo)
+
+pwFlag :: String
+pwFlag = "--persistent_worker"
+
+main :: IO ()
+main = do
+
+    -- redirect stdout to stderr
```
aherrmann: Maybe add a comment what this is for and why it's uncommented.

haskell/nixpkgs.bzl
```diff
@@ -83,7 +83,6 @@ haskell_toolchain(
     # hack in Nixpkgs.
     {locale_archive_arg}
     locale = {locale},
-    use_worker = {use_worker},
```
aherrmann: That's neat that the toolchain macro argument is no longer needed.

haskell/defs.bzl
```diff
@@ -91,6 +91,12 @@ _haskell_common_attrs = {
         cfg = "host",
         default = Label("@rules_haskell//haskell:ghc_wrapper"),
     ),
+    "worker": attr.label(
+        default = None,
```
aherrmann: It's sad that Bazel doesn't allow `select` as a `default`.

haskell/defs.bzl
```diff
@@ -253,27 +228,83 @@ haskell_library = rule(
...
+def _haskell_worker_wrapper(rule_type, **kwargs):
+    defaults = dict(
+        worker = select({
+            "@rules_haskell//haskell:use_worker": Label("@rules_haskell//tools/worker:bin"),
+            "//conditions:default": None,
+        }),
+    )
+    defaults.update(kwargs)
+
+    if rule_type == 1:
```
aherrmann: Better use strings as tags, say `"binary"`, `"test"`, and `"library"` rather than numbers. `rule_type = "binary"` is much more descriptive than `rule_type = 1`.

aherrmann:
> Yes, if you enable worker dependencies -- it's all good! But if you don't, plain bazel test //... fails (for me, not sure if it's some local issue). Instead, bazel test //tests/... succeeds. Could you try it? 

Yes, same for me. That makes sense as `//...` includes `//tools/worker:bin`. `bazel test //...` will also try to build (not execute) non-test targets. You can observe the same if you turn e.g. `//tests/binary-simple:binary-simple` into a `haskell_binary` instead of `haskell_test` and run `bazel test //tests/binary-simple/...`.

haskell/defs.bzl
aherrmann: I guess you're referring to [this bit](https://github.com/digital-asset/daml/blob/bd686aa6a0151e5cb4074ddf3ac24ceee13e7227/bazel_tools/haskell.bzl#L246). I think it's not so much a workaround for the fact that `select` is not allowed as a `default`. It's more generally a macro wrapper around the `haskell_repl` rule to define some custom defaults in that repository. Since we're not patching the rule definition directly, we don't have access to the the attribute's default value. 

That `select` is not allowed as an attribute's default value, but is allowed in a macro wrapping a rule seems like an arbitrary restriction, indeed. Turns out there's an open issue about it: https://github.com/bazelbuild/bazel/issues/287

tools/worker/Compile.hs
```diff
@@ -0,0 +1,98 @@
...
+          parseDynamicFlags dflags (map noLoc flags)
+
+        -- Normilize paths
+        let
+          normalise_hyp fp
+            | strt_dot_sl && "-" `isPrefixOf` nfp = cur_dir ++ nfp
+            | otherwise                           = nfp
+            where
+# if defined(mingw32_HOST_OS)
+          strt_dot_sl = "./" `isPrefixOf` fp || ".\\" `isPrefixOf` fp
```
aherrmann:
```suggestion
              strt_dot_sl = "./" `isPrefixOf` fp || ".\\" `isPrefixOf` fp
```
This should be aligned for the `where` clause.

tools/worker/Proto/Worker.hs
avdv, line 1: Hi @ulysses4ever!

I know, it's been a while... but do still have the worker.proto file laying around somewhere? It would be nice if you could provide it to us. :smile:

ulysses4ever: Hey! I'll look into it today. I'm surprised that something is missing: I'd expect that any missing file should be easily generated or downloaded or something... But it's been four years since I touched it so I may very well be wrong.

It's kinda ironic that a server machine I was using for it (and which I was using since) was recently (this July) decommissioned, and I had to move to another one, and in the process I nuked a lot of stuff. Probably including the playground for this project. But let's be optimistic and wait until I try to look it up locally.

ulysses4ever: Oh, I see, it's the generated file that I committed instead of the source. Shame on me!

ulysses4ever: All good. I developed this code as a separate repo first: https://github.com/tweag/bazel-worker/ (should have been a git submodule, perhaps, but they're such a pain...) and it does have `worker.proto`:

- https://github.com/tweag/bazel-worker/blob/master/src/worker.proto

Note that it's a file provided by Bazel itself:

- https://github.com/bazelbuild/bazel/blob/19293673be14e68598f032b1c61fa991884305ad/src/main/protobuf/worker_protocol.proto

It's fun to see the diff between the two: they extended the protocol quite a bit over the years.

PR #1061, opened 2019-08-27
Improve handling of C library dependencies of haskell_cabal_library|binary

aherrmann:
- Skip Haskell libraries in `--extra-lib-dirs` flags.
    These are redundant, as they are already defined in the corresponding package databases.
    This is to avoid command-line length overflow on collect2. E.g.
    ```
    gcc: error trying to exec '/nix/store/3ajmsz0ab2dybi1qm7nk0izabyqwfl4r-gcc-7.4.0/libexec/gcc/x86_64-unknown-linux-gnu/7.4.0/collect2': execv: Argument list too long
    ```
- Forward all C library include directories.
    I.e. not just `CompilationContext.includes`, but also `quote_includes`, and `system_includes`.
    `cc_library` targets typically expose their include directories in the `system_includes` attribute instead of `includes`.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -116,7 +116,11 @@ def _prepare_cabal_inputs(hs, cc, dep_info, cc_info, package_id, tool_inputs, to
     args = hs.actions.args()
     package_databases = dep_info.package_databases
     extra_headers = cc_info.compilation_context.headers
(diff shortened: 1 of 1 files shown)


regnat:
Awesome! I just hit this when trying to migrate a repository to `stack_snapshot`. I didn't look at the patch itself but I can confirm that it worked for me

PR #1065, opened 2019-09-01
Bump stackage nixpkgs + bazel 0.29

guibou:
- Stackage + nixpkgs are bumped together to get similar version of `proto-lens` and `proto-lens-protoc`.

- Bazel 0.29 comes from the current PR: https://github.com/NixOS/nixpkgs/pull/67835

Note: this is a work in progress, because nixpkgs introduced a regression in the way bazel handle python path and this should be fixed first. From my point of view, rules_haskell is ready for bazel 0.29, but not with the bazel from nixpkgs.


Profpatsch:
The `MAINTAINERS` file has a checklist for bumping bazel:

- [ ] Bump bazel download link for bazel in `.circleci/config.yml`
- [ ] Bump bazel download link for bazel in `azure-pipelines.yml`
- [ ] Update all bazel rules dependencies in `WORKSPACE` (e.g.
      `io_bazel_skydoc`)
- [ ] Update bazel in nixpkgs and bump `nixpkgs/default.nix`
- [ ] Bump MAX_BAZEL_* in `start`
- If we are updating to a new LTS:
  - Bump `MIN_BAZEL_*` in `start`
  - TODO
- [ ] Add update notice to `CHANGELOG`

mboes:
Are we there yet?

guibou:
@mboes Unfortunately no. I did not find the time to work on that, especially considering that the issue won't appear if we are using an official bazel release or if we are using bazel from nixpkgs in a nix context.

Profpatsch:
Can we get this merged? Or does it make sense to just wait for 1.0 by now?

guibou:
@Profpatsch This is not mergable in the current state. We either need to:

- Fix bazel in nixpkgs to work propertly with system provided python (and others)
- Conditionnaly use the python toolchain on linux_nixpkgs and use the system toolchain in other context. I still don't know how to do that.

guibou:
Please note that the issue is just with bazel from nixpkgs. Rules_haskell is compatible with bazel 0.29 without this pull request.

PR #1066, opened 2019-09-02
Add initial Bazel CI presubmit config.

fweikert:
The configuration will be used by the Bazel CI pipeline at https://buildkite.com/bazel/rules-haskell-haskell
This commit is only a first step towards testing rules_haskell on CI. The configuration will likely require several more iterations until the build actually passes.

Progress towards https://github.com/bazelbuild/continuous-integration/issues/235


fweikert:
FYI @laurentlb

Profpatsch:
Thank you!

Gladly merging when @laurentlb says that works (how do we confirm?)

fweikert:
Usually the process looks like this:
1. You give us the email addresses of several maintainers 
2. We make them admins of the Buildkite pipeline
3. One of these maintainers adds the Buildkite pipeline as a GitHub PR check

We can (and should) execute steps 1 & 2. However, step 3 doesn't make sense yet since I don't think that the presubmit will pass, given the current state of the pipeline.
My recommendation for now: Merge this PR, then iterate on the config until the build passes, then enable it as a GitHub PR check.

PR #1068, opened 2019-09-03
stack_snapshot: Precise extra dependencies

### LINKED ISSUE #1062: stack_snapshot can exceed command-line length limit on C pre-processor with include paths
**Describe the bug**
Consider the following `stack_snapshot`:
```
stack_snapshot(
     name = "stackage",
     packages = ["pkg-a", "pkg-b", "pkg-c"],
     deps = ["//:cdep"],  # cdep is a cc_library
)
```
and the following `haskell_library`:
```
haskell_library(
    name = "hslib",
    srcs = ["UsesCPP.hs"],  # A Haskell source file using `LANGUAGE CPP`
    deps = ["@stackage//:pkg-a", "@stackage//:pkg-b", "@stackage//:pkg-c"],
    compiler_flags = ["-v"],
)
```
Then the include paths for `//:cdep` will appear three times on the command-line for the C pre-processor when building `//:hslib`.

Each package in the snapshot adds the include paths due to `deps` to it's package-db entry and GHC does not deduplicate these paths before forwarding them to the C pre-processor down the line. GHC passes the include paths on the command-line, not in a response file. If a package depends on many snapshot packages and the snapshot has a few C library dependencies, then this can exceed the command-line length limit and cause the following type of error:
```
    gcc: error trying to exec '/nix/store/3ajmsz0ab2dybi1qm7nk0izabyqwfl4r-gcc-7.4.0/libexec/gcc/x86_64-unknown-linux-gnu/7.4.0/cc1': execv: Argument list too long
    `cc_wrapper' failed in phase `C pre-processor'. (Exit code: 1)
```

**To Reproduce**
Check out digital-asset/daml@0e67f30a5fd9d09b9cc80b34835d8854bac90066 and execute
```
bazel build //libs-haskell/da-hs-base:da-hs-base
```

https://github.com/tweag/rules_haskell/commit/c89f628039532c036bf54ce00c5bd3371ef775c3 demonstrates the root of the issue in a simplified form.

**Expected behavior**
C library dependency include directories should not be multiplied on the command-line.

**Environment**
 - OS name + version: openSUSE Tumbleweed 20190529
 - Bazel version: Build label: 0.28.1- (@non-git)
 - Version of the rules: 627b329065a5c10e2f56303c38d684b364344b9b

**Additional context**
It seems that passing the include paths to Cabal as `--extra-include-dirs` flags cause Cabal to write them into the resulting package configuration file and package-db entry.

aherrmann:
Precisely control which packages to add additional dependencies to.

This PR changes `stack_snapshot` so that the `deps` attribute no longer takes a list of targets to add as dependencies to all packages in the snapshot. Instead, it takes a dictionary mapping Stack package names to lists of targets to precisely control which dependencies to add to which package in the snapshot.

### Motivation

This change is motivated by #1062. Simply adding all `deps` to all packages in the snapshot has the effect that the library-dirs and include-dirs of these `deps` will be added to the package-db entry of each package in the snapshot. GHC doesn't deduplicate the include-dirs and this leads to overflowing the maximum command-line length, even on Linux.

Unfortunately, we cannot avoid adding to the library-dirs and include-dirs of snapshot packages that actually depend on extra libraries. Their cabal files will list `extra-libraries` and `includes`, and `Setup.hs` enforces that these can be found in the library-dirs and include-dirs of the resulting package configuration.

An advantage is that this thins the dependency graph and gives the user more control. The downside is that `stack_snapshot` becomes more cumbersome to use.

### Alternatives

An alternative approach would be to automatically detect which snapshot packages depend on which extra libraries and only add the extra dependencies where necessary. This would require either parsing the Cabal files, or for `stack dot` (or `stack ls dependencies json`) to list `extra-libraries` dependencies. However, that would also require some user input or heuristics to map from `extra-libraries` to Bazel targets, and would assume that all `includes` are provided by targets providing `extra-libraries`.

Closes #1062

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -721,6 +721,9 @@ def _invert(d):
 
 def _from_string_keyed_label_list_dict(d):
     """Convert string_keyed_label_list_dict to label_keyed_string_dict."""
+
+    # TODO Remove _from_string_keyed_label_list_dict once following issue
+    # is resolved: https://github.com/bazelbuild/bazel/issues/7989.
     out = {}
     for (string_key, label_list) in d.items():
         for label in label_list:
@@ -732,6 +735,9 @@ def _from_string_keyed_label_list_dict(d):
 
 def _to_string_keyed_label_list_dict(d):
     """Convert label_keyed_string_dict to string_keyed_label_list_dict."""
+
+    # TODO Remove _to_string_keyed_label_list_dict once following issue
+    # is resolved: https://github.com/bazelbuild/bazel/issues/7989.
     out = {}
     for (label, string_key_list) in d.items():
         for string_key in string_key_list.split(" "):
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -719,6 +719,25 @@ def _invert(d):
     """Invert a dictionary."""
     return dict(zip(d.values(), d.keys()))
 
+def _from_string_keyed_label_list_dict(d):
+    """Convert string_keyed_label_list_dict to label_keyed_string_dict."""
+    out = {}
+    for (string_key, label_list) in d.items():
+        for label in label_list:
+            if label in out:
+                out[label] += " " + string_key
+            else:
+                out[label] = string_key
+    return out
+
+def _to_string_keyed_label_list_dict(d):
+    """Convert label_keyed_string_dict to string_keyed_label_list_dict."""
+    out = {}
+    for (label, string_key_list) in d.items():
+        for string_key in string_key_list.split(" "):
+            out.setdefault(string_key, []).append(label)
+    return out
+
 def _label_to_string(label):
     return "@{}//{}:{}".format(label.workspace_name, label.package, label.name)
 
@@ -757,7 +776,7 @@ def _stack_snapshot_impl(repository_ctx):
         vendored_packages,
     )
 
-    extra_deps = [_label_to_string(label) for label in repository_ctx.attr.deps]
+    extra_deps = _to_string_keyed_label_list_dict(repository_ctx.attr.deps)
     tools = [_label_to_string(label) for label in repository_ctx.attr.tools]
 
     # Write out dependency graph as importable Starlark value.
@@ -829,7 +848,10 @@ haskell_cabal_library(
                     version = package.version,
                     flags = package.flags,
                     dir = package.sdist,
-                    deps = package.deps + extra_deps,
+                    deps = package.deps + [
+                        _label_to_string(label)
+                        for label in extra_deps.get(package.name, [])
+                    ],
                     tools = tools,
                     visibility = visibility,
                 ),
@@ -853,7 +875,7 @@ _stack_snapshot = repository_rule(
         "packages": attr.string_list(),
         "vendored_packages": attr.label_keyed_string_dict(),
         "flags": attr.string_list_dict(),
-        "deps": attr.label_list(),
+        "deps": attr.label_keyed_string_dict(),
         "tools": attr.label_list(),
         "stack": attr.label(),
     },
@@ -932,7 +954,7 @@ _fetch_stack = repository_rule(
 )
 """Find a suitably recent local Stack or download it."""
 
-def stack_snapshot(stack = None, vendored_packages = {}, **kwargs):
+def stack_snapshot(stack = None, deps = {}, vendored_packages = {}, **kwargs):
     """Use Stack to download and extract Cabal source distributions.
 
     Args:
@@ -945,7 +967,7 @@ def stack_snapshot(stack = None, vendored_packages = {}, **kwargs):
         unpacked source distribution. Each package must contain a Cabal file
         named `<package-name>.cabal` in the package root.
       flags: A dict from package name to list of flags.
-      deps: Dependencies of the package set, e.g. system libraries or C/C++ libraries.
+      deps: Extra dependencies of packages, e.g. system libraries or C/C++ libraries.
       tools: Tool dependencies. They are built using the host configuration, since
         the tools are executed as part of the build.
       stack: The stack binary to use to enumerate package dependencies.
@@ -959,7 +981,7 @@ def stack_snapshot(stack = None, vendored_packages = {}, **kwargs):
           vendored_packages = {"split": "//split:split"},
           tools = ["@happy//:happy", "@c2hs//:c2hs"],
           snapshot = "lts-13.15",
-          deps = ["@zlib.dev//:zlib"],
+          deps = {"zlib": ["@zlib.dev//:zlib"]},
       )
       ```
       defines `@stackage//:conduit`, `@stackage//:lens`,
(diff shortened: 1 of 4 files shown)


regnat:
Really nice! This fixes a similar error on my side.
But for some reason, the error reappears if I apply this patch on top of 844a4e2b7f7dba2f2b7811ad47c01a181109e691, any idea what might cause the incompatibility?

regnat:
> Really nice! This fixes a similar error on my side.
> But for some reason, the error reappears if I apply this patch on top of [844a4e2](https://github.com/tweag/rules_haskell/commit/844a4e2b7f7dba2f2b7811ad47c01a181109e691), any idea what might cause the incompatibility?

This is actually fixed by https://github.com/tweag/rules_haskell/pull/1081, so it's not too much of an issue

aherrmann:
> Really nice! This fixes a similar error on my side.
> But for some reason, the error reappears if I apply this patch on top of [844a4e2](https://github.com/tweag/rules_haskell/commit/844a4e2b7f7dba2f2b7811ad47c01a181109e691), any idea what might cause the incompatibility?

GHC doesn't normalize paths. E.g. due to package environment files and `${pkgroot}` we might get a path like `path/to/pkg-env/../../../path/to/pkgroot/../../../path/to/lib`. These can grow longer than the absolute paths, so that's how #1063 can introduce that issue. #1039 would work around that by normalizing the paths.

aherrmann:
rebased on master

haskell/cabal.bzl
```diff
@@ -719,6 +719,25 @@ def _invert(d):
     """Invert a dictionary."""
     return dict(zip(d.values(), d.keys()))
 
+def _from_string_keyed_label_list_dict(d):
```
regnat: Nit: It's probably worth mentioning here that this function and the following ones exist only to work around https://github.com/bazelbuild/bazel/issues/7989

aherrmann: Good point, there's a comment down in `def stack_snapshot`, but it doesn't hurt to have it in these functions, too.

mboes:
> Unfortunately, we cannot avoid adding to the library-dirs and include-dirs of snapshot packages that actually depend on extra libraries. Their cabal files will list extra-libraries and includes, and Setup.hs enforces that these can be found in the library-dirs and include-dirs of the resulting package configuration.

The problem is that `--extra-include-dirs` seems to be cumulative. They will appear in the `include-dirs` line of the package's `.conf` file, and it appears that downstream packages are allowed to include headers from directories listed there. Take for example `integer-gmp`. The `include-dirs` will list the include folder where `gmp.h` can be found. Any package that depends on `integer-gmp` is allowed to `#include <gmp.h>` in source files. So if the user says `deps = {"integer-gmp": "@gmp//:include"}` we must allow  `#include <gmp.h>` in downstream packages. We would otherwise not be compatible with Cabal in that case. I can't remember in which package I first noticed this, but packages in the wild do exploit this Cabal "feature".

mboes:
#385 and #1008 showed how dangerous it can be to invoke precedent to name fields, but to use a different attribute type. It's usually a sign that the intended semantics is different from the one in precedents, in which case we should really use a different attribute name. So given your new semantics, I'm wary of keeping `deps` as the attribute name.

Since include dirs (and probably library dirs) are transitive, as noted in my previous comment, we're not gaining much cachability by letting the user state dependencies in a more fine grained way. The original motivation for this change was the absence of dedup was leading to long command-line length. Couldn't we instead try deduping ourselves in some place appropriate, instead of changing the API?

aherrmann:
> Couldn't we instead try deduping ourselves in some place appropriate, instead of changing the API?

We can try to deduplicate the list in the `cc_wrapper`. The downside is that the include paths semantics (`-I`/`-isystem`/...) are a bit complicated, so we need to be careful not to change the meaning of the command-line.

> So given your new semantics, I'm wary of keeping deps as the attribute name.

I'm happy to choose a different name. How about `extra_deps`?

> The problem is that `--extra-include-dirs` seems to be cumulative. They will appear in the `include-dirs` line of the package's `.conf` file, and it appears that downstream packages are allowed to include headers from directories listed there. [...] So if the user says `deps = {"integer-gmp": "@gmp//:include"}` we must allow `#include <gmp.h>` in downstream packages. We would otherwise not be compatible with Cabal in that case.

Note, this PR does not change those semantics. Cabal will still accumulate the include paths from all transitive package dependencies. 

> Since include dirs (and probably library dirs) are transitive, as noted in my previous comment, we're not gaining much cachability by letting the user state dependencies in a more fine grained way.

That depends on the particular library dependency. E.g. `grpc-haskell-core` depends on `libgrpc` and appears relatively close to the leaves of the dependency graph. With more finely grained dependencies changing `libgrpc` would not cause the rest of the stack snapshot to be rebuilt. 

Granted, most use-cases don't depend on this feature, and if it's really required the same can be achieved using `vendored_packages`.

mboes:
> Granted, most use-cases don't depend on this feature, and if it's really required the same can be achieved using vendored_packages.

Hm, that's a good point. Now that we have `vendored_packages`, what about deprecating `stack_snapshot.deps` altogether?

aherrmann:
> Now that we have `vendored_packages`, what about deprecating `stack_snapshot.deps` altogether?

That's an interesting idea. It places `stack_snapshot` further away from what a similar `stack.yaml` would look like. Defining C library dependencies is a common use-case and currently `vendored_packages` is a bit verbose, due to the `http_arcive` side. So, I think if we want to go that way we'd need to provide some helpers to reduce boilerplate. In particular, fetching the correct version according to the chosen snapshot and dependency resolution. We could define `stack_unpack` as a replacement for `http_archive` which takes care of fetching the correct version. That could look something like this:

```
stack_unpack(
    name = "zlib",
    snapshot = "lts-14.1",  # snapshot determines package version
    deps_snapshot = "stackage",  # loads dependencies of `zlib` from this `stack_snapshot`
    extra_deps = ["@z//:z"],
)
stack_snapshot(
    name = "stackage",
    snapshot = "lts-14.1",
    packages = [
        ...
    ]
    vendored_packages = { "zlib": ["@zlib//:zlib"] },
)
```

mboes:
> We could define stack_unpack as a replacement for http_archive which takes care of fetching the correct version.

I like the idea. But it looks like we'll run into other difficulties if we take that route. We have to state the snapshot twice. But worse, we need a circular dependency between the `stack_unpack` rule and the `stack_snapshot` one. That's pretty horrid. The way you work around it is by cunningly not making the `deps_snapshot` field a true label. But that's not very intuitive.

I guess `extra_deps` (of type dictionary), is the least bad option.

aherrmann:
> But worse, we need a circular dependency between the `stack_unpack` rule and the `stack_snapshot` one. That's pretty horrid. The way you work around it is by cunningly not making the `deps_snapshot` field a true label. But that's not very intuitive.

Agreed, that's a bit of a hack.

> I guess `extra_deps` (of type dictionary), is the least bad option.

Okay, I've renamed the `deps` attribute to `extra_deps`.

PR #1070, opened 2019-09-03
Allow multiple haskell_cabal_* targets per package

aherrmann:
Makes the output of `haskell_cabal_library|binary` depend on the target name. This is to avoid collisions with other targets in the same Bazel package and allow multiple Cabal targets per Bazel package.

Adds regression tests for two `haskell_cabal_library` targets within one Bazel package, and for a `haskell_cabal_library` and `haskell_cabal_binary` within the same Bazel package. The latter has a real world use-case in Hackage packages that have a library and an executable component where the user wants to access both from Bazel.


aherrmann:
CI was red after rebase. The type of `x` in `//tests/haskel_cabal_library:lib` had changed from `Integer` to `Int`, causing a type error in `//tests/haskell_cabal_library`. Fixed and force pushed.

PR #1073, opened 2019-09-03
CHANGELOG: update for 0.10

Profpatsch:
Towards https://github.com/tweag/rules_haskell/issues/1040

PR #1076, opened 2019-09-05
Remove `haskell.bzl`

guibou:
`haskell.bzl` is deprecated, but still wrong.


haskell/haskell.bzl
```diff
@@ -13,9 +13,12 @@ load(
     _haskell_repl_aspect = "haskell_repl_aspect",
     _haskell_test = "haskell_test",
     _haskell_toolchain = "haskell_toolchain",
-    _haskell_toolchain_libraries = "haskell_toolchain_libraries",
     _haskell_toolchain_library = "haskell_toolchain_library",
 )
+load(
+    ":private/haskell_impl.bzl",
+    _haskell_toolchain_libraries = "haskell_toolchain_libraries",
```
aherrmann: It's an internal rule that shouldn't be user facing. Why not remove it from this file altogether?
See https://github.com/tweag/rules_haskell/pull/1045/files

guibou: I totally agree to remove it if you want.

aherrmann: Yes, I think that's better.

guibou: Done.

aherrmann: > Why not remove it from this file altogether?

I meant to remove `haskell_toolchain_libraries` from `haskell/haskell.bzl`. On the other hand, [release 0.10](https://github.com/tweag/rules_haskell/pull/1073/files) will contain the deprecation of `haskell/haskell.bzl`, so it seems fine to me to have that file removed completely in the release that follows after. cc @Profpatsch

Profpatsch: Should this PR here go into 0.10 then? Or 0.11?

aherrmann: This PR should then go into 0.11.

PR #1085, opened 2019-09-10
Windows CI: Install stack & update index once

aherrmann:
Work around flakiness of `stack update` on Windows.

The rules_haskell workspace contains two separate `stack_snapshot` calls. This can lead to parallel calls to `stack update` which fails with "lock busy" errors on Windows CI. This PR works around that issue by installing stack globally and running `stack update` before the Bazel run.

See https://github.com/tweag/rules_haskell/commit/81e6b3ac27128c0ec1d980df9381b4dd7161cb13#r35008281

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -24,6 +24,20 @@ jobs:
     displayName: "Enable da long paths"
 
   - bash: |
+        # rules_haskell can automatically install stack if it is not in $PATH,
+        # yet. However, if the index has not been updated before building. Then
+        # multiple parallel invocations of `stack` will both attempt to update
+        # the index during build. This will fail with the following error.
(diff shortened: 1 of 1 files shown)


mboes:
Do we have some idea of how we could avoid the race condition in general?

aherrmann:
@mboes I was wondering if we could extract the `stack update` into a separate repository rule and make every instance of `stack_snapshot` depend on that. The `stack_snapshot` macro could take care of instantiating the `stack update` workspace if it doesn't exist, yet. Though, I guess for this to work right there would have to be some file dependency between the `stack update` and the `stack_snapshot` instances.

mboes:
I think that makes a lot of sense. Worth a try.

PR #1086, opened 2019-09-10
Fix index out-of-bounds in cabal wrapper

aherrmann:
- The `relative_to` function in the Cabal wrapper had an index out of bounds error in the case where one argument was a prefix of the other.
- This issue was not triggered so far. This PR adds a C dependency to `//tests/haskell_cabal_library:lib` that would have triggered that issue as a regression test.


aherrmann:
CI failed because of stack flakiness on Windows. Rebased on master where that's fixed.

PR #1087, opened 2019-09-10
cc_wrapper shorten include paths

aherrmann:
Add shortening of include paths (not just library search paths) to the cc_wrapper.

GHC doesn't normalize paths and concatenates the path components obtained from package environment files and `${pkgroot}` references in package configuration files. The result paths can exceed the maximum path lengths on some systems or combined the maximum argument length. The cc_wrapper already normalizes library search paths. This extends it to also normalize include directories.

See https://github.com/tweag/rules_haskell/issues/1062 and https://github.com/tweag/rules_haskell/pull/1068#issuecomment-529373727.


aherrmann:
@Profpatsch Sounds good. I'll have to fix some warnings so I'll do that in a separate PR.

aherrmann:
@Profpatsch mergify was stuck on the merge step. I've force-pushed to restart.

PR #1094, opened 2019-09-11
Add shellcheck for cc_wrapper_windows

aherrmann:
As requested [here](https://github.com/tweag/rules_haskell/pull/1087#pullrequestreview-286739452)

Depends on #1087


aherrmann:
@Profpatsch force pushed to trigger a rerun.

PR #1095, opened 2019-09-11
Only build required cabal component

aherrmann:
Before, e.g. `haskell_cabal_library` would also build `exe` components in a package. This is wasted build time and also requires all package components to build correctly even if only the library component is required. The latter can be an issue in case of vendored (patched) packages.


mboes:
@regnat does #1097 take into account this change?

PR #1099, opened 2019-09-13
WIP bazelci: fix initial build errors

Profpatsch:
Getting https://buildkite.com/bazel/rules-haskell-haskell/ to pass.

### DIFF
--- a/tests/library-linkstatic-flag/BUILD.bazel
+++ b/tests/library-linkstatic-flag/BUILD.bazel
@@ -94,11 +94,22 @@ sh_inline_test(
     script = """
 set -euo pipefail
 is_dynamic () {
-    # taken from https://github.com/NixOS/nixpkgs/blob/0b3f50f844e2a6b507b18d7c5259bb850b382f87/pkgs/build-support/setup-hooks/auto-patchelf.sh#L167-L170
-    readelf -l -- "$1" | grep -q "^ *INTERP\\>"
+    local res=1
+    if type file >/dev/null; then
+        file --dereference -- "$1"
+        file --dereference -- "$1" | grep -q "dynamically linked"
+        res=$?
+    else
+        echo 'file found in environment, please install either' >&2
+        exit 127
+    fi
+    return $res
 }
 
 for f in "$@"; do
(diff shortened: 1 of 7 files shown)


tests/library-linkstatic-flag/BUILD.bazel
```diff
@@ -94,11 +94,22 @@ sh_inline_test(
     script = """
 set -euo pipefail
 is_dynamic () {
-    # taken from https://github.com/NixOS/nixpkgs/blob/0b3f50f844e2a6b507b18d7c5259bb850b382f87/pkgs/build-support/setup-hooks/auto-patchelf.sh#L167-L170
-    readelf -l -- "$1" | grep -q "^ *INTERP\\>"
+    local res=1
+    if type file >/dev/null; then
+        file --dereference -- "$1"
+        file --dereference -- "$1" | grep -q "dynamically linked"
```
aherrmann: Why the double `file --dereference` instead of `set -o pipefail`?

tests/library-linkstatic-flag/BUILD.bazel
```diff
@@ -94,11 +94,22 @@ sh_inline_test(
...
+    else
+        echo 'file found in environment, please install either' >&2
+        exit 127
+    fi
+    return $res
 }
 
 for f in "$@"; do
+    if [[ ! -e "$f" ]]; then
+      printf "does not exist: %s" "$f" >&2
```
aherrmann: Should this cause a failure?

tests/library-linkstatic-flag/BUILD.bazel
Profpatsch: Makes it easier to debug if it should fail again in the future (it’s just a test, so the output doesn’t matter for anything else)

tests/library-linkstatic-flag/BUILD.bazel
Profpatsch: Good catch

Profpatsch:
Hm, now of course the Windows test is failing. I’ll revert to not exiting when a file is missing.

PR #1102, opened 2019-09-16
haskell_cabal_library: Add haddocks

regnat:
Build and export the haddocks for `haskell_cabal_library`es

This builds on top of #1097 because the cabal_wrapper part would have to
be rewritten otherwise (though it's minor and easy to port to master if
 #1097 gets delayed)

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -9,12 +9,14 @@ load(":private/expansions.bzl", "expand_make_variables")
 load(":private/mode.bzl", "is_profiling_enabled")
 load(":private/path_utils.bzl", "join_path_list", "truly_relativize")
 load(":private/set.bzl", "set")
+load(":haddock.bzl", "generate_unified_haddock_info")
 load(
     ":private/workspace_utils.bzl",
     _execute_or_fail_loudly = "execute_or_fail_loudly",
 )
 load(
     ":providers.bzl",
+    "HaddockInfo",
     "HaskellInfo",
     "HaskellLibraryInfo",
     "get_ghci_extra_libs",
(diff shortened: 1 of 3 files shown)


Profpatsch:
Rebased on master.

regnat:
@Profpatsch I've fixed the failure. Waiting for the CI to confirm, but it should be good now

Profpatsch:
@regnat do you still want to merge this?

regnat:
Well that's up to you ;)

I just raised the fact that it makes the build of the cabal packages slower. But if nobody's stepping against it then yes, please merge

mboes:
Haddock does add a fair amount of time to build times. Is there some way to make this optional?

regnat:
> Haddock does add a fair amount of time to build times. Is there some way to make this optional?

I don't think we can't make it like `haskell_library` where haddocks are built on-demand (because all the build is happening in a single bazel action and I'm not sure we can reasonably split this as that would mean messing up with cabal's internal buisness). But we can add an extra argument to `haskell_cabal_library` to indicate whether or not we want the haddocks to be built − at the cost of a bit of complexity in the code as this has some repercussions both in the build script and in the exported providers

mboes:
mmh, let's start by merging this. Provided we have an idea of the impact. Could you collect numbers about how much slower `bazel build @stackage//stack` is?

regnat:
So I benchmarked building `@stackage//:proto-lens` (because stack isn't exported in rules_haskell) both on this branch and on master, and strangely enough it's notably faster on this branch:

| Branch | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| haskell_cabal_library | 135.147 ± 4.914 | 131.490 | 148.002 | 1.0 |
| master  | 186.213 ± 3.220 | 183.164 | 193.916 | 1.00 |

I guess there has been another change on master making the build slower, I'll see whether I get the same results after a rebase

regnat:
(ftr the above results have been generated with `hyperfine --export-markdown "benchs_on_branch_$(git rev-parse --abbrev-ref HEAD).md" -p "bazel clean --expunge && rm -rf ~/.cache/bazel_disk_cache" "bazel build @stackage//:proto-lens"`)

regnat:
I did some new benchmarks after rebasing on top of master, and indeed we now have a slight slowdown on this branch:

| Branch | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| haskell_cabal_library (rebased) | 199.437 ± 3.021 | 196.493 | 206.374 | 1.00 |
| master  | 186.213 ± 3.220 | 183.164 | 193.916 | 1.00 |

So this makes a 7% slowdown on this example, which I think is acceptable

(on a side note, I'd like to know why master seems to be much slower than what it was some times ago. Maybe that's because of #1156 )

PR #1110, opened 2019-09-22
Python / Haskell tests: fix the type of imported functions

guibou:
This fixs a longstanding issue. hs_init was called without arguments,
which is wrong because the right prototype is:

int hs_init(int argc, char** argv)

This was leading to some invalid read in memory and sometimes it was
segfaulting.

I force the type in the prototype of hs_init and during the call to
add_one_hs


Profpatsch:
Oh wow, this would explain why it would segfault if there was no `print` statement before `hs_init`.

PR #1113, opened 2019-10-02
Fix haddock with package reexports

regnat:
- Don't try to run haddock again on the sources of the reexported packages
- Propagate the haddocs of the re-exported packages


Profpatsch:
I can’t say I have a great deal of insight in this, but looks good in general.

Issue #1096, opened 2019-09-12
`stack_snapshot` get invalidated too often


regnat:
**Describe the bug**

Dependencies built with `stack_packages` or `haskell_cabal_package` gets
invalidated at least each time the `PATH` changes due to the use of
`use_default_shell_env`.

**To Reproduce**

From the `examples` directory:

```sh
# Build a random cabal package
nix-shell --pure --run "bazel build @stackage//:random"
# Rebuild it again and see that it's cached
nix-shell --pure --run "bazel build @stackage//:random"
# Add `hello` to the `buildInputs` list in `shell.nix`
# Rebuild the random package and see that it isn't cached anymore
nix-shell --pure --run "bazel build @stackage//:random"
```

**Expected behavior**

A change in the environment shouldn't trigger a cache miss.

**Environment**
 - OS name + version: nixos unstable
 - Bazel version: 0.28.0 (from the example's shell.nix)
 - Version of the rules: latest master (79e2c5316bef9873f6d6a155f1f14f3d3ae39967)

mboes:
Did you mean `stack_snapshot`?

regnat:
> Did you mean `stack_snapshot`?

Oh yes, sorry, I'll update the title

mboes:
> each time the PATH changes due to the use of use_default_shell_env.

This is a side-effect of not having a shell toolchain. Cabal assumes quite a bit from the environment (in terms of available commands).

jkachmar:
Should #1097 and #1117 (which have since been merged) automatically address the problem, and should this issue be closed accordingly?

Or is there a need for some sort of regression test to make sure `stack_snapshot` dependencies are always cached?

aherrmann:
This should be fixed now. I've verified that the following does not cause a rebuild, while it did before those PRs were merged.
```
$ nix-shell --pure --run 'bazel build @stackage//:void'
$ nix-shell --pure --run 'PATH=foo:$PATH bazel build @stackage//:void'
```

PR #1117, opened 2019-10-07
Define Unix shell toolchain

aherrmann:
As became apparent in https://github.com/tweag/rules_haskell/pull/1097 Cabal requires standard Unix shell tools in `$PATH`. Packages with `./configure` scripts even more so. So far these standard tools were brought into scope by setting `use_default_shell_env=True`. However, as pointed out in https://github.com/tweag/rules_haskell/issues/1096 this can cause frequent cache invalidation.

This PR adds a new `unix_toolchain` which captures standard Unix shell tools as defined by [IEEE Std 1003.1-2008](https://en.wikipedia.org/wiki/List_of_Unix_commands). It comes with two configuration mechanisms.
* `unix_configure`: Which just discovers the available shell tools in `PATH` in the loading phase. Note, this suffers the same issue as described in #1096. This is called automatically with `haskell_register_ghc_bindists`.
* `unix_nixpkgs`: Which obtains the required tools from nixpkgs. This is called automatically with `haskell_register_ghc_nixpkgs`.

`unix_toolchain` assumes that the Unix tools are globally installed (Nix store, or `$PATH`).

`unix_toolchain` is used to amend `$PATH` in the Cabal wrapper. To prevent `./configure` scripts and Cabal builds from using the wrong C compiler, we point Cabal to `cc.tools.cc`. (This currently doesn't work on Windows, see comment in code)

@regnat @Profpatsch If this works then you could rebase https://github.com/tweag/rules_haskell/pull/1097 on this PR and use the `unix_toolchain` in the new Python Cabal wrapper.


haskell/cabal.bzl
```diff
@@ -109,6 +109,7 @@ def _prepare_cabal_inputs(hs, cc, dep_info, cc_info, component, package_id, tool
             "%{ghc_pkg}": hs.tools.ghc_pkg.path,
             "%{runghc}": hs.tools.runghc.path,
             "%{ar}": cc.tools.ar,
+            "%{cc}": cc.tools.cc,
```
guibou: Why is this necessary in the context of an unix toolchain? Is this an "unrelated but fixed at the same time" issue?

haskell/nixpkgs.bzl
```diff
@@ -242,6 +249,19 @@ def haskell_register_ghc_nixpkgs(
...
 
+    # Unix tools toolchain required for Cabal packages
+    unix_nixpkgs_kwargs = dict(
+        nixopts = nixopts,
+        repositories = repositories,
+        repository = repository,
+    )
+    if unix_attributes != None:
+        unix_nixpkgs_kwargs["packages"] = unix_attributes
+    unix_nixpkgs(
```
regnat: https://github.com/tweag/rules_haskell/pull/1117/files#diff-85408e6f9e4818174d8baf3df5946e07R456 guards the call to `unix_configure` depending on whether the toolchain already exists. Is there a reason why it isn't the case here too or is it just an overlook?

haskell/private/unix/unix_nixpkgs.bzl
```diff
@@ -0,0 +1,116 @@
+load("@bazel_tools//tools/cpp:lib_cc_configure.bzl", "get_cpu_value")
+load(
+    "@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl",
+    "nixpkgs_package",
+)
+
+def _unix_nixpkgs(name, packages, **kwargs):
```
regnat: nixpkgs already has `stdenv` which (I assume) contains a standard-compliant toolchain (as `stdenv.initialPath`). It would be nice to be able to directly reuse it, but that requires changing the interface a little here as `stdenv.initialPath` is a list of derivations. Any opinion on that?

haskell/private/unix/unix_toolchain.bzl
```diff
@@ -0,0 +1,49 @@
...
+        cmd: "Absolute path to the {} command or None.".format(cmd)
+        for cmd in unix_commands
+    },
+)
+
+def _unix_toolchain_impl(ctx):
+    commands = {}
+    for cmd in unix_commands:
+        cmd_path = getattr(ctx.attr, cmd, None)
+        if not cmd_path:
```
regnat: Maybe we should warn if some command is missing as this can lead to hard-to-understand failures at a later stage

regnat:
> maybe we should move it to its own repo

Or if it makes things easier, we could have it be a different workspace in the `rules_haskell` tree − but I'm not sure that bazel's handling would make this easier

haskell/cabal.bzl
aherrmann: Yes, the comment that @regnat quoted is the reason for this change.

haskell/nixpkgs.bzl
aherrmann: Yes, there's a reason. In this case `_ghc_nixpkgs_toolchain` is created with a hard-coded name, so this repository macro can only be invoked once. Therefore there's no need for this check.

However in `haskell_register_ghc_bindists` the workspace name contains the GHC version number, so it could be called multiple times for different GHC versions. In that case we need to check to avoid name collisions.

haskell/private/unix/unix_nixpkgs.bzl
aherrmann: Oh, I didn't know that, yeah it's probably better to reuse that.

haskell/private/unix/unix_toolchain.bzl
aherrmann: I think that would be pretty noisy. [That list (IEEE Std 1003.1-2008)](haskell/private/unix/unix_commands.bzl) contains a good number of tools that are not commonly installed on systems. E.g. `fort77`. I expect in most cases missing tools will cause failures in `./configure` scripts which should produce a reasonably clear error message.

aherrmann:
@guibuo
> Actually, I think you must alsa open an issue in bazel issue tracker to propose this API as well.

There is an open issue about this on [the Bazel tracker](https://github.com/bazelbuild/bazel/issues/5265) and a related discussion on [the mailing list](https://groups.google.com/d/msg/bazel-dev/Lo_DKft4Cpc/yiENPDUaAQAJ).

> Is this needed only because of nix? How is this done in other projects which are not dependent on nix but are all dependent on an unix toolchain?

To my knowledge there is no good solution to this problem right now, the issue above confirms that. Most projects simply assume that such tools are in `PATH`.

@regnat
>  wonder though whether rules_haskell is the right place to put it. [...]

I agree this would better fit into a standalone repository in the long run. I started it here since it's needed in the context of rules_haskell for now, and I think it's easier to iterate on this while it's in the same repository until we know that it meets our needs.

haskell/private/unix/unix_nixpkgs.bzl
```diff
@@ -0,0 +1,116 @@
...
+    nixpkgs_package(
+        name = name,
+        nix_file_content = """
+with import <nixpkgs> {{ config = {{}}; overlays = []; }};
```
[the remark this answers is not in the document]
aherrmann: `rules_nixpkgs` does not inherit `NIX_PATH` instead it constructs its own `NIX_PATH` based on the `repositories` or `repository` attribute. I based this code on the `nixpkgs_cc_configure` rule which also imports `<nixpkgs>`. That one does indeed `buildEnv`, so that might be a better approach.

mboes:
I have not reviewed this PR yet. I share @regnat's concern about how this should toolchain should be shared among all rule sets. IIRC there is a ticket about defining such a toolchain somewhere in the opened by @nlopezgi. I'd even suggest coordinating with upstream about what would be best to do here (how about a question on bazel-discuss@ ?).

The answer will likely be, go ahead with a private unix toolchain for now. In which case, it's okay to keep it in this repo for now.

haskell/private/unix/unix_nixpkgs.bzl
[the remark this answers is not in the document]
regnat: I didn't know that `nixpkgs_cc_configure` was using `<nixpkgs>`. Let's keep it that way then

haskell/private/unix/unix_toolchain.bzl
regnat: Weh didn't look closely enough at the list indeed :astonished:

aherrmann:
@Profpatsch CI fails with the following
```
Failures:
--
  |  
  | tests/RunTests.hs:70:7:
  | 1) failures //tests/failures/transitive-deps:lib-dFailure
  | uncaught exception: IOException of type InvalidArgument
  | fd:15: hGetContents: invalid argument (invalid byte sequence)
  |  
  | To rerun use: --match "/failures///tests/failures/transitive-deps:lib-dFailure/"
  |  
  | tests/RunTests.hs:70:7:
  | 2) failures //tests/failures/transitive-deps:lib-cFailure
  | uncaught exception: IOException of type InvalidArgument
  | fd:15: hGetContents: invalid argument (invalid byte sequence)
  |  
  | To rerun use: --match "/failures///tests/failures/transitive-deps:lib-cFailure/"
```

That failure already exists on `master` since #1115 see CI runs [#424](https://buildkite.com/tweag-1/rules-haskell/builds/424#39bec1bc-59c7-4ec6-8e47-64f5daac7e2f) and [#426](https://buildkite.com/tweag-1/rules-haskell/builds/426#b1430639-7b7e-4d12-a368-202587369698). So, it seems to be unrelated to this PR.

mboes:
Looping in @laszlocsomor as well, the author of https://github.com/bazelbuild/bazel/issues/5265.

laszlocsomor:
Thanks for looping me in! IIUC the toolchain has an attribute for every Unix tool's absolute path (or `None`). This allows calling those tools as Bazel actions. Neat!

I wonder if we could use this toolchain in Bazel, maybe move it into [Skylib](https://github.com/bazelbuild/bazel-skylib). The toolchain would supersede `ctx.actions.run_shell` in some cases. Has anyone an idea how we could use it for genrules?

aherrmann:
> Has anyone an idea how we could use it for genrules?

Toolchains can expose custom make variables. The toolchain could set one for each existing tool, e.g. `$(UNIX_GREP)`.

mboes:
@laszlocsomor it would be mucho good to move this upstream somewhere. As pointed out upthread in this discussion, the extra code introduced in this PR is basically all to address a very generic concern in Bazel, namely that `run_shell` with `use_default_shell_env = True` is non-hermetic but a necessary evil in very many cases in Bazel today.

Skylib looks like a good place to me, though I hope that the existence of this toolchain there won't be hard to discover for new rule developers. @aherrmann shall we merge this but immediately thereafter submit upstream there? @laszlocsomor let us know if on second thought there are better places for this.

aherrmann:
> @aherrmann shall we merge this but immediately thereafter submit upstream there?

Sounds good.

PR #1121, opened 2019-10-10
Start script 0.11

Profpatsch:


Profpatsch:
Note that I already pushed this to the `release` branch (and thus to the website), so that new users at HaskellX don’t have a broken experience (and that they can do `--use-nix` if they are on NixOS).

Issue #1126, opened 2019-10-16
Non reproducible manifest


guibou:
**Describe the bug**

The content of manifest generated by rules Haskell is not reproducible.

This technically leads to different input for all of the Haskell rules hence more unnecessary rebuilds and more cache misses.

**To Reproduce**

That's a bit tricky. I'm just explaining the reason.

We use `find` in a few places to list files (such as `.o` files). `find` will return files in an order which is unspecified and depends mostly on the underlying file system implementation and the history of file creation / modifications. See for yourself:

```
λ paddle test_dir → ls                                                                                                                                                                                               ~/test_dir
λ paddle test_dir → touch a b c                                                                                                                                                                                      ~/test_dir
λ paddle test_dir → find                                                                                                                                                                                             ~/test_dir
.
./a
./b
./c
λ paddle test_dir → rm a                                                                                                                                                                                             ~/test_dir
λ paddle test_dir → touch a                                                                                                                                                                                          ~/test_dir
λ paddle test_dir → find                                                                                                                                                                                             ~/test_dir
.
./b
./c
./a
```

**Expected behavior**

I propose to pipe the output of all these command to `sort`.

I'll soon open a PR with a solution.

guibou:
@Profpatsch Should we schedule an important release for this issue now that #1128 is merged?

Rules haskell main (if only) selling points are that it is reproducible and fast (thank to bazel caching).

Due to this issue, rules haskell was not reproducible (the order of files in the manifest influence the resulting binary) and not fast (because bazel caching was not working).

Profpatsch:
I wanted to make a point-release anyway (to support 1.0). We should do that.

aherrmann:
#1128 was merged but didn't auto close this. Closing manually.

mboes:
@guibou let's use more precise terms than "reproducibility". A build is reproducible if it is hermetic and deterministic. The problem here is non-determinism specifically. It's been present in the rules_haskell codebase since August 2018, so over a year. Clearly, it hasn't been a blocker for anyone in all this time, or we would have noticed it sooner. So the P0 "blocker: fix immediately!" priority label and a request for an immediate new release seems excessive to me, even if I'm happy to see that we managed to fix this in short order.

[...]

PR #1128, opened 2019-10-17
Sorted manifest files

guibou:
Fix for #1126.

`find` output is not stable, resulting on generated files with similar
content but in a different order.

This is not good for caching. The trivial solution is to use `sort` to
fix the output.


mboes:
The addition of `sort` can seem useless at first blush, but is actually an important fix. You should therefore add a comment just above it to clarify that this is necessary to make the action output deterministic and reference #1126.

guibou:
@mboes see https://github.com/tweag/rules_haskell/pull/1135

PR #1133, opened 2019-10-23
Windows support for Cabal python wrapper

aherrmann:
- Support Windows paths in Cabal python wrapper (e.g. `C:\foo\bar`)
- Look for executables using bazel-runfiles in Cabal python wrapper. Necessary on Windows when no symlinks are created in the sandbox and executables need to be looked up in the runtime manifest instead.
- Use Bazel 1.1.0 on Windows - required for `cc_wrapper` on Windows, see https://github.com/bazelbuild/bazel/issues/9390.
- Test `network` and `language-c` on Windows - these use `./configure` scripts and are more problematic on Windows.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -86,7 +86,8 @@ def _cabal_tool_flag(tool):
         return "--with-{}={}".format(tool.basename, tool.path)
 
 def _make_path(hs, binaries):
-    return ":".join([binary.dirname for binary in binaries.to_list()])
+    path_list_sep = ";" if hs.toolchain.is_windows else ":"
+    return path_list_sep.join([binary.dirname for binary in binaries.to_list()])
 
 def _prepare_cabal_inputs(hs, cc, unix, dep_info, cc_info, component, package_id, tool_inputs, tool_input_manifests, cabal, setup, srcs, flags, cabal_wrapper, package_database):
     """Compute Cabal wrapper, arguments, inputs."""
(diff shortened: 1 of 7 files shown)


haskell/cabal.bzl
```diff
@@ -86,15 +86,17 @@ def _cabal_tool_flag(tool):
...
+    return path_list_sep.join([binary.dirname for binary in binaries.to_list()])
 
 def _prepare_cabal_inputs(hs, cc, unix, dep_info, cc_info, component, package_id, tool_inputs, tool_input_manifests, cabal, setup, srcs, flags, cabal_wrapper, package_database):
     """Compute Cabal wrapper, arguments, inputs."""
     with_profiling = is_profiling_enabled(hs)
 
     (ghci_extra_libs, env) = get_ghci_extra_libs(hs, cc_info)
     env.update(**hs.env)
-    env["PATH"] = _make_path(hs, tool_inputs) + ":" + ":".join(unix.paths)
+    path_list_sep = ";" if hs.toolchain.is_windows else ":"
```
regnat: nit: You can move this to the toplevel to avoid having to repeat it

aherrmann: Good point, turns out this is a good opportunity to also refactor the path separator in `LD_LIBRARY_PATH`, so I'll do that in a separate PR.

aherrmann: Done in https://github.com/tweag/rules_haskell/pull/1134

PR #1141, opened 2019-11-13
Pipeline defined Bazel version

aherrmann:
- The bindists pipeline now defines the Bazel version it will use.
    Bazel is fetched as a single binary from GitHub.
- The nixpkgs pipeline confirms that its Bazel version matches.
- The bindists pipeline is defined in its own script file.
    This is to avoid shell escaping issues.
- Enables strict action env in the bindists pipeline.
    This avoids changes in `$PATH` invalidating cached builds with
    `use_default_action_env`. This works with bindists targets, but the
    nixpkgs pipeline still contains some targets that are incompatible
    with strict action env.


.buildkite/bindists-pipeline
```diff
@@ -0,0 +1,12 @@
...
+
+BAZEL_DIR="$(.buildkite/fetch-bazel-bindist)"
+trap "rm -rf '$BAZEL_DIR'" EXIT
+export PATH="$BAZEL_DIR:$PATH"
+echo "common:ci --build_tag_filters -requires_lz4,-requires_proto,-requires_zlib,-requires_doctest,-requires_c2hs,-requires_threaded_rts,-dont_test_with_bindist" > .bazelrc.local
+# XXX: See .bazelrc [backward compatible options] for the the rational behind this flag
+echo "build --incompatible_use_python_toolchains=false" >> .bazelrc.local
+echo "build:ci --experimental_strict_action_env" >> .bazelrc.local
+./tests/run-start-script.sh --use-bindists
+bazel build --config ci //tests/...
```
aherrmann: This bit is copied from `.buildkite/pipeline.yml`

.buildkite/bindists-pipeline
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+BAZEL_DIR="$(.buildkite/fetch-bazel-bindist)"
+trap "rm -rf '$BAZEL_DIR'" EXIT
+export PATH="$BAZEL_DIR:$PATH"
+echo "common:ci --build_tag_filters -requires_lz4,-requires_proto,-requires_zlib,-requires_doctest,-requires_c2hs,-requires_threaded_rts,-dont_test_with_bindist" > .bazelrc.local
+# XXX: See .bazelrc [backward compatible options] for the the rational behind this flag
+echo "build --incompatible_use_python_toolchains=false" >> .bazelrc.local
+echo "build:ci --experimental_strict_action_env" >> .bazelrc.local
```
mboes: There ought to be a note here that

> Enables strict action env in the bindists pipeline.

is a temporary workaround. We should no longer do anything that would require `strict_action_env`, now that we have the means to achieve that, via the rules_sh toolchain.

aherrmann: I added a comment to that effect.

.buildkite/pipeline.yml
```diff
@@ -5,6 +5,8 @@ steps:
       echo "build:ci --host_platform=@rules_haskell//haskell/platforms:linux_x86_64_nixpkgs" > .bazelrc.local
       nix-shell --arg docTools false --pure --run '
       set -e
+      # Ensure that the Nixpkgs and bindists Bazel versions match.
```
flokli: what about

```suggestion
      # Ensure that the Nixpkgs bazel version matches with what's specified in .buildkite/bazel-version
```

?

flokli: Do we need to run the same test in `.buildkite/bindists-pipeline` too?

flokli: I mean, we don't have to, because we do the fetching there, so maybe update the comment to

```suggestion
      # Ensure that the Nixpkgs bazel version matches with what's specified in .buildkite/bazel-version and fetched in `.buildkite/bindists-pipeline` for the bindists version
```

.buildkite/bindists-pipeline
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
flokli: This could have a comment stating it calls the `fetch-bazel-bindist` script, configures bazelrc, and adds fetched `bazel` to $PATH.

.buildkite/bindists-pipeline
```diff
@@ -0,0 +1,14 @@
...
+trap "rm -rf '$BAZEL_DIR'" EXIT
+export PATH="$BAZEL_DIR:$PATH"
+echo "common:ci --build_tag_filters -requires_lz4,-requires_proto,-requires_zlib,-requires_doctest,-requires_c2hs,-requires_threaded_rts,-dont_test_with_bindist" > .bazelrc.local
+# XXX: See .bazelrc [backward compatible options] for the the rational behind this flag
+echo "build --incompatible_use_python_toolchains=false" >> .bazelrc.local
+# XXX: Remove once all rules using `use_default_shell_env = True` have been
+#   converted to using `rules_sh`'s POSIX toolchain.
+echo "build:ci --experimental_strict_action_env" >> .bazelrc.local
+./tests/run-start-script.sh --use-bindists
+bazel build --config ci //tests/...
```
flokli: If the environment changes propagate to the next buildkite step, we probably should move this to `pipeline.yml`, like in the nixpkgs case.

aherrmann: I tried that initially and found that
1) You have to be careful to escape variables. E.g. `\$PATH` instead of `$PATH`.
2) `export PATH=...` didn't seem to be picked up and `bazel` called the globally installed version instead.

So, putting the steps into a dedicated script made things much easier and less error prone.

FWIW the [examples](https://github.com/buildkite/bash-example/blob/aa4f98e314c0350edc6b5bd6858c072ccc93f2ac/.buildkite/pipeline.yml#L3) recommend the same thing.

.buildkite/pipeline.yml
aherrmann: The version is used to construct the URL from which Bazel is fetch, so checking the version against that file once again is redundant.

I've expanded the comment as suggested.

.buildkite/bindists-pipeline
aherrmann: I've added a comment to that effect.

.buildkite/bindists-pipeline
flokli: I'm mostly concerned about maintaining those three lines:
```
./tests/run-start-script.sh --use-(bindists,nix)
bazel build --config ci //tests:run-tests
```

in two places.

What about moving that into a `build.sh` script, which assumes a right `bazel` to be in `$PATH`, and calling it from both pipelines?

We could have `build.sh` accepting the `--use-?` as a parameter.

aherrmann: Those lines are different between `nixpkgs` and bindists.

In the `nixpkgs` pipeline they are:
```
      ./tests/run-start-script.sh --use-nix
      bazel build --config ci //tests:run-tests
      bazel coverage //tests/... --config ci --build_tag_filters "coverage-compatible" --test_tag_filters "coverage-compatible" --test_output=all
```

And in the bindists pipeline they are:
```
./tests/run-start-script.sh --use-bindists
bazel build --config ci //tests/...
```

Once we've fixed what's missing for the bindist so it can run the same steps as nixpkgs then we should indeed unify these steps into something like `build.sh`.

flokli: Can you create a followup issue for that?

flokli:
I'd suggest to squash the commits into one, as the first one failed CI.

PR #1136, opened 2019-10-31
Import Unix toolchain from rules_sh

aherrmann:
- Replaces the Unix toolchain defined in `haskell/private/unix` by `rules_sh`.
- Updates `rules_nixpkgs` to use `nixpkgs_sh_posix_configure`.
- Moves `rules_haskell_toolchains` out of `haskell/repositories.bzl`
    Otherwise we get an import cycle:
    - `haskell/repositories.bzl` loads `ghc_bindist.bzl`
    - `ghc_bindist.bzl` requires `rules_sh` for `rules_haskell_toolchains`
    - `rules_sh` is defined in `haskell/repositories.bzl`

PR #1142, opened 2019-11-13
Update to Bazel 1.1

aherrmann:
**depends on #1141 change base branch before merging**

This is based on and replaces https://github.com/tweag/rules_haskell/pull/1065.

- Updates nixpkgs to include Bazel 1.1.
- Updates the Stackage snapshot as in #1065 .
- Updates `rules_nixpkgs` to include `nixpkgs_python_configure`.
- Uses `nixpkgs_python_configure` to avoid issues with the Python toolchain as described in https://github.com/tweag/rules_haskell/pull/1065.
- Uses the nixpkgs platform constraint defined by rules_nixpkgs instead of defining its own.
    Defines an alias to be backwards compatible to users depending on that platform constraint.

Closes https://github.com/tweag/rules_haskell/pull/1065.

PR #1143, opened 2019-11-14
Remove use_default_shell_env

aherrmann:
**depends on #1136 change base branch before merging**

- Remove all instances of `use_default_shell_env = True`.
  Replaces them using the POSIX toolchain provided by `rules_sh`.
- On Windows some tools are found in paths containing spaces, e.g. `C:\Program Files\...`.
  This requires additional quoting in some places.

### DIFF
--- a/.buildkite/bindists-pipeline
+++ b/.buildkite/bindists-pipeline
@@ -9,8 +9,8 @@ BAZEL_DIR="$(.buildkite/fetch-bazel-bindist)"
 trap "rm -rf '$BAZEL_DIR'" EXIT
 export PATH="$BAZEL_DIR:$PATH"
 echo "common:ci --build_tag_filters -requires_lz4,-requires_proto,-requires_zlib,-requires_doctest,-requires_c2hs,-requires_threaded_rts,-dont_test_with_bindist" > .bazelrc.local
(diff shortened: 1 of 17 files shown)


aherrmann:
> Depending on whether this or #1142 gets merged first,

This PR transitively depends on #1142 through #1136 

> the comment in the CI setup for the bindist case should be updated accordingly.

Yes, I'll rebase and adapt the comment. Unfortunately, we cannot remove the flag, yet. [`@com_google_protobuf` sets `use_default_shell_env = True`](https://github.com/protocolbuffers/protobuf/blob/d2d6ff51a8fb67391c44223c6b0b523110ad8d21/protobuf.bzl#L147).

mboes:
We should send them a PR to get them to use `rules_sh`. ;)

aherrmann:
Turns out they're not using `use_default_shell_env` for `PATH`, but for `LD_LIBRARY_PATH`, so `rules_sh` won't help there. See https://github.com/protocolbuffers/protobuf/pull/2933.

mboes:
I opened https://github.com/protocolbuffers/protobuf/issues/6886 to propose reverting protocolbuffers/protobuf#2933.

aherrmann:
netlify got stuck. Force pushed to kick it.

PR #1146, opened 2019-11-15
Disable warnings when building stackage dependencies

### LINKED ISSUE #1026: Is it possible to build stackage dependencies with warnings disabled?
The output of `stack_snapshot` is easily polluted by warnings (find an random example of one screen of messages during build).

So I'm wondering: is it possible to build the dependencies from the `stack_snapshot` macro with all warnings deactivated? With the reasoning that warnings of dependencies are never relevant.

```
INFO: From HaskellCabalLibrary @stackage//:http-api-data-0.4:

/home/aspiwack/.cache/bazel/_bazel_aspiwack/bd7cdf0b0195cd3d7f8d5aa71bc1df4e/sandbox/linux-sandbox/300/execroot/my_workspace/external/stackage/http-api-data-0.4/Setup.lhs:24:2: error:
     warning: #warning You are configuring this package without cabal-doctest installed. The doctests test-suite will not work as a result. To fix this, install cabal-doctest before configuring. [-Wcpp]
     #warning You are configuring this package without cabal-doctest installed. \
      ^~~~~~~
   |
24 | #warning You are configuring this package without cabal-doctest installed. \
   |  ^

/home/aspiwack/.cache/bazel/_bazel_aspiwack/bd7cdf0b0195cd3d7f8d5aa71bc1df4e/sandbox/linux-sandbox/300/execroot/my_workspace/external/stackage/http-api-data-0.4/Setup.lhs:24:2: error:
     warning: #warning You are configuring this package without cabal-doctest installed. The doctests test-suite will not work as a result. To fix this, install cabal-doctest before configuring. [-Wcpp]
     #warning You are configuring this package without cabal-doctest installed. \
      ^~~~~~~
   |
24 | #warning You are configuring this package without cabal-doctest installed. \
   |  ^

/home/aspiwack/.cache/bazel/_bazel_aspiwack/bd7cdf0b0195cd3d7f8d5aa71bc1df4e/sandbox/linux-sandbox/300/execroot/my_workspace/external/stackage/http-api-data-0.4/Setup.lhs:24:2: error:
     warning: #warning You are configuring this package without cabal-doctest installed. The doctests test-suite will not work as a result. To fix this, install cabal-doctest before configuring. [-Wcpp]
     #warning You are configuring this package without cabal-doctest installed. \
      ^~~~~~~
   |
24 | #warning You are configuring this package without cabal-doctest installed. \
   |  ^
INFO: From HaskellCabalLibrary @stackage//:warp-3.2.27:

Network/Wai/Handler/Warp/HTTP2/Request.hs:53:9: warning: [-Wdeprecations]
    In the use of `requestBody' (imported from Network.Wai.Internal):
    Deprecated: "requestBody's name is misleading because it only gets a partial chunk of the body. Use getRequestBodyChunk instead."
   |
53 |       , requestBody = body
   |         ^^^^^^^^^^^

Network/Wai/Handler/Warp/HTTP2/Worker.hs:250:25: warning: [-Wdeprecations]
    In the use of `requestBody'
    (imported from Network.Wai, but defined in Network.Wai.Internal):
    Deprecated: "requestBody's name is misleading because it only gets a partial chunk of the body. Use getRequestBodyChunk instead."
    |
250 |                 !body = requestBody req
    |                         ^^^^^^^^^^^
```

jcpetruzza:
Fixes #1026, following @mboes' suggestion there.


haskell/private/expansions.bzl
```diff
@@ -0,0 +1,20 @@
+def expand_make_variables(name, ctx, strings, extra_label_attrs):
+    # All labels in all attributes should be location-expandable.
+    label_attrs = [
+        ctx.attr.deps,
```
aherrmann: Why assume that all rules have a `deps` attribute instead of taking `label_attrs` as a parameter to `expand_make_variables`?

haskell/cabal.bzl
```diff
@@ -116,6 +117,7 @@ def _prepare_cabal_inputs(hs, cc, posix, dep_info, cc_info, component, package_i
     ]
     args.add_all([component, package_id, setup, cabal.dirname, package_database.dirname])
     args.add("--flags=" + " ".join(flags))
+    args.add_all(sorted(compiler_flags), format_each = "--ghc-option=%s")
```
aherrmann: Why sort the flags? Is that generally safe?

haskell/private/expansions.bzl
jcpetruzza: I was under the impression that these particular attributes are common to all build rules (https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes), so it felt redundant/error-prone to pass them all the time.

Can change it back if you'd rather this to be explicit!

haskell/cabal.bzl
jcpetruzza: I was trying to avoid recompilation on cosmetic reordering of flags, but you are right, this could change the meaning of the flags! Will fix

haskell/private/expansions.bzl
aherrmann: Fair enough, thanks for explaining. I have no strong preference either way.

jcpetruzza:
Not sure I understand why mergify complains here. I thought it was missing a rebase but that didn't help. Do I need to fix something on our side?

aherrmann:
Tried a re-approve and re-run on the "checks" tab. Neither worked so I merged manually.

PR #1149, opened 2019-11-18
Make cat_hs example independent of Nix

aherrmann:
The `cat_hs` example required Nix to fetch `zlib`. However, `zlib` can also be built from source using Bazel. This PR does that to make the `cat_hs` example available to users who don't have Nix installed.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -10,8 +10,7 @@ for building Haskell code.
   with multiple third-party dependencies downloaded from Hackage,
   C library dependencies and split up into multiple libraries and
   a binary. We use a rule wrapping Cabal to build the Hackage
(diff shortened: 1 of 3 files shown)


mboes:
See https://github.com/protocolbuffers/protobuf/blob/master/protobuf_deps.bzl#L16-L23.

aherrmann:
> See https://github.com/protocolbuffers/protobuf/blob/master/protobuf_deps.bzl#L16-L23.

Interesting, though that one is [subject to Bazel library name mangling](https://github.com/protocolbuffers/protobuf/blob/3cc55d6167e086aef90710874d8de4ca68197f2f/third_party/zlib.BUILD#L32), which would break Cabal.

PR #1155, opened 2019-11-21
Copy Cabal sources on Windows

aherrmann:
This is to avoid spurious rebuilds due to changes in the source tree during configure steps invalidating the cache.

>    On Windows Bazel builds are not sandboxed and operations like executing
    configure scripts can modify the source tree. If the `srcs` attribute uses
    a glob like `glob(["**"])`, then these modified files will enter `srcs` on
    the next execution and invalidate the cache. To avoid this we copy the
    source tree on Windows.

**EDIT:** Not quite right yet. Currently, this only copies the directory containing the Cabal file, which breaks if the package depends on files (e.g. headers) from higher up in the source tree. E.g. `grpc-haskell-core` does this.

### DIFF
--- a/haskell/private/cabal_wrapper.py.tpl
+++ b/haskell/private/cabal_wrapper.py.tpl
@@ -97,6 +97,42 @@ def recache_db():
 recache_db()
 
 @contextmanager
+def in_srctree():
+    """Change into the source directory.
+
+    Cabal really wants the current working directory to be directory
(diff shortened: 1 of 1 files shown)


aherrmann:
A lower cost work-around for the issue is to run `Setup.hs clean` after the build on Windows. This will clean up files generated during `configure` and `build`.

Closing in favor of https://github.com/tweag/rules_haskell/pull/1156

PR #1156, opened 2019-11-22
Cabal: Run setup clean on Windows

aherrmann:
This is to avoid spurious rebuilds due to changes in the source tree during configure steps invalidating the cache.

>       Bazel builds are not sandboxed on Windows and operations like
>       executing `configure` scripts can modify the source tree. If the
>       `srcs` attribute uses a glob like `glob(["**"])`, then these modified
>       files will enter `srcs` on the next execution and invalidate the
>       cache. To avoid this we remove generated files.

### DIFF
--- a/haskell/private/cabal_wrapper.py.tpl
+++ b/haskell/private/cabal_wrapper.py.tpl
@@ -105,7 +105,7 @@ def tmpdir():
     try:
         yield distdir
     finally:
-        shutil.rmtree(distdir, ignore_errors = "%{is_windows}" == "True")
+        shutil.rmtree(distdir, ignore_errors = True)
 
 with tmpdir() as distdir:
     enable_relocatable_flags = ["--enable-relocatable"] \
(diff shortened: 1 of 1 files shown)


mboes:
> I think ctx can tell you whether we are in sandboxed mode or not.

Perhaps not. Can't find it anywhere.

aherrmann:
> > I think ctx can tell you whether we are in sandboxed mode or not.
> 
> Perhaps not. Can't find it anywhere.

Yeah, I can't find it anywhere either.

aherrmann:
> Absence of sandboxing isn't just a Windows issue. Linux non-sandboxed builds are possible too (it's the so called "standalone" mode I think). 

Good point, I've changed it to generally run `Setup.hs clean`.

mboes:
It's a small pessimization because `Setup.hs clean` is slow, but I agree - running it always is the least bad option.

PR #1161, opened 2019-11-25
haskell/gen_ghc_bindist.py: use python3 in shebang

flokli:
We use python3 in all other python shebangs in the repo anyways, so it's
reasonable to assume we require python3 to be present.

Debian-style distros didn't yet alias python to python3, but provide it
in the python2 package.

I don't really think we should require Python 2 too, but just Python 3
here aswell.

### DIFF
--- a/haskell/gen_ghc_bindist.py
+++ b/haskell/gen_ghc_bindist.py
@@ -1,10 +1,8 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # This is a happy-path tool to download the bindist
 # download paths and hashes, for maintainers.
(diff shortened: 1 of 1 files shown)


haskell/gen_ghc_bindist.py
```diff
@@ -1,10 +1,4 @@
-#!/usr/bin/env python
-
-# This is a happy-path tool to download the bindist
-# download paths and hashes, for maintainers.
-# It uses the hashes provided by download.haskell.org.
```
aherrmann: Why remove that comment?

flokli: That was a mistake. Updated.

Issue #1160, opened 2019-11-25
document dependencies of GHC bindist


flokli:
When building GHC with bindists on a vanilla ubuntu, there's a lot of assumptions about what packages need to be installed  (`gcc`, `make`, `build-essential`, `libtinfo`).

If it's not possible to provide these via bazel, so we should at least document what needs to be installed.

flokli:
We also probably shouldn't list just `build-essential`, that's just a meta-packages and won't help non-debian distros much.

PR #1162, opened 2019-11-25
Document system dependencies for bindist

aherrmann:
Closes https://github.com/tweag/rules_haskell/issues/1160

Documents required system dependencies for GHC bindist on Unix and Windows.

I've also matched the Bazel version requirement in the README with that of the start script.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,10 @@ You'll need [Bazel >= 0.29][bazel-getting-started] installed.
 
 If you are on NixOS, skip to the [Nixpkgs](#Nixpkgs) section.
 
+### System dependencies
+
+Refer to the "Before you begin" section in [the documentation](docs/haskell.rst).
+
 ### The easy way
 
 In a fresh directory, run:
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ The full reference documentation for rules is at https://haskell.build.
 
 ## Setup
 
-You'll need [Bazel >= 0.27][bazel-getting-started] installed.
+You'll need [Bazel >= 0.29][bazel-getting-started] installed.
 
 If you are on NixOS, skip to the [Nixpkgs](#Nixpkgs) section.
 
--- a/docs/haskell.rst
+++ b/docs/haskell.rst
@@ -25,9 +25,26 @@ In this tutorial you'll learn how to:
 Before you begin
 ----------------
 
-To prepare for the tutorial, first `install Bazel`_ if you don't have
-it installed already. Then, retrieve the ``rules_haskell`` GitHub
-repository::
+On a Unix system you will need the following tools installed.
+
+* ``gcc``
+* ``libffi``
+* ``libgmp``
+* ``libtinfo5``
+* ``make``
+* ``python3`` (``python`` also needs to be available in ``$PATH``. Depending on your distro, this might require installing the ``python`` meta-package, which might use Python 2 or 3, ``rules_haskell`` works with both.)
+
+On Ubuntu you can obtain them by installing the following packages. ::
+
+  build-essential libffi-dev libgmp-dev libtinfo5 libtinfo-dev python python3
+
+On Windows you will need.
+
+- ``msys2``
+- ``python3``
+
+Next, `install Bazel`_ if you don't have it installed already. Then, retrieve
+the ``rules_haskell`` GitHub repository::
 
   git clone https://github.com/tweag/rules_haskell/
 


README.md
```diff
@@ -22,7 +22,24 @@ The full reference documentation for rules is at https://haskell.build.
 
 ## Setup
 
-You'll need [Bazel >= 0.27][bazel-getting-started] installed.
+You'll need [Bazel >= 0.29][bazel-getting-started] installed.
+
+If you are on NixOS, skip to the [Nixpkgs](#Nixpkgs) section.
```
flokli: The exact same line below the windows section should probably be removed.

README.md
```diff
@@ -22,7 +22,24 @@ The full reference documentation for rules is at https://haskell.build.
...
+
+### System dependencies
+
+On a Unix system you will need the following tools installed.
+
+- `gcc`
```
flokli: On Ubuntu, I ended up with the following packages:

`build-essential libtinfo5 libtinfo-dev python libgmp-dev libffi-dev`

Not included are the ones silently pulled in by the base distribution, obviously.

flokli:
We should also add this to `docs/haskell.rst`, as it gets rendered to https://rules-haskell.readthedocs.io/en/latest/haskell.html (and the `README` doesn't).

README.md
aherrmann: Good catch. Removed.

README.md
aherrmann: Thanks, I've added a line with the Ubuntu packages.

aherrmann:
> We should also add this to `docs/haskell.rst`, as it gets rendered to https://rules-haskell.readthedocs.io/en/latest/haskell.html (and the `README` doesn't).

Good point. I've added it to the "before you begin" section.

README.md
```diff
@@ -22,10 +22,32 @@ The full reference documentation for rules is at https://haskell.build.
...
+- `libffi`
+- `libgmp`
+- `libtinfo5`
+- `make`
+- `python3`
+
+On Ubuntu you can obtain them by installing the following packages.
+
+```
+build-essential libffi-dev libgmp-dev libtinfo5 libtinfo-dev python
```
flokli: On Ubuntu, `python` means `python2`. `python3` seems to be already installed (and looking at the shebangs in the repo, I assume we also need it ;-)).

Should we explicitly list Python 2 and Python 3 here? Or how do the bindists work? Do they ship their own Python?

aherrmann: @flokli I took that list of Ubuntu packages from here https://github.com/tweag/rules_haskell/pull/1162#discussion_r350675766.
I've replaced `python` with `python3`.

docs/haskell.rst
```diff
@@ -25,9 +25,26 @@ In this tutorial you'll learn how to:
...
-it installed already. Then, retrieve the ``rules_haskell`` GitHub
-repository::
+On a Unix system you will need the following tools installed.
+
+* ``gcc``
+* ``libffi``
+* ``libgmp``
+* ``libtinfo5``
+* ``make``
+* ``python3`` (``/usr/bin/env python`` needs to be available)
```
flokli:
```suggestion
* ``python3`` (``python`` needs to be available in ``$PATH`` as well, depending on your distro, this might require ``python2`` too)
```

docs/haskell.rst
```diff
@@ -25,9 +25,26 @@ In this tutorial you'll learn how to:
...
-it installed already. Then, retrieve the ``rules_haskell`` GitHub
-repository::
+On a Unix system you will need the following tools installed.
+
+* ``gcc``
+* ``libffi``
+* ``libgmp``
+* ``libtinfo5``
+* ``make``
+* ``python3`` (``python`` also needs to be available in ``$PATH``, depending on your distro, this might require ``python2`` too)
```
flokli:
```suggestion
* ``python3`` (``python`` also needs to be available in ``$PATH``, depending on your distro, this might require installing  the ``python`` meta-package too (which might be Python 2 or 3, ``rules_haskell`` works with both)
```

docs/haskell.rst
```diff
@@ -25,9 +25,26 @@ In this tutorial you'll learn how to:
...
+
+On Ubuntu you can obtain them by installing the following packages. ::
+
+  build-essential libffi-dev libgmp-dev libtinfo5 libtinfo-dev python python3
```
flokli:
```suggestion
  build-essential libffi-dev libgmp-dev libtinfo5 libtinfo-dev python python3
```

docs/haskell.rst
aherrmann: There was a double parenthesis issue, so I've addressed it in another commit https://github.com/tweag/rules_haskell/pull/1162/commits/e335753457d08aa6204cd5a1ad70d56cd31cbc5e

README.md
flokli: This got out of sync with `docs/haskell.rst`. Can we only describe it in `docs/haskell.rst`, and refer to that file in here?

aherrmann: Good point, done.

aherrmann:
Merging manually as mergify is not picking this up.

PR #1166, opened 2019-12-03
Add fail package to the empty package blacklist

jcpetruzza:
`fail` is an adapter package backporting `MonadFail`. On newer versions of `base`, it will produce no output, so it needs to be added to the `EMPTY_PACKAGES_BLACKLIST` in order to build any of its reverse dependencies from `stack_snapshot()`


regnat:
Thanks!

Issue #1165, opened 2019-12-02
Support package database containing long module name in GHC 8.8


guibou:
**Describe the bug**

Since GHC 8.8 the package database format seems to have changed a bit and contains wrapped lines.

For example, here is a chunk of the entries for `data-default-instances-old-locale` in ghc `8.6.5` and `8.8.1`

Dump generation:

```
λ paddle ~ → nix-shell -p "haskell.packages.ghc865.ghcWithPackages(p: [p.data-default-instances-old-locale])"
[nix-shell:~]$ ghc-pkg dump > pkg865
```
```
λ paddle ~ → nix-shell -p "haskell.packages.ghc881.ghcWithPackages(p: [p.data-default-instances-old-locale])"
[nix-shell:~]$ ghc-pkg dump > pkg881
```

```
λ paddle ~ → cat pkg865| grep -A5 'name:.*data-default-instances-old-locale'                        
name: data-default-instances-old-locale
version: 0.0.1
id: data-default-instances-old-locale-0.0.1-GRzcBvmea5LLGj3CNDMpIw
key: data-default-instances-old-locale-0.0.1-GRzcBvmea5LLGj3CNDMpIw
license: BSD-3-Clause
maintainer: <l.mai@web.de>
```

```
λ paddle ~ → cat pkg881| grep -A5 'name:.*data-default-instances-old-locale'
name:                 data-default-instances-old-locale
version:              0.0.1
visibility:           public
id:
    data-default-instances-old-locale-0.0.1-26yQMJ7x0xCF8E8NsOHD2F
```

You can observe the newly introduced spaces and line break.

This breaks the assumptions done here: 

https://github.com/tweag/rules_haskell/blob/master/haskell/private/ls_modules.py#L30-L32

Especially, this file supposes that the content of the `id` field appears on the same line.

**To Reproduce**

I pushed in commit df22035013857f420df95e75a349617d1caa5a8d a reproduction.

**Expected behavior**

This is not supposed to crash.

**Additional context**

There are two solutions to fix the issue:

- Find a way to reset `ghc-pkg dump` to the old behavior. I haven't investigated yet.
- Write a more robust parser in `ls_modules.py`. I have a diff for that ready and running in production on a client codebase. I'll investigate the reasons of this change in `ghc-pkg dump` tomorrow and based on the results, I'll submit a PR.

guibou:
[...]

PR #1169, opened 2019-12-06
Support for package database with long package name

guibou:
This is a fix for #1165

GHC 8.8 changed the output format of `ghc-pkg dump` and introduce line
breaks in `id:` fields.

The parser was simply using `.split(':')` on line containing `id:`. This
no longer work.

This commit introduces a new parser which parses line by line the file
until entering a field description.

The error only appears with GHC 8.8, so I have no reproduction code
available on the current codebase. However the GHC 8.8 branch for
rules_haskell #1170 contains a non regression test for this bug.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -174,10 +174,16 @@ load(
 )
 
 haskell_register_ghc_nixpkgs(
-    attribute_path = "ghc",
+    attribute_path = "",
     compiler_flags = test_compiler_flags,
     haddock_flags = test_haddock_flags,
     locale_archive = "@glibc_locales//:locale-archive",
+
+    # The data-default-instances-old-locale package was is a reproduction sample for #1165
+    nix_file_content = """
+    with import <nixpkgs> {};
+    haskellPackages.ghcWithPackages(p: [p.data-default-instances-old-locale])
+    """,
     repl_ghci_args = test_repl_ghci_args,
     repository = "@nixpkgs_default",
     version = test_ghc_version,
--- a/haskell/private/ls_modules.py
+++ b/haskell/private/ls_modules.py
@@ -26,10 +26,60 @@ hidden_modules_file = sys.argv[4]
 reexported_modules_file = sys.argv[5]
 results_file = sys.argv[6]
 
+def read_config_file(f):
+    """
+    Read GHC package database format, which is mostly key:values lines
+
+    `f` is an open file descriptor
+    it returns a generator of (key, [values])
+    """
+
+    current_field_name = None
+    current_field_values = []
+
+
+    # we parse the file line by line, looking for "fieldName:" markers
+    # and then consuming all the following line if they are indented.
+    for line in f:
+        # we have an indented line, that's a field
+        if line.startswith("    "):
+            line = line.strip()
+            if line:
+                current_field_values.append(line.strip())
+        else:
+            # Dump the currently read field
+            if current_field_name:
+                yield (current_field_name, current_field_values)
+
+            # this is a new field
+            if ':' in line:
+               items = line.split(':')
+               current_field_name = items[0].strip()
+
+               if len(items) == 2:
+                   maybefield = items[1].strip()
+
+                   if maybefield:
+                       current_field_values = [maybefield]
+                   else:
+                       current_field_values = []
+            else:
+                current_field_name = None
+                current_field_values = []
+
 with io.open(global_pkg_db_dump, "r", encoding='utf8') as f:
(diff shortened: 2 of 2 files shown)


mboes:
We already have code to unfold multi-line fields into a single line (see `unfold_fields()`). Can that not be reused? Also, add a test. Even if so long as we don't switch to GHC 8.8, the test will always pass even without the fix.

guibou:
@mboes test added

About `unfold_fields` versus my implementation of `read_config_file`, this function is:

- badly documented: it says that it returns pairs of key value when it is not the case
- not accessible from outside: `pkgdb_to_bzl.py` is not importable as a module. To do so, the top level instructions must be gathered together and protected by an `if __name__ == '__main__'`.
- less robust: may file if first line is indented
- does not return sanitized output (`.strip()`)

Doing this refactor is outside of the scope of this PR from my point of view. I'm opening #1174 to track this refactoring opportunity.

WORKSPACE
```diff
@@ -174,10 +174,16 @@ load(
...
+    attribute_path = "",
     compiler_flags = test_compiler_flags,
     haddock_flags = test_haddock_flags,
     locale_archive = "@glibc_locales//:locale-archive",
+
+    # The data-default-instances-old-locale package was is a reproduction sample for #1165
+    nix_file_content = """
+    with import <nixpkgs> {};
+    haskellPackages.ghcWithPackages(p: [p.data-default-instances-old-locale])
+    """,
```
aherrmann: Would it work to pull in `data-default-instances-old-locale` through `stack_snapshot` instead and have a proper test case under `//tests` for this? That would also make the test case available in the bindist version and avoid increasing the difference between the nix and bindist versions.

aherrmann: Nevermind, just realised, that `ls_modules` is not used on `stack_snapshot`...

guibou: Well, for the "//tests" case, I have no idea, this is supposed to fail when parsing the package database, so I don't really know how we can write a "test" for it.
About `stack_snapshot`, I don't know either. Does `stack_snapshot` uses the package database?

mboes:
@guibou of the four points you make about `unfold_fields`, only one (outdated return type in documentation) seems relevant, and is easily fixed with a single line documentation patch. The `ghc-pkg` database format doesn't allow for fields to start indented, so I'm not sure I understand that point. I don't see why we care about importing `pkgdb_to_bzl.py` as a module, and sanitizing output (whatevery is really necessary here) doesn't like the job of a simple line unfolding function. Crucially, neither of the four points seem to support replacing a simple 7 line function with a 30 line function featuring 4 levels of `if-then-else` nesting (inside a `for`-loop).

Does simply reusing `unfold_fields` work well enough for your test case? If so, can we do that? Which of course doesn't preclude making improvements to that function later. I'd prefer to think harder before we say yes to adding more complicated control flow to this file.

guibou:
> @guibou of the four points you make about unfold_fields, only one (outdated return type in documentation) seems relevant, and is easily fixed with a single line documentation patch.

Done in #1197

> The ghc-pkg database format doesn't allow for fields to start indented, so I'm not sure I understand that point.

You are right.

> I don't see why we care about importing pkgdb_to_bzl.py as a module

Do you suggest we move `unfold_fields` to a proper module instead? I'm totally fine with that solution too.

PR closed because I'm the only interested user of bug #1165, and the discussion is taking more time than the free time I have available to fix the issue.

mboes:
> Do you suggest we move unfold_fields to a proper module instead? I'm totally fine with that solution too.

I'm confused by this question. Could you remind me again what problem we're trying to solve with a module? Currently `pkgdb_to_bzl.by` is a straight line script with no modules. Do you need to reuse part of the code in another script somewhere?

> I'm the only interested user of bug #1165

I likewise want to see this fixed. But not at the cost of more code to maintain with complex control flow when a simpler solution works just fine.

guibou:
> I'm confused by this question. Could you remind me again what problem we're trying to solve with a module? Currently pkgdb_to_bzl.by is a straight line script with no modules. Do you need to reuse part of the code in another script somewhere?

I'm confused too, didn't you suggest to use `unfold_fields` from `pkgdb_to_bzl.py` into `ls_modules.py`?

mboes:
Oh I see. This is modifying `ls_modules.py`, not `pkgdb_to_bzl.py`. Understood. Well, duplicating a 7 line function is fine too. But this points to the need to introduce down the line a simple "ghc db processing library" in Python, which both scripts could call.

Issue #1079, opened 2019-09-05
Changes in stack setup are not visible without bazel clean --expunge


guibou:
**Describe the bug**

When adding:

```
load("@rules_haskell//tools:repositories.bzl", "rules_haskell_worker_dependencies")
rules_haskell_worker_dependencies()
```

In my repository, `bazel fetch //...` tries to fetch the `stack` tool, succeed at fetching it, but does not do anything to patch the library loader to make it compatible with nixos, resulting in the classic "no such file or directory" error.

A workaround on nixos is to provide a working `stack` executable in the current `PATH`. However when adding `stack` in the environment, I need to `bazel clean --expunge` so that the change can be observed.

**To Reproduce**

Change the `stack` in your environment (e.g. `PATH`) (remove it, make it point to another binary) and observe that the next bazel call won't take that change into account.

Profpatsch:
I don’t think we support stack on NixOS.

PR #1170, opened 2019-12-06
Support for GHC 8.8

guibou:
Work in progress for GHC 8.8 support in rules_haskell.

Rules_haskell support GHC 8.8 natively (if #1165 is fixed), but the test suite does not.

This also includes a reproduction test for #1165

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -41,9 +41,9 @@ http_archive(
 load("@rules_haskell//haskell:cabal.bzl", "haskell_cabal_binary")
 haskell_cabal_binary(name = "happy", srcs = glob(["**"]), visibility = ["//visibility:public"])
     """,
-    sha256 = "22eb606c97105b396e1c7dc27e120ca02025a87f3e44d2ea52be6a653a52caed",
-    strip_prefix = "happy-1.19.10",
-    urls = ["http://hackage.haskell.org/package/happy-1.19.10/happy-1.19.10.tar.gz"],
+    sha256 = "fb9a23e41401711a3b288f93cf0a66db9f97da1ce32ec4fffea4b78a0daeb40f",
+    strip_prefix = "happy-1.19.12",
+    urls = ["http://hackage.haskell.org/package/happy-1.19.12/happy-1.19.12.tar.gz"],
 )
 
 http_archive(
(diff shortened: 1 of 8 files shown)


mboes:
It's useful to open these as a [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/).

guibou:
I've upgraded the PR.

Note that rules_haskell is still affected by #1165 , but I removed the testcase from this PR.

During the developement of this PR, I had many occurences of #1079 and I had to call `bazel clean --expunge` many times.

guibou:
Note: I hope for the best, but I also observed surprising behaviors:
- `bazel test` needs to be called many times to get a 100% test success
- `//tests/cc_haskell_import:hs-lib-b` and `//tests/haskell_import:add-one` are sometime stuck indefinitly. Killing the `bazel` process and running it again fixs it.

This should be investigated, but I still wanted to push the updates so it may

aherrmann:
This now also includes a Bazel update to 2.1. Is this necessary for GHC 8.8, or could that be done separately in its own PR? In any case, the bindist tests are failing on CI because the sha256 are not up-to-date, see [`.ci/update-bazel-version`](https://github.com/tweag/rules_haskell/blob/d59ccf9e7785553fd392924fb083d926c7e01134/.ci/update-bazel-version). 

---

Darwin with nixpkgs is now [failing with](https://circleci.com/gh/tweag/rules_haskell/8773?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link)
```
error: unknown argument: '-traditional'
error: unknown argument: '-Xpreprocessor'
```

---

> Note: I hope for the best, but I also observed surprising behaviors:
> 
> * `bazel test` needs to be called many times to get a 100% test success
> * `//tests/cc_haskell_import:hs-lib-b` and `//tests/haskell_import:add-one` are sometime stuck indefinitly. Killing the `bazel` process and running it again fixs it.
> 
> This should be investigated, but I still wanted to push the updates so it may

That's disconcerting. Can you isolate if this is due to the GHC 8.8 update or due to the Bazel update?

guibou:
@aherrmann The nixpkgs bump got ghc 8.8 and bazel 2.1 in scope roughly at the same time. I can try to find if there's a nixpkgs commit which does not includes both changes at the same time.

Or we can use override, it was not possible initially because of the policy we had on circle CI of not building any nix target, I'm not sure if that's still the case with buildkite.

aherrmann:
> @aherrmann The nixpkgs bump got ghc 8.8 and bazel 2.1 in scope roughly at the same time. I can try to find if there's a nixpkgs commit which does not includes both changes at the same time.
> 
> Or we can use override, it was not possible initially because of the policy we had on circle CI of not building any nix target, I'm not sure if that's still the case with buildkite.

Thanks for explaining. I think in this case an easy way to go about this would be to do the Bazel update first in a separate PR. GHC 8.6.5 is still available in nixpkgs with Bazel 2.1, so this doesn't require any overrides or additional Nix builds.

guibou:
I've restored the branch at the same status as when I opened it in december. However it does not work anymore. I'm hitting a bug similar to https://github.com/tweag/rules_haskell/issues/620 but for haddock instead (i.e `haddock: internal error: /run/user/1000/.haddock-2: createDirectory: already exists (File exists)`).

This PR was opened since too long. I tried to fix it this week-end on my free time, but I won't have time to allocate to it until next week-end. I'm closing it to avoid polluting the PR stack.

PR #1172, opened 2019-12-16
Use remote HTTP cache on CI

mrkkrp:
Once the corresponding infrastructure changes are merged (https://github.com/tweag/infra/pull/65), this enables using
google storage bucket for caching instead of caching to the local disk.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -5,7 +5,8 @@
 # https://bazel.build/roadmaps/platforms.html#replace---cpu-and---host_cpu-flags.
 build:windows --crosstool_top=@rules_haskell_ghc_windows_amd64//:cc_toolchain -s --verbose_failures --sandbox_debug
 
-build:ci --disk_cache=/tmp/bazel-disk-cache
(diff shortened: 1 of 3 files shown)


mboes:
Have you looked at performance impact, if any?

mrkkrp:
First run from scratch takes about 60 minutes (actually I might need to bump the limits). Subsequent runs are between 4 and 5 minutes and the state of the running instance does not affect the times, so everything is taken from the remote cache.

mrkkrp:
That said, we'll need to setup NixOS cache before we can switch to ephemeral instances, so it's just the first step towards that goal.

mboes:
60 minutes every 24 hours is a tough price to pay (ephemeral instances get scratched at least once every 24 hours). The number of PR's per day isn't high, so we'll be hitting that common case pretty frequently.

mrkkrp:
This is now only blocked on the Windows failure. I do not understand what is going wrong there.

mrkkrp:
As discussed, the windows failure is unrelated to this PR, so we might as well merge it.

PR #1194, opened 2019-12-30
Depend on cc toolchain

### LINKED ISSUE #1185: Flag --incompatible_enable_cc_toolchain_resolution will break rules_haskell in Bazel 1.2.1
Incompatible flag --incompatible_enable_cc_toolchain_resolution will break rules_haskell once Bazel 1.2.1 is released.

Please see the following CI builds for more information:

* [:ubuntu: 18.04 (OpenJDK 11)](<a href="https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/342#0508fe4d-40ae-4452-8ca8-1019fd5ca4a6" target="_blank">:ubuntu: 18.04 (OpenJDK 11)</a>)

Questions? Please file an issue in https://github.com/bazelbuild/continuous-integration

**Important**: Please do NOT modify the issue title since that might break our tools.

aherrmann:
Closes #1185.

Fixes `--incompatible_enable_cc_toolchain_resolution`
Also removes no longer required cc toolchain dependencies.

### DIFF
--- a/haskell/defs.bzl
+++ b/haskell/defs.bzl
@@ -180,6 +180,7 @@ def _mk_binary_rule(**kwargs):
             "repl_deprecated": "%{name}-repl",
         },
         toolchains = [
+            "@bazel_tools//tools/cpp:toolchain_type",
             "@rules_haskell//haskell:toolchain",
             "@rules_sh//sh/posix:toolchain_type",
         ],
@@ -225,6 +226,7 @@ _haskell_library = rule(
         "repl_deprecated": "%{name}-repl",
     },
     toolchains = [
+        "@bazel_tools//tools/cpp:toolchain_type",
         "@rules_haskell//haskell:toolchain",
         "@rules_sh//sh/posix:toolchain_type",
     ],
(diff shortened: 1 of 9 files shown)


mboes:
@aherrmann does this mean users also need to turn on `--incompatible_enable_cc_toolchain_resolution` ahead of the next Bazel release, or is the code backwards compatible?

haskell/defs.bzl
```diff
@@ -343,7 +342,6 @@ haskell_toolchain_library = rule(
             default = Label("@rules_haskell//haskell:toolchain-libraries"),
         ),
     ),
-    fragments = ["cpp"],
```
guibou: Why `fragment` was removed in this rule but not the others?

aherrmann:
> @aherrmann does this mean users also need to turn on `--incompatible_enable_cc_toolchain_resolution` ahead of the next Bazel release, or is the code backwards compatible?

This should be backwards compatible. At least the tests pass for me without passing that flag on Bazel 1.1.0.

haskell/defs.bzl
aherrmann: I haven't removed `fragment` and `current_cc_toolchain`, yet, so that rules_haskell still works without setting `incompatible_enable_cc_toolchain_resolution`.

This instance was removed because it is not needed. The implementation of `haskell_toolchain_library` only forwards some providers and doesn't lookup the cc toolchain.

PR #1196, opened 2019-12-30
Use platforms repo for constraints

aherrmann:
See #1182 

- Uses `@platforms//<setting>:<value>` instead of `@bazel_tools//platforms:<value>`.
- The `ghc_toolchain` macros expose users to `@platforms`. Therefore it's added to `haskell/repositories.bzl`. In principle `@platforms` is builtin, but the docs say it's preferred to explicitly import the repository.
- Similar changes are required in rules_nixpkgs and rules_sh before that flag can be set on CI and the issue is fully resolved.


aherrmann:
> So it's preferred to explicitly override the builtin one with a different version?

Yes, see https://github.com/bazelbuild/bazel/issues/8622
> Ideally, declare an explicit dependency on https://github.com/bazelbuild/platforms, name the repository as @platforms, and use constraints from this repository.

mboes:
The thing I'm confused about is - if Bazel implicitly provides `@platforms` already, doesn't that mean that the provided repository is never used (since the repository named `"platforms"` already exists)?

aherrmann:
> The thing I'm confused about is - if Bazel implicitly provides `@platforms` already, doesn't that mean that the provided repository is never used (since the repository named `"platforms"` already exists)?

Ah, that's a good point. I don't know how this works under the hood, but it seems `@platforms` is magical. Some print debugging shows that the `http_archive` for `platforms` is called. So, Bazel is not using the builtin, but the user defined repository. However, when I remove the `http_archive`, then the build still passes.

Issue #1174, opened 2019-12-17
Refactor `unfold_fields`


guibou:
**Is your feature request related to a problem? Please describe.**

The `unfold_fields` function in `pkgdb_to_bzl.py` suffers from a few limitation and may be refactored:

- [X] badly documented: it says that it returns pairs of key value when it is not the case. Done in #1197 
- [ ] not accessible from outside: `pkgdb_to_bzl.py` is not importable as a module. To do so, the top level instructions must be gathered together and protected by an `if __name__ == '__main__'`.
- [X] ~less robust: may fail if first line is indented~ (see discussion in #1169`, @mboes make the right point that the format of ghc pkg does not contain any indented first line)
- [ ] does not return sanitized output (lack of `.strip()`)

PR #1169 introduces a similar function in `ls_modules.py` and it may be a good idea to refactor both modules.

guibou:
I'm closing. The most important issue has been fixed (i.e. documentation, in #1197). The rest may be part of future refactoring, but no need to track all simple refactoring in different issues.

PR #1197, opened 2019-12-31
Fix documentation for unfold_fields

guibou:
Fix the documentation for unfold_fields by discussing the return type.

This is first step of #1174

### DIFF
--- a/haskell/private/pkgdb_to_bzl.py
+++ b/haskell/private/pkgdb_to_bzl.py
@@ -25,7 +25,12 @@ def unfold_fields(content):
     """Unfold fields that were split over multiple lines.
 
     Returns:
-        A list of field name/value pairs.
+        A list of string, each string containing multiples lines
+        grouped by indentation level. Indentation (spaces) is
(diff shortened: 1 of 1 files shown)


haskell/private/pkgdb_to_bzl.py
```diff
@@ -25,7 +25,12 @@ def unfold_fields(content):
     """Unfold fields that were split over multiple lines.
 
     Returns:
-        A list of field name/value pairs.
+        A list of string, each string containing multiples lines
```
mboes: Fewer words is better than more words. Here's a suggestion:

> A list of strings. Each string represents one field (a name/value pair separated by a colon).

Nothing else is particularly relevant to the user. Think of it as the `lines` function in Haskell (which likewise doesn't fiddle with whitespace, hence gets to likewise keep the documentation simple).

guibou: I do not agree with your suggestion because the function is technically more generic than that and does not care at all about colon or name/value pair.

Suggestion applied.

Issue #1090, opened 2019-09-11
Work around `stack update` race condition on Windows


Profpatsch:
I was wondering if we could extract the stack update into a separate repository rule and make every instance of stack_snapshot depend on that. The stack_snapshot macro could take care of instantiating the stack update workspace if it doesn't exist, yet. Though, I guess for this to work right there would have to be some file dependency between the stack update and the stack_snapshot instances.

_Originally posted by @aherrmann in https://github.com/tweag/rules_haskell/pull/1085_

PR #1199, opened 2019-12-31
Call stack update once

aherrmann:
Closes https://github.com/tweag/rules_haskell/issues/1090

As suggested in #1090 (see https://github.com/tweag/rules_haskell/pull/1085#issuecomment-529912386 for related discussion) this factors out the call to `stack update` into a separate repository rule `_stack_update`. Every instance of `_stack_snapshot` depends on the same instance of `_stack_update`, so that `stack update` is only called once.

This removes the need to call `stack update` leniently within `_stack_snapshot` and the need to preinstall `stack` and run `stack update` manually on Windows CI.

I have used Bazel's [`--experimental_workspace_rules_log_file`](https://docs.bazel.build/versions/master/workspace-log.html#finding-non-hermetic-rules) to verify the following.
- `stack update` is executed on every fetch.
- `stack update` is run before calls to `stack` in `_stack_snapshot`.
- `_stack_snapshot` is not invalidated by `_stack_update`.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -24,20 +24,6 @@ jobs:
     displayName: "Enable da long paths"
 
   - bash: |
-        # rules_haskell can automatically install stack if it is not in $PATH,
-        # yet. However, if the index has not been updated before building. Then
-        # multiple parallel invocations of `stack` will both attempt to update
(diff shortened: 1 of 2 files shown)


aherrmann:
Merging manually as mergify is not picking this up.

PR #1202, opened 2020-01-02
Use remote cache in the Haskell test script

mrkkrp:
Alas, previously the Haskell test scirpt did not use remote cache settings.
I do not have a hard proof, but it looks very probable that this is the
reason for the remarkably slow execution of this part of the pipeline.


mrkkrp:
It looks like the Haskell scirpt is run not only in Buildkite pipelines, but also at least on CircleCI which won't have the access to the bucket. Perhaps it should be configured a little bit differently. Will try an alternative approach.

.buildkite/pipeline.yml
```diff
@@ -3,16 +3,17 @@ steps:
     command: |
       . /var/lib/buildkite-agent/.nix-profile/etc/profile.d/nix.sh
       echo "build:ci --host_platform=@rules_haskell//haskell/platforms:linux_x86_64_nixpkgs" > .bazelrc.local
+      echo "build:ci --remote_http_cache=https://storage.googleapis.com/tweag-bazel-cache-buildkite --remote_upload_local_results=true --google_default_credentials" > .bazelrc.local
```
aherrmann: This duplicates the remote cache flags defined in `.bazelrc` and is likely to cause divergence and confusion.

How about the following instead:
```
echo "build --config=ci --config=ci_linux" >> .bazelrc.local
```

I did some tests with this locally, and it did enable the flags defined under `build:ci` and `build:ci_linux`.

mrkkrp: Actually I'm going to drop the flags from `.bazelrc` so this will be the only place where they are set. I think it's better this way because the flags are only for Buildkite builds and this way they will be "localized" and there won't be any danger that other scripts will pick them up.

PR #1208, opened 2020-01-07
Allow making the cabal builds totally silent

regnat:
Give an option in `cabal_haskell_library` and `stack_snapshot` to make the builds totally silent (except in case of failure).

This is useful to enforce quietness even for build steps for which cabal can't enforce it (like `cabal configure` or `cabal haddock`).

I didn't add a test for it because I don't know how to test for the console output of a rule in bazel. If someone knows a way, I'd be happy to add a test for it.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -113,7 +113,7 @@ def _prepare_cabal_inputs(hs, cc, posix, dep_info, cc_info, direct_cc_info, comp
         env["SDKROOT"] = "macosx"  # See haskell/private/actions/link.bzl
 
     if verbose:
-      env["CABAL_VERBOSE"] = "True"
+        env["CABAL_VERBOSE"] = "True"
 
     args = hs.actions.args()
     package_databases = dep_info.package_databases
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -932,6 +932,7 @@ haskell_cabal_library(
     tools = {tools},
     visibility = {visibility},
     compiler_flags = ["-w", "-optF=-w"],
+    verbose = {verbose},
 )
 """.format(
                     name = package.name,
@@ -945,6 +946,7 @@ haskell_cabal_library(
                     ],
                     tools = tools,
                     visibility = visibility,
+                    verbose = repr(repository_ctx.attr.verbose),
                 ),
             )
         if package.versioned_name != None:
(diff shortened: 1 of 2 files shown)


haskell/cabal.bzl
```diff
@@ -961,6 +974,10 @@ _stack_snapshot = repository_rule(
         "tools": attr.label_list(),
         "stack": attr.label(),
         "stack_update": attr.label(),
+        "verbose": attr.bool(
+            default = False,
+            doc = "Whether to show the output of the build",
+        ),
```
aherrmann: This needs to be forwarded to the generated calls to `haskell_cabal_library` to have any effect.

regnat: Indeed, just noticed that. Should be good now

haskell/private/cabal_wrapper.py.tpl
```diff
@@ -29,12 +29,21 @@ import sys
...
         sys.stderr.flush()
-    subprocess.check_call(cmd, *args, **kwargs)
+    if verbose:
+        subprocess.run(cmd, check=True, *args, **kwargs)
+    else:
+        try:
+            subprocess.run(cmd, check=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, *args, **kwargs)
+        except subprocess.CalledProcessError as err:
+            print(err.stdout.decode())
+            print(err.stderr.decode(), file=sys.stderr)
```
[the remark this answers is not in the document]
regnat: Oh thanks, didn't know that. I'll try this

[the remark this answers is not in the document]
regnat: Nice, that works indeed.

haskell/private/cabal_wrapper.py.tpl
```diff
@@ -29,12 +29,23 @@ import sys
...
+    if verbose:
+        subprocess.run(cmd, check=True, *args, **kwargs)
+    else:
+        try:
+            subprocess.run(cmd, check=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, *args, **kwargs)
+        except subprocess.CalledProcessError as err:
+            sys.stdout.buffer.write(err.stdout)
+            sys.stderr.buffer.write(err.stderr)
+            # print(err.stdout.decode())
+            # print(err.stderr.decode(), file=sys.stderr)
```
aherrmann: Remove those?

regnat: Oops :innocent: 

Done

PR #1223, opened 2020-01-27
Generate docker image on the fly and do not assume fixed user id

mrkkrp:


.buildkite/image/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:xenial-20200114
+
+WORKDIR /home/
+
+RUN apt-get --quiet update
+RUN apt-get -y upgrade
+RUN apt-get --quiet install -y sudo build-essential libtinfo5 libtinfo-dev python python3 libgmp-dev libffi-dev curl
```
aherrmann: The [Docker docs](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get) say
> Avoid `RUN apt-get upgrade`

and

> Always combine `RUN apt-get update` with `apt-get install` in the same RUN statement.

Maybe better
```
RUN apt-get update
 && apt-get install -y sudo build-essential libtinfo5 libtinfo-dev python python3 libgmp-dev libffi-dev curl
 && rm -rf /var/lib/apt/lists/*
```

PR #1228, opened 2020-01-31
Fix API documentation for common rules

### LINKED ISSUE #1122: The api documentation don't document the args for some rules
**Describe the bug**

The documentation of the core haskell rules doesn't describe the args, but just displays them as `**kwargs | Unknown; Optional`

For example, the documentation of [haskell_binary](https://api.haskell.build/haskell/defs.html#haskell_binary) shows as:

image

**Expected behavior**

The argument list should be clearly documented as it is for the other rules (like [here](https://api.haskell.build/haskell/defs.html#haskell_import))

**Environment**

N/A

**Additional context**

This is due to the fact that these rules are effectively macros taking `**kwargs` as argument (the actual rules are called inside the macros and do the actual arg-checking, but skylib can't see that).

aherrmann:
Closes #1122 

The rules `haskell_library|binary|test` are exposed through wrapper macros, the rules themselves are hidden. As a consequence the API documentation doesn't show the `doc` attribute of the rule attributes. This PR makes the attributes explicit in the wrapper macros, and moves the attribute documentation into the docstring of the wrapper macros. Unfortunately, wrapper macros cannot inherit rule documentation. See  https://github.com/bazelbuild/stardoc/issues/27

### DIFF
--- a/haskell/defs.bzl
+++ b/haskell/defs.bzl
@@ -330,7 +330,28 @@ def haskell_binary(
         **kwargs
     )
 
-def haskell_test(**kwargs):
+def haskell_test(
+        name,
+        src_strip_prefix = "",
+        srcs = [],
+        extra_srcs = [],
+        deps = [],
+        data = [],
+        compiler_flags = [],
+        repl_ghci_args = [],
+        runcompile_flags = [],
+        plugins = [],
+        tools = [],
+        worker = None,
+        linkstatic = True,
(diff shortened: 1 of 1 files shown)


aherrmann:
[Rendered](https://5e341462c8d3a7000967a98d--api-haskell-build.netlify.com/haskell/defs.html)

PR #1230, opened 2020-01-31
Replace @repl output by haskell_repl

### LINKED ISSUE #1189: Repl targets fail with runfiles-related error
**Describe the bug**
The repl targets don't work on my machine.
 
**To Reproduce**

```
$ bazel run tests/binary-simple:binary-simple@repl
INFO: Analyzed target //tests/binary-simple:binary-simple@repl (0 packages loaded, 3 targets configured).
INFO: Found 1 target...
Target //tests/binary-simple:binary-simple@repl up-to-date:
  bazel-bin/tests/binary-simple/binary-simple@repl
INFO: Elapsed time: 0.149s, Critical Path: 0.01s
INFO: 1 process: 1 local.
INFO: Build completed successfully, 6 total actions
INFO: Build completed successfully, 6 total actions
GHCi, version 8.6.5: http://www.haskell.org/ghc/  :? for help
Loaded package environment from bazel-out/k8-fastbuild/bin/tests/binary-simple/repl-package_env-binary-simple
Loaded package environment from bazel-out/k8-fastbuild/bin/tests/binary-simple/repl-package_env-binary-simple
  File "bazel-out/host/bin/haskell/cc_wrapper-python", line 349, in <module>
    Main()
  File "bazel-out/host/bin/haskell/cc_wrapper-python", line 272, in Main
    module_space = FindModuleSpace()
  File "bazel-out/host/bin/haskell/cc_wrapper-python", line 111, in FindModuleSpace
    raise AssertionError('Cannot find .runfiles directory for %s' % sys.argv[0])
AssertionError: Cannot find .runfiles directory for bazel-out/host/bin/haskell/cc_wrapper-python
`cc_wrapper-python' failed in phase `gcc'. (Exit code: 1)
```
**Expected behavior**

**Environment**
 - Ubuntu Rodete
 - Bazel version 2.0.0
 - Version of the rules: 3fc77108396c99d89772170b040fb8ff65df8148
- Python 2.7.17rc1

aherrmann:
This turns `@repl` into a regular Bazel target which can properly declare its runfiles dependencies. Previously, `@repl` was just an additional output and as such could not define a `DefaultInfo` provider.

Closes #1189

This change is backwards compatible in that the names `<name>@repl` and `<name>-repl` are still available to call a REPL. However, `haskell_repl` passes `-i` flags to GHCi so that the new `@repl` targets will load more modules by source. E.g. `bazel run //tests/repl-targets:hs-bin@repl` loads `Main` and `QuuxLib` from source, while the old `@repl` only loaded `Main` from source.


haskell/defs.bzl
```diff
@@ -294,6 +290,26 @@ def haskell_binary(
         **kwargs
     )
 
+    repl_kwargs = {
+        attr: kwargs[attr]
+        for attr in ["testonly", "tags"]
+        if attr in kwargs
+    }
+    native.alias(
```
guibou: Do we really care about backward compatibility here?

It won't break dramatically and if the release note contains the right lines, the migration is trivial.

haskell/defs.bzl
```diff
@@ -380,6 +396,28 @@ def haskell_test(
         **kwargs
     )
 
+    repl_kwargs = {
+        attr: kwargs[attr]
+        for attr in ["tags"]
+        if attr in kwargs
+    }
+    native.alias(
```
guibou: Same thing about backward compatibility.

haskell/defs.bzl
```diff
@@ -468,6 +506,26 @@ def haskell_library(
         **kwargs
     )
 
+    repl_kwargs = {
+        attr: kwargs[attr]
+        for attr in ["testonly", "tags"]
+        if attr in kwargs
+    }
+    native.alias(
```
guibou: Same here.

tests/RunTests.hs
```diff
@@ -64,7 +64,7 @@ main = hspec $ do
       assertSuccess (bazel ["run", "//tests/multi_repl:c_multi_repl", "--", "-ignore-dot-ghci", "-e", ":load BC.C", "-e", "c"])
 
   describe "failures" $ do
-    all_failure_tests <- bazelQuery "kind(rule, //tests/failures/...) intersect attr('tags', 'manual', //tests/failures/...)"
+    all_failure_tests <- bazelQuery "kind('haskell_library|haskell_binary|haskell_test', //tests/failures/...) intersect attr('tags', 'manual', //tests/failures/...)"
```
guibou: Why?

haskell/defs.bzl
aherrmann: We've been getting issues referring to the `@repl` targets, and they are documented in a couple places. So, removing them would likely affect a good number of users. The goal of this PR is only to change implementation, not interface.

haskell/defs.bzl
aherrmann: Same reasoning as in https://github.com/tweag/rules_haskell/pull/1230#discussion_r373990927

tests/RunTests.hs
aherrmann: The `@repl` and `-repl` are now separate targets, so that a query `kind(rule, ...)` returns those as well. But, those targets are not broken, while this test case expects all returned targets to fail. So, we need to be more specific in the query.

haskell/defs.bzl
guibou: Thank you.

tests/RunTests.hs
guibou: Ok. If you feel it, please add a comment.

Thank you.

aherrmann:
@mboes 

> At this point, this change is purely an implementation change, which just fixes a bug but otherwise offers the same behaviour, right?

Yes, this fixes a bug and removes redundant code. The behavior is the same up to the `-i` flag as described in the PR message itself.

> Or we may want to deprecate the implicit repl targets, using the [`deprecation` attribute](https://docs.bazel.build/versions/master/be/common-definitions.html). But in that case it should be clear what we recommend. hrepl? `haskell_repl`? Sometimes the former and sometimes the latter?

I'd be in favor of deprecating them. However, as you say, that requires documentation change as well, which is out of scope for this PR.

tests/RunTests.hs
aherrmann: Good point, I've added a comment.

mboes:
> I'd be in favor of deprecating them. However, as you say, that requires documentation change as well, which is out of scope for this PR.

Agreed. To be clear, if there's a way to change the implementation without the interface, then documentation changes and deprecations I expect to happen as future PR's.

PR #1233, opened 2020-02-04
Add explanations for the new Buildkite NixOS pipelines

mrkkrp:


aherrmann:
mergify choked on the merge commit so I've rebased on master instead and force pushed.

PR #1235, opened 2020-02-05
Source nodejs from nixpkgs

aherrmann:
The order of imports was wrong in the WORKSPACE file, such that
rules_sass implicitly caused rules_nodejs to fetch a bindist of node.
This fails on NixOS where the bindist does not work because of the
different location of the dynamic loader on NixOS.


aherrmann:
@Mergifyio rebase

PR #1236, opened 2020-02-06
Add haskell-indexer aspect, to generate Kythe entries.

robinp:
The docstring contains an example for processing the entries into a
serving table. The produced table can be browsed using Underhood, for
example.

For now the indexer binary needs to be defined using a Bazel repo with
fixed name. Will add a toolchain for sourcing it eventually.

Note: PR in progress to make haskell-indexer buildable using nix easily, so it can be sourced with `rules_nixpkgs`.

### DIFF
--- a/haskell/indexer.bzl
+++ b/haskell/indexer.bzl
@@ -1,0 +1,168 @@
+"""haskell-indexer support"""
+
+# With a fair amount of copy-paste from haddock.bzl.
+
+load("@bazel_skylib//lib:dicts.bzl", "dicts")
+load("@bazel_skylib//lib:paths.bzl", "paths")
+load(
+    "@rules_haskell//haskell:providers.bzl",
+    "HaskellInfo",
+    "HaskellLibraryInfo",
+    "HaskellToolchainLibraryInfo",
+    "HaskellIndexerInfo",
+    "get_ghci_extra_libs",
+)
+load(":private/context.bzl", "haskell_context", "render_env")
+load(":private/set.bzl", "set")
+
+def _get_named_path(suffix, unique_id):
+    """Get path to a named file related to a package given its id.
+
+    Args:
+      suffix: string.
+      unique_id: string.
+
+    Returns:
+      string: relative path to file.
+    """
+    return unique_id + "." + suffix
+
+def _haskell_indexer_aspect_impl(target, ctx):
+    if not (HaskellInfo in target):
+        return []
+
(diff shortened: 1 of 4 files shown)


haskell/indexer.bzl
```diff
@@ -0,0 +1,168 @@
...
+    _haskell_indexer_aspect_impl,
+    attrs = {
+        "_indexer_wrapper_tpl": attr.label(
+            allow_single_file = True,
+            default = Label("@rules_haskell//haskell:private/indexer_wrapper.sh.tpl"),
+        ),
+        # TODO(robinp): use a toolchain to get indexer binary
+        "_indexer_binary": attr.label(
+            allow_single_file = True,
+            default = Label("@haskell_indexer_ghc//:bin/ghc_kythe_wrapper"),
```
aherrmann: Do you have an instance of `@haskell_indexer_ghc` somewhere? I'd like to try it out to see what this looks like.

haskell/indexer.bzl
```diff
@@ -0,0 +1,168 @@
...
+    )
+
+    indexer_info = HaskellIndexerInfo(
+        entries = entries_file,
+    )
+    output_files = OutputGroupInfo(default = [entries_file, log_file])
+
+    return [indexer_info, output_files]
+
+haskell_indexer_aspect = aspect(
```
aherrmann: I think we'll also need some tests for this. Similar aspects/rules like haddock, docstring, or REPL, often break when we make changes that touch compiler or linker flags or inputs.

haskell/indexer.bzl
```diff
@@ -0,0 +1,168 @@
...
+        "@rules_haskell//haskell:toolchain",
+        "@rules_sh//sh/posix:toolchain_type",
+    ],
+)
+
+"""Output haskell-indexer entries in Kythe format.
+
+Example usage:
+
+    $ bazel build -c opt //... --aspects @rules_haskell//haskell:indexer.bzl%haskell_indexer_aspect
```
aherrmann: This requires `@haskell_indexer_ghc` (or later on the toolchain). Please also explain how to import those into a Bazel project.

haskell/indexer.bzl
```diff
@@ -0,0 +1,168 @@
...
+    $ find bazel-out/k8-opt/ -type f -name '*.kythe.entries'
+
+Serving with Kythe:
```
aherrmann: Can you also point to some docs on how to install Kythe?

PR #1238, opened 2020-02-10
Simplify transitive dynamic library dependencies on macOS

aherrmann:
On Darwin GHC [will pass the `-dead_strip_dylibs` flag](https://gitlab.haskell.org/complyue/ghc/commit/b592bd98ff25730bbe3c13d6f62a427df8c78e28) to the linker. This flag will remove any shared library loads from the binary's header that are not directly resolving undefined symbols in the binary. I.e. any indirect shared library dependencies will be removed. This conflicts with Bazel's builtin cc rules, which assume that the final binary will load all transitive shared library dependencies. In particlar shared libraries produced by Bazel's cc rules never load shared libraries themselves, which causes missing symbols at runtime on macOS. https://github.com/tweag/rules_haskell/pull/627 worked around this by forcing the linker to resolve symbols in all transitive dynamic library dependencies using `-u` flags.

Now that we have `cc_wrapper` a much simpler and less costly solution is to just drop the `-dead_strip_dylibs` linker flag.

I've confirmed that removing `_darwin_create_extra_linker_flags_file` alone causes [`//tests/indirect-link:indirect-link-dynamic` to fail](https://circleci.com/gh/tweag/rules_haskell/8392). Dropping `-dead_strip_dylibs` makes it succeed again.

This PR also refactors linker flag parsing in `cc_wrapper` to avoid duplication.

### DIFF
--- a/haskell/private/actions/compile.bzl
+++ b/haskell/private/actions/compile.bzl
@@ -152,12 +152,6 @@ def _compilation_defaults(hs, cc, java, posix, dep_info, plugin_dep_info, srcs, 
     compile_flags += hs.toolchain.compiler_flags
     compile_flags += user_compile_flags
 
-    # Work around macOS linker limits.  This fix has landed in GHC HEAD, but is
-    # not yet in a release; plus, we still want to support older versions of
-    # GHC.  For details, see: https://phabricator.haskell.org/D4714
-    if hs.toolchain.is_darwin:
-        compile_flags += ["-optl-Wl,-dead_strip_dylibs"]
-
     package_ids = []
     for plugin in plugins:
         package_ids.extend(all_dependencies_package_ids(plugin.deps))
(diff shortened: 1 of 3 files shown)


mboes:
Adding @judah as a second reviewer. It would be nice to get his perspective on this kind of macOS linking changes.

aherrmann:
> I'm fine with this change even if it elevates the lower bound of supported GHC versions on macOS. But if that's the case, we should call this out. Will builds work with this change even on older GHC's?

Yes, this should not raise the lower bound of compatible GHC versions. According to [gitlab](https://gitlab.haskell.org/complyue/ghc/commit/b592bd98ff25730bbe3c13d6f62a427df8c78e28) GHC passes `-dead_strip_dylibs` since version 8.6.1, prior versions don't. We do not depend on that flag being set, we just filter it out if it is set.

GHC's reason for setting `-dead_strip_dylibs` was macOS's MACH-O header size limitation. In rules_haskell we employ other workarounds to avoid exceeding that limit that are compatible with Bazel's linking conventions.

aherrmann:
> Adding @judah as a second reviewer. It would be nice to get his perspective on this kind of macOS linking changes.

@judah is this something you would like to take a look at?

judah:
Sorry for not responding earlier.  I don't have anything specific to add to this review, but from what I understand this change seems reasonable.

PR #1239, opened 2020-02-12
Avoid shadowing of cc_info in haskell_library_impl

aherrmann:
The `compile_info_output_groups` function expects a `CcInfo` of the
targets compilation dependencies. However, in `haskell_library_impl` it
was accidentally given the output `CcInfo` of the target itself due to
name shadowing. This change fixes that by not shadowing the name
`cc_info`. See `haskell_binary_impl` for reference, where no such
shadowing occurs.

This did not cause any issues because `compile_info_output_groups` only
takes C libraries from the given `CcInfo` and ignores Haskell libraries.

### DIFF
--- a/haskell/private/haskell_impl.bzl
+++ b/haskell/private/haskell_impl.bzl
@@ -592,7 +592,7 @@ def haskell_library_impl(ctx):
     linking_context = cc_common.create_linking_context(
         libraries_to_link = libraries_to_link,
     )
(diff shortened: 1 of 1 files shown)


Profpatsch:
I guess this was dependent on #1242 so if you rebase it should pass?

Profpatsch:
Maybe mergify does it for us.

aherrmann:
@Mergifyio rebase

PR #1240, opened 2020-02-12
.bazelrc: improve try-import of .bazelrc.local

Profpatsch:
Importing the relative path fails in some cases. %workspace% always
points to the absolute directory of the WORKSPACE file.


aherrmann:
@Mergifyio rebase

PR #1241, opened 2020-02-13
Mangle static C libraries at their defining target

aherrmann:
**depends on #1239**

Bazel mangles dynamic libraries by creating symbolic links underneath `_solibs_<cpu>`, while the static libraries remain unmangled. However, GHC assumes that the static and dynamic versions of a C library share the same library name, which is in conflict with Bazel approach. To work around this we create symbolic links for static libraries whose library name differs from the dynamic library name.

Previously, these symbolic links were generated at the Haskell target transitively depending on the respective C library. There are two downsides to this approach:

1. The same symbolic links are created repeatedly, wasting resources.
2. The library search paths in the package configuration of one Haskell    library are no longer valid for another Haskell library that depends    on the former.

Number 2 is the reason why we need to define `(LD_)LIBRARY_PATH` environment variables for various actions. Because GHC only uses `-L` parameters for direct C library dependencies of the current target, not transitive C library dependencies through other Haskell libraries.

This change instead uses symbolic links created directly at the C library target through an aspect `haskell_cc_libraries_aspect`. This way each such symlink is only created once, resolving 1. And library search paths at a Haskell library's package configuration remain valid for other Haskell targets depending on it, resolving 2.

Avoiding `(LD_)LIBRARY_PATH` will be implemented in a follow-up PR. 

* Refactors code for linking C libraries to accommodate for `haskell_cc_libraries_aspect`.
* Defines the `haskell_cc_libraries_aspect` which mangles static libraries where necessary.
* Use the mangled static libraries produced by the new aspect, instead of mangling at each Haskell target.
* The way that `_haskell_proto_library_aspect` is implemented requires working around cyclic aspect dependencies by extending the `HaskellProtoInfo` provider.
* More robust check in `is_hs_library` to avoid doubly linking `ffi` with GHC bindist.

This PR is best reviewed commit by commit. See the respective commit messages for further details.


aherrmann:
Docs generation fails with
```
3:20:53 PM: TypeError: aspect() got an unexpected keyword argument 'required_aspect_providers'
```
I'm trying if updating from Skydoc 0.3.0 to Stardoc 0.4.0 fixes this.

haskell/haddock.bzl
```diff
@@ -326,7 +333,10 @@ haskell_doc = rule(
     _haskell_doc_rule_impl,
     attrs = {
         "deps": attr.label_list(
-            aspects = [haskell_doc_aspect],
+            aspects = [
+                haskell_cc_libraries_aspect,
```
regnat: Do we really need to access the cc libraries here?

aherrmann: Yes, the haddock aspect depends on the C libraries [here](https://github.com/tweag/rules_haskell/blob/513e775f00fcddbc5d2fc1a62e6818e7ea5025f0/haskell/haddock.bzl#L136). If we don't add the `haskell_cc_libraries_aspect` here, then that aspect will fail to find the `HaskellCcLibrariesInfo` provider. E.g. for `//tests:toto`
```
<target //tests:utils> (rule '_haskell_library') doesn't contain declared provider 'HaskellCcLibrariesInfo'
```

regnat: Oh indeed, should have looked better. Now I wonder why haddock needs these, but that's a whole other topic

aherrmann: No worries, it's a good question.

PR #1244, opened 2020-02-14
Update from Skydoc to Stardoc

aherrmann:
**depends on https://github.com/tweag/rules_sh/pull/11 and https://github.com/tweag/rules_nixpkgs/pull/104. Update the respective imports to master after these PRs have been merged, before merging this PR.**

The Bazel team has [deprecated Skydoc in favor of Stardoc](https://github.com/bazelbuild/stardoc/blob/master/docs/skydoc_deprecation.md). Skydoc is no longer extended to support new Bazel features, e.g. the `required_aspect_providers` attribute to `aspect`, see https://github.com/tweag/rules_haskell/pull/1241. In order to be able to use such features and still generate API documentation we need to to upgrade to Stardoc.

 However, Stardoc is not a one-to-one replacement for Skydoc. Some migration effort is required:

* Special sections in docstrings (`Args:`) have to appear at the end of   the docstring. The regular summary has to appear before. Docstrings in   `rules_haskell` have been shuffled accordingly.
* Stardoc does not operate on the `.bzl` files as inputs alone.   Instead, it [expects `bzl_library` targets as inputs](https://github.com/bazelbuild/stardoc/blob/master/docs/skydoc_deprecation.md#starlark-dependencies). These   `bzl_library` targets have to capture all `load`ed modules as `deps`.
  * `rules_sh` and `rules_nixpkgs` have to be updated to expose     `bzl_library` targets.
  * A `bzl_library` target has to be defined for `@bazel_tools`.
  * Imports of modules from the same ruleset have to be local. I.e.     `rules_haskell` modules cannot import other `rules_haskell` imports     as `@rules_haskell//...`, only as `//...` or `:...`.
* Stardoc [no longer generates a standalone html page](https://github.com/bazelbuild/skydoc/issues/148) for the API   documentation. Instead, it accepts templates for definitions and the   module header. Stardoc provides markdown and html templates. However,   for our use-case of generating a standalone html page it is easier to   define our own custom templates based on the templates that Skydoc   used before.
* Stardoc no longer generates an overview page and global navigation   page. Instead we use dedicated metadata templates to extract the necessary   metadata and generate an overview page and navigation from this   metadata.
* Stardoc [no longer contains a powerful markdown to html converter](https://github.com/bazelbuild/skydoc/issues/148).   Instead, it is mostly targeted at generated markdown API   documentation. We use pandoc with custom templates to convert that   markdown documentation to html.
* On a Nix platform we import pandoc from nixpkgs. On other Linux or   macOS we use the standalone binary distributions that the pandoc   project publishes.
* Stardoc newly follows `load` commands and generates documentation for   exported symbols. That makes it redundant to generate documentation   for some modules, e.g. `haskell/repl.bzl`, as these are already   exposed through other modules, e.g. `haskell/defs.bzl`.
* Stardoc [no longer recognizes the `Example:` special docstring section](https://github.com/bazelbuild/stardoc/issues/47).   Instead, we change the docstrings to use a markdown level 3 heading to   match the generated `Args:` section.
* Stardoc is stricter in which markup it allows in docstrings. E.g. it   requires a space between an argument and it's documentation, or a   newline around code snippets.
* Stardoc changed the way that rules define their docstring. In Skydoc   it was a multiline string literal following the rule definition. In   Stardoc it is the value of the `doc` attribute.


mboes:
This sounds fine to me, provided that Stardoc can provide the same things as Skydoc:

* the ability to serve documentation as a bunch of HTML on a static website,
* said HTML should be not too ugly looking and ideally the CSS should be customizable,
* proper support for example sections, module level documentation, rule documentation and macro documentation. In short, none of the information that we're able to convey via Skydoc should be lost.

aherrmann:
> This sounds fine to me, provided that Stardoc can provide the same things as Skydoc:
> 
> * the ability to serve documentation as a bunch of HTML on a static website,

Stardoc does not itself produce a standalone HTML page, however it exposes enough that we can generate one with pandoc.

> * said HTML should be not too ugly looking and ideally the CSS should be customizable,

I've taken the Skydoc templates and ported them to pandoc so the resulting page looks very similar.

> * proper support for example sections, module level documentation, rule documentation and macro documentation. In short, none of the information that we're able to convey via Skydoc should be lost.

The page generated with Stardoc as of this PR should not lose any of the information conveyed by the previous Skydoc documentation. The one difference I'm aware of is that Skydoc categorized symbols by whether they are rules, macros, or repository rules. Stardoc no longer does that.

Profpatsch:
It looks extremely similar to the old version; great work!

Issue #1248, opened 2020-02-19
Error building `path-io` package.


joneshf:
**Describe the bug**

There appears to be an error when building the [`path-io`](https://hackage.haskell.org/package/path-io) package: `hGetContents: invalid argument (invalid byte sequence)`. When I've seen this failure in the past, it appears to be a locale issue. It might not be a locale issue, though. I'm not sure how to fix this issue, in any case. Any ideas how to fix things so `path-io` might compile?

**To Reproduce**

I've put together a minimal reproduction at [joneshf/rules_haskell_invalid_byte_sequence](https://github.com/joneshf/rules_haskell_invalid_byte_sequence). For posterity, the files included are:

<details>
<summary>BUILD.bazel</summary>

```bazel
load(
    "@rules_haskell//haskell:defs.bzl",
    "haskell_library",
)

haskell_library(
    deps = [
        "@stackage//:path-io",
    ],
    name = "invalid_byte_sequence",
)
```
</details>

<details>
<summary>WORKSPACE.bazel</summary>

```bazel
workspace(
    name = "rules_haskell_invalid_byte_sequence",
)

load(
    "@bazel_tools//tools/build_defs/repo:http.bzl",
    "http_archive",
)

http_archive(
    name = "rules_haskell",
    sha256 = "5778f69448b775ee96ee26439e3624db540a9e515caa46aa2eff99f9c91ab849",
    strip_prefix = "rules_haskell-1fedad7e64ea00ecf1bcb4be70ca51a11dc48312",
    urls = [
        "https://github.com/tweag/rules_haskell/archive/1fedad7e64ea00ecf1bcb4be70ca51a11dc48312.zip",
    ],
)

load(
    "@rules_haskell//haskell:repositories.bzl",
    "rules_haskell_dependencies",
)

rules_haskell_dependencies()

load(
    "@rules_haskell//haskell:toolchain.bzl",
    "rules_haskell_toolchains",
)

rules_haskell_toolchains(
    version = "8.8.2",
)

load(
    "@rules_haskell//haskell:cabal.bzl",
    "stack_snapshot",
)

stack_snapshot(
    name = "stackage",
    packages = [
        "path-io",
    ],
    snapshot = "lts-15.0",
)
```
</details>

Running `bazel build` shows the failure:

```console
$ bazel build //...
[...]

PR #1249, opened 2020-02-20
Allow GHC bindists to set locale

joneshf:
In #1248, it was discovered that the locale for GHC bindists was not able to be overridden. It would default to `C.UTF-8`. If a system didn't have the `C.UTF-8` locale, it could lead to (unclear) encoding failures later on. In particular, `haddock` might fail in an unclear way:

```
haddock: internal error: /tmp/tmputn68mya/doc/html/path-io/haddock-response300-1.txt: hGetContents: invalid argument (invalid byte sequence)
```

This PR allows setting the locale for GHC bindists and also writes up a little bit in the troubleshooting section. I wasn't 100% sure how to make this happen, but this seems to allow my [minimal reproduction](https://github.com/joneshf/rules_haskell_invalid_byte_sequence) to [build](https://github.com/joneshf/rules_haskell_invalid_byte_sequence/tree/fix-set-locale). Lemme know if anything should change, or if this is not the right approach.

I couldn't find any tests to extend for the changes I made here. Should a new test be created, or is this not an area to test?

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,36 @@ protects against this problem. If sandboxing is not an option, simply
 put the source files for each target in a separate directory (you can
 still use a single `BUILD` file to define all targets).
 
+### hGetContents: invalid argument (invalid byte sequence)
+
+If you are using the GHC bindists and see an error message like this:
+
+```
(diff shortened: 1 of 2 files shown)


joneshf:
Gotcha! Thanks for the quick review!

PR #1252, opened 2020-02-24
Define Bazel version in one shared location

aherrmann:
Instead of each CI pipeline separately defining the Bazel version, we define it in one shared location `.ci/bazelversion`. Ideally, we would use `.bazelversion` in the top-level which would be compatible with [bazelisk](https://github.com/bazelbuild/bazelisk). Unfortunately, this is incompatible with the way Bazel is packaged in nixpkgs right now. See https://github.com/NixOS/nixpkgs/issues/80950.

The pipelines that use nixpkgs to provision Bazel check that the Bazel version matches using `.ci/check-bazel-version`. This was already the case on Linux, but has been added on macOS. The pipelines that fetch a Bazel binary distribution use `.ci/fetch-bazel-bindist` which in turn consults the `.ci/bazelversion` file. This was already the case on buildkite, but has been added to netlify and Azure. To avoid issues with corrupted downloads we check the downloaded bindist against expected sha256 hashes defined in `.ci/bazel-*.sha256`.

The files `.ci/bazelversion` and `.ci/bazel-*.sha256` need to be kept in sync. The script `.ci/update-bazel-version` helps with this. Given a Bazel version number it will update both files. The nixpkgs revision needs to be updated separately. The `MAINTAINERS.md` file has been updated accordingly.


.ci/check-bazel-version
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -euo pipefail
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+# Ideally we would place .bazelversion in the top-level so it could be used by
+# tools like bazelisk. However, this breaks the nixpkgs provided Bazel
+# installation, see https://github.com/NixOS/nixpkgs/issues/80950.
+VERSION_EXPECTED="bazel $(cat "$DIR/bazelversion")"
+VERSION_ACTUAL=$(bazel version --gnu_format)
+# nixpkgs Bazel version ends on '- (@non-git)'.
```
Profpatsch: Should we fix this in nixpkgs?

aherrmann: I think that would also fix [this issue](https://github.com/NixOS/nixpkgs/issues/80950#issuecomment-590337389), so that would be good. Though, I don't think we need to block this PR on it.

Profpatsch: I guess that’s already fixed by the changes in https://github.com/NixOS/nixpkgs/pull/80739#pullrequestreview-364046797

aherrmann: IIUC that PR changes `bazel-real` for `bazel-VERSION...`, so it fixes https://github.com/NixOS/nixpkgs/issues/80950, but doesn't change the `- (@non-git)` part. (Unless Bazel 2.1 does that, I haven't checked).

PR #1254, opened 2020-02-26
Add a CI pipeline for macOS bindist

aherrmann:
We had a few issues in the past reported by users that tried to use or develop rules_haskell on macOS outside of Nix. On our CI it has been a blind spot so far. This PR adds a CircleCI job for testing rules_haskell on macOS using the GHC bindist outside of Nix.

In setting up the pipeline this work surfaced and fixed a few more macOS issues:
- The script to fetch a Bazel bindist uses the `OSTYPE` variable to determine the OS. On macOS it was incorrectly checking for the value `darwin` when it should be checking for the pattern `darwin*`.
- The script to fetch a Bazel bindist required `sha256sum` to verify the downloaded artifact. On macOS `coreutils` (which include `sha256sum`) are typically not pre-installed. In that case we can fall back to `openssl dgst -sha256`.
- macOS users also need to set `BAZEL_USE_CPP_ONLY_TOOLCHAIN=1` outside of Nix. This adds a corresponding entry to the troubleshooting section in the `README.md`.
- rules_haskell requires Python 3. But, macOS only pre-installs Python 2.7. Users often install Python 3 using homebrew or pyenv. However, these tools don't install Python into a system path, but `/usr/local/bin` or a user specific path. The Bazel provided [autodetecting toolchain](https://github.com/bazelbuild/bazel/blob/ed0865c3c95bcd7c9b3dd3059e5f444c48e59660/tools/python/BUILD.tools#L137) fails to find such Python installations. This PR adds a autoconfigure toolchain for the GHC bindist case similar to the POSIX autoconfigure toolchain, which looks for a Python 3 interpreter in a repository rule. Closes https://github.com/tweag/rules_haskell/pull/1250.
- rules_haskell defaults to the locale `C.UTF-8` since Haskell packages often require unicode support. However, macOS typically doesn't provide that locale. In this case we default to `en_US.UTF-8`.
- This newly defined pipeline tests the intersection of targets tested by the Linux bindist pipeline and the nixpkgs macOS pipeline. Additionally it skips the `extra-source-files` test because `ld` on macOS does not understand response files (`-Wl,@...`).

### DIFF
--- a/.ci/fetch-bazel-bindist
+++ b/.ci/fetch-bazel-bindist
@@ -11,7 +11,20 @@ esac
 VERSION=$(cat "$TOP/.ci/bazelversion")
 INSTALL="$(mktemp -d)"
 (cd "$INSTALL" && curl -LO "https://github.com/bazelbuild/bazel/releases/download/$VERSION/bazel-$VERSION-$OS$EXT" >&2)
-(cd "$INSTALL" && sha256sum --quiet -c "$TOP/.ci/bazel-$VERSION-$OS$EXT.sha256") >&2
+if command -v sha256sum &>/dev/null; then
+  (cd "$INSTALL" && sha256sum --quiet -c "$TOP/.ci/bazel-$VERSION-$OS$EXT.sha256") >&2
+elif command -v openssl &>/dev/null; then
+  # macOS typically doesn't have sha256sum (coreutils) installed.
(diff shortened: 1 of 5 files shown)


aherrmann:
A cached run of this new pipeline [took ~7min](https://circleci.com/gh/tweag/rules_haskell/8489) compared to the preexisting nixpkgs macOS pipeline which [took ~13min](https://circleci.com/gh/tweag/rules_haskell/8490).

mboes:
cc @jamesthompson

PR #1262, opened 2020-03-04
Setup ghcide for rules_haskell repository

aherrmann:
Depends on #1350.

This configures `hie-bios` for `ghcide` for `rules_haskell`.
- Adds [`ghcide`](https://github.com/digital-asset/ghcide#readme) as built by `stack_snapshot` and `haskell_cabal_binary` to the repository. https://github.com/tweag/rules_haskell/pull/1304 will make this easier in the future.
- Adds a `haskell_repl` target to determine GHC flags required for the REPL session.
- Adds a `.hie-bios` script to the repository that is picked up by `hie-bios`'s `bios` cradle. When executed that should write all GHCi flags to a file specified by the `HIE_BIOS_OUTPUT` environment variable. See [hie-bios README](https://github.com/mpickering/hie-bios#readme). Once multi-component support is implemented in ghcide this should become more flexible. See https://github.com/digital-asset/ghcide/pull/522#issuecomment-611421786
- Adds a simple test case that builds `ghcide` and checks that `ghcide --version` works. This is run on nixpkgs and bindist configurations.
- Adds a more involved test case to `RunTests` that invokes `ghcide` on a target and checks if it can successfully determine the required GHC flags. This is only run on `nixpkgs` configurations, since `run-tests` is only executed there.
- Adds a section to the use-case documentation explaining how to setup `ghcide` with `rules_haskell`. The setup is still pretty involved, https://github.com/tweag/rules_haskell/pull/1304 will simplify it a fair bit. We may want to hold off wider announcement, e.g. a tweet, until this is merged.

---

In order to make use of `ghcide` on `rules_haskell` you need to configure your editor to use it. See [setup instructions](https://github.com/digital-asset/ghcide#using-it). For example, the following works for [neovim](https://neovim.io/) with [coc](https://github.com/neoclide/coc.nvim):
`.vim/coc-settings.json`
```
{
    "languageserver": {
        "haskell": {
            "command": "nix-shell",
            "args": [
                "--run",
                "./.ghcide --lsp"
            ],
            "rootPatterns": [
                ".hie-bios"
            ],
            "filetypes": [
                "hs",
                "lhs",
                "haskell"
            ]
        }
    }
}
```

PR #1263, opened 2020-03-04
Add optional hie_bios output group to haskell_repl

aherrmann:
This allows users to configure [`hie-bios`](https://github.com/mpickering/hie-bios#readme) to use these outputs to determine GHC flags. See https://github.com/tweag/rules_haskell/pull/1262 (WIP) for an example setup with `ghcide`.

* Factors out the common flags for `ghci` and `hie-bios` in the `haskell_repl` implementation.
* Factors out the GHC flags required to configure `cc_wrapper` to avoid duplication.
* Creates an output group for `haskell_repl` that produces a file holding all relevant ghci flags for that target. (Excludes flags that are ghci specific, e.g. `repl_ghci_args` or flags required for `-ghci-script`)

A downside of this approach, compared to e.g. [hrepl](https://github.com/google/hrepl), is that `haskell_repl` targets are not generated on-the-fly but have to be defined in `BUILD` files. However, this is a general limitation of `haskell_repl`.

### DIFF
--- a/haskell/cc.bzl
+++ b/haskell/cc.bzl
@@ -159,3 +159,32 @@ def cc_interop_info(ctx):
             if CcInfo in dep
         ]).linking_context.libraries_to_link.to_list(),
     )
+
+def ghc_cc_program_args(cc):
+    """Retruns the -pgm* flags required to override cc.
+
+    Args:
+      cc: string, path to the C compiler (cc_wrapper).
+
+    Returns:
+      list of string, GHC arguments.
+    """
+    return [
+        # GHC uses C compiler for assemly, linking and preprocessing as well.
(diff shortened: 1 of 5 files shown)


haskell/cc.bzl
```diff
@@ -159,3 +159,32 @@ def cc_interop_info(ctx):
             if CcInfo in dep
         ]).linking_context.libraries_to_link.to_list(),
     )
+
+def ghc_cc_program_args(cc):
+    """Retruns the -pgm* flags required to override cc.
```
mboes: typo

aherrmann: Thanks! Fixed.

PR #1269, opened 2020-03-06
Add tests/pre-commit to automate `bazel test //...` and `bazel run //:buildifier-fix`

smelc:
This pre-commit should avoid pushing commits that will break CI. It should be fast in case you executed `bazel test //...` already (and if you didn't, the hook has your back!).

Nevertheless, I don't know if using a pre-commit hook is a usual Tweag workflow, so please discard this PR without providing details if you dislike this.

### DIFF
--- a/tests/pre-commit
+++ b/tests/pre-commit
@@ -1,0 +1,25 @@
+#!/bin/bash
+#
+# Hook that executes the tests and format *.bz,*.bazel files automatically
+
+set -e
+
+function format_bazel() {
+  # Adapted from https://prettier.io/docs/en/precommit.html
+  FILES=$(git diff --cached --name-only --diff-filter=ACMR "*.bz" "*.bazel" | sed 's| |\\ |g')
+  [ -z "$FILES" ] && return 0
+
+  # Format all files (it's imprecise as we know only $FILES need modifications, but it'll do)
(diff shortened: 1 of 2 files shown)


mboes:
Not a fan of hooks. They don't compose, they are hard to synchronize, and
complicate a bunch of workflows for little gain.

tests/pre-commit
```diff
@@ -0,0 +1,25 @@
...
+
+set -e
+
+function format_bazel() {
+  # Adapted from https://prettier.io/docs/en/precommit.html
+  FILES=$(git diff --cached --name-only --diff-filter=ACMR "*.bz" "*.bazel" | sed 's| |\\ |g')
+  [ -z "$FILES" ] && return 0
+
+  # Format all files (it's imprecise as we know only $FILES need modifications, but it'll do)
+  nix-shell --pure --command "bazel run //:buildifier-fix" || return 1
```
aherrmann: That's risky. There have been bugs in buildifier in the past that caused lossy transformations or changed semantics. E.g. https://github.com/bazelbuild/buildtools/issues/645 or https://github.com/bazelbuild/buildtools/issues/766. I would generally recommend reviewing the changes performed by buildifier.

smelc:
Given the remarks above, the hook won't do good, hence closing without merging.

Issue #1211, opened 2020-01-13
Type-check macro arguments manually


Profpatsch:
**Describe the bug**

We don’t check the types of arguments passed to macros we define, so users can get very confusing error messages when they update.

For example, the `extra_deps` field to `stack_snapshot` changed to be an attrset instead of a list, leading to the following error when updating:

```
ERROR: /home/philip/.cache/bazel/_bazel_philip/ec4bffc753a4896ad5f0595f4cbba0c3/external/rules_haskell/haskell/cabal.bzl:725:37: Traceback (most recent call last):
	File "/home/philip/kot/work/tweag/zeuslogics/WORKSPACE", line 108
		stack_snapshot(<5 more arguments>)
	File "/home/philip/.cache/bazel/_bazel_philip/ec4bffc753a4896ad5f0595f4cbba0c3/external/rules_haskell/haskell/cabal.bzl", line 1036, in stack_snapshot
		_stack_snapshot(stack = stack, <3 more arguments>)
	File "/home/philip/.cache/bazel/_bazel_philip/ec4bffc753a4896ad5f0595f4cbba0c3/external/rules_haskell/haskell/cabal.bzl", line 1040, in _stack_snapshot
		_from_string_keyed_label_list_dict(<1 more arguments>)
	File "/home/philip/.cache/bazel/_bazel_philip/ec4bffc753a4896ad5f0595f4cbba0c3/external/rules_haskell/haskell/cabal.bzl", line 725, in _from_string_keyed_label_list_dict
		d.items()
type 'list' has no method items()
ERROR: error loading package 'external': Package 'external' contains errors
```

This can be circumvented if we defensively type-check our macro’s inputs.

**To Reproduce**

Write a `WORKSPACE` file like

[WORKSPACE.txt](https://github.com/tweag/rules_haskell/files/4054012/WORKSPACE.txt)

and run `bazel build @stackage://...`

**Expected behavior**

The error thrown should be something like:

“stackage_snapshot `extra_deps` requires a dict from dependency name to list of targets, but was given a list of targets” 

**Environment**
 - OS name + version: nixos
 - Bazel version: 2.0.0
 - Version of the rules: head (see attached WORKSPACE file)

smelc:
Do you have a particular set of rules for which you'd like to have better error messages thanks to typing tests?

For example, we could add tests for the rules for which there are examples (output shows the number of `Examples` section within each file):

```
[...]

PR #1270, opened 2020-03-09
Typecheck stack_snapshot's extra_deps field

smelc:
I've added typechecking of stackage's extra_deps field, as described in #1220 

### Usual tests

```
bazel test //...
bazel run //:buildifier
```

### Tests specific to this issue

Using the `WORKSPACE` file described in #1211, with the following `stack_snapshot`:

``` bazel
stack_snapshot(
    name = "stackage",
    packages = [],
    extra_deps = [
        "foobar",
    ],
)
```

yields this error message: `stack_snapshot extra_deps requires a dict from dependency name to list of targets, but was given: list`

Using this `stack_snapshot`:

``` bazel
stack_snapshot(
    name = "stackage",
    packages = [],
    extra_deps = {
        1: ["zlib"]
    },
)
```

yields this error message: `stack_snapshot extra_deps's dict requires string keys, but key "1" has type int`

Using this `stack_snapshot`:

``` bazel
stack_snapshot(
    name = "stackage",
    packages = [],
    extra_deps = {
        "zlib" : { "foo" : "bar" }
    },
)
```

yields this error message: `stack_snapshot extra_deps's dict requires list values, but value "{"foo": "bar"}" has type dict`

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -11,6 +11,7 @@ load(":private/expansions.bzl", "expand_make_variables")
 load(":private/mode.bzl", "is_profiling_enabled")
 load(":private/path_utils.bzl", "join_path_list", "truly_relativize")
 load(":private/set.bzl", "set")
+load(":private/typing.bzl", "typecheck_stackage_extradeps")
 load(":haddock.bzl", "generate_unified_haddock_info")
 load(
     ":private/workspace_utils.bzl",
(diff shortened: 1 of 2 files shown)


smelc:
> It’s a bit weird to just check one argument, but it’s better than no check.

I stuck to @mboes's recommandation in https://github.com/tweag/rules_haskell/issues/1211#issuecomment-596137674

PR #1275, opened 2020-03-11
Join `package_root` paths with path separator

joneshf:
This fixes an issue I ran into when trying to depend on `@rules_haskell//tools/runfiles:runfiles` while also checking coverage. It was generating a flag to `hpc` like:
```
-hpcdir
bazel-out/k8-fastbuild/bin/external/rules_haskelltools/runfiles/.hpc
```
But, it seems like it should have been generating a flag to `hpc` like:
```
-hpcdir
bazel-out/k8-fastbuild/bin/external/rules_haskell/tools/runfiles/.hpc
```

The commit message has more details.

There's also a reproduction repo with a [failing commit](https://github.com/joneshf/rules_haskell_coverage_failure/tree/74ca9d8d24e8f3c53282bcf1d3b698f7c423445a) and a [fixed commit](https://github.com/joneshf/rules_haskell_coverage_failure/tree/b37142bffb3dd35153a7d2742920c5715b568d10).

Some things I was unsure about:
* [ ] I wasn't really sure how to write a regression test for this. It seems like it only happens if you depend on `runfiles` from elsewhere, but `runfiles` is part of this workspace. Might need some thinking. More than happy to add a test if there's a way to trigger the failure.
* [ ] I was unsure if the `src_root` should be built on top of the `package_root` (naively, it seems like yes), but I couldn't really tell if these two values are similar by coincidence, or if they necessarily have to be this way. I.e. `src_root` is `paths.join(package_root, src_strip_prefix)`, but not sure if the idea of a "src_root" (whatever that is) is always a child path of a "package_root" (whatever that is). More than happy to change this, if it makes sense.


aherrmann:
Thank you for fixing this, I also just bumped into this in a different context. :slightly_smiling_face: 

> * [ ]  I wasn't really sure how to write a regression test for this. It seems like it only happens if you depend on `runfiles` from elsewhere, but `runfiles` is part of this workspace. Might need some thinking. More than happy to add a test if there's a way to trigger the failure.

We have a [test-case that tests code in external repositories](https://github.com/tweag/rules_haskell/tree/42fcc5bb209fd87d767d0958b16297bff1cf1661/tests/external-haskell-repository), maybe the issue could be triggered in there?

> * [ ]  I was unsure if the `src_root` should be built on top of the `package_root` (naively, it seems like yes), but I couldn't really tell if these two values are similar by coincidence, or if they necessarily have to be this way. I.e. `src_root` is `paths.join(package_root, src_strip_prefix)`, but not sure if the idea of a "src_root" (whatever that is) is always a child path of a "package_root" (whatever that is). More than happy to change this, if it makes sense.

Yes, with this PR `src_root = paths.join(package_root, src_strip_prefix)` would be correct. Though, `src_strip_prefix` is marked as deprecated. So, I wouldn't worry about it in this PR.

joneshf:
> We have a [test-case that tests code in external repositories](https://github.com/tweag/rules_haskell/tree/42fcc5bb209fd87d767d0958b16297bff1cf1661/tests/external-haskell-repository), maybe the issue could be triggered in there?

Nice! I'll give that a whirl. Seems like it might have to grow a `coverage-compatible` tag to be included in the `bazel coverage` runs, that right?

aherrmann:
> Nice! I'll give that a whirl.

Thank you!

> Seems like it might have to grow a `coverage-compatible` tag to be included in the `bazel coverage` runs, that right?

Yes, that's right.

PR #1276, opened 2020-03-11
Follow http 301 when installing nix on the CI

regnat:
For a couple of hours, curl-ing the nix installer at
https://nixos.org/nix/install returned a 301 to another location, which
broke the darwin CI because the script wasn't following the redirect.

This has apparently been reverted upstream, but in case the revert gets
reverted again, better to tell curl to follow 301s

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
       - run:
           name: Install Nix
           command: |
-            curl https://nixos.org/nix/install | sh
+            curl -L https://nixos.org/nix/install | sh
 
       - run:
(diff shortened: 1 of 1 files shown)


mboes:
@aherrmann any idea why this one is wedged in the merge queue?

aherrmann:
> @aherrmann any idea why this one is wedged in the merge queue?

Looking at the merge queue it seems to be queued behind https://github.com/tweag/rules_haskell/pull/1273 which was stuck because Azure didn't pick it up. I've started an Azure run and will keep an eye on it.

Issue #741, opened 2019-03-12
Check bazel version


facundominguez:
Check that rules haskell is being used with a version of bazel that is known to work.

A lower bound check should ensure a not-too-old bazel is used. And an upper bound check should ensure that a not-to-new version of bazel is used with an old version of `rules_haskell`. Ideally, the error message should hint what to do: either try a different version of bazel, or a newer version of rules_haskell, or lift the upper bound.

facundominguez:
In other languages:
https://bazelbuild.github.io/rules_nodejs/common/check_bazel_version.html
https://github.com/bazelbuild/rules_go/issues/1108

facundominguez:
This is what we might want to use here:
https://github.com/bazelbuild/bazel-skylib/blob/master/docs/versions_doc.md

Got it from https://github.com/bazelbuild/bazel/issues/1014#issuecomment-436734881

smelc:
Closed by #1277

PR #1277, opened 2020-03-11
Check bazel version: print a warning if version not in range

smelc:
PR for issue #741

As discussed on slack, we don't to fail if the actual bazel version is not the expected one, we prefer to issue a warning. As a consequence, the code doesn't write anything if it cannot find the version (don't want to issue a warning about a possible warning..). 

For a reason detailed here: https://github.com/bazelbuild/bazel/issues/8305 checking bazel's version cannot be done anywhere, that is why it is done in [haskell/repositories.bzl](https://github.com/tweag/rules_haskell/blob/master/haskell/repositories.bzl).

I did not find a preferred way how to issue a warning message, so I'm simply using Starlark's `print` function for our messages, please reconsider this.

### In `rules_haskell/tutorial`, with bazel from APT

```
> sudo apt install bazel=1.2.1
> bazel --version
bazel 1.2.1
> bazel build //...
Starting local Bazel server and connecting to it...
DEBUG: /home/churlin/.cache/bazel/_bazel_churlin/b8c8fe86b6cf2c554ee68454e2b527a7/external/rules_haskell/haskell/private/versions.bzl:28:9: WARNING: bazel version is too old. Expected 2.0.0, but found: 1.2.1
... usual output
```

```
> sudo apt purge bazel; sudo apt install bazel=2.0.0
> bazel --version
bazel 2.0.0
> bazel build //...
Starting local Bazel server and connecting to it...
... usual output
```

```
> sudo apt purge bazel; sudo apt install bazel=2.1.0
> bazel --version
bazel 2.1.0
> bazel build //...
DEBUG: /home/churlin/.cache/bazel/_bazel_churlin/b8c8fe86b6cf2c554ee68454e2b527a7/external/rules_haskell/haskell/private/versions.bzl:31:9: WARNING: bazel version is too recent. Expected 2.0.0, but found: 2.1.0
... usual output
```

### At the repo's root, with bazel from nix-shell

```
> bazel --version
bazel 2.0.0- (@non-git)
> bazel test //...
# tests pass
> bazel run //:buildifier
# nothing to format
```

Then with a modified `nixkpgs/default.nix`, to have bazel `2.1.0`:

```
-  # 2020-01-29
-  sha256 = "1iy3iz91xx58k9wgswalipa8gxxdafqq82lisg6s16ml8y232f00";
-  rev = "42a195919a10ce70309c3666dfe8206e380fae1b";
+  # 2020-03-10
+  sha256 = "1kvyrs8g1zidrcs714k8g1gawrqln9x1nndpbfnblkglnzvvrac5";
+  rev = "3eea746d74d1fdfa57eef62b9159f5a51f23562a";
```

```
> bazel --version
bazel 2.1.0- (@non-git)
> bazel test //...
Starting local Bazel server and connecting to it...
DEBUG: /media/crypt1/repos/rules_haskell_741/haskell/private/versions.bzl:31:9: WARNING: bazel version is too recent. Expected 2.0.0, but found: 2.1.0- (@non-git)
# tests fail, because of an unexpected version of ghc (8.8.2 available, 8.6.5 expected)
> bazel run //:buildifier
DEBUG: /media/crypt1/repos/rules_haskell_741/haskell/private/versions.bzl:31:9: WARNING: bazel version is too recent. Expected 2.0.0, but found: 2.1.0- (@non-git)
# nothing to format
```

### DIFF
--- a/haskell/private/versions.bzl
+++ b/haskell/private/versions.bzl
@@ -1,0 +1,34 @@
+# check_version below cannot be called from anywhere
+# because native.bazel_version's availability is restricted:
+#   https://github.com/bazelbuild/bazel/issues/8305
+# An alternative hacky way is as follows:
+#   https://github.com/protocolbuffers/upb/blob/master/bazel/repository_defs.bzl
+#
+# There are utility functions for parsing versions numbers here:
+#   load("@bazel_skylib//lib:versions.bzl", "versions")
+# But we don't want to use them, as skylib is not yet loaded when code
+# in this file executes (and there's no way to execute it later; see first
+# paragraph above).
+
+def check_version(actual_version):
+    if type(actual_version) != "string" or len(actual_version) < 5:
+        return  # Unexpected format
+
+    expected_version = "2.0.0"  # Change THIS LINE when changing bazel version
+
+    actual_major = int(actual_version[0])
+    actual_minor = int(actual_version[2])
+    actual_patch = int(actual_version[4])
+    actual = (actual_major, actual_minor, actual_patch)
+
+    expected_major = int(expected_version[0])
+    expected_minor = int(expected_version[2])
+    expected_patch = int(expected_version[4])
+    expected = (expected_major, expected_minor, expected_patch)
+
+    if actual < expected:
+        print("WARNING: bazel version is too old. Expected {}, but found: {}".format(expected_version, actual_version))
+        return
+    if expected < actual:
+        print("WARNING: bazel version is too recent. Expected {}, but found: {}".format(expected_version, actual_version))
+        return
--- a/haskell/private/versions.bzl
+++ b/haskell/private/versions.bzl
@@ -52,8 +52,8 @@ def check_version(actual_version):
     if (min_bazel <= actual) and (actual <= max_bazel):
         return  # All good
 
-    min_bazel_string = ".".join([str(x) for x in list(min_bazel)])
-    max_bazel_string = ".".join([str(x) for x in list(max_bazel)])
+    min_bazel_string = ".".join([str(x) for x in min_bazel])
+    max_bazel_string = ".".join([str(x) for x in max_bazel])
 
     adjective = "old" if actual < min_bazel else "recent"
 
--- a/haskell/private/versions.bzl
+++ b/haskell/private/versions.bzl
@@ -10,25 +10,51 @@
 # in this file executes (and there's no way to execute it later; see first
 # paragraph above).
 
+def _parse_version_chunk(version_chunk):
+    """
+    Args:
+      version_chunk: a chunk of a semantic version, possibly with trailing
+                     content at the end; such as "29", "2", or "3 foobar".
+                     We handle trailing content because
+                     native.bazel_version can: try printing actual_version
+                     in check_version below (got for example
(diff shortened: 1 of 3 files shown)


haskell/private/versions.bzl
```diff
@@ -0,0 +1,34 @@
...
+#   load("@bazel_skylib//lib:versions.bzl", "versions")
+# But we don't want to use them, as skylib is not yet loaded when code
+# in this file executes (and there's no way to execute it later; see first
+# paragraph above).
+
+def check_version(actual_version):
+    if type(actual_version) != "string" or len(actual_version) < 5:
+        return  # Unexpected format
+
+    expected_version = "2.0.0"  # Change THIS LINE when changing bazel version
```
aherrmann: In `start` we define a range of valid Bazel versions. Why restrict it to a single version here, rather than using the same range?

smelc: Yes indeed, at the time I wrote this code it was unclear which version we're supported; and then we had the Bazel meeting (last week) and I made the release which made things clearer. I will update the PR accordingly.

smelc: I used the same range in `versions.bzl` and `start` in fccd1ae824c5a0e6a3747edde5cc34fec25390a9.

haskell/private/versions.bzl
```diff
@@ -0,0 +1,60 @@
...
+    min_bazel = (0, 29, 0)  # Change THIS LINE when changing bazel min version
+    max_bazel = (2, 0, 0)  # Change THIS LINE when changing bazel max version
+
+    actual = tuple(_parse_bazel_version(actual_version))
+
+    if (min_bazel <= actual) and (actual <= max_bazel):
+        return  # All good
+
+    min_bazel_string = ".".join([str(x) for x in list(min_bazel)])
+    max_bazel_string = ".".join([str(x) for x in list(max_bazel)])
```
aherrmann: Is the explicit `list` conversion necessary? IIRC tuples are directly iterable.

haskell/private/versions.bzl
```diff
@@ -0,0 +1,60 @@
...
+
+def _parse_version_chunk(version_chunk):
+    """
+    Args:
+      version_chunk: a chunk of a semantic version, possibly with trailing
+                     content at the end; such as "29", "2", or "3 foobar".
+                     We handle trailing content because
+                     native.bazel_version can: try printing actual_version
+                     in check_version below (got for example
+                     "2.0.0- (@non-git)").
```
aherrmann: Nothing to change, just FYI, this is an artifact of the Nix provided Bazel, see https://github.com/NixOS/nixpkgs/issues/80950.

haskell/private/versions.bzl
smelc: Indeed you are right, I removed the useless conversion in 9caec1b5b94374853603dddff84b31762bece19f and will merge into master once the CI validates the change.

aherrmann: @smelc Please set the `merge-queue` label instead of merging manually. This way mergify ensure that merge doesn't introduce CI failures on master and it will merge automatically once CI is green.

smelc: It was my intent to use mergify :smiley: but thanks for detailing the workflow.

PR #1267, opened 2020-03-05
Add relative RUNPATH entries in haskell_cabal_* targets

aherrmann:
`haskell_cabal_binary|library` targets that depended on Bazel managed dynamic C libraries (e.g. the nixpkgs provided `libz` for `@stackage//:zlib`) were missing relative `RUNPATH` entries. Instead, Cabal only generated absolute `RUNPATH` entries pointing to the sandbox's execroot, which are invalid outside that sandbox, see https://github.com/tweag/rules_haskell/issues/1130. This can cause build failures with Cabal packages that depend transitively on dynamic C libraries, e.g. `hie-bios` failed to build in https://github.com/tweag/rules_haskell/pull/1262.

This PR adds relative `RUNPATH` entries in the same fashion as is done for regular Haskell targets.

This PR also adds a regression test that checks for the presence of a relative `RUNPATH` entry.

PR #1282, opened 2020-03-16
Cabal: Add RUNPATH entries for transitive C library dependencies

aherrmann:
Add `RUNPATH` entries for all transitive C library dependencies to Cabal targets. GHC passes `-l` flags for all transitive C library dependencies when linking. Accordingly, we need to provide `RUNPATH` entries for all transitive C libraries. Also, to ensure that binaries can find their dynamic library dependencies, we add them to the `runfiles` attribute of the `DefaultInfo` provider.

Also, make static Haskell libraries available in Cabal build action, so that Cabal binaries can link statically against Haskell libraries.

Extends the test-case introduced in https://github.com/tweag/rules_haskell/pull/1267 to test for `RUNPATH` entries to transitive C library dependencies in Cabal binary targets.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -32,7 +32,9 @@ load(
 load(
     ":private/cc_libraries.bzl",
     "deps_HaskellCcLibrariesInfo",
+    "get_cc_libraries",
     "get_ghci_library_files",
+    "get_library_files",
     "haskell_cc_libraries_aspect",
 )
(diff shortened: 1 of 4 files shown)


regnat:
@Profpatsch @aherrmann, any idea why this isn't picked-up by the merge bot?

aherrmann:
Azure didn't pick it up.

PR #1285, opened 2020-03-17
shellcheck.bzl: remove unused arg + more uniform handling of quoting of arguments

smelc:
This PR:

- removes the unused `exclude_args` argument from the `shellcheck` rule
- quotes more uniformly arguments to the `shellcheck` executable, to be more future-proof, using skylib's [shell.quote](https://github.com/bazelbuild/bazel-skylib/blob/master/lib/shell.bzl#L37-L49)

##  Tests:

* `bazel test //...`
* `bazel run //:buildifier`
* Break shellchecking of some scripts files, observe that `bazel test //...` fails
* Check that rule `shellcheck`'s argument `sh_flavor` is still honored by fiddling `tests/shellcheck/BUILD.bzl` `sh_flavor` values, and observing that it breaks tests.

### DIFF
--- a/tests/shellcheck/shellcheck.bzl
+++ b/tests/shellcheck/shellcheck.bzl
@@ -1,18 +1,17 @@
 load("//tests:inline_tests.bzl", "sh_inline_test")
-load("@bazel_skylib//lib:shell.bzl", "shell")
 
 def shellcheck(name, args, data, sh_flavor = "sh", visibility = None):
-    cmd = ["shellcheck", "--color=always"]
+    fst_args = ["--color=always"]
     if sh_flavor != None:
-        cmd += ["--shell", sh_flavor]
+        fst_args += ["--shell", sh_flavor]
     sh_inline_test(
         name = name,
         visibility = visibility,
         size = "small",
-        args = args,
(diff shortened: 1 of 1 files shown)


tests/shellcheck/shellcheck.bzl
```diff
@@ -1,16 +1,18 @@
...
     sh_inline_test(
         name = name,
         visibility = visibility,
         size = "small",
         args = args,
         data = data,
         script = """\
-shellcheck --color=always {} {} "$1"
-""".format(shell_arg, excludes_arg),
+{} "$1"
```
aherrmann:
```suggestion
shellcheck "$@"
```
`args` is a list and could hold multiple values, correct? It just happens to only be called with a single argument at the time, so this never caused any issues.

Also, why not pass `--color=always` and optionally `--shell ...` in the `args` attribute to `sh_inline_test`? Then you don't have to worry about quoting.

smelc: Indeed, I've implemented your suggestion in dcac0a8. I believe the code has the simplest form possible now :+1:

Issue #978, opened 2019-06-29
Add stack_snapshot to start script


mboes:
**Is your feature request related to a problem? Please describe.**

Most users most of the time will need a third-party dependency hosted on Hackage. The preferred way to build those is using `stack_snapshot` and `haskell_cabal_library`. It is not obvious to the user that this is so just by running the `start` script alone.

**Describe the solution you'd like**

Since the above is the common case, we should include a dummy call to `stack_snapshot` in the `start` script. It could expose just `base` using that rule, even if in the case of `base`, use of this workspace rule is not strictly necessary.

One downside to doing this is that we'll be making Bazel projects depend on Stack by default. I don't think this is too onerous a dependency.

smelc:
An ambitious way to help users would be to make `start` generate most of the boilerplate.

A minimal way to do that would be to:

- Look for `stack.yaml` to infer `stack_snapshot`
- Look for `*.cabal` files to infer `haskell_cabal_*`
- Crawl subdirectories, to infer mimimal `BUILD.bazel` files for splitting the project bazel-package wise

If we don't invest too much in handling cabal subtleties, such a tool would have a good `output / effort` ratio (given that we appropriately advertise it's a helper tool, not an automatic translator..) and we could test its coverage easily by testing against public packages.

aherrmann:
I think what @mboes had in mind here was much simpler. 

> include a dummy call to `stack_snapshot` in the `start` script

We can make this a bit more featureful and include a dependency on a tool like `alex` (showcasing `haskell_cabal_binary`) and a C library like `zlib` just to demonstrate common use-cases, and to avoid people getting stuck early on.

---

The tool you're proposing, is more of a `stack2bazel` tool than a `start` script, I think.

smelc:
> I think what @mboes had in mind here was much simpler.

Yes of course, my intent is not to do `stack2bazel` for this task; I wanted to highlight a more ambitious way to make `rules_haskell` easier for newcomers :smiley:

mboes:
I think @aherrmann's suggested examples are great. Let's go for that.

smelc:
> I think @aherrmann's suggested examples are great. Let's go for that.

Implemented @aherrmann's examples:

- Introduced use of `alex` in `start` in 9e6943d
- Introduced use of `zlib` in `start` in e4cd57d.

PR #1286, opened 2020-03-19
start script: add stack_snapshot example

smelc:
PR for issue #978

### DIFF
--- a/start
+++ b/start
@@ -256,6 +256,36 @@ haskell_binary(
     srcs = [":Example.hs"],
     deps = [":base", "@stackage//:time"],
 )
+
+# Load cabal rules
+load(
+    "@rules_haskell//haskell:cabal.bzl",
+    "haskell_cabal_binary",
+)
+
+# An alternative way to compile Example.hs, using cabal to provide
+# the dependencies: note how neither base nor time is specified here. It's
+# specified in example.cabal instead
+# Use 'bazel run //:example' to build and execute it.
+haskell_cabal_binary(
+    name = "cabalexample",
+    srcs = [":Example.hs", "example.cabal"], # Include your Setup.hs
+    # file if you don't want to use the default
+)
+EOF
+
+echo "Creating example.cabal" >&2
+cat > example.cabal <<"EOF"
+cabal-version:       >=1.10
+name:                afancyname
+version:             0.1.0.0
+build-type:          Simple
+
+executable cabalexample
+  main-is:             Example.hs
+  build-depends:       base
+                     , time
+  default-language:    Haskell2010
 EOF
 
 echo "Creating Example.hs" >&2
@@ -274,5 +304,6 @@ EOF
 cat >&2 <<"EOF"
 WORKSPACE and initial BUILD files created. To run Bazel and build the example:
 
-    $ bazel run //:example
+    $ bazel run //:example # Build using stackage
+    $ bazel run //:cabalexample # Build using cabal
 EOF
--- a/start
+++ b/start
@@ -110,6 +110,7 @@ nix_toolchain=$(cat <<EOF
 load(
     "@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl",
     "nixpkgs_git_repository",
+    "nixpkgs_package",
     "nixpkgs_python_configure",
 )
 
--- a/start
+++ b/start
@@ -169,6 +169,31 @@ EOF
 )
 
 nix_zlib=$(cat <<EOF
+nixpkgs_package(
+    name = "nixpkgs_zlib",
+    attribute_path = "zlib",
+    repository = "@nixpkgs",
+)
+
+nixpkgs_package(
+    name = "zlib.dev",
+    build_file_content = """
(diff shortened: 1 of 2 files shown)


start
```diff
@@ -181,6 +181,19 @@ load(
...
 
+load(
+    "@rules_haskell//haskell:cabal.bzl",
+    "stack_snapshot"
+)
+
+stack_snapshot(
+    name = "stackage",
+    packages = [
+        "time", # Provides the dependency @stackage//:time
```
mboes: Redundant comment.

smelc: Comment has been removed in c40197be5973b2e70bb0f7a33f95e388cbd25962.

smelc:
> `time` is not a great example, because `time` is not in Stackage per say. It's part of the core libraries that ship with GHC. See the output of `ghc-pkg list`.

I didn't know this difference as I had only used packages through stack :+1: 

I used `zlib` instead of `time` starting at e4cd57d.

start
```diff
@@ -142,13 +142,76 @@ haskell_register_ghc_nixpkgs(
...
+    name = "zlib",
+    # Import ':z' as 'srcs' to enforce the library name 'libz.so'. Otherwise,
+    # Bazel would mangle the library name and e.g. Cabal wouldn't recognize it.
+    srcs = [":z"],
+    hdrs = glob(["*.h"]),
+    includes = ["."],
+    visibility = ["//visibility:public"],
+)
+cc_library(name = "z", srcs = glob(["*.c"]), hdrs = glob(["*.h"]))
+""",
```
mboes: Using `build_file` instead of `build_file_content` would be more readable. There's no syntax highlighting inside a string. You could store the content in `zlib.BUILD.bazel`.

start
```diff
@@ -181,7 +245,47 @@ load(
...
+    "stack_snapshot"
+)
+
+stack_snapshot(
+    name = "stackage",
+    extra_deps = {"zlib": ["@zlib.dev//:zlib"]},
+    packages = ["zlib"],
+    snapshot = "lts-14.27" # Last snapshot published for ghc-8.6.5
+    # the default version picked up by rules_haskell
+)
```
aherrmann: Sorry, I should have been more clear what I meant with the Alex example. What I had in mind was to use it as a `tool` in `stack_snapshot`, like we do in `rules_haskell`: https://github.com/tweag/rules_haskell/blob/827d654235189242b0a9d45df8bc8754469b4579/WORKSPACE#L105-L108

Such tool dependencies are relatively common as a project's dependency graph grows. So, it seems reasonable to include an example in the `start` script to avoid people getting stuck when they reach that point.

start
```diff
@@ -237,25 +341,43 @@ haskell_toolchain_library(name = "base")
...
+    "haskell_cabal_binary",
+)
+
+# Example usage of alex, use 'bazel builds //:words' to execute it
+genrule(
+    name = "words",
+    srcs = ["@alex", "@alex_example//file"],
+    outs = ["Words.hs"],
+    cmd = "$(location @alex) $(location @alex_example//file) -o $(location Words.hs)",
 )
```
aherrmann: Using Alex directly in a project, on the other hand, is much less common. So, I don't think there is much value in including such an example in the `start` script. I expect for most users it will just be a slight annoyance that they have to remove this code in their fresh project.

smelc: I get your point, as a bazel newcomer I'd surely like having a bit more than a bit less in the `start` script, but that's not the targeted audience.

start
smelc: Issue addressed in 8f723ad7bd326ddadea066e252385c1dafa527a8

start
smelc: I've moved inlined BUILD content in standalone files (alex.BUILD.bazel and zlib.BUILD.BAZEL) 
 in 8f723ad7bd326ddadea066e252385c1dafa527a8. I've removed alex's genrule too.

start
smelc: Issue addressed in 8f723ad

start
```diff
@@ -181,7 +263,45 @@ load(
...
+
+echo "Creating $ALEX_BUILD_FILE" >&2
+cat > "$ALEX_BUILD_FILE" <<EOF
+load("@rules_haskell//haskell:cabal.bzl", "haskell_cabal_binary")
+
+haskell_cabal_binary(
+    name = "alex",
+    srcs = glob(["**"]),
+    visibility = ["//visibility:public"],
+)
```
aherrmann:
```suggestion
haskell_cabal_binary(
    name = "alex",
    srcs = glob(["**"]),
    verbose = False,
    visibility = ["//visibility:public"],
)
```
It makes sense to set `verbose = False`, since it's an external dependency

smelc: Done in 7c8135e023f585da4c17e05d17ad0664c2ce69e3

start
mboes: But it's also not @aherrmann's point. The `start` script is not meant to show you the 10% case. It's meant to get you started with the 80% case. Other places than this are more appropriate to document uncommon cases (such as using `alex` in a genrule), e.g. as an example in the examples directory.

PR #1291, opened 2020-03-24
Tutorial: highlight necessary content of the WORKSPACE file

### LINKED ISSUE #780: Create tutorial for setting up a haskell toolchain (bindists & nixpkgs)
The current tutorial focuses on what to do *once* you have a working toolchain setup.

As we can see in issues like https://github.com/tweag/rules_haskell/issues/776, bazel’s error messages are extremely poor when it comes to workspace rules and toolchains, plus the logic on what goes where is *extremely* convoluted, especially for a new user.

We must provide a step-by-step example and self-contained setups that can be copied for each supported path and decouple them from `rules_haskell` in a way that they are self-contained (which is not extremely easy, because `bazel //...` happily tries to build all subdirectories, even those with `WORKSPACE` files in them.

smelc:
PR for issue #780 

Highlight that the way to obtain rules_haskell in tutorial/WORKSPACE is limited to the tutorial's case, otherwise better refer to the template created by start

also link to Common Haskell Build Use Cases to highlight the other mandatory thing in WORKSPACE file: obtaining GHC

I tried to make a minimal addition, so as to avoid polluting the tutorial; yet giving the sufficient pointers to make users able to set up a toolchain

### DIFF
--- a/docs/haskell.rst
+++ b/docs/haskell.rst
@@ -32,7 +32,7 @@ On a Unix system you will need the following tools installed.
 * ``libgmp``
 * ``libtinfo5``
 * ``make``
-* ``python3`` (``python`` also needs to be available in ``$PATH``. Depending on your distro, this might require installing the ``python`` meta-package, which might use Python 2 or 3, ``rules_haskell`` works with both.)
+* ``python3`` (``python`` also needs to be available in ``$PATH``. Depending on your distribution, this might require installing the ``python`` meta-package, which might use Python 2 or 3, ``rules_haskell`` works with both.)
 
 On Ubuntu you can obtain them by installing the following packages. ::
 
@@ -89,13 +89,31 @@ special:
   contains a ``BUILD.bazel`` file is a *package*. You will learn about
   packages later in this tutorial.)
 
-To designate a directory as a Bazel workspace, create an empty file
+To designate a directory as a Bazel workspace, create a file
 named ``WORKSPACE`` in that directory.
+This file defines `external dependencies`_.
 
 When Bazel builds the project, all inputs and dependencies must be in
 the same workspace. Files residing in different workspaces are
 independent of one another unless linked, which is beyond the scope of
 this tutorial.
+
+Understand the WORKSPACE file
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+File ``tutorial/WORKSPACE`` defines how to obtain ``rules_haskell``
+and a compiler toolchain.  Because this tutorial lives in ``rules_haskell``'s
+repository, ``rules_haskell`` is made available as follows::
+
+  local_repository(
+      name = "rules_haskell",
+      path = "..",
+  )
+
+We refer to the template created by the start_ script
+to make ``rules_haskell`` available outside a local clone.
+The ``WORKSPACE`` file also defines how to obtain the compiler toolchain,
+as detailed in section `picking a compiler`_.
 
 Understand the BUILD file
 ^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


docs/haskell.rst
```diff
@@ -89,14 +89,32 @@ special:
...
 the same workspace. Files residing in different workspaces are
 independent of one another unless linked, which is beyond the scope of
 this tutorial.
 
+Understand the WORKSPACE file
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+File ``tutorial/WORKSPACE`` defines how to obtain ``rules_haskell``
+and a compiler toolchain.  Because this tutorial lives in ``rules_haskell``'s
+repository, ``rules_haskell`` is made available as follows::
```
Profpatsch:
```suggestion
repository, we reference ``rules_haskell`` as follows::
```

docs/haskell.rst
```diff
@@ -89,14 +89,32 @@ special:
...
+
+  local_repository(
+      name = "rules_haskell",
+      path = "..",
+  )
+
+We refer to the template created by the start_ script
+to make ``rules_haskell`` available outside a local clone.
```
Profpatsch:
```suggestion
If you want to set up the `WORKSPACE` file for your own repository,
run the start_ script and it will generate a fitting configuration.
```

aherrmann:
> Maybe we should make it clearer in the tutorial, that users should not copy the `tutorial` files, but run the `start` script.

Agreed, and I'd go a step further and remove the `local_repository` snippet. A good guideline: If you show a code snippet, then it should be correct to copy-paste it.

Instead, you can have a sentence pointing out that `tutorial/WORKSPACE` only works within the `rules_haskell` repository. For user's own projects point to the `start` script, and maybe also include a `http_archive` snippet. Most rules sets have a copy-pasteable `http_archive` snippet in their setup instructions. E.g. [`rules_sh`](https://github.com/tweag/rules_sh#setup), [`rules_go`](https://github.com/bazelbuild/rules_go#setup), [`rules_nodejs`](https://bazelbuild.github.io/rules_nodejs/install.html).

Profpatsch:
> A good guideline: If you show a code snippet, then it should be correct to copy-paste it.

Haha, I had a sentence like this, but then deleted it again. So yeah, +1

Profpatsch:
> and maybe also include a `http_archive` snippet. Most rules sets have a copy-pasteable `http_archive` snippet in their setup instructions. E.g. [`rules_sh`](https://github.com/tweag/rules_sh#setup), [`rules_go`](https://github.com/bazelbuild/rules_go#setup), [`rules_nodejs`](https://bazelbuild.github.io/rules_nodejs/install.html).

Even if you get the `http_archive` stuff right, it’s still *way* too easy to miss some implicit invocation and have nothing work with abysmal error messages. I would put heavy focus on telling people to *always* use the start script.

smelc:
> Agreed, and I'd go a step further and remove the `local_repository` snippet. A good guideline: If you show a code snippet, then it should be correct to copy-paste it.

I agree, but the solutions we're discussing are working around the fact that the tutorial fetches `rules_haskell` in a way that doesn't work for production.

Is there a good reason for that? Could we change the tutorial it so that it fetches `rules_haskell` using `http_archive` like the `start` script does? We could still leave the `local_archive` example commented in `WORKSPACE`, as I understand it can be convenient to work against the local clone for experimenting.

smelc:
To proceed, in ea30ef6a3fd1714814cd40aff2deac9a648f036b, I added a `http_archive` snippet in the _use cases_ page and referenced it in the tutorial. I also removed the `local_repository` snippet I had added in my first commit in the tutorial.

docs/haskell.rst
```diff
@@ -89,14 +89,27 @@ special:
...
 
 When Bazel builds the project, all inputs and dependencies must be in
 the same workspace. Files residing in different workspaces are
 independent of one another unless linked, which is beyond the scope of
 this tutorial.
 
+Understand the WORKSPACE file
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+File ``tutorial/WORKSPACE`` defines how to obtain ``rules_haskell``.
```
aherrmann:
```suggestion
The file ``tutorial/WORKSPACE`` defines how to obtain ``rules_haskell``.
```

docs/haskell.rst
```diff
@@ -89,14 +89,27 @@ special:
...
+This file works within the ``rules_haskell``
```
aherrmann:
```suggestion
This file only works within the ``rules_haskell``
```

smelc: Done in bec27b4.

docs/haskell.rst
smelc: Done in bec27b4.

PR #1300, opened 2020-04-16
Set runfiles attribute in haskell_doc

aherrmann:
This is to make generated Haddocks available in the runfiles tree of targets that have a `data` dependency on a `haskell_doc` target.

Adds a regression test to `//tests/haddock`.

### DIFF
--- a/haskell/haddock.bzl
+++ b/haskell/haddock.bzl
@@ -321,9 +321,10 @@ def _haskell_doc_rule_impl(ctx):
         arguments = [args],
     )
 
+    files = html_dict_copied.values() + [index_root]
     return [DefaultInfo(
-        files = depset(html_dict_copied.values() + [index_root]),
-        runfiles = ctx.runfiles(html_dict_copied.values() + [index_root]),
+        files = depset(files),
+        runfiles = ctx.runfiles(files),
     )]
 
 haskell_doc = rule(
--- a/haskell/haddock.bzl
+++ b/haskell/haddock.bzl
@@ -323,6 +323,7 @@ def _haskell_doc_rule_impl(ctx):
 
     return [DefaultInfo(
         files = depset(html_dict_copied.values() + [index_root]),
+        runfiles = ctx.runfiles(html_dict_copied.values() + [index_root]),
     )]
 
 haskell_doc = rule(
--- a/tests/haddock/BUILD.bazel
+++ b/tests/haddock/BUILD.bazel
@@ -56,8 +56,32 @@ haskell_doc(
     deps = [":haddock-lib-b"],
 )
 
+sh_test(
+    name = "data-dependency",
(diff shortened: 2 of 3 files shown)


tests/haddock/BUILD.bazel
```diff
@@ -56,8 +56,32 @@ haskell_doc(
...
 )
 
+sh_test(
+    name = "data-dependency",
+    srcs = ["data_dependency.sh"],
+    args = ["$(rootpaths :haddock)"],
+    data = [":haddock"],
+    tags = [
+        # Fails in profiling mode due to missing haddocks for Deep.
+        "requires_dynamic",
```
aherrmann, line 66: This is not a regression introduced with this PR but already the case on `master`. I.e. the following fails on master:
```
$ nix-shell --pure --run 'bazel test //tests/haddock/... -c dbg'
...
ERROR: /.../rules_haskell/tests/haddock/BUILD.bazel:18:1: HaskellHaddock //tests/haddock:haddock-lib-a failed (Exit 1) haddock_wrapper-haddock-lib-a failed: error executing command bazel-out/k8-dbg/bin/tests/haddock/haddock_wrapper-haddock-lib-a '--package-name=testsZShaddockZShaddock-lib-a' '--package-version=0' -D ... (remaining 96 argument(s) skipped)
...
tests/haddock/LibA.hs:10:1: error:
    Could not find module ‘Deep’
    There are files missing in the ‘haddock-lib-deep’ package,
    try running 'ghc-pkg check'.
    Use -v to see a list of the files searched for.
   |
10 | import Deep (deep_lib)
   | ^^^^^^^^^^^^^^^^^^^^^^
INFO: Elapsed time: 1.354s, Critical Path: 0.57s
INFO: 0 processes.
```
This does not cause a CI failure on master, because we pass the flag `--build_tests_only` in profiling mode. This PR introduces a test case that depends on the haddock targets, which means they are now included despite `--build_tests_only`. We set the tag to disable the test on profiling builds.

haskell/haddock.bzl
```diff
@@ -323,6 +323,7 @@ def _haskell_doc_rule_impl(ctx):
 
     return [DefaultInfo(
         files = depset(html_dict_copied.values() + [index_root]),
+        runfiles = ctx.runfiles(html_dict_copied.values() + [index_root]),
```
guibou, line 326: The `html_dict` is huge and here we are building two temporaries copies of it.

I do wonder what can be the impact on performance. Something such as:

```
    l = html_dict_copied.values() + [index_root]
    return [DefaultInfo(
￼        files = depset(l),
￼        runfiles = ctx.runfiles(l),
```

may be better.

aherrmann: Good catch. It's tracking directories, not individual files, so one entry per target, but you're right, better not copy it unnecessarily.

PR #1304, opened 2020-04-20
stack_snapshot build executable components

aherrmann:
Depends on #1303.

Adds a new attribute `components` to `stack_snapshot` which allows users to specify which Cabal components to build for a given package in the snapshot (defaults to building the library component). E.g. for the `proto-lens-protoc` package a user could specify the following to build both the library and executable components.
```
    components = {
        "proto-lens-protoc": [
            "lib",
            "exe",
        ],
    },
```
This reduces the need for dedicated `http_archive` definitions for executable components of Cabal packages.

The executable targets are available under labels of the form `@<snapshot>-exe//<package_name>:<exe_name>`, as discussed in https://github.com/tweag/rules_haskell/issues/1306. For executable components that share the package's name the short form `@<snapshot>-exe//<package>` applies. For technical reasons (described in #1306) these executable targets are `alias`es pointing to `haskell_cabal_binary` targets of the form `@<snapshot>//:_<package>_exe_<exe_name>`. These labels are considered an implementation detail and not publicly visible.

The dependency graph generated by `stack` captures the library and executable dependencies of a package. When generating `haskell_cabal_binary` targets based on declared `exe` components, we pass in all package dependencies as well as the library component (if present).

The dependency graph generated by `stack` does not provide information on which components a package includes. Therefore, there is no easy way to automatically determine the set of components. However, only few packages define executable components, so it is feasible for users to manually define them as needed.

A set of known executable components is included in `rules_haskell` for common packages such as `alex` or `happy`.

`proto-lens-protoc` is converted from a dedicated `http_archive` to use the `components` attribute instead. This functions as a test-case for this feature.

`ghcide` is also converted from a dedicated `http_archive` to use the `components` attribute. The `ghcide` use-case documentation is updated accordingly.

Note, this PR does not yet inject executable components into the `tools` attribute of generated Cabal targets. E.g. `alex` and `happy` still have to be passed into `stack_snapshot` using the `tools` attribute. This will be addressed in a follow-up PR.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -1102,6 +1102,8 @@ def _stack_snapshot_impl(repository_ctx):
             package.name: struct(
                 name = package.name,
                 version = package.version,
+                library = package.components.lib,
+                executables = package.components.exe,
                 deps = [Label("@{}//:{}".format(repository_ctx.name, dep)) for dep in package.deps],
                 flags = package.flags,
             )
@@ -1202,17 +1204,18 @@ haskell_cabal_library(
                 build_file_builder.append(
                     """
 haskell_cabal_binary(
-    name = "{name}_exe_{exe}",
+    name = "_{name}_exe_{exe}",
     exe_name = "{exe}",
     flags = {flags},
     srcs = glob(["{dir}/**"]),
     deps = {deps},
     tools = {tools},
-    visibility = {visibility},
+    visibility = ["@{workspace}-exe//{name}:__pkg__"],
     compiler_flags = ["-w", "-optF=-w"],
     verbose = {verbose},
 )
 """.format(
+                        workspace = repository_ctx.name,
                         name = package.name,
                         exe = exe,
                         flags = package.flags,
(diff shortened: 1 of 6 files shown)


haskell/cabal.bzl
```diff
@@ -845,6 +900,10 @@ def _compute_dependency_graph(repository_ctx, snapshot, core_packages, versioned
         is_core_package = name in _CORE_PACKAGES
         all_packages[name] = struct(
             name = name,
+            components = _parse_components(
+                name,
+                remaining_components.pop(name, ["lib"]),
```
guibou, line 905: Why only `"lib"`?

aherrmann: That forms the default set of package components. I.e. by default we say a package only has a library component. ([`dict.pop(key, default=unbound)`](https://docs.bazel.build/versions/master/skylark/lib/dict.html#pop))

Profpatsch:
> broken_heart Pull request #1303's state different from open.

Hm, what does this message mean?

aherrmann:
> > broken_heart Pull request #1303's state different from open.
> 
> Hm, what does this message mean?

I updated the PR message and didn't remove the "depends on" line. The PR that this one depends on has since been merged, so dpulls complains about depending on a closed PR. However, the status dpulls emits is green, so it can be ignored.

PR #1305, opened 2020-04-20
stack_snapshot use executable components in tools attribute

### LINKED ISSUE #1306: Build executables in stack_snapshot and use them as tools

Some of the packages defined by a call to `stack_snapshot` depend on executables (tools) that themselves are Hackage packages. Examples include libraries like [`language-c`](http://hackage.haskell.org/package/language-c) which depends on [`alex`](http://hackage.haskell.org/package/alex) and [`happy`](http://hackage.haskell.org/package/happy), or [`grpc-haskell-core`](http://hackage.haskell.org/package/grpc-haskell-core) which depends on [`c2hs`](http://hackage.haskell.org/package/c2hs). Furthermore, some packages provide executables that the user of `stack_snapshot` will likely wish to invoke directly, e.g. [`doctest`](http://hackage.haskell.org/package/doctest) or [`ghcide`](http://hackage.haskell.org/package/ghcide).

Since https://github.com/tweag/rules_haskell/pull/907 `stack_snapshot` provides the `tools` attribute which allows to pass executables that are built outside of `stack_snapshot` in to be available as build tools for the packages defined by `stack_snapshot`. This approach is very limited:
1) Every package depends on the same set of tools, which means that changing any tools or adding additional tools requires a full rebuild of all packages.
2) This does not scale well for tools that appear in the middle of the dependency graph. `c2hs` is such an example, it itself depends on Hackage packages like `language-c`, which in turn depends on `alex` and `happy`. I.e. to build a package that depends on `c2hs`, say `haskell-grpc-core` we need to provide `alex` and `happy`, then a `stack_snapshot` for `c2hs` dependencies, then build `c2hs` itself, and then finally a `stack_snapshot` for `haskell-grpc-core` and other packages. I.e. in a typical use-case we end up with external workspaces `@alex`, `@happy`, `@c2hs_deps`, `@c2hs`, and `@stackage`.
3) This is very inconvenient for packages with many dependencies that provide executables. E.g. `ghcide` has many dependencies and provides a library and an executable. If we wish to build it with `stack_snapshot` we need to define a `stack_snapshot` exporting `ghcide` the library and all dependencies of `ghcide` the binary, and then define a separate external workspace for the `ghcide` binary where we import all these dependencies. This incurs a lot of boilerplate and we have to be careful to keep the versions/revisions in sync between `stack_snapshot` and the `http_archive` used for the `haskell_cabal_binary` call. Alternatively, we could build both the library and binary in one `http_archive` but then we have to manually perform the dependency resolution for `ghcide`.

I'd like `stack_snapshot` to build both library and executable components of packages and integrate them into the dependency graph as appropriate. I.e. we shouldn't have to provide `alex`, `happy`, `c2hs`, ... to the `tools` attribute of `stack_snapshot`. And we shouldn't have to resort to intermediate `stack_snapshot` invocations as required for tools like `c2hs`.

As of now `stack` does not provide us any information about the components provided by a package. And we would prefer to avoid implementing a Cabal file parser in Starlark. For that reason I'd propose to have users define the components of the packages they expect to be built, with a predefined set for common packages like `alex`, `happy`, .... In the end this would look something like this

```
stack_snapshot(
    name = "stackage",
    packages = [
        "haskell-grpc-core",
        "ghcide",
    ]
    components = {
        "ghcide": ["lib", "exe"],  # Where `exe` is short for `exe:<package_name>`
        # The following are common and included in `rules_haskell`
        # "alex": ["exe"],
        # "happy": ["exe"],
        # "c2hs": ["exe"],
    },
)
```

* Hazel used to parse Cabal files to automatically determine which components are available. However, this introduces a fair amount more complexity as seems necessary that we would prefer to avoid. In any case this only addresses the question of whether the `components` attribute should be user defined or inferred.
* `stack ls dependencies json` could possibly be patched to include information on package components. However, it is unclear how much effort that would require. Also, parsing JSON into Starlark is not without issues either. As above this only addresses whether `components` should be user defined or inferred. 

**Additional context**

* The issue was discussed briefly in https://github.com/tweag/rules_haskell/pull/907
* https://github.com/tweag/rules_haskell/pull/1305, https://github.com/tweag/rules_haskell/pull/1304, https://github.com/tweag/rules_haskell/pull/1303 provides a possible implementation.
* A discussion around the naming of the generated library and executable targets was started in https://github.com/tweag/rules_haskell/pull/1303 and will be continued here.

aherrmann:
Closes #1306 

Uses executable components of package dependencies in the `tools` attribute in Cabal targets generated by `stack_snapshot`. E.g. we no longer need to build `alex` and `happy` outside of `stack_snapshot` to then pass them in via `stack_snapshot`'s `tools` attribute. Instead, `stack_snapshot` will generate targets `@stackage-exe//alex` and `@stackage-exe//happy` and pass them into the `tools` attribute of any generated `haskell_cabal_*` target that depends on `alex` or `happy` respectively.

A use-case where this becomes very useful is the tool `c2hs`. Contrary to `alex` or `happy` it does itself depend on Hackage packages, e.g. `language-c`. Meaning, if someone wants to include a package that depends on `c2hs` in a `stack_snapshot` call, say [`grpc-haskell-core`](http://hackage.haskell.org/package/grpc-haskell-core), then one has to first build `c2hs` separately. To achieve that one has to define a separate `stack_snapshot` for the dependencies of `c2hs`. This approach does not scale very well. With this PR the issue is avoided, as `stack_snapshot` will generate `@stackage-exe//c2hs` itself and pass it to `grpc-haskell-core` automatically.

The `tools` attribute to `stack_snapshot` is still available in case users do wish to provide external tools, e.g. a Nix provided `alex` binary. We keep an externally provided `alex` in `rules_haskell`'s `stack_snapshot` as a test-case for this.

This also simplifies the `start` script, as it is no longer a common use-case to have to provide an externally built tool.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -20,6 +20,22 @@ bazel_skylib_workspace()
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
+http_archive(
+    name = "alex",
+    build_file_content = """
+load("@rules_haskell//haskell:cabal.bzl", "haskell_cabal_binary")
+haskell_cabal_binary(
+    name = "alex",
+    srcs = glob(["**"]),
+    verbose = False,
+    visibility = ["//visibility:public"],
+)
+    """,
+    sha256 = "d58e4d708b14ff332a8a8edad4fa8989cb6a9f518a7c6834e96281ac5f8ff232",
+    strip_prefix = "alex-3.2.4",
+    urls = ["http://hackage.haskell.org/package/alex-3.2.4/alex-3.2.4.tar.gz"],
+)
+
 load(
     "@rules_haskell//:constants.bzl",
     "test_ghc_version",
@@ -30,6 +46,7 @@ load("@rules_haskell//haskell:cabal.bzl", "stack_snapshot")
 stack_snapshot(
     name = "stackage",
     components = {
+        "alex": [],
         "proto-lens-protoc": [
             "lib",
             "exe",
@@ -67,6 +84,13 @@ stack_snapshot(
     ],
     setup_deps = {"polysemy": ["cabal-doctest"]},
     snapshot = test_stack_snapshot,
+    tools = [
+        # This is not required, as `stack_snapshot` would build alex
+        # automatically, however it is used as a test for user provided
+        # `tools`. We also override alex's components to avoid building it
+        # twice.
+        "@alex",
+    ],
 )
 
 # In a separate repo because not all platforms support zlib.
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -20,32 +20,6 @@ bazel_skylib_workspace()
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
-http_archive(
-    name = "alex",
-    build_file_content = """
-load("@rules_haskell//haskell:cabal.bzl", "haskell_cabal_binary")
-haskell_cabal_binary(
-    name = "alex",
-    srcs = glob(["**"]),
-    visibility = ["//visibility:public"],
-)
-    """,
-    sha256 = "d58e4d708b14ff332a8a8edad4fa8989cb6a9f518a7c6834e96281ac5f8ff232",
-    strip_prefix = "alex-3.2.4",
-    urls = ["http://hackage.haskell.org/package/alex-3.2.4/alex-3.2.4.tar.gz"],
-)
-
-http_archive(
-    name = "happy",
-    build_file_content = """
-load("@rules_haskell//haskell:cabal.bzl", "haskell_cabal_binary")
-haskell_cabal_binary(name = "happy", srcs = glob(["**"]), visibility = ["//visibility:public"])
-    """,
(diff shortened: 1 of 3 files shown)


aherrmann:
Tested on the `daml` repository, see https://github.com/digital-asset/daml/pull/5631/commits/04f38dce94a09f9585a893fa8c2b0c389a8f1e45.

aherrmann:
> This looks fine, but I don't see in the PR where these executables are being called. What does their label look like?

The user facing labels are the ones generated by https://github.com/tweag/rules_haskell/pull/1304, e.g. `@stackage-exe//alex`. These user facing labels are only generated for packages that are listed in the `packages` attribute. I.e. binaries of private packages are not exposed. Within `stack_snapshot` itself tools are referenced by [this code](https://github.com/tweag/rules_haskell/pull/1305/files#diff-b8d0916c99703c813c8bcf9b8585a5cfR1163-R1166) using the internal labels of the form `@stackage//:_alex_exe_alex`. This is so that dependencies on executables of private packages work as well. These labels are not user facing.

PR #1307, opened 2020-04-20
Update bazel 2.0.0 --> 2.1.0

aherrmann:


aherrmann:
It looks like repository rules became a little more lazy with the Bazel update. The [`stack_update` mechanism](https://github.com/tweag/rules_haskell/pull/1199) used to prevent a race on the Hackage lock [stopped working on Windows](https://dev.azure.com/tweag/rules_haskell/_build/results?buildId=4204&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=a8035825-ecc2-53a4-dc8f-13b782b0dc20&l=3415). Adding a [`read` of the `stack_update` file](https://github.com/tweag/rules_haskell/pull/1307/commits/86339f365babfc650cf617be0b9fd905abe04d4e) was sufficient to enforce the dependency again. Long-term I'd like to find a better solution, ideally some form of pinning similar to `rules_jvm_external`. But that's out of scope for this PR, I'll add a separate issue about that idea.

PR #1316, opened 2020-04-30
Make the haddock files of the deps available for cabal builds

### LINKED ISSUE #1292: Docstrings are lost for cross-package reexports in `cabal_package`
**Describe the bug**

If a module `Foo` reexports some symbol `x` from another (module in another) package, then the doctsring for `x` will be missing in the haddocks of `Foo` − `x` will be present but without any documentation that might be attached to it.
For example, `Test.Hspec` (from `hspec`) reexports `it` (amongst others) from `Test.Hspec.Core.Spec` (from `hspec-core`), and the documentation of `Test.Hspec` doesn't show the documentation for `it`:

In `Test.Hspec.Core.Spec`

image

In `Test.Hspec`

image

**To Reproduce**

From current's `rules_haskell` master (9a52a45410e0aa0e7005a572b40bf1320fd65220):

```shell
bazel build @hackage//:hspec
firefox bazel-bin/external/stackage/hspec-2.7.1/_install/hspec-2.7.1_haddock_html/Test-Hspec.html
firefox bazel-bin/external/stackage/hspec-core-2.7.1/_install/hspec-core-2.7.1_haddock_html/Test-Hspec-Core-Spec.html
```

**Expected behavior**

The reexported value should be fully documented (like it is the case with stack or cabal)

**Environment**
 - OS name + version: NixOS 20.03
 - Bazel version: 2.0.0 (from the nix-shell)
 - Version of the rules: 9a52a45410e0aa0e7005a572b40bf1320fd65220

regnat:
Fix #1292

I didn't add a test for this as I'm not sure how to do that without introspecting either the generated html or the haddock stderr (to ensure that there's no warning wrt a missing identifier), both of which would be painful and brittle.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -135,6 +135,7 @@ def _prepare_cabal_inputs(
         cabal_wrapper,
         package_database,
         verbose,
+        transitive_haddocks,
         dynamic_binary = None):
     """Compute Cabal wrapper, arguments, inputs."""
     with_profiling = is_profiling_enabled(hs)
@@ -216,6 +217,7 @@ def _prepare_cabal_inputs(
             transitive_headers,
             depset(transitive_compile_libs),
             depset(transitive_link_libs),
+            depset(transitive_haddocks),
             dep_info.interface_dirs,
             dep_info.static_libraries,
             dep_info.dynamic_libraries,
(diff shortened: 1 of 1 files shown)


aherrmann:
@regnat Looks like the linting step failed on CI.

regnat:
> @regnat Looks like the linting step failed on CI.

Thanks for the notice, should be good now

aherrmann:
@Mergifyio rebase

aherrmann:
I don't know what mergify is choking on. Local rebase worked just fine. I've force-pushed it.

PR #1325, opened 2020-05-07
Properly pass the CC toolchain to cabal

cocreature:
Cabal’s with-gcc option does not apply to the cases where GHC itself
calls the C compiler which actually happens all the time. We noticed
this in daml and tested this patch there where it works successfully.

The flags here are the same as in ghc_cc_program_args. I’m not quite
sure what the best way to deduplicate this is or if this is worth
doing.

### DIFF
--- a/haskell/private/cabal_wrapper.py.tpl
+++ b/haskell/private/cabal_wrapper.py.tpl
@@ -117,7 +117,13 @@ def tmpdir():
     # Build into a sibling path of the final binary output location.
     # This is to ensure that relative `RUNPATH`s are valid in the intermediate
     # output in the `--builddir` as well as in the final output in `--bindir`.
-    distdir = tempfile.mkdtemp(dir=os.path.dirname(pkgroot))
+    # Executables are placed into `<distdir>/build/<package-name>/<binary>`.
+    # Libraries are placed into `<distdir>/build/<library>`. I.e. there is an
+    # extra subdirectory for libraries.
+    if component.startswith("exe:"):
+        distdir = tempfile.mkdtemp(dir=os.path.dirname(os.path.dirname(pkgroot)))
+    else:
+        distdir = tempfile.mkdtemp(dir=os.path.dirname(pkgroot))
     try:
         yield distdir
     finally:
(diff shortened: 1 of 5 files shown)


haskell/private/cabal_wrapper.py.tpl
```diff
@@ -143,6 +143,18 @@ with tmpdir() as distdir:
         "--with-hc-pkg=" + ghc_pkg,
         "--with-ar=" + ar,
         "--with-gcc=" + cc,
+        "--ghc-option=-pgmc=" + cc,
```
aherrmann, line 146: Could this re-use [`ghc_cc_program_args`](https://github.com/cocreature/rules_haskell/blob/a3d4a5759c31380d810a0916e248cb9a6a6fda49/haskell/cc.bzl#L164) to avoid these flags diverging in future?
Using something like `"%{ghc_cc_args}": repr(ghc_cc_program_args(cc))` [here](https://github.com/cocreature/rules_haskell/blob/a3d4a5759c31380d810a0916e248cb9a6a6fda49/haskell/cabal_wrapper.bzl#L31).

cocreature: Ah that looks good, I’ll give it a shot. Wasn’t quite sure how to best pass this down.

cocreature: I’ve tried it but just passing it on doesn’t quite work. I’m missing the `find_exe` logic from the python script in that case so `cc` points to the wrong location. I’m not really sure what to do about this. I could imagine some logic in the python script that applies `find_exe` to flag that don’t start with `-` but that seems worse than inlining the flags here. Any ideas?

aherrmann: Indeed, that's unfortunate. `ghc_cc_program_args` just expects a string representing cc, so we could also use a place holder and fill it in on the python side. Something like this:
`bzl`:
```
    "%{ghc_cc_args}": repr(ghc_cc_program_args("$CC"))
```
`python`:
```
    ] +
    [ arg.replace("$CC", cc) for arg in %{ghc_cc_args} ] +
    [
```

That's getting on the verge, but seems still better than duplicating the flags.

cocreature: Clever, I like it! I’ve made the change.

aherrmann:
The CI failure is because the `cc_wrapper` assumes to be running in the execroot when producing the final artifact and optimizing the `RUNPATH`. But, the Cabal wrapper changes into the directory that contains the `.cabal` file.

aherrmann:
This is finally working, though I've had to make a fair bit of additions, so it would be good to get someone else's review on this.

Issue #1323, opened 2020-05-04
GHC > 8.6. Profiling build error with -optP-traditional


guibou:
**Describe the bug**

Profiling build with GHC 8.8 and 8.10 are not working anymore on my codebase.

`build build -c dbg //...` fails with the following:

```
cc1: error: command line option ‘-traditional’ is valid for the driver but not for C
cc1: error: command line option ‘-Xpreprocessor -DPROFILING’ is valid for the driver but not for C
`cc_wrapper-python' failed in phase `C Compiler'. (Exit code: 1)
```

**To Reproduce**

I rebased my `wip-ghc88` branch on top of `rules_haskell` master, I'm observing different errors, as well as the one depicted here:

```
$ git checkout wip-ghc88
$ nix-shell
$ bazel build -c dbg //tests/target-name/...
INFO: Analyzed 6 targets (0 packages loaded, 0 targets configured).
INFO: Found 6 targets...
ERROR: /home/guillaume/tweag/rules_haskell/tests/target-name/BUILD.bazel:17:1: HaskellBuildLibrary //tests/target-name:lib0-a_Aa failed (Exit 1) ghc_wrapper failed: error executing command bazel-out/host/bin/haskell/ghc_wrapper bazel-out/k8-dbg/bin/tests/target-name/compile_flags_lib0-a_Aa_HaskellBuildLibraryProf ... (remaining 1 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
cc1: error: command line option ‘-traditional’ is valid for the driver but not for C
cc1: error: command line option ‘-Xpreprocessor -DPROFILING’ is valid for the driver but not for C
`cc_wrapper-python' failed in phase `C Compiler'. (Exit code: 1)
INFO: Elapsed time: 0.368s, Critical Path: 0.28s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
```

I'm also observing:

```
$ bazel build -c dbg @haskell_package_repository_dummy//:library-with-cbits
INFO: Analyzed target @haskell_package_repository_dummy//:library-with-cbits (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
[...]

PR #1330, opened 2020-05-11
Bump GHC/Stackage in examples

hanshoglund:
This bumps the `examples` workspace to use https://www.stackage.org/lts-15.12.

Using the Nix shell most targets in the examples workspace build fine, but `bazel build //rts/...` hangs forever. Any pointers on how to debug would be very welcome.

```
[nix-shell] $ bazel build //rts/...
INFO: Analyzed 6 targets (0 packages loaded, 0 targets configured).
INFO: Found 6 targets...
[1 / 9] HaskellBuildLibrary //rts:add-one-hs; 3s linux-sandbox
^C
```

### DIFF
--- a/examples/WORKSPACE
+++ b/examples/WORKSPACE
@@ -13,14 +13,14 @@ rules_haskell_dependencies()
 load("@rules_haskell//haskell:nixpkgs.bzl", "haskell_register_ghc_nixpkgs")
 
 haskell_register_ghc_nixpkgs(
-    attribute_path = "haskell.compiler.ghc865",
+    attribute_path = "haskell.compiler.ghc883",
     repository = "@rules_haskell//nixpkgs:default.nix",
-    version = "8.6.5",
+    version = "8.8.3",
 )
 
 load("@rules_haskell//haskell:toolchain.bzl", "rules_haskell_toolchains")
 
-rules_haskell_toolchains(version = "8.6.5")
+rules_haskell_toolchains(version = "8.8.3")
 
 load(
     "@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl",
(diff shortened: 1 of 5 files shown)


examples/WORKSPACE
```diff
@@ -106,6 +106,6 @@ stack_snapshot(
         "text",
         "text-show",
     ],
-    snapshot = "lts-14.0",
-    vendored_packages = {"split": "@split//:split"},
+    snapshot = "lts-15.2",
+    # vendored_packages = {"split": "@split//:split"},
```
hanshoglund, line 110: Any pointers on how to regenerate this patch would be most welcome :).

examples/WORKSPACE
```diff
@@ -64,25 +64,25 @@ cc_library(name = "z", srcs = glob(["*.c"]), hdrs = glob(["*.h"]))
...
+# haskell_cabal_library(
+#     name = "split",
+#     version = packages["split"].version,
+#     srcs = glob(["**"]),
+#     deps = packages["split"].deps,
+#     visibility = ["//visibility:public"],
+# )
+#     """,
+#     patch_args = ["-p1"],
+#     patches = ["@rules_haskell_examples//:split.patch"],
```
hanshoglund, line 81: Any pointers on how to regenerate this patch would be most welcome :).

aherrmann:
This includes an update of GHC to 8.8 and is hitting the same issue as reported here https://github.com/tweag/rules_haskell/issues/1323.

examples/WORKSPACE
aherrmann: The following steps should lead to a working patch:
* Find out which version of `split` is contained in the new snapshot.
* Checkout the `split` source repository at the corresponding git revision.
* Increase the version bound on `base` in that version's Cabal file.
* Generate a patch.

aherrmann:
#1563 updated the examples to a later GHC and Stackage snapshot version.

PR #1338, opened 2020-05-22
Ghci rebindable syntax

aherrmann:
`haskell_repl` automatically executes a ghci script for module loading. This script failed if `-XRebindableSyntax` was enabled because it uses do notation. This change fixes this by binding `(>>=)` etc. to the corresponding `Prelude` functions in the ghci script. This will not affect the user's ghci session since the script calls `:reload` at the end.

This also includes a regression test that invokes the repl on a module that uses rebindable syntax.

### DIFF
--- a/haskell/assets/ghci_script
+++ b/haskell/assets/ghci_script
@@ -3,6 +3,11 @@
 import qualified GHC.IO.Handle as Handle
 import qualified System.IO as IO
 import qualified System.Directory as Dir
+:set -Wno-name-shadowing
+(>>=) = (Prelude.>>=)
+(>>) = (Prelude.>>)
+return = Prelude.return
+fromString = Prelude.id
 rules_haskell_stdout_dupe <- Handle.hDuplicate IO.stdout
 :{
 (rules_haskell_stdout_copy_file, rules_haskell_stdout_copy_h) <- do
--- a/haskell/assets/ghci_script
+++ b/haskell/assets/ghci_script
@@ -3,7 +3,7 @@
 import qualified GHC.IO.Handle as Handle
 import qualified System.IO as IO
 import qualified System.Directory as Dir
-:set -Wno-name-shadowing
+:seti -Wno-name-shadowing
 (>>=) = (Prelude.>>=)
 (>>) = (Prelude.>>)
 return = Prelude.return
--- a/tests/RunTests.hs
+++ b/tests/RunTests.hs
@@ -42,6 +42,10 @@ main = hspec $ do
 
       assertSuccess (bazel ["run", "//tests/binary-indirect-cbits:binary-indirect-cbits@repl", "--", "-ignore-dot-ghci", "-e", ":main"])
 
+    it "with rebindable syntax" $ do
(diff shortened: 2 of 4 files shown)


haskell/assets/ghci_script
```diff
@@ -3,6 +3,11 @@
 import qualified GHC.IO.Handle as Handle
 import qualified System.IO as IO
 import qualified System.Directory as Dir
+:set -Wno-name-shadowing
```
regnat, line 6: Isn't this `:set` kept even after reload?

aherrmann: That is correct. I've changed it to use `:seti` instead of `:set`. That way any shadowing in source files will still cause a warning while the warning will only be disabled in the interactive session itself.

I haven't found a way to avoid affecting that warning altogether. `ghci_script` requires bind syntax, so we have to bind `(>>=)` and friends to the `Prelude`'s definitions, this requires `-Wno-name-shadowing` to work if the module being loaded defines different bindings for `(>>=)` etc. I haven't found a way to only temporarily `:seti`.

regnat: A solution would be to not use the do-notation in the script, but explicitely call `Prelude.>>=` everywhere. But that's quite painful

aherrmann: I've tried that but sadly it's not enough. We need to bind the outcome of some IO operations using bind syntax in the REPL directly, e.g. 
```
rules_haskell_stdout_dupe <- Handle.hDuplicate IO.stdout
```
That implicitly requires `do` notation.

ghorn:
I don't like disabling warnings in GHCI because one of the main GHCI workflows is to get everything working before you actually compile. Instead could you get rid of `do` in the ghci script and instead use `foo Prelude.>>= $ \x ->`

aherrmann:
> I don't like disabling warnings in GHCI because one of the main GHCI workflows is to get everything working before you actually compile. Instead could you get rid of `do` in the ghci script and instead use `foo Prelude.>>= $ \x ->`

@ghorn I should clarify this better. This change only applies `-Wno-name-shadowing` via `:seti`. `:seti` only affects code typed directly into the REPL and GHCi scripts as we use for setup of `haskell_repl`. The warning is not disabled for Haskell modules loaded into GHCi. I.e. if any of the Haskell sources loaded by the REPL contain name shadowing, then this will still raise a warning.

In short, the use-case of making sure everything works before you actually compile is not affected by this change.

---

> Instead could you get rid of `do` in the ghci script and instead use `foo Prelude.>>= $ \x ->`

Unfortunately not, the GHCi setup script needs to bind the result of an `IO` action to a name. In particular it does
```
rules_haskell_stdout_dupe <- Handle.hDuplicate IO.stdout
```
This uses do notation under the hood. You can emulate the issue in a GHCi session as follows:
```
$ ghci -XRebindableSyntax
> import qualified System.IO as IO
IO> x <- IO.getLine
<interactive>:2:1: error: Not in scope: ‘>>=’
```
I'm not aware of a way to bind the result of an `IO` action to a name at the REPL prompt without using `<-` syntax.

PR #1343, opened 2020-05-26
Coverage: prefix .hpc directory with rule name

guibou:
Right now coverage files are generated in a `.hpc/` directory. If two rules in the same bazel package have the same input file, they will generate the same `.mix` file in the `.hpc` directory.

This leads to build failure. Actually, bazel detect this during its analysis phase and fails.

I've added tests in ` tests/two-same-file` to show this behavior.

The fix consist of moving the `.hpc` file to `RULE_NAME_.hpc`. The name is a bit convoluted, it may have been `.hpc_RULE_NAME`, but it simplifies the logic.

### DIFF
--- a/tests/two-same-file/BUILD.bazel
+++ b/tests/two-same-file/BUILD.bazel
@@ -1,0 +1,34 @@
+load(
+    "@rules_haskell//haskell:defs.bzl",
+    "haskell_library",
+    "haskell_test",
+)
+
+package(
+    default_testonly = 1,
+    default_visibility = ["//visibility:public"],
+)
+
+haskell_test(
+    name = "one",
+    srcs = ["Main.hs"],
+    coverage_report_format = "html",
+    expected_covered_expressions_percentage = 55,
+    expected_uncovered_expression_count = 4,
+    tags = ["coverage-compatible"],
+    deps = [
+        "//tests/hackage:base",
+    ],
+)
+
+haskell_test(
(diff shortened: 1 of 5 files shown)


tests/two-same-file/BUILD.bazel
```diff
@@ -0,0 +1,34 @@
...
+    "@rules_haskell//haskell:defs.bzl",
+    "haskell_library",
+    "haskell_test",
+)
+
+package(
+    default_testonly = 1,
+    default_visibility = ["//visibility:public"],
+)
+
```
aherrmann, line 11:
```suggestion
```
This seems redundant. Both are `haskell_test` so they should already be `testonly = 1`, and nothing seems to depend on these targets, so they don't need to be `//visibillity:public`.

tests/two-same-file/BUILD.bazel
```diff
@@ -0,0 +1,34 @@
+load(
+    "@rules_haskell//haskell:defs.bzl",
+    "haskell_library",
```
aherrmann, line 3:
```suggestion
```
Unused import.

guibou:
I've done the fix asked. But do not merge yet, I'm having a doubt about a possible regression.

zimbatm:
> I've done the fix asked. But do not merge yet, I'm having a doubt about a possible regression.

@guibou told me that his reservations are gone :)

PR #1352, opened 2020-06-04
Make haskell_repl work with remote_download_toplevel

### LINKED ISSUE #1334: haskell_repl doesn't work with `--remote_download_toplevel`
**Describe the bug**

Running a `haskell_repl` target with `--remote_download_toplevel` causes it to crash because it can't find the `.so` of some dependencies

**To Reproduce**

```sh
# Build it once to populate the remote cache. Everything should work so far
bazel run //tests/multi_repl:c_only_repl --disk_cache=~/.cache/bazel_disk_cache
# Clean the local cache
bazel clean
# Re-run it with `--remote_download_toplevel`
bazel run //tests/multi_repl:c_only_repl --disk_cache=~/.cache/bazel_disk_cache --remote_download_toplevel
```

The second `bazel run` should fail with

```
<command line>: can't load .so/.DLL for: libHStestsZSmultiZUreplZSaZSa.so (libHStestsZSmultiZUreplZSaZSa.so: cannot open shared object file: No such file or directory)
```

**Expected behavior**

`--remote_download_toplevel` shouldn't prevent the repl targets from working

**Environment**
 - OS name + version:
 - Bazel version:
 - Version of the rules:

**Additional context**

I guess that's somehow due to these dependencies not being properly declared, although a quick glance at the code doesn't uncover anything obvious

regnat:
- Add the needed .so and .dyn_hi files to the inputs of the repl
- Add a regression test for #1334

Fix #1334

### DIFF
--- a/haskell/private/cc_libraries.bzl
+++ b/haskell/private/cc_libraries.bzl
@@ -49,7 +49,7 @@ def get_cc_libraries(cc_libraries_info, libraries_to_link):
         if not cc_libraries_info.libraries[cc_library_key(lib_to_link)].is_haskell
     ]
 
-def get_ghci_library_files(hs, cc_libraries_info, libraries_to_link):
+def get_ghci_library_files(hs, cc_libraries_info, libraries_to_link, follow_symlinks = False):
     """Get libraries appropriate for loading with GHCi.
 
     See get_library_files for further information.
@@ -60,10 +60,11 @@ def get_ghci_library_files(hs, cc_libraries_info, libraries_to_link):
         libraries_to_link,
         dynamic = not hs.toolchain.is_static,
         pic = True,
+        follow_symlinks = follow_symlinks,
     )
     return static_libs + dynamic_libs
 
-def get_library_files(hs, cc_libraries_info, libraries_to_link, dynamic = False, pic = None):
+def get_library_files(hs, cc_libraries_info, libraries_to_link, dynamic = False, pic = None, follow_symlinks = False):
     """Get libraries appropriate for linking with GHC.
 
     Takes a list of LibraryToLink and returns a list of the appropriate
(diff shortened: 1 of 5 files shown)


haskell/private/cc_libraries.bzl
```diff
@@ -79,6 +80,7 @@ def get_library_files(hs, cc_libraries_info, libraries_to_link, dynamic = False,
       libraries_to_link: list of LibraryToLink.
       dynamic: Whether dynamic libraries are preferred.
       pic: Whether position independent code is required.
+      follow_symlinks: Whether to return the actual path of the libraries or their symbolic link in the _solib_k8 directory
```
aherrmann, line 83: This sounds a bit misleading. IIUC this will return both the actual library file and the symbolic link, not either or.

haskell/private/cc_libraries.bzl
```diff
@@ -114,6 +116,14 @@ def get_library_files(hs, cc_libraries_info, libraries_to_link, dynamic = False,
             static_libs.append(static_lib)
         elif dynamic_lib:
             dynamic_libs.append(dynamic_lib)
+            extra_lib_files = \
+                [
+                    lib
+                    for lib in [lib_to_link.resolved_symlink_dynamic_library, lib_to_link.resolved_symlink_interface_library]
+                    if lib != None
+                ]
```
aherrmann, line 124:
```suggestion
            extra_lib_files = [
                lib
                for lib in [lib_to_link.resolved_symlink_dynamic_library, lib_to_link.resolved_symlink_interface_library]
                if lib != None
            ]
```
Nit: To be consistent with the formatting of similar expressions throughout rules_haskell.

haskell/private/cc_libraries.bzl
regnat: Indeed. I've renamed it to `include_real_paths`. Not really pretty, but should be less misleading

aherrmann: Thanks, I think that clarifies it well.

PR #1298, opened 2020-04-02
[WIP] Start fix of --experimental_use_sandboxfs

smelc:
Fixes #1265 

This draft PR starts implementing the solution drafted by @aherrmann in https://github.com/tweag/rules_haskell/issues/1265#issuecomment-607196384 to make `--experimental_use_sandboxfs` work when using bindists (no `NixOS`, no `nix-shell` context). Let me record that the intended setup to test this PR is [described here](https://github.com/tweag/rules_haskell/issues/1173#issuecomment-606723360).

It fixes the issue that executables - within `ghc`'s install `bin` directory - need content in the sibling `lib` directory. Because this content is not tracked precisely right now, it is missing when executing within [sandboxfs](https://github.com/bazelbuild/sandboxfs) with `--experimental_use_sandboxfs` (see [bazel+sandboxfs](https://blog.bazel.build/2018/04/13/preliminary-sandboxfs-support.html) usage).

The first commit fixes failures of the form `../lib/path` being missing. I made a checkpoint here because with this commit, `bazel test //tests/...` now fails in a different manner:

```
  File "bazel-out/host/bin/haskell/cc_wrapper-python", line 111, in FindModuleSpace
    raise AssertionError('Cannot find .runfiles directory for %s' % sys.argv[0])
AssertionError: Cannot find .runfiles directory for bazel-out/host/bin/haskell/cc_wrapper-python
compiling bazel-out/k8-fastbuild/bin/tests/library-with-cbits/_hsc/library-with-cbits-static/AddOne_hsc_make.c failed (exit code 1)
command was: bazel-out/host/bin/haskell/cc_wrapper-python -c bazel-out/k8-fastbuild/bin/tests/library-with-cbits/_hsc/library-with-cbits-static/AddOne_hsc_make.c -o bazel-out/k8-fastbuild/bin/tests/library-with-cbits/_hsc/library-with-cbits-static/AddOne_hsc_make.o -fno-stack-protector -iquote . -iquote bazel-out/k8-fastbuild/bin -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer -fno-canonical-system-headers -Wno-builtin-macro-redefined -D__DATE__="redacted" -D__TIMESTAMP__="redacted" -D__TIME__="redacted" -Iexternal/rules_haskell_ghc_linux_amd64/bin/../lib/include/
```

[...]

PR #1363, opened 2020-06-11
sandboxfs compatibility

regnat:
This is yet-another attempt at getting rules_haskell to support

This partially overlaps with #1326 and #1298, but not fully as it hacks to get the nixpkgs toolchain working with sandboxfs (by bind-mounting the store in the sandbox), so that we can track and find the issues in the build rules rather than being stuck at getting the toolchain to work properly.

The blockers atm to get everything to build are 
- `ghc-paths` tries to generate files in the working directory (in the `Setup.hs`) which isn't allowed
- `c2hs` fails because happy can't find a template file anymore


regnat:
Not sure why I get `Failed to get sandboxfs version from sandboxfs` from buildkite :thinking: Sandboxfs is indeed there and `--version` works fine. I'll investigate a bit more

Issue #1310, opened 2020-04-22
Enable pinning and better caching with stack_snapshot


aherrmann:
**Is your feature request related to a problem? Please describe.**

`stack_snapshot` calls out to `stack` at every fetch multiple times:
* `stack update` in order to update the Hackage database (factored out into `stack_update`).
* `stack unpack` to download the package sources.
* `stack ls dependencies` and `stack dot` to determine the dependency graph.

This causes some issues:
* Concurrency issues with `stack update` which does not support multiple simultaneous invocations. We can work around this within one Bazel repository, see https://github.com/tweag/rules_haskell/pull/1199. But, this also causes issues with multiple simultaneous jobs on one CI runner which we cannot address well within `rules_haskell`, see https://github.com/tweag/rules_haskell/issues/1167.
* What `stack unpack` fetches is opaque to Bazel and is not cached well in Bazel's repository cache, see https://github.com/tweag/rules_haskell/issues/1112 and relatedly https://github.com/tweag/rules_haskell/issues/1168.

**Describe the solution you'd like**

I would like `stack_snapshot` to be more transparent for Bazel. The new [`stack ls dependencies json`](https://github.com/commercialhaskell/stack/pull/4424) feature provides most of the required information: The package name, version, dependencies, and the package location (Hackage, URL, repository). We could then use [`download_and_extract`](https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#download_and_extract) or [`git_repo`](https://github.com/bazelbuild/bazel/blob/0a0a13bd4eaeefc3cdf17e8e34dcca4e1453a491/tools/build_defs/repo/git_worker.bzl#L32) to let Bazel fetch the sources in a way that Bazel can cache.

Unfortunately, `stack ls dependencies json` does not provide the `sha256`. However, Bazel gives access to the `sha256` after the download, so we could generate a lock file containing the `sha256` (or the `commit` and `shallow_since` in case of `git_repo`).

Similar to [`rules_jvm_external`](https://github.com/bazelbuild/rules_jvm_external#pinning-artifacts-and-integration-with-bazels-downloader) I would like `stack_snapshot` to provide a pinning mechanism. Given a lock file `stack_snapshot` would not have to call out to `stack` at all, avoiding any of the issues around `stack update`.

**Describe alternatives you've considered**

[...]

PR #1364, opened 2020-06-12
Use stack ls dependencies json

aherrmann:
This is a step towards https://github.com/tweag/rules_haskell/issues/1310

Version 2.3.1 of `stack` introduced [`stack ls dependencies json`](https://github.com/commercialhaskell/stack/pull/4424). With this PR we make use of that feature to determine the package dependency graph and package versions, instead of the previous approach of parsing the `dot` formatted dependency graph.

Following the example of `rules_jvm_external` this vendors the Starlark JSON parser from https://github.com/erickj/bazel_json.

This PR also separates the step that determines the dependency graph from the step that fetches the package sources. This is in preparation for #1310 where we'll want to store the dependency graph and package locations in a lock file and perform the fetching in a separate repository rule that can depend on a lock file alone.

I have verified that the resulting external workspaces are equivalent
up to expected differences by comparing the results of `bazel fetch
@stackage//:all` in the `example` directory and `@stackage//:all` and `@ghcide//:all` in the project root. Note, dictionaries are iterated in insertion order. To compare the results before and after this commit one has to sort the contents of `package.bzl` and `BUILD.bazel` by package name first.

~The old mechanism did not respect Cabal revisions. This new implementation fetches the correct Cabal revisions as defined in the Stackage snapshot.~
(This difference was due to not passing `--resolver` to `stack unpack` as is required.)


third_party/bazel_json/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright 2019 Erick Johnson <ejohnson82@gmail.com>
```
hanshoglund, line 1: I've no objections to using this library, nor to committing it in the repo, but:

- `vendor` might be a better name for the top-level directory
- For better Git history, consider doing a subtree merge of the original repo rather than a copy-paste

Neither point is blocker for the PR.

haskell/cabal.bzl
```diff
@@ -916,67 +927,93 @@ def _compute_dependency_graph(repository_ctx, snapshot, core_packages, versioned
...
+        stack + ["ls", "dependencies", "json", "--global-hints", "--external"],
     )
-    transitive_unpacked_sdists = []
-    indirect_unpacked_sdists = []
-    remaining_components = dicts.add(_default_components, user_components)
-    for package in exec_result.stdout.splitlines():
-        name = _chop_version(package)
-        version = _version(package)
+    package_specs = json_parse(exec_result.stdout)
+
```
hanshoglund, line 985: Maybe future proof? Something like:

```suggestion
if type(package_specs) != "list":
    fail(msg="Unexpected output format for `stack ls dependencies json.`")
```

haskell/cabal.bzl
```diff
@@ -916,67 +927,93 @@ def _compute_dependency_graph(repository_ctx, snapshot, core_packages, versioned
...
+    # Collect package metadata
+    remaining_components = dict(**user_components)
+    for package_spec in package_specs:
+        name = package_spec["name"]
```
hanshoglund, line 989: Similarly we could throw a high-level message if the name/version fields are missing.

haskell/cabal.bzl
aherrmann: That's a good idea, I've added a validation to both `package_specs` and the individual `package_spec`.

haskell/cabal.bzl
aherrmann: Likewise addressed by adding a validation step.

aherrmann:
@Mergifyio refresh

aherrmann:
Mergify seems to be stuck showing the following conflicting queues on this PR and #1367.

> The following pull requests are queued:
>    #1367, #1364 (priority: medium)

> The following pull requests are queued:
>    #1364, #1367 (priority: medium)

Merging manually.

PR #1366, opened 2020-06-16
Allow including extra files with Haddocks

### LINKED ISSUE #1056: `haskell_doc` should allow including extra files

The haddock markup allows inserting urls to raw documents, as well as images, but this doesn't work with `haskell_doc` because these files aren't bundled into the `doc-` bazel output.

Add a `data` argument to `haskell_doc` which will list some files to be included in the documentation output.

We could instead add a `haddock_data` argument to `haskell_library` so that the files are directly added to the output of the `haskell_doc` aspect

hanshoglund:
Allow including extra files with Haddocks

Fixes #1056


haskell/haddock.bzl
```diff
@@ -257,16 +257,26 @@ def _haskell_doc_rule_impl(ctx):
...
 
         ctx.actions.run_shell(
-            inputs = [html_dir],
+            inputs = [html_dir] + ctx.files.extra_files,
             outputs = [output_dir],
             command = """
       mkdir -p "{doc_dir}"
       # Copy Haddocks of a dependency.
       cp -R -L "{html_dir}/." "{target_dir}"
+      {extra}
```
regnat, line 266: Putting this here means that the extra files will end-up in the documentation of every package included in this `haskell_doc` target (which can mean a lot if you have enough cabal dependencies).

I guess a way to do this in a more granular way would be to add a `haddock_extra_files` argument to `haskell_{library,binary}` and propagate these extra files in the `haskell_doc_aspect`

guibou:
What is the status of this PR? It solves an important problem. It may be imperfect, but that's the only proposed solution for an issue opened in 2019.

hanshoglund:
Hello @guibou. I will close the PR as I'm not actively working on it.

If anyone wants to use the code, feel free to copy it!

PR #1367, opened 2020-06-16
Reverse order of `-fplugin` parameters

### LINKED ISSUE #1255: Reverse order of `-fplugin` parameters
`plugins = [":foo", ":bar"]` results in `:bar` getting installed before `:foo`. rules_haskell is ostensibly doing the right thing (it creates the flags in the same order they are in the list). However, [GHC currently installs plugins in the reverse order that they're provided](https://gitlab.haskell.org/ghc/ghc/issues/17884), which is non-intuitive. The order is significant as plugins can affect each other during installation.

It would be helpful if rules_haskell could ensure that the order of the list is the same as the order of installation. On GHC up to at least 8.8.2, that would mean passing them in reverse order, then when the GHC issue is hopefully fixed, pass the flagis in the same order as the list for future GHC versions.

Currently I just add a comment before each `plugins` list to explain that the behavior is the opposite of what's implied by the list.

hanshoglund:
Fixes #1255

### DIFF
--- a/tests/plugin-install-order/BUILD.bazel
+++ b/tests/plugin-install-order/BUILD.bazel
@@ -46,6 +46,7 @@ haskell_library(
         ":process",
     ],
 )
+
 haskell_library(
     name = "plugin-lib2",
     srcs = ["Plugin2.hs"],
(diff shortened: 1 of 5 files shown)


tests/plugin-install-order/BUILD.bazel
```diff
@@ -0,0 +1,90 @@
+"""
+Tests that GHC plugins are applied in the order declared in the BUILD file.
+
+Note [Plugin order]
+~~~~~~~~~~~~~~~~~~~
+
+We reverse the (counter-intuitive) order used by GHC. This test is
```
aherrmann, line 7: IIUC with this change we use the same order as GHC which is the counter-intuitive choice of applying the plugins in reverse order than defined in the attribute.
```suggestion
We use the same (counter-intuitive) order of plugins as used by GHC. This test is
```

hanshoglund: No, the GHC order is used in master. This patch reverses the GHC order and adds a test which will break when the order is changed upstream (so that we can revert this).

hanshoglund: The TODO comments they were just redundant `ghc_plugin` options. I've removed them.

aherrmann: > No, the GHC order is used in master. This patch reverses the GHC order and adds a test which will break when the order is changed upstream (so that we can revert this).

Oh, I see, I misread the issue before. Thanks for clarifying!



> The TODO comments they were just redundant `ghc_plugin` options. I've removed them.

Thanks!

Issue #1368, opened 2020-06-16
Haskell modules generated from Protobuf imports are hidden by default


hanshoglund:
**Describe the bug**
Haskell modules generated from Protobuf imports (such as `google.protobuf.FileDescriptorSet`) are hidden by default, requiring ugly "reexports" in the Protobuf code.

**To Reproduce**
WIP

**Expected behavior**
WIP

**Environment**
 - OS name + version:
 - Bazel version:
 - Version of the rules:

**Additional context**

Issue #1111, opened 2019-10-01
Protobuf generated module conflicts with proto-lens


jpgneves:
**Describe the bug**
Some protobuf files import `google/protobuf/descriptor.proto`, such that for a `proto_library` target for such a file to build, I have to include `@com_google_protobuf//:descriptor_proto` dependency.

However, `proto-lens` already includes a module with the same name generated from similar definitions, which results in an ambiguous module name error:

```
   Ambiguous module name `Proto.Google.Protobuf.Descriptor':
     it was found in multiple packages:
       descriptor-proto proto-lens-0.4.0.1
   |
25 | import qualified Proto.Google.Protobuf.Descriptor
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Example of a file I'm attempting to generate Haskell from: https://github.com/googleforgames/open-match/blob/master/third_party/google/api/annotations.proto#L20

I've filed an issue with `proto-lens` (https://github.com/google/proto-lens/issues/344), and they suggest `rules_haskell` might be a better place to fix the clash, as a suggested option would be that maybe `rules_haskell` could provide the required `.proto` files, as `proto-lens` does itself: https://github.com/google/proto-lens/blob/5e5bec02a4193d0a0d797da3551fc0893adb836e/proto-lens/package.yaml#L27

**To Reproduce**
Create a `.proto` file which has a dependency on `google/protobuf/descriptor.proto`:

```
package foo;

import "google/protobuf/descriptor.proto";
```

Create a `BUILD.bazel` file using `rules_haskell` with appropriate targets:

```
load("@rules_haskell//haskell:protobuf.bzl",
     "haskell_proto_library",
     "haskell_proto_toolchain",
)

haskell_proto_toolchain(
    name = "protobuf-toolchain",
    protoc = "@com_google_protobuf//:protoc",
    plugin = "@proto-lens-protoc//:bin/proto-lens-protoc",
    deps = [
        "@stackage//:base",
        "@stackage//:bytestring",
        "@stackage//:containers",
        "@stackage//:data-default-class",
        "@stackage//:deepseq",
        "@stackage//:lens-family",
        "@stackage//:lens-labels",
        "@stackage//:proto-lens",
        "@stackage//:text",
    ],
)

proto_library(
     name = "foo",
     srcs = ["foo.proto"],
     deps = [ "@com_google_protobuf//:descriptor_proto" ],
)

[...]

PR #1369, opened 2020-06-16
Add test for Protobuf name clash issue

hanshoglund:
Add a test showing how the the name clash issue reported in #1111 does not appear with the recent proto-lens (e.g. the version of Stackage pinned in master/this commit).

Also introduces a workaround (but not a proper fix) for #1368.

### DIFF
--- a/tests/haskell_proto_simple/BUILD.bazel
+++ b/tests/haskell_proto_simple/BUILD.bazel
@@ -2,18 +2,17 @@
 Minimal reproduction of https://github.com/tweag/rules_haskell/issues/1111
 """
 
-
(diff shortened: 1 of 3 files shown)


nlewo:
I tried to apply your patch on the release v0.10 (which was used in issue #1111) to see if it fails. Unfortunately, it fails for an unrelated reason... :(

hanshoglund:
@nlewo I asked the reporter for feedback on the original ticket, but haven't heard anything yet. 

Do you have any objections to merging this?

nlewo:
> Do you have any objections to merging this test?
No, but I really don't have a strong opinion on it :(

PR #1371, opened 2020-06-19
Allow Bazel 2.2.0 in start script

hanshoglund:


hanshoglund:
Hold that thought, seems I need to bump `versions.bzl` as well.

Profpatsch:
Hm, what’s the status?

hanshoglund:
Sorry, this fell off my radar. Will close as there is no real progress. Let's create a new ticket for the next version bump.

aherrmann:
https://github.com/tweag/rules_haskell/pull/1395 would cover 2.2.0 as well.

Issue #1365, opened 2020-06-16
__STDC_VERSION__ regression in master


liam-ly:
**Describe the bug**
In v0.12, this all works, but in master, compiling zlib using stack_snapshot with centos7 gives
```
ERROR: /root/.cache/bazel/_bazel_root/f8087e59fd95af1ae29e8fcb7ff1a3dc/external/stackage/BUILD.bazel:12:1: HaskellCabalLibrary @stackage//:zlib failed (Exit 1) cabal_wrapper failed: error executing command bazel-out/host/bin/external/rules_haskell/haskell/cabal_wrapper lib:zlib zlib-0.6.2.1 false external/stackage/zlib-0.6.2.1/Setup.hs external/stackage/zlib-0.6.2.1 ... (remaining 8 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox

In file included from ../../../../../../../../external/rules_haskell_ghc_linux_amd64/lib/include/Rts.h:29:0: error:
    0,
                     from /root/.cache/bazel/_bazel_root/f8087e59fd95af1ae29e8fcb7ff1a3dc/sandbox/processwrapper-sandbox/3/execroot/TestProject/bazel-out/k8-fastbuild/bin/external/stackage/zlib-0.6.2.1/tmpykxn9foy/tmp/ghc17613_0/ghc_7.c:2:

../../../../../../../../external/rules_haskell_ghc_linux_amd64/lib/include/Stg.h:29:3: error:
     error: #error __STDC_VERSION__ does not advertise C99, C++11 or later
     # error __STDC_VERSION__ does not advertise C99, C++11 or later
       ^
   |
29 | # error __STDC_VERSION__ does not advertise C99, C++11 or later
   |   ^

In file included from ../../../../../../../../external/rules_haskell_ghc_linux_amd64/lib/include/Rts.h:183:0: error:
    0,
                     from /root/.cache/bazel/_bazel_root/f8087e59fd95af1ae29e8fcb7ff1a3dc/sandbox/processwrapper-sandbox/3/execroot/TestProject/bazel-out/k8-fastbuild/bin/external/stackage/zlib-0.6.2.1/tmpykxn9foy/tmp/ghc17613_0/ghc_7.c:2:
../../../../../../../../external/rules_haskell_ghc_linux_amd64/lib/include/rts/storage/ClosureMacros.h: In function ‘overwritingClosure_’:

[...]

PR #1374, opened 2020-06-22
Pass CC/GHC toolchain flags to Cabal builds

aherrmann:
Closes https://github.com/tweag/rules_haskell/issues/1365

* `compiler_flags` defined at `haskell_register_ghc_*` will also be forwarded to Cabal builds using `haskell_cabal_binary|library`. This allows users to pass compiler flags that are required for their platform, e.g. #1365, and is also in line with what Hazel used to do.
* This adds a regression test for Cabal builds that ensures that GHC toolchain flags are forwarded to Cabal builds.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -211,7 +211,8 @@ def _prepare_cabal_inputs(
         for arg in ["-package-db", "./" + _dirname(package_db)]
     ], join_with = " ", format_each = "--ghc-arg=%s", omit_if_empty = False)
     args.add("--flags=" + " ".join(flags))
-    args.add_all(compiler_flags, format_each = "--ghc-option=%s")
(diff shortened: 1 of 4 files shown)


aherrmann:
This fails on CI because we explicitly set the default `-XNoOverloadedStrings` in the GHC toolchain in `WORKSPACE` for testing purposes. For regular Haskell targets this doesn't cause any issues, however for Cabal targets `--ghc-option(s)` are passed after `extensions` and other GHC flags specified in the cabal file, i.e. this way the toolchain options override package specific options defined in the cabal file. Cabal itself has a notion of ["default" GHC arguments](http://lambda.inf.elte.hu/haskell/doc/libraries/Cabal-2.2.0.1/Distribution-Simple-Program-GHC.html#v:ghcOptExtraDefault) that should precede the flags specified in the Cabal file and via `--ghc-option`. However, this doesn't seem to be exposed to the user through a command-line option.

PR #1376, opened 2020-06-26
Enable pinning with stack_snapshot

### LINKED ISSUE #1112: `stack_snapshot` leads to a big download that can't be cache by bazel
**Describe the bug**

When using `stack_snapshot`, the repository rule downoalds all the (transitive) haskell dependencies using `stack unpack`.
This download isn't cached by bazel (because it's in a repository rule). It is however cached by stack, but in a different (and local-only) cache.

**To Reproduce**

- build the examples repository (with a `disk_cache` and `repository_cache` defined)
- remove `~/.stack/pantry`
- rebuild the examples repository

**Expected behavior**

The second build shouldn't trigger a re-download of all the haskell deps

**Environment**

OS name + version: nixos unstable
Bazel version: 0.28.0 (from the example's shell.nix)
Version of the rules: latest master (79e2c53)

### LINKED ISSUE #1167: Concurrency issues with `stack_snapshot()`
**Describe the bug**
We are seeing build failures on CI, with an error message that looks as follows:

```
Command failed: /home/daniel/.cache/bazel/_bazel_daniel/526b0fb4c773181d605f6e65995536aa/external/haskell_stack/bin/stack --resolver /home/daniel/repos/habito-2/bazel/repo/stackage-snapshot.yaml ls dependencies --global-hints --separator=-
/home/daniel/.stack/pantry/global-hints-cache.yaml.tmp: renameFile:renamePath:rename: does not exist (No such file or directory)
```

It turns out this happens when two CI jobs are running simultaneously on the same box and it is likely due to `stack ls dependencies` not protecting certain operations with a lock.

Even though there is probably a bug in `stack` here, the fact that all bazel workspaces depend on artifacts under `~/.stack` make isolation and even hermeticity arguably harder to enforce. Would it make sense to run `stack` with `--stack-root` pointing to a directory controlled by the `stack_snapshot()` rule?

**To Reproduce**
I managed to reproduce the error locally by doing the following:
  1. `rm -Rf ~/.stack/`
  1. On two different copies of the bazel workspace, run simultaneously `bazel clean --expunge; bazel build //...`

I imagine that a relatively large number of stackage dependencies may be needed for the race condition to manifest.

**Expected behavior**
Concurrent builds don't interfere with each other.

**Environment**
 - OS name + version: Linux, several variants
 - Bazel version: `bazel-1.1.0` and `stack-2.1.3.1`, both provisioned from `nixpkgs`
 - Version of the rules: Latest `master`  (`c169112800280cdc87964e9cad1752f643b603dd`)

**Additional context**

At the moment we don't have a workaround for this, ideas are welcome!

### LINKED ISSUE #1168: Cache the stackage snapshot repository

We were trying to move from a setup where we use a `ghcWithPackages` with all our external dependencies plus invocations to `haskell_toolchain_library` to one based on `stackage_snapshot`. The main blocker we have at the moment is that fetching this repository takes either 1 or 5 minutes (depending on whether `~/.stack` already exists, I think), which is a bit too long. IIUC, the source of the slowness is `stack unpack` having to handle 450+ packages, and possibly doing it one at a time.

On the other hand, If I make a tarball of `external/stackage` it takes 45MB, which I can download in a few seconds.

I'd like `stack_snapshot()` to try to download a content-indexed tarball of the repository from a WebDAV cache, given as an optional attribute (usually the same url used for bazel's remote cache). In case of a cache-miss, build the repository just like it is done now, but at the end make a tarball with its contents and try  to upload it to the cache.

For creating the hash to index the cache, most other inputs to the `stack_snapshot()` rule could be considered verbatim. The exceptions would be:  `local_snapshot`, whose content would have to be used; and `stack`, where the version is probably a better value than its location.

We are not aware of any good workarounds at the moment (oth, if we are the only ones experiencing this, maybe we are doing something wrong?)

**Additional context**

The reason this is an issue for us is because, IIUC, it means that every time we update any dependency, every developer will have to wait a few minutes for their next build to happen. With our current setup, all haskell targets get invalidated (which is also annoying, and what we were trying to improve), but at least one only pays for what you are building locally and we can let CI pre-populate bazel's remote cache. Additionally, the 5 minute delay is particularly problematic on CI, where it messes with timeouts we have for our builds there.

### LINKED ISSUE #1310: Enable pinning and better caching with stack_snapshot

aherrmann:
depends on #1379

Closes #1310

This implements pinning for `stack_snapshot` similar to `rules_jvm_external`'s `maven_install`. By default `stack_snapshot` works as before, calling `stack ls dependencies json` to determine package versions and the dependency graph on the fly and calling `stack unpack` to fetch the package sources. However, if the optional attribute `stack_snapshot_json` is specified then it is read as a lock file that stores all relevant information to be able to fetch and build the snapshot packages without invoking stack. The lock file can be generated or updated with the script `@<workspace>-unpinned//:pin`.

This makes re-fetching `stack_snapshot` repositories faster as Hackage and archive dependencies can be cached in Bazel's repository cache, and `stack` does not need to update it's database to determine package versions and dependencies. 
Closes #1112
Closes #1168

Additionally, this avoids issues due to concurrent invocations of `stack` which can fail due to a race condition on the Hackage security lock.
Closes #1167

The generated `snapshot.json` file is meant to be checked into the user's repository. It should not be updated manually, only using `bazel run @<workspace>-unpinned//:pin`. Similar to `rules_jvm_external` we validate the contents of the lock file and store and check a checksum to detect a corrupted lock file.

This PR extends the `stack_snapshot` API docs and adds a comment to `examples/WORKSPACE` and the `WORKSPACE` generated by the `start` script instructing the user how to pin `stack_snapshot` dependencies. Additionally, this pins the `stack_snapshot` repositories `@stackage`, `@stackage-zlib`, and `@ghcide` in `rules_haskell`. The `@stackage` repository in `examples` remains unpinned and serves as a test-case for an unpinned `stack_snapshot`.

Notes
- The resulting `stack_snapshot` workspaces are identical with the exception of revisions of Cabal files.
- `stack ls dependencies json` does not expose the Hackage revision of any given package used in the snapshot, only the version. The original Cabal file may not work in the given Stackage snapshot, meaning a later revision is required, but downloading the latest revision is not reproducible. To work around this we pin the current git revision of the [all-cabal-hashes][all-cabal-hashes] repository and download Cabal files from that git revision. This may potentially use a later Hackage revision than used in the Stackage snapshot.
- `stack ls dependencies json` does not give us access to `sha256` of Hackage or archive dependencies. See https://github.com/commercialhaskell/stack/issues/5274 and https://github.com/commercialhaskell/stack/pull/5280. We use the [all-cabal-hashes][all-cabal-hashes] repository to obtain `sha256` of Hackage dependencies, and fetch archive dependencies during pinning to determine the hash.
- `stack ls dependencies json` does not provide subdirectory information of `archive` dependencies. This change includes a heuristic which expects the Cabal file at the archive root or underneath a top-level directory and defines `stripPrefix` accordingly.
- Bazel cannot cache `git` (or `hg`) dependencies in the repository cache as of now, see https://github.com/bazelbuild/bazel/issues/5086. This implementation keeps using `stack unpack` for these kinds of dependencies. To avoid issues around `stack update` this should also be switched to a Bazel implementation in the future.

[all-cabal-hashes]: https://github.com/commercialhaskell/all-cabal-hashes


Profpatsch:
Hyped for this. It removes the 5–10 second stack overhead that happened every time the bazel daemon was started, right?

aherrmann:
> It removes the 5–10 second stack overhead that happened every time the bazel daemon was started, right?

That's right, to be precise it removes any calls to `stack` for `stack_snapshot` that use the lock file and, for now, don't use `git` or `hg` dependencies.

examples/WORKSPACE
```diff
@@ -107,5 +107,8 @@ stack_snapshot(
         "text-show",
     ],
     snapshot = "lts-14.0",
+    # Pin Stackage dependencies with `bazel run @stackage-unpinned//:pin`.
```
hanshoglund, line 110:
```suggestion
    # This example uses an unpinned version of Stackage, meaning that stack is invoked on every build.
    # To switch to pinned stackage dependencies, run `bazel run @stackage-unpinned//:pin` and
    # uncomment the following line. 
```

start
```diff
@@ -271,8 +271,12 @@ stack_snapshot(
     name = "stackage",
     extra_deps = {"zlib": ["@zlib.dev//:zlib"]},
     packages = ["zlib"],
-    snapshot = "lts-14.27", # Last snapshot published for ghc-8.6.5
-    # the default version picked up by rules_haskell
+    # Last snapshot published for ghc-8.6.5 the default version picked up by
+    # rules_haskell
+    snapshot = "lts-14.27",
+    # Pin Stackage dependencies with \`bazel run @stackage-unpinned//:pin\`.
```
hanshoglund, line 277: Similar to my comment in `examples/WORKSPACE`.

haskell/cabal.bzl
```diff
@@ -1089,7 +1089,155 @@ library
     # Sort the items to make sure that generated outputs are deterministic.
     return {name: resolved[name] for name in sorted(resolved.keys())}
 
-def _download_packages(repository_ctx, snapshot, resolved):
+def _pin_packages(repository_ctx, resolved):
+    """Pin resolved packages.
```
hanshoglund, line 1093: Maybe a stupid question, but does this show up in the generated API docs? If not, we should probably have equivalent information somewhere `docs`. Not a blocker of course.

haskell/cabal.bzl
```diff
@@ -1556,6 +1916,16 @@ def stack_snapshot(
...
 
+    This rule invokes the `stack` tool for version and dependency resolution
+    based on the specified snapshot. You can generate a `stack_snapshot.json`
+    file to avoid invoking `stack` on every fetch and instead pin the outcome
+    in a file that can be checked into revision control. Execute the following
+    command and then specify the `stack_snapshot_json` attribute:
+
+    ```
+    bazel run @stackage-unpinned//:pin
+    ```
```
Profpatsch, line 1927: We should add some docs on how to update the snapshot.

Profpatsch: Ah, it’s in the pin script, but mentioning it here as well can’t hurt.

examples/WORKSPACE
aherrmann: Thanks! I've rephrased it slightly. The phrasing "version of Stackage" is a bit confusing as it may be interpreted as "Stackage snapshot version". The Stackage snapshot is already pinned with the `snapshot` and `local_snapshot` attribute. `stack_snapshot_json` additionally pins the package metadata that `stack` already knows in a way that's accessible to Bazel so that `stack` no longer needs to be invoked.

start
aherrmann: done

haskell/cabal.bzl
```diff
@@ -1089,7 +1089,155 @@ library
...
+
+    # Determine current git revision of all-cabal-hashes.
+    repository_ctx.download(
+        "https://api.github.com/repos/commercialhaskell/all-cabal-hashes/git/ref/heads/hackage",
+        output = "all-cabal-hashes-hackage.json",
+        executable = False,
+    )
+    hashes_json = json_parse(repository_ctx.read("all-cabal-hashes-hackage.json"))
+    hashes_object = _parse_json_field(hashes_json, "object", "dict", "all-cabal-hashes hackage branch")
+    hashes_commit = _parse_json_field(hashes_object, "sha", "string", "all-cabal-hashes hackage branch")
```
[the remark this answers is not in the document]
aherrmann: done

haskell/cabal.bzl
```diff
@@ -1187,21 +1335,149 @@ def _parse_packages_list(packages, vendored_packages):
...
+    )
+
+def _snapshot_json_checksum(all_cabal_hashes, resolved):
+    return hash(repr({
+        "all-cabal-hashes": all_cabal_hashes,
+        "resolved": resolved,
+    }))
+
+def _write_snapshot_json(repository_ctx, all_cabal_hashes, resolved):
+    """Write a snapshot.json file.
```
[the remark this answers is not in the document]
aherrmann: It calls `repository_ctx.file`, so it writes into the repository root of the current remote repository. I've added a few words to clarify that.

haskell/cabal.bzl
aherrmann: I've expanded the docs a bit.

aherrmann:
> It would be good to add more regression tests for the _unpinned_ use case, as this is what current projects will see.
> 
> Otherwise my concerns are mostly about documentation. I suggest opening an issue to track improvements there.

That's a good idea, I've opened https://github.com/tweag/rules_haskell/issues/1385 to that end. Feel free to edit and add to the list.

PR #1379, opened 2020-06-29
Refactor implementation of `stack_snapshot`

aherrmann:
This PR extracts the refactorings of the `stack_snapshot` implementation that occurred in https://github.com/tweag/rules_haskell/issues/1310. It does not change the functionality of `stack_snapshot` at this point. I have confirmed that the results of `bazel fetch` don't change between `master` and this PR on `@stackage`, `@stackage-zlib`, `@ghcide`, and `@stackge` in `examples`.

- Parse instead of validating the output of `stack ls dependencies json`. In particular, consistently introduce a `location.type` field into a package specification. Before, core packages had no `location` field at all.
- Factor out checks on the `snapshot`/`local_snapshot` attributes to make them re-usable between pinned and non-pinned `stack_snapshot`.
- Factor out parsing of the user provided package list to make them re-usable between pinned and non-pinned `stack_snapshot`.
- Split `_compute_dependency_graph` into two separate functions `_resolve_packages` and `_download_packages`. The pinned version will only need to call a version of `_download_packages` but not `_resolve_packages`. This also simplifies the transformations on the output of `stack ls dependencies json` that `_compute_dependency_graph` used to perform for historical reasons. The output of `stack ls dependencies json` is already pretty close to what we need in `stack_snapshot`. The new format is also easier to de-/serialize through JSON. Notably, `_resolve_packages` sorts the packages by name to ensure that the generated `BUILD.bazel` file and `packages.bzl` are reproducible.
- The generated `packages.bzl` file is modified to contain one package per line instead of all packages in one line. This is to make the file easier to read for humans and friendlier towards diffing.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -1149,6 +1149,50 @@ def _parse_stack_snapshot(repository_ctx, snapshot, local_snapshot):
         fail("Please specify one of snapshot or local_snapshot")
     return snapshot
 
+def _parse_packages_list(packages, vendored_packages):
+    """Parse the `packages` attribute to `stack_snapshot`.
+
+    Validates that there are no duplicates between `packages` and
+    `vendored_packages`.
+
+    Args:
+      packages: The list of requested packages versioned or unversioned.
+      vendored_packages: The dict of provided vendored packages.
+
+    Returns:
+      struct(all, core, versioned, unversioned):
+        all: The unversioned names of all requested packages.
+        core: The unversioned names of requested core packages.
+        versioned: The versioned names of requested versioned packages.
+        unversioned: The unversioned names of requested unversioned packages.
+    """
+    all_packages = []
+    core_packages = []
+    versioned_packages = []
+    unversioned_packages = []
+
+    for package in packages:
+        has_version = _has_version(package)
+        unversioned = _chop_version(package) if has_version else package
+        if unversioned in vendored_packages:
+            fail("Duplicate package '{}'. Packages may not be listed in both 'packages' and 'vendored_packages'.".format(package))
+        all_packages.append(unversioned)
+        if unversioned in _CORE_PACKAGES:
+            if has_version:
+                fail("{} is a core package, built into GHC. Its version is determined entirely by the version of GHC you are using. You cannot pin it to {}.".format(unversioned, _version(package)))
+            core_packages.append(unversioned)
+        elif has_version:
+            versioned_packages.append(package)
+        else:
+            unversioned_packages.append(package)
+
+    return struct(
+        all = all_packages,
+        core = core_packages,
+        versioned = versioned_packages,
+        unversioned = unversioned_packages,
+    )
+
 def _stack_snapshot_impl(repository_ctx):
     snapshot = _parse_stack_snapshot(
         repository_ctx,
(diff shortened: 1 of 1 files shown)


haskell/cabal.bzl
```diff
@@ -892,23 +892,71 @@ def _get_components(components, package):
...
+    actual_ty = type(json[field])
+    if actual_ty != ty:
+        fail(fmt % "Expected field '{field}' of type '{expected}', but got '{got}'.".format(
+            field = field,
+            expected = ty,
+            got = actual_ty,
+        ))
+    return json[field]
+
+def _parse_package_spec(package_spec):
```
Profpatsch, line 920: A link to the definition of the package spec format would be nice (though I don’t have the illusion that there exists one)

haskell/cabal.bzl
```diff
@@ -892,23 +892,71 @@ def _get_components(components, package):
...
+        field: _parse_json_field(package_spec, field, ty, errmsg.format(context = "package description"))
+        for (field, ty) in [("name", "string"), ("version", "string"), ("dependencies", "list")]
+    }
+
+    # Parse location field.
+    location = {}
+    if parsed["name"] in _CORE_PACKAGES or not "location" in package_spec:
+        location["type"] = "core"
+    else:
+        location_type = _parse_json_field(package_spec["location"], "type", "string", errmsg.format(context = "location description"))
```
Profpatsch, line 934: Some really long lines from here

haskell/cabal.bzl
```diff
@@ -892,23 +892,71 @@ def _get_components(components, package):
...
+        if location_type == "project package":
+            location["type"] = "vendored"
+        elif location_type == "hackage":
+            location["type"] = location_type
+            url_prefix = _parse_json_field(package_spec["location"], "url", "string", errmsg.format(context = location_type + " location description"))
+            location["url"] = url_prefix + "/{name}-{version}.tar.gz".format(**parsed)
+            # stack does not expose sha-256, see https://github.com/commercialhaskell/stack/issues/5274
```
Profpatsch, line 941: why is that important?

Profpatsch:
I didn’t review the second-to-last commit “Separate resolving and downloading packages”, as it was kind of hard to follow which parts went where, but I guess we’ll see if the changes are correct. It would probably be easier to review if it separated the parts that remove the legacy transformations from the pure refactor code.

haskell/cabal.bzl
aherrmann: That's a good idea. I'm not aware of a standalone definition, however the `stack` sources certainly define the current format. I've added a link to the docstring.

haskell/cabal.bzl
aherrmann: Fixed.

haskell/cabal.bzl
aherrmann: Once we download Hackage and archive dependencies using Bazel rather than `stack unpack` we'll want the SHA-256 to make sure the download is reproducible and caches well in Bazel's repository cache. However, since `stack ls dependencies json` doesn't expose the SHA-256 we have to determine it in other ways. https://github.com/tweag/rules_haskell/pull/1376 implements that.

If `stack` implements https://github.com/commercialhaskell/stack/issues/5274 then we can directly extract the SHA-256 here instead of determining it ourselves later on.

PR #1383, opened 2020-07-01
haskell_repl: Track data and runfiles explicitly

### LINKED ISSUE #1380: haskell_repl fails to load on compiler errors when collect_data = True
**Describe the bug**
A `haskell_repl` target fails to open the REPL if there are compiler errors in one of the modules being loaded from source. This happens only if `collect_data = True`.

**To Reproduce**
* Introduce an error in `tests/binary-simple/Main.hs`, e.g. delete the definition of `main`.
* Try `bazel run //tests/binary-simple:binary-simple@repl` and observe a Bazel build error that prevents the REPL from starting.
* Then set the default of `collect_data` to `False` in `haskell/repl.bzl`, retry the above, and observe the REPL opening and only then reporting an error.

**Expected behavior**
The REPL itself should build and open without errors, an compiler errors in modules loaded from source should only be reported by GHCi in the REPL session.

**Environment**
 - OS name + version: Ubuntu 19.10
 - Bazel version: 2.1.0
 - Version of the rules: 0b66b7bc8d6dd9e118ba4dabcf65be99022171d0

**Additional context**
This seems to be a consequence of setting `collect_data = True` in [`ctx.runfiles`](https://github.com/tweag/rules_haskell/blob/0b66b7bc8d6dd9e118ba4dabcf65be99022171d0/haskell/repl.bzl#L394). The [Bazel runfiles guide](https://docs.bazel.build/versions/master/skylark/rules.html#runfiles-features-to-avoid) recommends to avoid that feature and instead track runfiles explicitly. This requires some care for targets that depend rules from other rule sets, e.g. a `scala_binary` in a `data` dependency, to ensure that their runtime dependencies are also tracked correctly.

aherrmann:
In case of `haskell_binary` or `haskell_test` targets the `DefaultInfo.runfiles` will include the binary itself. This meant that if `collect_data == True` then the REPL depended on the compiled binary of targets that should be loaded into the REPL by source. I.e. if a module that should be loaded into the REPL by source cannot be compiled due to a compiler error then it can also not be loaded into a REPL with `collect_data = True`. This is very inconvenient as the REPL might help to fix the compiler error.

This changes fixes this by tracking data dependencies and runfiles explicitly rather than just relying on `ctx.runfiles(collect_data = True)`. In particular, in case of targets that should be loaded by source we will not include the full `DefaultInfo.runfiles` (which would include the final binary of a `haskell_binary|test`), but instead we only include the files and runfiles of `data` dependencies.

Closes #1380

### DIFF
--- a/haskell/repl.bzl
+++ b/haskell/repl.bzl
@@ -97,6 +97,22 @@ def _merge_runfiles(runfiles_list):
             result = result.merge(runfiles)
     return result
 
+def _data_runfiles(ctx, rule, attr):
+    """Generate runfiles for a data attribute.
+
+    Attrs:
+      ctx: The rule context.
+      rule: The rule object, `ctx` for a rule, `ctx.rule` for an aspect.
+      attr: The attribute name of the data attribute.
+
+    Returns:
+      A runfiles object capturing data files and data runfiles.
+    """
+    return _merge_runfiles(
+        [ctx.runfiles(files = getattr(rule.files, attr, []))] +
+        [data[DefaultInfo].default_runfiles for data in getattr(rule.attr, attr, [])],
+    )
+
 def _merge_HaskellReplLoadInfo(load_infos):
     source_files = depset()
     import_dirs = depset()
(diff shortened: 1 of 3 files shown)


aherrmann:
rebased on master to fix merge conflict.

PR #1388, opened 2020-07-07
Fix linker errors on Ubuntu >= 18.04 with GHC bindist

### LINKED ISSUE #1349: rules_haskell buildkite broken
It looks like rules_haskell is broken both HEAD and in-project, starting May 26.
Example Failure: https://buildkite.com/bazel/rules-haskell-haskell/builds/372#020c5972-dec5-4c7c-bff0-b4aa8cf2e372

Snip:
```
/usr/bin/ld.gold: error: /tmp/ghc3_0/ghc_2.o: requires dynamic R_X86_64_32 reloc against 'ZCMain_main_closure' which may overflow at runtime; recompile with -fPIC
--
  | /usr/bin/ld.gold: error: ../../../../../external/rules_haskell_ghc_linux_amd64/lib/base-4.13.0.0/libHSbase-4.13.0.0.a(Base.o): requires unsupported dynamic reloc 11; recompile with -fPIC
```

aherrmann:
Closes #1349

On some more recent Linux distributions (e.g. Ubuntu 18.04) `gcc` defaults to linking with `-pie`. At the same time GHC passes `-r` to the linker when joining object files. In order to avoid errors of the form 

     /usr/bin/ld: -r and -pie may not be used together

we need to pass `-no-pie` to the linker. Additionally, GHC compiles the `main` function during linking with non-position independent code. Attempting to link this code with `-pie` will cause the following error:

    /usr/bin/ld.gold: error: /tmp/ghc3_0/ghc_2.o: requires dynamic R_X86_64_32 reloc against 'ZCMain_main_closure' which may overflow at runtime; recompile with -fPIC
    /usr/bin/ld.gold: error: ../../../../../external/rules_haskell_ghc_linux_amd64/lib/base-4.13.0.0/libHSbase-4.13.0.0.a(Base.o): requires unsupported dynamic reloc 11; recompile with -fPIC

This was not caught on rules_haskell's bindist CI since it was still running on Ubuntu 16.04 which hadn't enabled `-pie` by default, yet. I've upgraded the rules_haskell bindist CI image to Ubuntu 18.04. I've [confirmed](https://buildkite.com/tweag-1/rules-haskell/builds/1560#c75bf4be-df7a-4f67-9499-c75b5c6c5326) that this does not break the build on Ubuntu 16.04.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -213,13 +213,7 @@ def _prepare_cabal_inputs(
     ], join_with = " ", format_each = "--ghc-arg=%s", omit_if_empty = False)
     args.add("--flags=" + " ".join(flags))
     if not hs.toolchain.is_darwin and not hs.toolchain.is_windows:
-        # On some more recent Linux distributions (e.g. Ubuntu 18.04) `gcc`
-        # defaults to linking with `-pie`. At the same time Cabal passes `-r`
-        # to the linker. In order to avoid errors of the form
-        #
-        #     /usr/bin/ld: -r and -pie may not be used together
-        #
-        # we need to pass `-no-pie` to the linker.
+        # See Note [No PIE when linking] in haskell/private/actions/link.bzl
         args.add("--ghc-option=-optl-no-pie")
     args.add_all(compiler_flags, format_each = "--ghc-option=%s")
     if dynamic_binary:
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -212,6 +212,15 @@ def _prepare_cabal_inputs(
         for arg in ["-package-db", "./" + _dirname(package_db)]
     ], join_with = " ", format_each = "--ghc-arg=%s", omit_if_empty = False)
     args.add("--flags=" + " ".join(flags))
+    if not hs.toolchain.is_darwin and not hs.toolchain.is_windows:
+        # On some more recent Linux distributions (e.g. Ubuntu 18.04) `gcc`
+        # defaults to linking with `-pie`. At the same time Cabal passes `-r`
+        # to the linker. In order to avoid errors of the form
+        #
+        #     /usr/bin/ld: -r and -pie may not be used together
+        #
+        # we need to pass `-no-pie` to the linker.
+        args.add("--ghc-option=-optl-no-pie")
     args.add_all(compiler_flags, format_each = "--ghc-option=%s")
     if dynamic_binary:
         args.add_all(
--- a/haskell/cc.bzl
+++ b/haskell/cc.bzl
@@ -5,7 +5,7 @@ These rules are deprecated.
 
 load(
     "@bazel_tools//tools/build_defs/cc:action_names.bzl",
-    "CPP_LINK_DYNAMIC_LIBRARY_ACTION_NAME",
+    "CPP_LINK_EXECUTABLE_ACTION_NAME",
     "C_COMPILE_ACTION_NAME",
 )
 load(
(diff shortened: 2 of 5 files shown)


haskell/cc.bzl
```diff
@@ -112,9 +113,6 @@ def cc_interop_info(ctx):
     cc_files = ctx.files._cc_toolchain + cc_wrapper.inputs.to_list()
     cc_manifests = cc_wrapper.manifests
 
-    # XXX Workaround https://github.com/bazelbuild/bazel/issues/6876.
-    linker_flags = [flag for flag in linker_flags if flag not in ["-shared"]]
```
aherrmann, line 116: While debugging #1349 I noticed that this could be simplified by following the suggestion from https://github.com/bazelbuild/bazel/issues/6876#issuecomment-452727209.

haskell/private/actions/link.bzl
```diff
@@ -108,6 +108,16 @@ def link_binary(
             print("WARNING: dynamic linking and profiling don't mix. Omitting -dynamic.\nSee https://ghc.haskell.org/trac/ghc/ticket/15394")
         else:
             args.add_all(["-pie", "-dynamic"])
+    elif not hs.toolchain.is_darwin and not hs.toolchain.is_windows:
+        # GHC compiles the `main` function during linking with non-position
+        # independent code. On some more recent Linux distributions (e.g.
+        # Ubuntu 18.04) `gcc` defaults to linking with `-pie`, so we need
```
hanshoglund, line 114: Is there an upstream ticket for why GHC is unable to produce PIEs?

haskell/cabal.bzl
```diff
@@ -212,6 +212,15 @@ def _prepare_cabal_inputs(
         for arg in ["-package-db", "./" + _dirname(package_db)]
     ], join_with = " ", format_each = "--ghc-arg=%s", omit_if_empty = False)
     args.add("--flags=" + " ".join(flags))
+    if not hs.toolchain.is_darwin and not hs.toolchain.is_windows:
+        # On some more recent Linux distributions (e.g. Ubuntu 18.04) `gcc`
```
hanshoglund, line 216: I'd suggest using [notes](https://ro-che.info/articles/2018-05-02-ghc-note-snippet) to link these comments together, and include a link to the original ticket.

dcreager:
We were seeing this on github/semantic#595, and I can confirm that this patch fixes our Linux builds.

haskell/cabal.bzl
aherrmann: Good idea, I've added a more detailed note and pointed to it next to the changes.

haskell/private/actions/link.bzl
aherrmann: There's a note in the GHC code base that explains this a bit further: https://gitlab.haskell.org/ghc/ghc/-/blob/d0bab2e3419e49cdbb1201d4650572b57f33420c/compiler/main/DynFlags.hs#L5551-5557.
I've added a pointer to it and related GHC tickets to the note in `haskell/private/actions/link.bzl`.

aherrmann:
> setting `no-pie` everywhere seems like a bit of a sledgehammer. It would be good to get more clarity on the upstream issue.

As pointed out in [this GHC note](https://gitlab.haskell.org/ghc/ghc/-/blob/d0bab2e3419e49cdbb1201d4650572b57f33420c/compiler/main/DynFlags.hs#L5551-5557) that corresponds to GHC's behavior provided that the C compiler supports the `-no-pie` flag. As pointed out in https://github.com/tweag/rules_haskell/pull/1388/commits/26cc476d81f746c0ac68fa851e28bb237c8f8d75 it's not so straight-forward to implement such autodetection in the case of rules_haskell. If necessary we could make this user-configurable or if really necessary perform auto-detection in a GHC wrapper.

PR #1389, opened 2020-07-09
Upgrade to ghcide 0.2.0

aherrmann:
Upgrades `ghcide` to version 0.2.0.

Extends the `hie_bios` output to include the list of source files that are covered by the cradle. This informs `ghcide` for which source files the provided flags are valid. Without this information `ghcide` would have to re-query the cradle for newly opened source files and then merge the corresponding `HscEnv` objects. After such a merge typechecking has to be repeated, which makes it costly.

`ghcide` 0.2.0 also adds multi-cradle support and now passes the current file name to the `hie-bios` script of a bios cradle. The latter can be used to query the Bazel target corresponding to a given source file and build `hie_bios` file directly from the aspect rather than having to manually define a dedicated `haskell_repl` target. However, this will likely trigger some `HscEnv` merges as described above and needs a bit more testing. I'll defer that approach to a separate PR.


teto:
cool change to pass target to hie_bios: `hie_bios_flags "$TARGET"`, I was thinking of doing the same to be able to generate multiple bioses on entering nix-shell and reference those in a multicradle hie.yaml. Let me know if you want help testing it (when it's ready)

hanshoglund:
LGTM.

Issue #379, opened 2018-07-30
Fully static binaries


mboes:
GHC by default creates "static" libraries. As explained in #378, this really means "mostly static" in Bazel terms, because some dependencies are still dynamically linked. It is possible, however, to create fully static binaries, provided there are static versions of all external libraries available. We should implement this link mode as well, and make it configurable per-target. We should probably stick to a similar API for this to what the [CC rules already provide](https://docs.bazel.build/versions/master/be/c-cpp.html#cc_binary.linkstatic).

cc @lunaris

Profpatsch:
nixpkgs has made major progress in this territory, meta issue: https://github.com/NixOS/nixpkgs/issues/43795

Here’s an example how dhall uses that feature: https://github.com/dhall-lang/dhall-haskell/commit/3e305ac373d7c8b2cda14d0b0b53437a3979343c

They add support to cabal to make this possible:
* https://github.com/haskell/cabal/pull/5451
* https://github.com/haskell/cabal/pull/5446

We don’t use cabal, but we can infer the GHC flags we need from these patches.

jkachmar:
Is there anything that interested parties like myself can help contribute here? 

Fully statically linked executables (in addition to everything else that's part of the `rules_haskell` project) would be a huge, huge win for me both personally and professionally so I'm significantly interested in helping this along.

aherrmann:
It's worth noting that since https://github.com/tweag/rules_haskell/pull/587 rules_haskell links binaries "mostly static" in the sense that both `haskell_library` and `cc_library` targets will be linked statically, if possible.

It seems that what's missing to achieve full parity with the [cc rules](https://docs.bazel.build/versions/master/be/c-cpp.html#cc_binary.linkstatic) is the `fully_static_link` feature. Maybe this could be achieved by passing `-optl-static` to GHC. However, this fails if any dependency is only available as dynamic library.

mboes:
Yes but a fully static mode would enable trivially deployable Haskell binaries, which is what I'm assuming @jkachmar is shooting for. I'd say the way to go about this is to enable `-optl-static` in `compiler_flags`, see what breaks, and take it from there (we might need to thread more information through the CC providers).

jkachmar:
Just so, @mboes. 

The added wrinkle is that it would also entail something like what @nh2 has done in the linked NixOS thread. 

[...]

PR #1390, opened 2020-07-10
Add support for building fully-static binaries

lunaris:
Off the back of much discussion in #379.

This commit implements support for fully-statically-linked binaries by
extending functionality linked to the `is_static` attribute of an active
Haskell toolchain. In particular:

* Packages built with Cabal will be passed fields that ensure their
  static artifacts are relocatable (`-fPIC` and
  `-fexternal-dynamic-refs`).

* Intermediate artifacts (such as binaries built by `hsc2hs` for the
  purposes of generating Haskell code) will also be built statically to
  avoid issues caused by the absences of dynamic libraries.

* Linker flags for REPLs will be generated so that they correctly find
  static C library dependencies instead of failing/finding dynamic
  counterparts.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -165,7 +165,7 @@ def _prepare_cabal_inputs(
         dynamic = True,
     )
 
-    # Executables build by Cabal will link Haskell libraries statically, so we
+    # Executables built by Cabal will link Haskell libraries statically, so we
     # only need to include dynamic C libraries in the runfiles tree.
     (_, runfiles_libs) = get_library_files(
         hs,
@@ -230,6 +230,34 @@ def _prepare_cabal_inputs(
             ],
             uniquify = True,
         )
+
+    # When building in a static context, we need to make sure that Cabal passes
+    # a couple of options that ensure any static code it builds can be linked
+    # correctly.
+    #
+    # * If we are using a static runtime, we need to ensure GHC generates
+    #   position-independent code (PIC). On Unix we need to pass GHC both
+    #   `-fPIC` and `-fexternal-dynamic-refs`: with `-fPIC` alone, GHC will
+    #   generate `R_X86_64_PC32` relocations on Unix, which prevent loading its
+    #   static libraries as PIC.
+    #
+    # * If we are building fully-statically-linked binaries, we need to ensure that
+    #   we pass arguments to `hsc2hs` such that objects it builds are statically
+    #   linked, otherwise we'll get dynamic linking errors when trying to
+    #   execute those objects to generate code as part of the build.  Since the
+    #   static configuration should ensure that all the objects involved are
+    #   themselves statically built, this is just a case of passing `-static` to
+    #   the linker used by `hsc2hs` (which will be our own wrapper script which
+    #   eventually calls `gcc`, etc.).
+    if hs.toolchain.static_runtime:
+        args.add("--ghc-option=-fPIC")
+
+        if not hs.toolchain.is_windows:
+            args.add("--ghc-option=-fexternal-dynamic-refs")
+
+    if hs.toolchain.fully_static_link:
+        args.add("--hsc2hs-option=--lflag=-static")
+
     args.add("--")
     args.add_all(package_databases, map_each = _dirname, format_each = "--package-db=%s")
     args.add_all(direct_include_dirs, format_each = "--extra-include-dirs=%s")
@@ -359,7 +387,7 @@ def _haskell_cabal_library_impl(ctx):
     else:
         profiling_library = None
         static_library = vanilla_library
-    if hs.toolchain.is_static:
+    if hs.toolchain.static_runtime:
         dynamic_library = None
     else:
         dynamic_library = hs.actions.declare_file(
--- a/haskell/nixpkgs.bzl
+++ b/haskell/nixpkgs.bzl
@@ -76,7 +76,8 @@ haskell_toolchain(
     tools = {tools},
     libraries = toolchain_libraries,
     version = "{version}",
-    is_static = {is_static},
+    static_runtime = {static_runtime},
+    fully_static_link = {fully_static_link},
     compiler_flags = {compiler_flags} + {compiler_flags_select},
     haddock_flags = {haddock_flags},
     cabalopts = {cabalopts},
@@ -90,7 +91,8 @@ haskell_toolchain(
             toolchain_libraries = toolchain_libraries,
             tools = ["@rules_haskell_ghc_nixpkgs//:bin"],
             version = repository_ctx.attr.version,
-            is_static = repository_ctx.attr.is_static,
+            static_runtime = repository_ctx.attr.static_runtime,
+            fully_static_link = repository_ctx.attr.fully_static_link,
             compiler_flags = repository_ctx.attr.compiler_flags,
             compiler_flags_select = compiler_flags_select,
             haddock_flags = repository_ctx.attr.haddock_flags,
@@ -107,7 +109,8 @@ _ghc_nixpkgs_haskell_toolchain = repository_rule(
         # These attributes just forward to haskell_toolchain.
         # They are documented there.
         "version": attr.string(),
-        "is_static": attr.bool(),
+        "static_runtime": attr.bool(),
+        "fully_static_link": attr.bool(),
         "compiler_flags": attr.string_list(),
         "compiler_flags_select": attr.string_list_dict(),
         "haddock_flags": attr.string_list(),
@@ -162,7 +165,9 @@ _ghc_nixpkgs_toolchain = repository_rule(_ghc_nixpkgs_toolchain_impl)
 
 def haskell_register_ghc_nixpkgs(
         version,
-        is_static = False,
+        is_static = None,  # DEPRECATED. See _check_static_attributes_compatibility.
+        static_runtime = None,
+        fully_static_link = None,
         build_file = None,
         build_file_content = None,
         compiler_flags = None,
@@ -209,6 +214,17 @@ def haskell_register_ghc_nixpkgs(
       ```
 
     Args:
+      is_static: Deprecated. The functionality it previously gated
+        (supporting GHC versions with static runtime systems) now sits under
+        static_runtime, a name chosen to avoid confusion with the new flag
+        fully_static_link, which controls support for fully-statically-linked
+        binaries. During the deprecation period, we rewrite is_static to
+        static_runtime in this macro as long as the new attributes aren't also
+        used. This argument and supporting code should be removed in a future release.
+      static_runtime: True if and only if a static GHC runtime is to be used. This is
+        required in order to use statically-linked Haskell libraries with GHCi
+        and Template Haskell.
+      fully_static_link: True if and only if fully-statically-linked binaries are to be built.
       compiler_flags_select: temporary workaround to pass conditional arguments.
         See https://github.com/bazelbuild/bazel/issues/9199 for details.
       sh_posix_attributes: List of attribute paths to extract standard Unix shell tools from.
@@ -218,6 +234,12 @@ def haskell_register_ghc_nixpkgs(
     nixpkgs_sh_posix_repo_name = "rules_haskell_sh_posix_nixpkgs"
     haskell_toolchain_repo_name = "rules_haskell_ghc_nixpkgs_haskell_toolchain"
     toolchain_repo_name = "rules_haskell_ghc_nixpkgs_toolchain"
+
+    static_runtime, fully_static_link = _check_static_attributes_compatibility(
+        is_static = is_static,
+        static_runtime = static_runtime,
+        fully_static_link = fully_static_link,
+    )
 
     # The package from the system.
     nixpkgs_package(
@@ -237,7 +259,8 @@ def haskell_register_ghc_nixpkgs(
     _ghc_nixpkgs_haskell_toolchain(
         name = haskell_toolchain_repo_name,
         version = version,
-        is_static = is_static,
+        static_runtime = static_runtime,
+        fully_static_link = fully_static_link,
         compiler_flags = compiler_flags,
         compiler_flags_select = compiler_flags_select,
         haddock_flags = haddock_flags,
@@ -264,6 +287,49 @@ def haskell_register_ghc_nixpkgs(
         name = nixpkgs_sh_posix_repo_name,
         **sh_posix_nixpkgs_kwargs
     )
+
+def _check_static_attributes_compatibility(is_static, static_runtime, fully_static_link):
+    """Asserts that attributes passed to `haskell_register_ghc_nixpkgs` for
+    controlling use of GHC's static runtime and whether or not to build
+    fully-statically-linked binaries are compatible.
+
+    Args:
+      is_static: Deprecated. The functionality it previously gated
+        (supporting GHC versions with static runtime systems) now sits under
+        static_runtime, a name chosen to avoid confusion with the new flag
+        fully_static_link, which controls support for fully-statically-linked
+        binaries. During the deprecation period, we rewrite is_static to
+        static_runtime in this macro as long as the new attributes aren't also
+        used. This argument and supporting code should be removed in a future release.
+      static_runtime: True if and only if a static GHC runtime is to be used. This is
+        required in order to use statically-linked Haskell libraries with GHCi
+        and Template Haskell.
+      fully_static_link: True if and only if fully-statically-linked binaries are to be built.
+
+    Returns:
+      A tuple of static_runtime, fully_static_link attributes, which are guaranteed
+      not to be None, taking into account the deprecated is_static argument.
+    """
+
+    # Check for use of the deprecated `is_static` attribute.
+    if is_static != None:
+        if static_runtime != None or fully_static_link != None:
+            fail("is_static is deprecated. Please use the static_runtime attribute instead.")
+
+        print("WARNING: is_static is deprecated. Please use the static_runtime attribute instead.")
+        static_runtime = is_static
+
+    # Currently we do not support the combination of a dynamic runtime system
+    # and fully-statically-linked binaries, so fail if this has been selected.
+    if not static_runtime and fully_static_link:
+        fail(
+            """\
+Fully-statically-linked binaries with a dynamic runtime are not currently supported.
+Please pass static_runtime = True if you wish to build fully-statically-linked binaries.
+""",
+        )
+
+    return bool(static_runtime), bool(fully_static_link)
 
 def _find_children(repository_ctx, target_dir):
     find_args = [
--- a/haskell/private/actions/compile.bzl
+++ b/haskell/private/actions/compile.bzl
@@ -56,6 +56,21 @@ def _process_hsc_file(hs, cc, hsc_flags, hsc_inputs, hsc_file):
     args.add_all(["--cflag=" + f for f in cc.compiler_flags])
     args.add_all(["--cflag=" + f for f in cc.include_args])
     args.add_all(["--lflag=" + f for f in cc.linker_flags])
+
+    # If are building fully-statically-linked binaries, we need to ensure that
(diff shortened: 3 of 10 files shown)


haskell/repl.bzl
```diff
@@ -316,12 +316,31 @@ def _compiler_flags_and_inputs(hs, repl_info, static = False, path_prefix = ""):
         cc_library_files = _concat(get_library_files(hs, cc_libraries_info, cc_libraries))
     else:
         cc_library_files = get_ghci_library_files(hs, cc_libraries_info, cc_libraries)
-    link_libraries(cc_library_files, args)
+
```
lunaris, line 319: @aherrmann thoughts on this? I considered patching `link_libraries` but couldn't see a nice way to do it that didn't make it kind of pointlessly general (e.g. I started by trying `lib_format = "-l%s"` / `dir_format = "-L%s"` flags to supplant `prefix_optl` but that doesn't really work because you need to conditionally format and if I pass in lambdas, what is the function really abstracting for me?). Thus I implemented a specialised version here. Moreover, the use of `bazel-out` feels bit ugly but I couldn't see an obviously nicer way -- perhaps if this issue only affects external repositories we could instead test for `.startswith("external")` and flip the condition? I don't have enough confidence to say either way; thoughts welcome.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -134,12 +142,22 @@ class Args:
...
+                # this explicitly.
                 self.rpaths.extend(self.library_paths)
+
+            # When in a static configuration, we need to ensure that the objects
+            # that hsc2hs builds in particular are statically linked. Since the
+            # static configuration should ensure that all the objects involved
+            # are themselves statically built, this is just a case of passing
+            # `-static` to the real C compiler to produce the final object.
+            if self.output.endswith("hsc_make") or self.output.endswith("hsc_make.o"):
+                self.args.extend(["-static"])
```
lunaris, line 160: As mentioned in https://github.com/tweag/rules_haskell/issues/379#issuecomment-657456884 I will see if this can be hoisted to a higher level.

haskell/repl.bzl
aherrmann: A good way to do this may be an additional `path_prefix = ""` optional argument to `link_libraries`. The `-L` format strings could then be extended with that, e.g.
```
    dirfmt = "-optl-L" + path_prefix + "%s"
    ...
    dirfmt = "-L" + path_prefix + "%s"
```
This doesn't have to be conditional on whether the path starts with `bazel-out` or `external`. Both these types of paths exist under the execroot.

lunaris:
@aherrmann I've pulled out the `hsc2hs` flags as suggested, which I agree is much nicer. Also I took your recommendation and added `path_prefix` to `link_libraries`. `bazel test //...` is all green for me locally though I'm not sure if CI can catch any more (e.g. Windows, Darwin) cases.

If you think this looks sensible, some ideas I have:

* Thoughts on adding a couple of `build @stackage//:<package>` s to the test suite? In particular I know that (currently at least) `hslua` exercises the `hsc2hs` code path and as seen `openssl` will trigger some of the GHCi linking code paths.

* I think we should hand-craft an `hsc2hs` example to make sure that the `--lflag=-static` works there (I've not tested this but assume it will work as it does in the Cabal wrapper).

* Naturally, lots of documentation for the static use case, which feels fairly complete now! (At least, if you're using Nix/Linux and not a bindist...)

lunaris:
Yea looks like we might need some guards for Windows at least -- the Azure failures seem to be linker errors, which I assume are genuine and caused by `is_static` having more aggressive semantics now.

aherrmann:
Right, `--lflag=-static` is too strict for the mostly static use-case with the static RTS, and probably also for Windows, as it completely disables linking against dynamic libraries. Sorry, I didn't think of that before. It looks like we'll have to separate the static RTS and fully static use-case.

IIUC the only difference between static RTS and fully static is whether `--lflag=-static` needs to be passed to `hsc2hs`, correct? I'm not sure that this justifies an additional dedicated toolchain attribute. Maybe a more generic `hsc2hsopt` toolchain attribute could handle this case? WDYT?

>     * Thoughts on adding a couple of `build @stackage//:<package>` s to the test suite? In particular I know that (currently at least) `hslua` exercises the `hsc2hs` code path and as seen `openssl` will trigger some of the GHCi linking code paths.

Yes, not opposed to that. If possible it's better to create a dedicated test-case in `tests`. That's more stable towards future updates of the Stackage snapshot and also doesn't introduce additional external dependencies. But, if that's not feasible then an additional `stack_snapshot` dependency is a good alternative. Though, we currently don't have a static RTS or fully static CI pipeline in rules_haskell. So, anything hidden behind `if is_static` won't be exercised. It would be good to add such a pipeline, though that sounds like a task for a separate PR.

>     * Naturally, lots of documentation for the static use case, which feels fairly complete now! (At least, if you're using Nix/Linux and not a bindist...)

Absolutely, a use-case section in `docs/haskell-use-cases.rst` would be a good start for that. I think it's fine that this only covers Nix/Linux for now.

lunaris:
To elaborate on your comment, as I understand it -- passing `-fPIC` / `-fexternal-dynamic-refs` to Cabal and path-prefixing `-L` flags in REPLs will work for static runtimes with mostly-statically-linked binaries, is that right? If so, then yes, that only leaves `--lflag=-static` to `hsc2hs`.

As to how we trigger this:

* I appreciate that renaming `is_static` is not really an option without an aggressive bump, but in an ideal scenario I feel like I'd lean towards toolchains having `static_runtime` and `static_build` attributes (or similarly named).

* Given we likely can't do that, I'd still be inclined to add an attribute (perhaps that implies `is_static`) to the toolchain. My reasoning here would be that, supposing we discover that we have missed a corner case (e.g. some TH/REPL/gnarly thing we have to patch around), I could see us _then_ needing something alongside the `hsc2hsopt` attribute, etc. Whereas if we added `builds_static` (for example, which could, for example, imply `is_static`) we might allow for that in future.

Just my 2c though -- generally happy to do whatever as long as it works and is documented :+1:

aherrmann:
> To elaborate on your comment, as I understand it -- passing `-fPIC` / `-fexternal-dynamic-refs` to Cabal and path-prefixing `-L` flags in REPLs will work for static runtimes with mostly-statically-linked binaries, is that right? If so, then yes, that only leaves `--lflag=-static` to `hsc2hs`.

Yes, that's what I meant.



>     * I'd still be inclined to add an attribute (perhaps that implies `is_static`) to the toolchain. My reasoning here would be that, supposing we discover that we have missed a corner case (e.g. some TH/REPL/gnarly thing we have to patch around), I could see us _then_ needing something alongside the `hsc2hsopt` attribute, etc. Whereas if we added `builds_static` (for example, which could, for example, imply `is_static`) we might allow for that in future.

That's a good point! Okay, let's go with a dedicated attribute.

AFAIK the only other user of the `is_static` attribute is daml, and a toolchain attribute name is pretty easy to change. So, I'm not opposed to changing it to a more descriptive name. `is_static` is ambiguous when it comes to distinguishing the static RTS and fully static builds use-cases. So, it makes sense to change it. I agree that `static_runtime` is a better name for this attribute.

Regarding the fully static mode, looking for precedent in Bazel there is the [`fully_static_link`](https://docs.bazel.build/versions/master/be/c-cpp.html#cc_library.linkstatic) feature flag. Following that we could name the new GHC toolchain attribute `fully_static_link`.

aherrmann:
> a. it seems to be something of a top-level feature switch that affects all Bazel

It's a recognized value for the [`features` attribute](https://docs.bazel.build/versions/3.4.0/be/common-definitions.html#common.features) of the `cc_*` rules. If present, then `-static` will [be set during linking](https://github.com/bazelbuild/bazel/blob/3.3.0/src/main/java/com/google/devtools/build/lib/rules/cpp/CppActionConfigs.java#L1376-L1391). Using it as an attribute name shouldn't interfere with that functionality. Given that fully static linking in Haskell requires support from the GHC toolchain, it makes sense to make this switch an attribute of the toolchain itself, rather than a feature than can be selected on individual targets.

> b. it may currently be broken (!) -- [bazelbuild/bazel#8672](https://github.com/bazelbuild/bazel/issues/8672)

That's good to know. It looks like building fully statically linked `cc_binary` targets requires extra flags and configuration next to `features = ["fully_static_link"]`. This shouldn't interfere with fully static linking of Haskell targets, though.

> That said, I do think it is morally the right choice, and if we're not against changing (or at least initially deprecating) `is_static` then that sounds great 

That sounds good. Such a deprecation can be implemented at the nixpkgs toolchain wrapper macro [`haskell_register_ghc_nixpkgs`](https://github.com/tweag/rules_haskell/blob/abaec6502a4474f10b3c367fb5e90173ee0e349c/haskell/nixpkgs.bzl#L160).

> `static_runtime` and `fully_static_link`, which implies `static_runtime`, correct?

Yes, though I think it's best if the user has to explicitly write
```
    static_runtime = True,
    fully_static_link = True,
```
and we raise an error in case of `static_runtime = False, fully_static_link = True,`. My reasoning is this: IIUC the nixpkgs implementation of fully static linking works with a dynamic RTS and links libraries dynamically for GHCi and Template Haskell. So, in principle the combination `static_runtime = False, fully_static_link = True,` is possible, it's just not supported by rules_haskell at this point. If support for this configuration is ever implemented in rules_haskell, then turning an implicit implication into an explicit setting would be a subtle breaking change.

haskell/repl.bzl
```diff
@@ -316,12 +315,25 @@ def _compiler_flags_and_inputs(hs, repl_info, static = False, path_prefix = ""):
...
+    # External static C library dependencies are an example of this -- unchanged
+    # we may end up with paths like `external/some_dependency/lib` and/or
+    # `bazel-out/k8-fastbuild/bin/_solib_k8/...`; the former containing the
+    # archive (`.a`) file we want, but being relative to the execroot, and the
+    # latter being relative to the workspace root but only containing dynamic
+    # libraries.
+    #
+    # We fix this by identifying these libraries when generating linker flags and
+    # prefixing them so that they will be visible when the REPL runs.
+    link_libraries(cc_library_files, args, path_prefix = "$RULES_HASKELL_EXEC_ROOT/")
```
aherrmann, line 330:
```suggestion
    link_libraries(cc_library_files, args, path_prefix = path_prefix + "/")
```
`path_prefix` is already an argument to `_compiler_flags_and_inputs`.

haskell/repl.bzl
```diff
@@ -301,7 +301,6 @@ def _compiler_flags_and_inputs(hs, repl_info, static = False, path_prefix = ""):
     for package_cache in repl_info.dep_info.package_databases.to_list():
         args.extend(["-package-db", paths.join(path_prefix, package_cache.dirname)])
 
-    # Load C library dependencies
```
aherrmann, line 304: Why remove this?

lunaris:
I've made all the changes you suggested -- great to see this patch getting smaller and smaller! I'll take up test cases and documentation when I next have time and then the only question I see left is deprecating `is_static`. Options I can see:

1. It's just not there in the next release. We have some documentation about it but if you use it you'll get a Bazel error.

2. It's there in the next release but if you use it we throw an error telling you to use `static_runtime` and perhaps explaining/linking to docs about `fully_static_build`.

3. It's there in the next release but if you use it we pass it as `static_runtime` and give you a warning letting you know that `is_static` is deprecated and that you should move away from it. Possibly also explaining/linking to docs about `fully_static_build`.

There's also the question of combinations of `static_runtime` and `fully_static_build` -- do we want to check and warn about these? If so, do we already have some sort of warnings for "bad" combinations (like bindist/Windows/dynamic runtime)? If not, do we want to add this while we're there? I'm guessing any such logic could go in `haskell_toolchain`, since I think all roads go through there.

More feedback welcome as always; cheers.

haskell/private/cc_libraries.bzl
```diff
@@ -95,7 +95,7 @@ def get_library_files(hs, cc_libraries_info, libraries_to_link, dynamic = False,
         pic = dynamic
 
     # PIC is irrelevant on static GHC.
```
lunaris, line 97: Not sure I understand this comment on reflection, given the way we use `-fPIC` and `-fexternal-dynamic-refs` in this PR -- @aherrmann thoughts and also whether we should elaborate on it?

haskell/private/actions/compile.bzl
```diff
@@ -258,7 +272,7 @@ def _compilation_defaults(hs, cc, java, posix, dep_info, plugin_dep_info, srcs,
         # to debug issues in non-sandboxed builds.
         "-Wmissing-home-modules",
     ])
-    if hs.toolchain.is_static and not hs.toolchain.is_windows:
+    if hs.toolchain.static_runtime and not hs.toolchain.is_windows:
         # A static GHC RTS requires -fPIC. However, on Unix we also require
         # -fexternal-dynamic-refs, otherwise GHC still generates R_X86_64_PC32
         # relocations which prevents loading these static libraries as PIC.
```
lunaris, line 278: Looking at this I think I should check for `is_windows` in the Cabal wrapper too, correct? I don't do this at present so will revisit.

haskell/private/actions/compile.bzl
```diff
@@ -56,6 +56,20 @@ def _process_hsc_file(hs, cc, hsc_flags, hsc_inputs, hsc_file):
...
+    # dynamic linking errors when trying to execute those objects to generate
+    # code as part of the build.  Since the static configuration should ensure
+    # that all the objects involved are themselves statically built, this is
+    # just a case of passing `-static` to the linker used by `hsc2hs` (which
+    # will be our own wrapper script which eventually calls `gcc`, etc.).
+    #
+    # Note that we also do this in our Cabal wrapper, where hsc2hs might be
+    # called by Cabal as part of the build process.
+    if hs.toolchain.static_runtime:
+        args.add("--lflag=-static")
```
lunaris, line 71: D'oh, missed this -- will hang off `fully_static_link` instead.

haskell/cabal.bzl
```diff
@@ -226,6 +226,33 @@ def _prepare_cabal_inputs(
...
+    #   statically linked, otherwise we'll get dynamic linking errors when
+    #   trying to execute those objects to generate code as part of the build.
+    #   Since the static configuration should ensure that all the objects
+    #   involved are themselves statically built, this is just a case of passing
+    #   `-static` to the linker used by `hsc2hs` (which will be our own wrapper
+    #   script which eventually calls `gcc`, etc.).
+    if hs.toolchain.static_runtime:
+        args.add_all([
+            "--ghc-option=-fPIC",
+            "--ghc-option=-fexternal-dynamic-refs",
```
lunaris, line 250: See my other comment -- should we check `is_windows` here? I don't think the same applies for `fully_static_link` since at present using that with `is_windows` is just simply not supported, right?

haskell/private/haskell_impl.bzl
```diff
@@ -362,7 +362,7 @@ def haskell_library_impl(ctx):
     srcs_files, import_dir_map = _prepare_srcs(ctx.attr.srcs)
 
     with_shared = not ctx.attr.linkstatic
-    if with_profiling or hs.toolchain.is_static:
+    if with_profiling or hs.toolchain.static_runtime:
```
lunaris, line 365: I don't _think_ it'll be an issue here based on previous comments, but if we ever _do_ want a dynamic runtime with a static binary set (honestly I still don't think I understand how this could work 😂 ) this will need to be unpicked/tested I think.

haskell/private/cc_libraries.bzl
aherrmann: Good question. I looked into the history, I introduced that comment [for Windows](https://github.com/tweag/rules_haskell/commit/16ed109d4659dec9f403d24a0080cd262f12bf79#diff-5fd52f58c9c22b2fe4b00f1b9d279ad0R269-R270) and later [generalized it to static RTS](https://github.com/tweag/rules_haskell/commit/fb37a1e4377621d96213746a596a99fb7bef8017#diff-5fd52f58c9c22b2fe4b00f1b9d279ad0L255-R256). But, it doesn't look quite right, indeed. `pic_required` controls whether a `LibraryToLink.static_library` can be preferred over a `.dynamic_library` or if it has to be a `.pic_static_library`. Even with a static RTS pic should be required for GHCi or Template Haskell. I think the trouble here is that too often a static archive is only assigned to `.static_library` even if `.pic_static_library` would be appropriate ([e.g. here](https://github.com/tweag/rules_haskell/blob/eb16a5401770098a801775ea3a893b09cafe054c/haskell/private/haskell_impl.bzl#L553)). It seems a better approach would be to assign generated static libraries to `.pic_static_library` if `-fPIC` is set and remove `pic_required` here. But, that seems outside the scope of this PR.

haskell/private/actions/compile.bzl
aherrmann: Yes, IIRC `-fexternal-dynamic-refs` doesn't work on Windows, or at least is not requried. Preferably, that would go into `haskell/cabal.bzl`.

haskell/cabal.bzl
aherrmann: Indeed, `-fexternal-dynamic-refs` should be Unix only.

aherrmann:
> I've made all the changes you suggested -- great to see this patch getting smaller and smaller!

Thank you!

> the only question I see left is deprecating `is_static`. Options I can see:
> 
>     1. It's just not there in the next release. We have some documentation about it but if you use it you'll get a Bazel error.
> 
>     2. It's there in the next release but if you use it we throw an error telling you to use `static_runtime` and perhaps explaining/linking to docs about `fully_static_build`.
> 
>     3. It's there in the next release but if you use it we pass it as `static_runtime` and give you a warning letting you know that `is_static` is deprecated and that you should move away from it. Possibly also explaining/linking to docs about `fully_static_build`.

In the past we've usually gone with 3, it sounds reasonably straight-forward to implement in `haskell_register_ghc_nixpkgs`. `is_static` is currently only exposed through nixpkgs GHC, so we don't need to worry about bindist.


> There's also the question of combinations of `static_runtime` and `fully_static_build` -- do we want to check and warn about these?

I'd say, `not static_runtime and fully_static_build` should raise an error (`fail()`) for now. As your work has shown, rules_haskell is not currently able to handle that combination. All the other combinations should be fine. `static_runtime` defaults to `False` on Unix and `True` on Windows, `fully_static_build` defaults to `False` on all platforms.

> If so, do we already have some sort of warnings for "bad" combinations (like bindist/Windows/dynamic runtime)? If not, do we want to add this while we're there?

I don't think we have checks for any such "bad" combinations, yet. TBH I don't know if `static_runtime = False` on Windows is bad. I've never tried to build a dynamic RTS GHC on Windows.

lunaris:
@aherrmann I implemented the compatibility checking/deprecation here: https://github.com/tweag/rules_haskell/pull/1390/files#diff-63173affa8c52d7a87df960435e8c641R275; see what you think. I've also pushed a chunk of documentation which hopefully looks OK and will see what the Windows tests say now that (hopefully) everything is gated by appropriate conditionals.

haskell/nixpkgs.bzl
```diff
@@ -159,7 +162,9 @@ _ghc_nixpkgs_toolchain = repository_rule(_ghc_nixpkgs_toolchain_impl)
 
 def haskell_register_ghc_nixpkgs(
         version,
-        is_static = False,
+        is_static = None, # DEPRECATED. See _check_static_attributes_compatibility.
+        static_runtime = None,
+        fully_static_link = None,
```
aherrmann, line 167: I think it would be good to add descriptions for these attributes to the `Args` section of the docstring. In particular, that `is_static` is deprecated, that `static_runtime = True` means that GHC has to be built with a static RTS and that it's needed to use static Haskell libs for GHCi/TH, and that `fully_static_link` will produce fully statically linked binaries.

lunaris:
Added the missing arguments documentation and hopefully have appeased Buildifier; :crossed_fingers: we get a green CI this time!

PR #1395, opened 2020-07-16
Upgrade Bazel 2.1.0 --> 3.3.1

aherrmann:
* Updates `.ci/bazelversion` and the corresponding sha-256 files.
* Updates nixpkgs to include bazel 3.3.1.


aherrmann:
This doesn't change anything apart from the Bazel version and doesn't break any compatibility with the previously established version range. So, if 2.1.0 was know to work before then it still does. See [history of `versions.bzl`](https://github.com/tweag/rules_haskell/commits/839c5c0ecd15e0f14d6f77d2fd0abecdea078e32/haskell/private/versions.bzl) for how this has been treated so far.

It's true that CI currently only tests one Bazel version. I've opened an issue to track this: https://github.com/tweag/rules_haskell/issues/1402

> But Bazel uses SemVer, so the `3.3.1` test should only give us confidence for the range `3.3 <= x < 4`.

That's fair enough, but it is a change of how the version check is implemented. The upper bound is inclusive, so `x < 4` can't be expressed right now. Changing this seems out of scope for this PR.

hanshoglund:
> The upper bound is inclusive, so x < 4 can't be expressed right now. 

That's fine. My point was simply that we should test all supported major versions. Let's track this in #1402.

Issue #1067, opened 2019-09-03
PATH environment variable for `actions.run` on nix


guibou:
**Describe the bug**

All the `ctx.actions.run` which call a python script in rules haskell are failing since the following change in nixpkgs bazel:  https://github.com/NixOS/nixpkgs/commit/c4813515ca34e7f2aef73280646c94c4c1b0eb86

This change removes a hack: previously, the path of the `python` interpreter was selected from an hardcoded value: 

https://github.com/NixOS/nixpkgs/commit/c4813515ca34e7f2aef73280646c94c4c1b0eb86#diff-9f6c1905e08b768cabebaa8921124660L9-L10

Since this patch, bazel from nixpkgs use bazel default behavior which is to find for `python` in the current `PATH`.

However it seems that `bazel` is cleaning the environment before executing a few of the actions needed in rules_haskell. At least `ls_module`, as shown in the following bazel incantation: 

```
[nix-shell:~/tweag/rules_haskell]$ bazel test //tests/library-with-sysdeps:bin -s
INFO: Analyzed target //tests/library-with-sysdeps:bin (0 packages loaded, 0 targets configured).
INFO: Found 1 test target...
SUBCOMMAND: # //tests/library-with-sysdeps:library-with-sysdeps [action 'HaskellLinkDynamicLibrary tests/library-with-sysdeps/libHStestsZSlibrary-with-sysdepsZSlibrary-with-sysdeps-ghc8.6.5.so']
(cd /home/guillaume/.cache/bazel/_bazel_guillaume/80819d925a91d7b49ed839165b536ee7/execroot/rules_haskell && \
  exec env - \
    LANG=en_US.UTF-8 \
    LOCALE_ARCHIVE=external/glibc_locales/lib/locale/locale-archive \
  bazel-out/host/bin/external/rules_haskell/haskell/ghc_wrapper bazel-out/k8-fastbuild/bin/tests/library-with-sysdeps/compile_flags_library-with-sysdeps_HaskellLinkDynamicLibrary bazel-out/k8-fastbuild/bin/tests/library-with-sysdeps/extra_args_library-with-sysdeps_HaskellLinkDynamicLibrary.and.library-with-sysdeps.dynamic.manifest)
SUBCOMMAND: # //tests/library-with-sysdeps:library-with-sysdeps [action 'Action tests/library-with-sysdeps/exposed-modules-library-with-sysdeps']
[...]

PR #1396, opened 2020-07-16
Fix ls_modules.py for overlong ids.

robinp:
For some reason, long ids are broken to the next line, so the current
line parsing logic fails.

For a few minutes, I was detoured to think this is https://github.com/tweag/rules_haskell/issues/1067, but then noticed this is an actual python script error.

TESTED=manually

### DIFF
--- a/haskell/private/ls_modules.py
+++ b/haskell/private/ls_modules.py
@@ -29,7 +29,22 @@ results_file = sys.argv[6]
 with io.open(global_pkg_db_dump, "r", encoding='utf8') as f:
     names = [line.split()[1] for line in f if line.startswith("name:")]
     f.seek(0)
-    ids = [line.split()[1] for line in f if line.startswith("id:")]
+    # Indicated if value of sought key (here, id) was reflowed to the next
+    # line. Probably happens for long values.
+    reflowed = False
+    ids = []
+    for line in f:
+        if reflowed:
+            ids.append(line.strip())
+            reflowed = False
+            continue
+        if not line.startswith("id:"):
+            continue
+        parts = line.split()
+        if len(parts) == 1:
+            reflowed = True
+        else:
+            ids.append(parts[1])
 
     # A few sanity checks.
     assert len(names) == len(ids)


robinp:
As I see this piece is not tested yet, and I wouldn't want to spend time adding scaffolding. If you added the mundane pieces, I would love to.

aherrmann:
I don't think this requires a lot of scaffolding. Under what use-case did you encounter this issue? I.e. what package had a sufficiently long id to trigger the line wrapping?
A simple regression test could be to add an example under `tests/` that defines a target with a similarly lengthy package id and a dependency on it to trigger the issue.

robinp:
I see - it was `data-default-instances-containers`, which got pulled in somehow for GHC 8.8.3. I'm not sure what generates that global package database - I have both a `ghcWithPackages` with quite some packages, and `haskell_toolchain_library` entries generated for them to access.

How would I bring in a `haskell_toolchain_library` with a long enough name?

aherrmann:
Oh, sorry I misread the change and didn't realize this was about the global package db. Indeed, the global package db dump is generated in the toolchain rule using `ghc-pkg dump --global` and the way that `data-default-instances-containers` ends up in there is through `ghcWithPackages`.

In that case, one way to test this would be a script that calls `ls_modules` directly on an example db-dump that exhibits the line wrap and compares the output against an expected output.

This leads me to a follow-up question. Couldn't the `name:` field be wrapped as well? I recall this issue has been discussed in the past, see https://github.com/tweag/rules_haskell/pull/1169. As mentioned in that thread there is already an implementation that handles wrapping more generally in [`pkgdb_to_bzl.py`](https://github.com/tweag/rules_haskell/blob/eb16a5401770098a801775ea3a893b09cafe054c/haskell/private/pkgdb_to_bzl.py#L24-L40). `pkgdb_to_bzl.py` is called in a repository rule context, so sharing the code in a module is a bit tricky to setup. But, using the same code in both with comments pointing to the other would already be an improvement and sounds more robust than ad-hoc handling of the `id:` field only.

Infinisil:
Just encountered this today. And indeed, I believe all package conf files (and probably the database subsequently) have this line wrapping behavior. In nixpkgs I fixed this for the Haskell builder using an AWK expression that turns it into a more easily processable form, see https://github.com/NixOS/nixpkgs/pull/78738

aherrmann (approved):
The issue came up again in #1453 so better to merge this to at least have the issue fix. We can do the factoring out and testing in a follow-up.

Closes #1453

aherrmann:
@Mergifyio rebase

aherrmann:
@robinp Looks like there is a permissions issue. Neither mergify nor I can push the rebase of this branch. Could you please rebase on master and push?

aherrmann:
I took a stab at implementing https://github.com/tweag/rules_haskell/pull/1396#issuecomment-660958851 in https://github.com/tweag/rules_haskell/pull/1457. The module import in `pkgdb_to_bzl.py` in repository rule context turned out to work easier than I feared.

So, we can use #1457 instead of this PR.

PR #1407, opened 2020-08-11
docs: remove redundant pthread linker flag

aherrmann:
https://github.com/tweag/rules_haskell/pull/1390 documents that fully static linking requires `-optl-pthread`. However, that flag is already set by `rules_haskell` [during linking](https://github.com/tweag/rules_haskell/blob/a67be06b3df851a3a01aebbe42b25583044a6bba/haskell/private/actions/link.bzl#L119). ~This PR removes it.~ This PR removes the flag from the docs.

@lunaris I saw that you also [removed that flag](https://github.com/lunaris/minirepo/commit/feb51aa5faec8c61339cf7d7d8041f6f580c1948#diff-35b9415092675c95c3395c9a2db3b346L44) in `minirepo`. Can this be removed from the docs or am I missing something?

### DIFF
--- a/docs/haskell-use-cases.rst
+++ b/docs/haskell-use-cases.rst
@@ -712,8 +712,8 @@ expression that supplies appropriate ``cc`` and ``binutils`` derivations::
   )
 
 With the toolchain taken care of, you can then create fully-statically-linked
-binaries by passing the ``-optl-static`` and ``-optl-pthread`` flags as
-``compiler_flags`` to GHC, e.g. in ``haskell_binary``::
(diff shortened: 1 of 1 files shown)


lunaris:
I _think_ it can, yes -- I believe that this is because I'm using Musl though (it might be that with glibc you get symbol warnings/errors -- haven't tested I'm afraid).

aherrmann:
Sorry, that was badly phrased on my part. The PR is removing the flag from docs, since it's automatically passed by `rules_haskell` during linking already.

lunaris:
Ah apologies -- makes sense then!

Issue #1408, opened 2020-08-12
User interface for fully statically linked Haskell binaries


aherrmann:
#1390 added support for fully statically linked Haskell binaries. As of that implementation users need to configure the GHC toolchain for fully static linking by setting `static_runtime = True` and `fully_static_link = True` on the toolchain *and* they need to configure `compiler_flags = ["-optl-static"]` on the `haskell_binary` targets that they want to have fully statically linked. @mboes raised whether this flag is necessary and whether we could build on Bazel precedent and use the [`fully_static_link`](https://docs.bazel.build/versions/master/be/c-cpp.html#cc_binary.linkstatic) feature flag instead. I figured it's best to have that discussion on this issue tracker.

A related question was previously discussed to decide on the interface of the GHC toolchain, see https://github.com/tweag/rules_haskell/pull/1390#issuecomment-658877757. As I argued in that discussion it is a property of the GHC distribution whether it is capable of fully static linking, so it makes sense to have a corresponding switch on the GHC toolchain rule. The remaining question is whether and how users control fully static linking on individual targets. These are the options I see:

1. Always link fully statically when the toolchain is configured with `fully_static_link = True`. I.e. always pass `-optl-static` to GHC (or `--ghc-option=-optl-static` to Cabal).
2. Let the user configure fully static linking per target using Bazel's `fully_static_link` feature flag.
3. Let the user configure fully static linking per target by passing `-optl-static` to GHC (or `--ghc-option=-optl-static` to Cabal).

---

Here's my view on each of these options:
1. I think this would be the ideal option if it was feasible, but I don't think that it is feasible in practice at this point. E.g. I tried this approach on [`minirepo`](https://github.com/lunaris/minirepo/tree/58c3a67a09e97d22995380bb97f9e8cef6215fa9) but noticed that it then fails to build `@ghcide-exe//ghcide`. The reason is that it depends on `libtinfo` of `ncurses` which is only available dynamically in the `static-haskell-nix` configuration. Trying to switch to a static `libtinfo` fails due to missing symbols such as `__memcpy_chk`. IIUC this is means that `libtinfo` depends on `glibc` and is incompatible with `musl`. This indicates that it is necessary to give users per target control of whether they want to link fully statically or not and rules this option out.
[...]

PR #1410, opened 2020-08-14
Configure `-optl-static` via feature flag

aherrmann:
As discussed in #1408 this replaces the need to configure `compiler_flags` directly by setting `features = ["fully_static_link"]` to configure a fully statically linked binary. This does not yet address how to configure fully static linking for binaries generated by `stack_snapshot`. However, configuring the feature flag globally via the command-line flag `--features fully_static_link` will already produce fully statically linked `stack_snapshot` binaries.

### DIFF
--- a/docs/haskell-use-cases.rst
+++ b/docs/haskell-use-cases.rst
@@ -712,8 +712,7 @@ expression that supplies appropriate ``cc`` and ``binutils`` derivations::
   )
 
 With the toolchain taken care of, you can then create fully-statically-linked
-binaries by passing the ``-optl-static`` flag as ``compiler_flags`` to GHC,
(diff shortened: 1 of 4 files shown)


mboes:
How bizarre. The force-push notification above attributes it to @TerrorJack, when it's likely from @aherrmann...

aherrmann:
> How bizarre. The force-push notification above attributes it to @TerrorJack, when it's likely from @aherrmann...

It's the rebase from mergify. For some reason mergify attributes that to some random user (random as in I don't know what determines it). For a while it was @mboes, e.g. [here](https://github.com/tweag/rules_haskell/pull/1300), recently it switched to being @TerrorJack.

PR #1417, opened 2020-09-11
Update ghcide to 0.4.0

teto:
...can  I reference it from my bazel repository ? `bazel build @rules_haskell//:ghcide-exe` wont reference it. so it's here just for testing/documenting ?


aherrmann:
> ...can I reference it from my bazel repository ? `bazel build @rules_haskell//:ghcide-exe` wont reference it. so it's here just for testing/documenting ?

It is not possible to reference dependencies that were defined in another project's `WORKSPACE` file. Bazel effectively ignores `WORKSPACE` files of external workspaces. Users of rules_haskell that would like to build `ghcide` should instead define a corresponding `stack_snapshot` in their own project's `WORKSPACE` file as documented [here](https://rules-haskell.readthedocs.io/en/latest/haskell-use-cases.html#configuring-ide-integration-with-ghcide).

teto:
ok I'll upgrade it to 0.4 then

Issue #1418, opened 2020-09-12
haskell_cabal_library linker errors when using GHC 8.10


mboes:
**Describe the bug**
Building even a simple `haskell_cabal_library` fails with linker errors when using GHC 8.10. The same code built without Cabal does not lead to linker errors.

**To Reproduce**
Switch `rules_haskell` to GHC 8.10.1 or 8.10.2. Then,

```
$ bazel build @stackage//:void
[undefined reference to various RTS symbols errors]
```

**Expected behavior**
I expect the build to succeed with no linker errors.

**Environment**
 - OS name + version: NixOS 20.03 (using a Nixpkgs toolchain) or Debian buster (using GHC bindists).

**Additional context**
This bug has already been reported multiple times: see #1401 #1413 #1327. I'm creating a new issue to gather what we know in a single place, with a more focused issue description.

According to https://github.com/tweag/rules_haskell/issues/1327#issuecomment-664378765, this is a regression introduced by #1388, which is itself necessary on some systems when using GHC 8.8. Since we don't use GHC 8.10 during CI testing, we didn't catch this.

On a PIE-by-default systems like Debian Buster,
* building a `haskell_library` using GHC 8.8 works fine, with or without #1388,
* building a `haskell_cabal_library` using GHC 8.8 works fine, with or without #1388,
* building a `haskell_library` using GHC 8.10 works fine, with or without #1388,
* building a `haskell_cabal_library` using GHC 8.10 works but only without #1388.

The latter case also fails on NixOS with a Nixpkgs toolchain.

So what's still unclear is why the result matrix for `haskell_cabal_library` is different on GHC 8.8 and on GHC 8.10. It's likely that https://gitlab.haskell.org/ghc/ghc/-/issues/15319 has something to do with it.

What's also unclear is why for GHC 8.10, the result matrix is different for `haskell_library` and `haskell_cabal_library`. The former works after #1388 while the latter does not.

cc @patrickt @robinbb

robinbb:
I noticed during my investigation of this issue (on top of `github/semantic` repo) that the `ghc-cabal` binary shipped with the GHC bindist was not linked with the GHC-supplied 'libffi'. I wonder if this is related; perhaps the bindists are improperly built?

mboes:
Additional findings:
[...]

PR #1421, opened 2020-09-15
doc(.github): Add a note of 8.10 linking issues to bug template

Profpatsch:
Until we solve the issues with 8.10, we can avoid some duplicate bug reports by linking to the WIP issue.


mboes:
#1418 fixed by #1424, so this is no longer necessary. But it was a good idea.

PR #1429, opened 2020-09-19
Check each file we might overwrite or clash with.

philderbeast:
This change is more explicit in its messaging about each file that exists and that we'd be overwriting. It avoids false positive clashes with directories such as the common `build` directory in many repos.

```
> ./rules-haskell-start
INFO: Creating a WORKSPACE file based on GHC bindists. If you want to use a nix-based setup (e.g. on NixOS), call with `--use-nix`. See `--help` for more info.
ERROR: The current directory already has a .bazelrc file we don't want to overwrite.
```

### DIFF
--- a/start
+++ b/start
@@ -48,19 +48,41 @@ parse_args () {
     esac
 }
 
-check_clash () {
-    if [ -f "$1" ] # directories don't clash, we're only writing files
+check_dir () {
+    if [ -d "$1" ] # we can't write a file when a directory of that name exists
     then
-        echo "ERROR: The current directory already has a $1 file we don't want to overwrite." >&2
+        echo "STOP: There's a directory named $1 but we want to write a file with the same name. Please delete or rename the $1 directory." >&2
+        exit 1
+    fi
+}
+
+check_alt () {
+    check_dir $2
+    if [ -f "$1" ]
+    then
+        echo "STOP: There's a $1 file but we intend to write a $2 file. When both exist, bazel will pick the one with the .bazel extension, ignoring the other one. Please delete the $1 file." >&2
+        exit 1
+    fi
+}
+
+check_clash () {
+    check_dir $1
+    if [ -f "$1" ]
+    then
+        echo "STOP: The current directory already has a $1 file and we don't want to overwrite it." >&2
         exit 1
     fi
 }
 
 check_files_dont_exist () {
-    # Writing BUILD.bazel does not overwrite a BUILD file but include it anyway.
-    # If we're uncomfortable with the slightly off clash message then write
-    # a separate test and message for that case.
-    for clash in .bazelrc WORKSPACE BUILD BUILD.bazel Example.hs
+    # A BUILD.bazel file takes precedence over a BUILD file and likewise
+    # a WORKSPACE.bazel file takes precedence over a WORKSPACE file. We write
+    # BUILD.bazel and WORKSPACE files. Seems an odd mix.
+    check_alt WORKSPACE.bazel WORKSPACE
+    check_alt BUILD BUILD.bazel
+    check_alt zlib.BUILD zlib.BUILD.bazel
+
+    for clash in .bazelrc WORKSPACE BUILD.bazel zlib.BUILD.bazel Example.hs
     do
         check_clash $clash
     done
--- a/start
+++ b/start
@@ -77,7 +77,9 @@ check_clash () {
 check_files_dont_exist () {
     # A BUILD.bazel file takes precedence over a BUILD file and likewise
     # a WORKSPACE.bazel file takes precedence over a WORKSPACE file. We write
-    # BUILD.bazel and WORKSPACE files. Seems an odd mix.
+    # BUILD.bazel and WORKSPACE files.
+    # Some Bazel tooling may fail on WORKSPACE.bazel files,
+    # e.g. https://github.com/bazelbuild/bazel-gazelle/issues/678
     check_alt WORKSPACE.bazel WORKSPACE
     check_alt BUILD BUILD.bazel
 
--- a/start
+++ b/start
@@ -57,7 +57,7 @@ check_dir () {
 }
 
 check_alt () {
-    check_dir $2
+    check_dir "$2"
     if [ -f "$1" ]
     then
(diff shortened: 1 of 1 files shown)


start
```diff
@@ -48,14 +48,24 @@ parse_args () {
...
+        echo "ERROR: The current directory already has a $1 file we don't want to overwrite." >&2
         exit 1
     fi
 }
 
+check_files_dont_exist () {
+    # Writing BUILD.bazel does not overwrite a BUILD file but include it anyway.
+    # If we're uncomfortable with the slightly off clash message then write
+    # a separate test and message for that case.
+    for clash in .bazelrc WORKSPACE BUILD BUILD.bazel Example.hs
```
aherrmann, line 63: It seems this bit has become out of sync with which files this script actually generates. E.g. we're not checking for files like `zlib.BUILD.bazel`. Though, I'm not sure what's the best option to avoid that kind of going out of sync in the future.

start
```diff
@@ -48,14 +48,24 @@ parse_args () {
...
-        echo "Current directory already has WORKSPACE and/or BUILD and/or BazelExample.hs files." >&2
+        echo "ERROR: The current directory already has a $1 file we don't want to overwrite." >&2
         exit 1
     fi
 }
 
+check_files_dont_exist () {
+    # Writing BUILD.bazel does not overwrite a BUILD file but include it anyway.
+    # If we're uncomfortable with the slightly off clash message then write
+    # a separate test and message for that case.
```
aherrmann, line 62: I think a separate error message would be worth the effort. The issue is that `BUILD.bazel` takes precedence and `bazel` would ignore a pre-existing `BUILD` file which may induce a fair bit of head-scratching if a user expects `bazel` to read that file.

> It avoids false positive clashes with directories such as the common build directory in many repos.

That check could also be more specific and use `[ -f` to avoid this edge-case.

For completeness we should also check for [`WORKSPACE.bazel`](https://docs.bazel.build/versions/master/build-ref.html#workspace) which would take precedence over the `WORKSPACE` file that this script generates.

start
```diff
@@ -48,14 +48,24 @@ parse_args () {
     esac
 }
 
-check_files_dont_exist () {
-    if [ -e WORKSPACE ] || [ -e BUILD ] || [ -e BazelExample.hs ]
+check_clash () {
+    if [ -f "$1" ] # directories don't clash, we're only writing files
```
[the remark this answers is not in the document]
philderbeast: I've added a commit checking -d first and then -f, directory first and then file.

start
philderbeast: I've added checks for files that can have alternate names.

start
philderbeast: I've added a check for `zlib.BUILD.bazel` and its alternative name `zlib.BUILD`.

philderbeast:
Here are the messages the script writes now:

```
> ./start
INFO: Creating a WORKSPACE file based on GHC bindists. If you want to use a nix-based setup (e.g. on NixOS), call with `--use-nix`. See `--help` for more info.
STOP: There's a WORKSPACE.bazel file but we intend to write a WORKSPACE file. When both exist, bazel will pick the one with the .bazel extension, ignoring the other one. Please delete the WORKSPACE.bazel file.
> ./start
...
STOP: There's a BUILD file but we intend to write a BUILD.bazel file. When both exist, bazel will pick the one with the .bazel extension, ignoring the other one. Please delete the BUILD file.
> ./start
...
STOP: The current directory already has a .bazelrc file and we don't want to overwrite it.
> ./start
...
STOP: The current directory already has a WORKSPACE file and we don't want to overwrite it.
> ./start
...
STOP: The current directory already has a BUILD.bazel file and we don't want to overwrite it.
> ./start
...
STOP: The current directory already has a zlib.BUILD.bazel file and we don't want to overwrite it.
> ./start
...
STOP: There's a zlib.BUILD file but we intend to write a zlib.BUILD.bazel file. When both exist, bazel will pick the one with the .bazel extension, ignoring the other one. Please delete the zlib.BUILD file.
> ./start
...
STOP: The current directory already has a Example.hs file and we don't want to overwrite it.
> ./start
INFO: Creating a WORKSPACE file based on GHC bindists. If you want to use a nix-based setup (e.g. on NixOS), call with `--use-nix`. See `--help` for more info.
Creating zlib.BUILD.bazel
Creating WORKSPACE
Creating .bazelrc
Creating BUILD.bazel
Creating Example.hs
WORKSPACE and initial BUILD files created. To run Bazel and build the example:

    $ bazel run //:example
```

start
```diff
@@ -48,14 +48,45 @@ parse_args () {
...
-        echo "Current directory already has WORKSPACE and/or BUILD and/or BazelExample.hs files." >&2
+        echo "STOP: The current directory already has a $1 file and we don't want to overwrite it." >&2
         exit 1
     fi
 }
 
+check_files_dont_exist () {
+    # A BUILD.bazel file takes precedence over a BUILD file and likewise
+    # a WORKSPACE.bazel file takes precedence over a WORKSPACE file. We write
+    # BUILD.bazel and WORKSPACE files. Seems an odd mix.
```
aherrmann, line 80:
```suggestion
    # BUILD.bazel and WORKSPACE files.
    # Some Bazel tooling may fail on WORKSPACE.bazel files,
    # e.g. https://github.com/bazelbuild/bazel-gazelle/issues/678
```
Clashes with `BUILD` are relatively common, while clashes with `WORKSPACE` aren't. I assume that is why most Bazel projects still use `WORKSPACE`. It seems that some Bazel tooling also has trouble with `WORKSPACE.bazel`, e.g. the cited issue. So, using `WORKSPACE` is the safer option for now.

aherrmann (approved):
@philderbeast Sorry for the long delay on this one.

Thank you for addressing the review comments!

I've rebased this on latest master and made a few small changes.
1) The check on `zlib.BUILD.bazel` is not needed. In that case Bazel will not look for a file with the `.bazel` extension.
2) Use `[ -e` in `check_clash`. Any kind of file with the same name would cause an error. It's unlikely that we'll encounter a named pipe or the like named `.bazelrc` etc. but it doesn't hurt to check.
3) The comment about WORSPACE.bazel vs WORKSPACE

Issue #1430, opened 2020-09-19
Docs on picking a compiler, 'haskell_register_ghc_bindists' unknown symbol.


philderbeast:
**Describe the bug**
I'm following along with the docs at use cases and [picking a compiler](https://rules-haskell.readthedocs.io/en/latest/haskell-use-cases.html#picking-a-compiler) is failing.

**To Reproduce**
Run the start script and then add these lines from the docs to the end of the `WORKSPACE` file.

```
# WORKSPACE
load(
    "@rules_haskell//haskell:defs.bzl",
    "haskell_register_ghc_bindists",
)

haskell_register_ghc_bindists(
    version = "8.2.2", # Any GHC version
)
```

```
> bazel run //:example
ERROR: /.../WORKSPACE:29:1: file '@rules_haskell//haskell:defs.bzl'
does not contain symbol 'haskell_register_ghc_bindists'
```

**Environment**
```
> sw_vers
ProductName:	Mac OS X
ProductVersion:	10.13.6
BuildVersion:	17G14019
> bazel --version
bazel 3.3.1
```

philderbeast:
From the docs for [stack_snapshot](https://api.haskell.build/haskell/cabal.html#stack_snapshot) it looks like there's another way to pick the compiler.

> NOTE: Make sure your GHC version matches the version expected by the snapshot. E.g. if you pass snapshot = "lts-13.15", make sure you use GHC 8.6.4 (e.g. by invoking rules_haskell_toolchains(version="8.6.4")).

mboes:
Thanks! The documentation was out-of-date.

PR #1431, opened 2020-09-19
Load ghc_bindists from ghc_bindist.bzl.

philderbeast:
See #1430.


mboes:
Fixed on master. Well spotted!

PR #1434, opened 2020-10-07
Lcov support for coverage

guibou:
This is a work in progress PR about improving code coverage.

The main feature of this branch is that it allows the generation of lcov coverage file, instead of hpc results. lcov is compatible with more coverage tools / hosted coverage service than hpc.

First commit: *Coverage: add provider for coverage files* can be merged as it is now. It just tell the coverage process to generate the `COVERAGE_FILE_OUTPUT` variable and read coverage result from that.
Second commit: 
*Example usage for the lcov merger* introduce `hpc2lcov`, a tool which converts hpc files to lcov. It also show how the coverage results can be post-processed and gathered using the bazel coverage API. We may imagine refactoring our current coverage implementation using a similar solution. For example, we are generating a wrapper, when actually the bazel coverage API may do that for us.
Third commit: *WIP: tentative to move the hpc2lcov utilities in bazel* is a work in progress where I'm trying to build hpc2lcov (which as an haskell binary) with bazel instead of nix.


guibou:
This (once rebased on top of #1594) is a partial implementation for lcov support (#1434{

guibou:
I have no feedback regarding this MR since 3 years. I'm not using rules_haskell anymore.

I'm closing. Feel free to ping me if you want to restart this work and want to have more details.

PR #1437, opened 2020-10-23
Migrate to the modern linker input API.

benjaminp:
See https://github.com/bazelbuild/bazel/issues/10860.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -3,6 +3,8 @@
 # Global Configuration
 # --------------------
 
+common --incompatible_require_linker_input_cc_api
+
 # test environment does not propagate locales by default some tests reads files
 # written in UTF8, we need to propagate the correct environment variables, such
 # as LOCALE_ARCHIVE We also need to setup an utf8 locale
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -498,18 +498,23 @@ def _haskell_cabal_library_impl(ctx):
         requested_features = ctx.features,
         unsupported_features = ctx.disabled_features,
     )
-    library_to_link = cc_common.create_library_to_link(
-        actions = ctx.actions,
-        feature_configuration = feature_configuration,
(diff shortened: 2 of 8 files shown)


mboes (approved):

mboes:
LGTM, but assigning @aherrmann for an additional review.

aherrmann (commented):
Thank you!

Should [this instance](https://github.com/benjaminp/rules_haskell/blob/318c530609fed1d7629746f9b705eed65e8e7abf/haskell/private/cc_libraries.bzl#L307) also be changed or is there a reason to keep using `libraries_to_link` in that case?

Did you try enabling `--incompatible_require_linker_input_cc_api` in [`.bazelrc`](https://github.com/benjaminp/rules_haskell/blob/318c530609fed1d7629746f9b705eed65e8e7abf/.bazelrc#L58)? Would be good to check against regressions on this on CI if possible.

benjaminp:
> Thank you!

Thank you for the review.

> 
> Should [this instance](https://github.com/benjaminp/rules_haskell/blob/318c530609fed1d7629746f9b705eed65e8e7abf/haskell/private/cc_libraries.bzl#L307) also be changed or is there a reason to keep using `libraries_to_link` in that case?

Yes, indeed! This makes me realize I accidentally pushed an old version of my branch that didn't fix that case through a git snafu. Should be okay now.

> 
> Did you try enabling `--incompatible_require_linker_input_cc_api` in [`.bazelrc`](https://github.com/benjaminp/rules_haskell/blob/318c530609fed1d7629746f9b705eed65e8e7abf/.bazelrc#L58)? Would be good to check against regressions on this on CI if possible.

Yes, that's how I was testing locally. I committed this to the branch now, too.

aherrmann (approved):
Thank you!

aherrmann:
@Mergifyio rebase

PR #1441, opened 2020-11-11
Upgrade to rules_nodejs 1.7.0

Wyverald:
rules_nodejs pre 1.6.0 had invalid escape sequences in its Starlark code, which was silently ignored but now throws an error (due to https://github.com/bazelbuild/bazel/commit/73402fa4aa5b9de46c9a4042b75e6fb332ad4a7f).


aherrmann (approved):
Thanks for the update!

PR #1446, opened 2020-11-17
Fix string escape in BUILD file

meteorcloudy:
Following up https://github.com/tweag/rules_haskell/pull/1441
Fixes required by bazelbuild/bazel@73402fa


aherrmann (approved):
Thank you!

PR #1449, opened 2020-11-26
Fix a regression of dylibs missing in runfiles

aherrmann:
https://github.com/tweag/rules_haskell/pull/1256 introduced a regression whereby `.so` libraries required by a dynamically linked binary were no longer contained in the runfiles. This PR fixes that. Additionally, this PR adds regression tests.


facundominguez (approved):
Works for me.

PR #1450, opened 2020-11-29
Add documentation use case for pre-processors

mboes:
Document how to use tools like `hspec-discover`, which are common.

cc @codygman


aherrmann (approved):
LGTM, thank you!

I've fixed the missing link for hspec-discover.

PR #1452, opened 2020-12-02
Fix broken links to api.haskell.build

avdv:
Also, use https for all links since, well, nobody uses http anymore.


aherrmann (approved):
Thank you!

PR #1440, opened 2020-11-09
Fix stack_snapshot pinning unsupported archive format

aherrmann:
Closes #1392 

stack_snapshot pinning needs to download the tarballs of archive dependencies and discover the location of the Cabal file within the tarball. So far this was done with a heuristic implemented in Starlark. Unfortunately, Starlark's IO API is very limited and it was not possible to discover Cabal files that are deeper than in the archive toplevel. This affects projects such as haskell-language-server, funflow, or hmatrix.

This PR changes this to shell out to `find` (`dir` on Windows) to recursively search for the relevant Cabal file in the unpacked tarball.

This PR does not add a test case. The issue occurs when running `bazel run @stackage-unpinned//:pin` which is difficult to do within a Bazel test-case. However, I have manually tested `bazel run @hmatrix-unpinned//:pin` on Linux, and I have tested the `dir` command on Windows in isolation. I am currently unable to test `bazel run @hmatrix-unpinned//:pin` on Windows due to the following error
```
Exception while reading snapshot from lts-15.4:
HttpExceptionRequest Request {
  host                 = "raw.githubusercontent.com"
  port                 = 443
  secure               = True
  requestHeaders       = [("User-Agent","Haskell pantry package")]
  path                 = "/commercialhaskell/stackage-snapshots/master/lts/15/4.yaml"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 (InternalException (HandshakeFailed (Error_Protocol ("certificate has unknown CA",True,UnknownCa))))
```

When we update from ghcide to haskell-language-server within rules_haskell in the future then this could serve as a test case for this feature.

PR #1454, opened 2020-12-03
Fixup for #1440

aherrmann:
The strip-prefix generated in #1440 still contains the filename of the Cabal file. This PR fixes this by applying `paths.dirname`.

I've been able to test pinning on Windows meanwhile and found that `dir` might accidentally execute a Git-bash or msys2 introduced `dir` alias for `ls` instead of `cmd`'s `dir`. This PR calls `cmd /c` to ensure that we're executing the correct `dir` on Windows. `paths.relativize` also doesn't handle Windows path separators, so we have to convert back and forth between `/` and `\`.

I didn't find a good way to test this in a regular `bazel test`. So, I added a test to `tests/RunTests.hs` and the various CI pipelines that tests pinning and whether the generated target builds. `hspec` turned out to be a package with nested packages that doesn't take too long to build.


teto:
tested both pinning and building. Mission accomplished it was successful thanks

mboes (approved):

PR #1455, opened 2020-12-04
stack unpack with versioned packages

### LINKED ISSUE #1447: Versioned packages (from hackage) don't work with stack_snapshot
**Describe the bug**
Doing
```
stack_snapshot(
    name = "stackage",
    packages = [
        "base"
        "haxl-2.3.0.0",
    ],
    snapshot = "lts-16.11",
)
```
Used to work in v0.12 and before, it doesn't work in 0.13 nor master.

I think the problem is that it runs `stack --resolver lts-16.11 unpack base haxl` instead of `stack --resolver lts-16.11 unpack base haxl-2.3.0.0` (note the version number)

**Expected behavior**
It pulls Haxl from hackage and it runs

**Environment**
 - OS name + version: Mac OS Catalina
 - Bazel version: 3.3.1
 - Version of the rules: master

aherrmann:
Closes #1447 

Calls `stack --resolver lts-16.11 unpack base haxl-2.3.0.0` instead of `stack --resolver lts-16.11 unpack base haxl`. Otherwise, `stack` will fail to download Hackage packages that are not part of the specified snapshot.

An unfortunate side effect is that fetches of Stackage packages become less reproducible as `stack` will now fetch the latest revision from Hackage instead of the revision that is pinned in the specified snapshot. However, this shouldn't break any builds in practice and a fully reproducible `stack_snapshot` is available via pinning.

Adds a non-Stackage dependency (`optparse-helper`) to the example as a regression test.

cc @jjant

### DIFF
--- a/examples/WORKSPACE
+++ b/examples/WORKSPACE
@@ -105,6 +105,8 @@ stack_snapshot(
         "optparse-applicative",
         "text",
         "text-show",
+        # Demonstrates a Hackage dependency that is not included in the Snapshot.
+        "optparse-helper-0.2.1.1",
     ],
     snapshot = "lts-14.0",
     # This example uses an unpinned version of stack_snapshot, meaning that stack is invoked on every build.
--- a/examples/cat_hs/lib/args/BUILD.bazel
+++ b/examples/cat_hs/lib/args/BUILD.bazel
@@ -11,6 +11,7 @@ haskell_library(
     deps = [
         "@stackage//:base",
         "@stackage//:optparse-applicative",
+        "@stackage//:optparse-helper",
     ],
 )
 
--- a/examples/cat_hs/lib/args/src/Args.hs
+++ b/examples/cat_hs/lib/args/src/Args.hs
@@ -11,6 +11,7 @@ module Args
   ) where
 
 import Options.Applicative
+import Options.Applicative.Helper (fpDesc, infoHelper)
 
 
 -- | A file argument.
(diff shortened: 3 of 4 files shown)


mboes:
I wonder why `stack --resolver lts-16.11 unpack haxl` doesn't do what we expect? Sounds like an upstream bug to me, which we should file in addition to working around it here.

mboes:
Oh I see - I guess this happens when the snapshot does not pin the version number, i.e. when the package is not in the snapshot.

mboes (commented):
If I understand correctly, with this change, the resolver is largely ignored. It's still used for core packages, but the core packages are pinned by the version of GHC anyways. So do we still need to pass the `--resolver` argument to unpack?

aherrmann:
@m Indeed, the resolver is largely ignored in this case. Thinking more about this, this is actually an issue. Take a custom snapshot like the following
```
resolver: lts-14.0
packages:
  - github: hspec/hspec
    commit: "9f3f4c38952f526701a67b6e26336a3a5aec0e89"
    sha256: "5233805b6370f488dd95149644d6689fde8e5dbddb3de3becb915409920ed2a0"
    subdirs:
      - .
      - hspec-contrib
      - hspec-core
      - hspec-discover
```
`stack --resolver ... unpack hspec-core` will fetch the version specified by the local snapshot, but `stack --resolver ... unpack hspec-core-2.7.1` will fetch the version from Hackage, which is different in this case.

So, we can only add the version suffix to Hackage dependencies not to snapshot dependencies.

mboes (approved):

PR #1460, opened 2020-12-27
Move macOS CI to GitHub Actions

mboes:
GitHub Actions has macOS runners available now. So we no longer need
to pay for an expensive macOS plan from CircleCI. The new setup using
GitHub actions is also simpler, because we don't need to save and
restore a disk cache anymore. Instead, we have BuildBuddy as the
remote cache. Where the old CircleCI jobs were clocking at 1h15 using
the disk cache, we're now at 16 minutes using the remote cache.

cc @siggisim


mboes:
The CircleCI error is expected, since we are removing the configuration file for it in this PR.

siggisim:
This looks great, let me know if I can be helpful!

mboes:
oh hey @siggisim! Thanks for stopping by. I was wondering about one thing. BuildBuddy has a cache view, which is very helpful, e.g. [here](https://app.buildbuddy.io/invocation/c656f749-c8b2-49db-bc00-3f765836ece4#cache). But does it also provide any tools to diagnose cache misses? The ideal would be to be able to download a list of actions, identified by their outputs, for which there was a cache miss.

siggisim:
Hey @mboes! At the moment we only have the high level cache overview stats. We're working on a couple of different methods to help make debugging cache hits easier.

In the meantime, [this guide](https://docs.bazel.build/versions/master/remote-execution-caching-debug.html) shows you how to get the execution logs from two runs, sort them, and compare which inputs changed.

.github/workflows/workflow.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+jobs:
+  test-macos-nixpkgs:
+    name: Build & Test - macOS (Nixpkgs)
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: cachix/install-nix-action@v12
+        with:
+          nix_path: nixpkgs=./nixpkgs/default.nix
```
aherrmann, line 12: IIUC this no longer enables Cachix caching itself. Looking at the docs it seems reasonably straight forward, assuming that the secrets are available to GH actions: https://nix.dev/tutorials/continuous-integration-github-actions.html.

aherrmann (commented):
Awesome, thank you for implementing this!

Just a few comments.

.bazelrc
```diff
@@ -52,8 +52,10 @@ build:ci-linux-bindist --repository_cache ~/repo_cache
 #   strict action env to avoid changes in `PATH` invalidating the cache.
 build:ci-linux-bindist --experimental_strict_action_env
 
-build:ci-macos-bindist --disk_cache=~/.cache/bazel/
-build:ci-macos-nixpkgs --disk_cache=~/.cache/bazel/
+build:ci-macos --bes_results_url=https://app.buildbuddy.io/invocation/
+build:ci-macos --bes_backend=grpcs://cloud.buildbuddy.io
+build:ci-macos --remote_cache=grpcs://cloud.buildbuddy.io
+build:ci-macos --remote_timeout=3600
```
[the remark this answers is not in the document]
siggisim: Not strictly required, but this is the timeout used by bazel for all remote execution and cache calls (including cache uploads and downloads). The default value of 60s can be hit when working with larger artifacts, and the error message in that case can be hard to interpret. 

The bazel project itself uses [600](https://github.com/bazelbuild/bazel/blob/master/.bazelrc#L7) seconds - which might be a more reasonable default.

.github/workflows/workflow.yaml
mboes: I'm not sure Cachix is necessary, since the official binary cache has a 100% cache hit rate for all the derivations in this PR.

.bazelrc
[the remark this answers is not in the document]
aherrmann: @siggisim Thanks for clarifying!

.github/workflows/workflow.yaml
aherrmann: Sounds good, we can still add it later if we need custom overrides at some point.

aherrmann:
I've also pushed an update to the mergify config. I tested it in the mergify config editor.

aherrmann (approved):

aherrmann:
One of the new GH actions [failed on a rate limit](https://github.com/tweag/rules_haskell/pull/1460/checks?check_run_id=1649282419#step:5:312)
```
ERROR: Traceback (most recent call last):
	File "/private/var/tmp/_bazel_runner/d1e3bdaa0c37ce8d00ecff62b05849c3/external/rules_haskell/haskell/cabal.bzl", line 1670
		_pin_packages(repository_ctx, <1 more arguments>)
	File "/private/var/tmp/_bazel_runner/d1e3bdaa0c37ce8d00ecff62b05849c3/external/rules_haskell/haskell/cabal.bzl", line 1191, in _pin_packages
		repository_ctx.download(<3 more arguments>)
java.io.IOException: Error downloading [https://api.github.com/repos/commercialhaskell/all-cabal-hashes/git/ref/heads/hackage] to /private/var/tmp/_bazel_runner/d1e3bdaa0c37ce8d00ecff62b05849c3/external/stackage-pinning-test-unpinned/all-cabal-hashes-hackage.json: GET returned 403 rate limit exceeded
```
I don't think I've seen this on CircleCI before. Is there a special limit for downloading from github through github actions?

Enabling GH actions disk cache together with Bazel repository cache may help avoid this.

mboes:
That's unfortunate. But I'm guessing that rate limiting applies to any machine.  The error you quote is in the unpinning test, which by design does network requests, right?

aherrmann:
> The error you quote is in the unpinning test, which by design does network requests, right?

Yes, that's right. Though the pinned ones also [download the Cabal files](https://github.com/tweag/rules_haskell/blob/d58e2318e84f0280002ae95569fa09f593c0482f/haskell/cabal.bzl#L1319-L1325) from GH. Presumably this contributes to the access count and might push it over the limit together with the unpinned test. With a repository cache enabled those pinned package Cabal files would be cached already in most cases.

mboes:
Is a repository cache something it makes sense for BuildBuddy to also provide (now or in the future)? The alternative would be to use GitHub's cache restore functionality, with all the downsides that has (manually keying an immutable cache correctly is a source of much brittleness).

aherrmann:
> Is a repository cache something it makes sense for BuildBuddy to also provide (now or in the future)? The alternative would be to use GitHub's cache restore functionality, with all the downsides that has (manually keying an immutable cache correctly is a source of much brittleness).

That would be a great feature for BuildBuddy or in general any remote cache to provide. However, the Bazel [feature request](https://github.com/bazelbuild/bazel/issues/6359) for a remote repository cache is still open. It looks like the flag [`--experimental_remote_downloader`](https://github.com/bazelbuild/bazel/pull/10622) enables this, though I'm not sure if this requires dedicated support from the remote cache.

Re GH cache, a cache setup similar to what we had with CircleCI can be modeled using [`restore-keys`](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key). An example of such a config is [here](https://github.com/aherrmann/example_servant_elm_bazel/blob/4288c1a4a21121864df883e38254074d11bb7469/.github/workflows/ci.yml#L16-L22).

siggisim:
We have "alpha" support for --experimental_remote_downloader, but its current implementation in bazel has quite a few limitations - I wouldn't recommend using it for anything mission critical.

Ultimately in order for the repo cache to be effective, it should be as close to the build machine as possible. Its main benefit comes from reducing the network bottleneck - so using a "remote" repository cache kind of defeats the purpose unless the remote cache is co-located or close to your build machines.

In our CI, we use GitHub's caching functionality:
https://github.com/buildbuddy-io/buildbuddy/blob/master/.github/workflows/main.yaml#L19

And then specify a repo cache directory in our .bazelrc:
https://github.com/buildbuddy-io/buildbuddy/blob/master/.bazelrc#L43

Just using a local repository cache is less brittle than using a full disk cache, because it's [designed](https://docs.bazel.build/versions/master/guide.html#the-repository-cache) to be used across workspaces and will only be pulled from if the download request has a SHA256 specified.

mboes:
@siggisim Interesting. Like the CircleCI cache, AFAIU GitHub's cache is immutable. Unlike the CircleCI cache, which appears to be forever, the [GitHub cache is short-lived](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy) (7 days if no cache hits during that time). So I guess you assume that every once in a while the cache will be dropped, for reasons, and therefore there will be neither unbounded growth of the repository cache (due to keeping only repositories forever in the cache)? Dropping the cache every once in a while is also crucial to keeping the cache effective. Since it's immutable, if the cache was never dropped, new repositories would never make it into the cache.

.github/workflows/workflow.yaml
```diff
@@ -7,6 +7,11 @@ jobs:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v2
+      - name: Mount Bazel cache
+        uses: actions/cache@v1
+        with:
+          path: ~/repo-cache
+          key: repo-cache-${{ secrets.REPO_CACHE_VERSION }}
```
aherrmann, line 14: Why a secret instead of a suffix configured in the `workflow.yaml`? IIUC this will need to be updated manually whenever the external dependencies are changed, so might just as well update it in the corresponding PR.

Also, why rely on a manual version bump, that is likely to be forgotten, instead of using something like [`hashFiles`](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#example-using-multiple-restore-keys) or commit sha with fall-through?

mboes: Because this is what is recommended by GitHub staff: https://github.community/t/how-to-clear-cache-in-github-actions/129038/5.

We could spend brain cycles working out exactly which files should go into `hashFiles()`, but even if we get that correct now we would have to make sure that list stays correct later. Hence not obviating the need for a manual mechanism to purge the cache, as above. Setting a cache version number is expedient and I submit sufficient given the short-lived nature of caches on GitHub anyways (7 days).

aherrmann: Fair enough, though the way I read that thread it's not a direct recommendation of this approach and more a mention that it exists. I'm all on-board for minimizing brain cycles on this aspect. But, I think that speaks in favor of a simple suffix in the code instead of requiring an out-of-band change through GH secrets. Case in point, I myself don't see the corresponding secret in the repository's secrets tab, so there's already some permissions issue to debug, and external contributors won't be able to access it for sure. I've opened a PR here: https://github.com/tweag/rules_haskell/pull/1472

mboes: FWIW you do have access to that secret. It just doesn't exist right now. I added it then removed it as an experiment.

PR #1462, opened 2020-12-29
Make macOS bindist configure more deterministic

### LINKED ISSUE #1461: Bindist builds on macOS are non-deterministic
**Describe the bug**
Remote caching of builds on macOS that use a GHC sourced from a bindist is ineffective. Cache misses for the exact same working tree can approach 30% when building the rules_haskell test suite.

**To Reproduce**

```
$ bazel build //tests/... --disk_cache /tmp/disk_cache
$ bazel clean --expunge
$ bazel build //tests/... --disk_cache /tmp/disk_cache
```

Notice the amount of code being rebuilt rather than fetched from the disk cache. This was first noticed when using BuildBuddy as the remote cache. See e.g. [this invocation](https://app.buildbuddy.io/invocation/77e259ff-fd39-45bb-9212-3d656a4d15df#cache), which is only partially cached despite being a rerun of a previous CI build job.

**Expected behavior**
In the presence of a remote cache, rebuilding all tests after expunging should lead to 100% cache hit rate.

**Environment**
 - OS name + version: macOS
 - Bazel version: 3.3.1

**Additional context**
The problem appears to be that many targets depend on `libHSrts.a` or `libCffi.a`. These static libraries come from the bindist. However, the `./configure; make install` of the bindist seems to modify timestamps in these libraries. As a result, their SHA256 hash is different each time, which has a knock-on effect on the hashes of downstream targets.

Related: https://github.com/bazelbuild/bazel/issues/10886.

mboes:
GHC from a bindist requires running `./configure` and then `make
install`. One would assume that the result is always the same, in
particular if we run this twice on the same host with the same
environment variables. However, this is not the case. See #1461.

The fix is to set an environment variable telling `ar` and friends to
not futz around with timestamps in static library archives. This
environment variable is only necessary and only has an effect on
macOS, but there's no harm in setting it always. If the bindists stop
having a configure stage, or start using `libtool` instead of `ar` and
`ranlib`, then the fix will better be pushed within the bindists
themselves, using the `-D` flag to `libtool`, introduced in XCode
10.2.

Fixes #1461.


mboes:
With this change, on GitHub Actions the build goes down from 36 minutes to less than 13 minutes. With a 99.9% cache hit rate on BuildBuddy.

aherrmann (approved):
Great, thanks for tracking this down!

aherrmann:
This makes me think that we should also be passing `D` to `ar` in the [static link actions](https://github.com/tweag/rules_haskell/blob/3987c494f7a3b72805855246a844a28880dd0b15/haskell/private/actions/link.bzl#L277-L299).

mboes:
If we used `libtool` instead of `ar` on macOS, then that would be possible. But I believe `ar` does not support the `-D` flag. In fact Bazel sometimes picks `libtool` over `ar`, which is deprecated on macOS. But we override that [here](https://github.com/tweag/rules_haskell/blob/master/haskell/cc.bzl#L125-L131). Perhaps migrating to `libtool` completely is what we should be aiming for. Or better yet, maybe it's possible now to leverage the CC toolchain API to give us a linking action that works, without us having to hardcode either `ar` or `libtool`.

aherrmann:
@mboes Good point about the CC Starlark API. I've opened #1469 to track this.

PR #1464, opened 2020-12-31
Move rules_haskell workspace to GHC 8.10.3

mboes:
We currently use 8.8.3 in the workspace. This moves us to 8.10.3.
Thanks to the improvements to Windows support, we can use the same
version of GHC across all platforms. As part of the move, we drop
support for 8.10.1 and 8.10.2, for reasons explained in the commit
messages, and add proper support for 8.10.3 on Linux.


haskell/ghc_bindist.bzl
```diff
@@ -373,10 +373,14 @@ haskell_toolchain(
...
     if os == "windows":
         # These libraries cause linking errors on Windows when linking
-        # pthreads, due to libwinpthread-1.dll not being loaded.
-        execute_or_fail_loudly(ctx, ["rm", "mingw/lib/gcc/x86_64-w64-mingw32/7.2.0/libstdc++.dll.a"])
-        execute_or_fail_loudly(ctx, ["rm", "mingw/x86_64-w64-mingw32/lib/libpthread.dll.a"])
-        execute_or_fail_loudly(ctx, ["rm", "mingw/x86_64-w64-mingw32/lib/libwinpthread.dll.a"])
+        # pthreads, due to libwinpthread-1.dll not being loaded. It's
+        # hard to guesss the paths of these libraries, and we can't
+        # assume `find` is available, so we delete the files on
+        # a best-effort basis (no error if we can't find them).
```
aherrmann, line 379: We could use `dir` instead of `find` similar to how it's done [here](https://github.com/tweag/rules_haskell/blob/3987c494f7a3b72805855246a844a28880dd0b15/haskell/cabal.bzl#L1267). E.g. `dir /sb mingw/libwinpthread.dll.a`.

azure-pipelines.yml
```diff
@@ -90,3 +90,4 @@ jobs:
       /c/bazel/bazel.exe build "@stackage-pinning-test//:hspec"
 
     displayName: 'Run Bazel'
+    timeoutInMinutes: 120
```
aherrmann, line 93: Is there an issue for this on the GHC tracker?

aherrmann (commented):
The CI failures seem unrelated to this change.
```
ERROR: /private/tmp/nix-build-bazel-3.3.1.drv-0/bazel_evDZYbOv/out/external/bazel_tools/third_party/ijar/BUILD:72:10: Linking of rule '@bazel_tools//third_party/ijar:zipper' failed (Exit 1): cc_wrapper.sh failed: error executing command 
...
0  0x108148a51  __assert_rtn + 129
1  0x108321444  ld::passes::stubs::x86_64::classic::LazyPointerAtom::LazyPointerAtom(ld::passes::stubs::Pass&, ld::Atom const&, bool, bool) + 420
2  0x10831ec6a  ld::passes::stubs::Pass::makeStub(ld::Atom const&, bool) + 2762
3  0x10831fc5f  ld::passes::stubs::Pass::process(ld::Internal&) + 2511
4  0x1083202c0  ld::passes::stubs::doPass(Options const&, ld::Internal&) + 128
5  0x108149ff4  main + 1860
A linker snapshot was created at:
        /tmp/zipper-2020-11-31-085219.ld-snapshot
ld: Assertion failed: (targetAtom != NULL), function Fixup, file ../../../../ld64/src/ld/ld.hpp, line 776.
clang-6.0: error: linker command failed with exit code 1 (use -v to see invocation)
```
on MacOS with nixpkgs and
```
ERROR: /var/lib/buildkite-agent-agent/.cache/bazel/_bazel_buildkite-agent-agent/8fa61af0b560ed001a9fcaeda5aedc45/external/local_jdk/BUILD.bazel:210:10: Label '@local_jdk//:lib/tools.jar' is invalid because '@local_jdk//lib' is a subpackage; perhaps you meant to put the colon here: '@local_jdk//lib:tools.jar'?
```
on Linux with nixpkgs.

aherrmann:
The MacOS with nixpkgs error seems to be related to the nixpkgs update, see https://github.com/NixOS/nixpkgs/issues/105573.

azure-pipelines.yml
mboes: It turns out that the initial run just happened to be a slow run. Subsequent runs were all of comparable length to previously. So it's unclear whether this is an upstream issue. But bumping up `timeoutInMinutes` should have the effect of making CI more reliable.

mboes:
> The MacOS with nixpkgs error seems to be related to the nixpkgs update, see NixOS/nixpkgs#105573.

Yes. Bazel does not build on macOS since that update. So this PR is currently blocked on upstream fixing the Bazel build.

PR #1465, opened 2021-01-02
Add documentation for haskell_import

mboes:
Modify haskell_toolchain_library documentation to make it clearer when
to use what rule.

### DIFF
--- a/haskell/defs.bzl
+++ b/haskell/defs.bzl
@@ -554,6 +554,15 @@ haskell_import = rule(
             default = Label("@rules_haskell//haskell:version_macros"),
         ),
     },
+    doc = """\
+Internal rule. Do not use.
+
+The attributes of this rule loosely correspond to the fields of the
+GHC package database. Refer to the [GHC User's Guide][ghc-doc-pkginfo]
+for documentation.
+
+[ghc-doc-pkginfo]: https://downloads.haskell.org/ghc/latest/docs/html/users_guide/packages.html#installedpackageinfo-a-package-specification
+""",
 )
 
 haskell_toolchain_library = rule(
@@ -570,7 +579,10 @@ haskell_toolchain_library = rule(
         "@rules_haskell//haskell:toolchain",
     ],
     doc = """\
-Import packages that are prebuilt outside of Bazel.
+Import prebuilt libraries supplied by the toolchain.
+
+Use this rule to make dependencies that are prebuilt (supplied as part
+of the compiler toolchain) available as targets.
 
 ### Examples
 
@@ -589,9 +601,6 @@ Import packages that are prebuilt outside of Bazel.
       ],
   )
   ```
-
-Use this rule to make dependencies that are prebuilt (supplied as part
-of the compiler toolchain) available as targets.
 """,
 )
 


haskell/defs.bzl
```diff
@@ -554,6 +554,16 @@ haskell_import = rule(
...
+    doc = """\
+Import libraries that are prebuilt outside of Bazel.
+
+The attributes of this rule loosely correspond to the fields of the
+GHC package database. Refer to the [GHC User's Guide][ghc-doc-pkginfo]
+for documentation. If a library is registered in GHC's global package
+database, then you can use `haskell_toolchain_library` instead.
+
+[ghc-doc-pkginfo]: https://downloads.haskell.org/ghc/latest/docs/html/users_guide/packages.html#installedpackageinfo-a-package-specification
+""",
```
aherrmann, line 566: The `haskell_import` rule only works for packages that are registered in the global package-db. It's not meant to be used by users directly. It's used to collect builtin packages and make them available to the toolchain. See https://github.com/tweag/rules_haskell/pull/925

mboes: It sounds like the contract of `haskell_import` slightly changed over time. Let's say I have a GHC package that I built with my own means (i.e. not with `cabal_haskell_library` or `haskell_library`). How do I expose that to Bazel? Can I use `haskell_import` for that the way I can use `cc_import` to expose any odd C/C++ library along with some metadata?

aherrmann: `haskell_import` assumes that the package is [listed in the global package db](https://github.com/tweag/rules_haskell/pull/859/files#diff-9b1fa0d6cbde73a03f524f92963b1b781de46bcd395e844ea3f3fc2ba5e10222R689), currently there is no way to import an additional package-db into Bazel. It is possible to import an externally built Haskell library as a C library via `cc_library` and use it as a dependency to regular Haskell targets. However, this won't include metadata like the package configuration.

The implementation of `haskell_import` is tailored towards toolchain packages and includes a few related workarounds. If we need a more generic `haskell_import` it may be better to rename the current one as `haskell_toolchain_library_import` or similar and implement the new `haskell_import` separately. AFAIK we didn't have a use-case so far that required a generic `haskell_import`.

mboes:
OK. I amended the commit to say that this an internal rule, which we're not documenting. We might be able to generalize `haskell_import` later, to make it justify its name.

haskell/ghc.BUILD.tpl
```diff
@@ -7,11 +7,6 @@ load(
     "@rules_haskell//haskell:cc_toolchain_config.bzl",
     "cc_toolchain_config",
 )
-load(
-    "@rules_haskell//haskell:defs.bzl",
-    "haskell_import",
-    "haskell_toolchain",
-)
```
aherrmann, line 14: Those are required, CI fails without them: https://buildkite.com/tweag-1/rules-haskell/builds/1757#b1b05522-9deb-45fb-8093-496baac8f385/35-839
```
ERROR: /home/buildkite-agent-agent/.cache/bazel/_bazel_buildkite-agent-agent/8b8a09a5fc28e77c5514319513adb8f9/external/rules_haskell_ghc_linux_amd64/BUILD:478:1: name 'haskell_import' is not defined
```

aherrmann (commented):
> OK. I amended the commit to say that this an internal rule, which we're not documenting. We might be able to generalize `haskell_import` later, to make it justify its name.

That sounds good, though I don't see it reflected on GH. Maybe the commit didn't get pushed? Or am I missing something?

mboes:
Bad rebase. I fixed that, so this PR is now only a documentation fix, as intended.

aherrmann (approved):
Thanks, LGTM

PR #1471, opened 2021-01-10
Move Linux CI jobs to GitHub Actions

mboes:
Hot on the heals of #1460, we move the Linux jobs to GitHub Actions as
well. In practice, BuildKite with custom runners wasn't buying us that
much time, though we may return to custom runners (via Actions) later.
The only CI job not on Actions is now the Windows one.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -12,6 +12,8 @@ test --test_env=LANG=C.UTF-8 --test_env=LOCALE_ARCHIVE
 
 # Platform / Toolchain Selection
 # ------------------------------
+build:linux-nixpkgs --config=nixpkgs
+build:macos-nixpkgs --config=nixpkgs
 build:nixpkgs --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host
 # Use this configuration when targeting Windows. Eventually this will
 # no longer be required:
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -24,10 +24,13 @@ jobs:
           nix_path: nixpkgs=./nixpkgs/default.nix
       - name: Configure
         run: |
-          [[ ${{ matrix.os }} == macos ]] && EXTRA_CONFIG=--config=macos-nixpkgs
+          case ${{ matrix.os }} in
+            macos) BUILD_CONFIG=macos-nixpkgs;;
+            ubuntu) BUILD_CONFIG=linux-nixpkgs;;
+          esac
           cat >.bazelrc.local <<EOF
           common --config=ci
-          build --config=nixpkgs $EXTRA_CONFIG
+          build --config=$BUILD_CONFIG
           EOF
       - name: Check Bazel version
         run: |
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,9 +1,10 @@
 pull_request_rules:
   - name: automatic merge
     conditions:
-      - 'status-success~=Build & Test - macOS \(Nixpkgs\).*'
-      - 'status-success~=Build & Test - macOS \(bindist\).*'
-      - "status-success~=buildkite/rules-haskell(|/pr)"
+      - 'status-success~=Build & Test - Nixpkgs \(ubuntu\).*'
+      - 'status-success~=Build & Test - Nixpkgs \(macos\).*'
+      - 'status-success~=Build & Test - bindist \(ubuntu\).*'
+      - 'status-success~=Build & Test - bindist \(macos\).*'
       - "status-success=tweag.rules_haskell"
       - "status-success=deploy/netlify"
       - "#approved-reviews-by>=1"
--- a/README.md
+++ b/README.md
@@ -254,7 +254,7 @@ This chooses the `cc_toolchain` bundled with GHC.
 ### Configuring your platform
 
 `rules_haskell` can be built and tested on Linux, MacOS, and Windows. Depending
-on the platform GHC can be provisioned using Nixpkgs or by downloading a binary
+on the platform GHC can be provisioned using nixpkgs or by downloading a binary
 distribution. In case of nixpkgs other toolchains (C compiler, Python, shell
 tools) will also be provided by nixpkgs, in case of bindist they will be taken
 from the environment (`$PATH`). The following `--config` options select the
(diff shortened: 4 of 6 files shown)


.bazelrc
```diff
@@ -12,8 +12,7 @@ test --test_env=LANG=C.UTF-8 --test_env=LOCALE_ARCHIVE
 
 # Platform / Toolchain Selection
 # ------------------------------
-build:linux-nixpkgs --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host
-build:macos-nixpkgs --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host
+build:nixpkgs --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host
```
aherrmann, line 15: This breaks the config flags documented in the README. With this change the last reference to `linux-nixpkgs` is gone and `--config=linux-nixpkgs` becomes invalid. I'd prefer to keep the config matrix in the README valid for symmetry with the bindist cases and in case we need platform specific flags in the future.

.github/workflows/workflow.yaml
```diff
@@ -17,33 +21,32 @@ jobs:
...
-          # Retry if needed due to network flakiness.
-          nix-shell --arg docTools false --pure --run \
-          'cmd="bazel fetch @stackage//..."; $cmd || $cmd || $cmd'
           # XXX 2019-01-22 Disable start script checking on Darwin
           # due to a clash between binutils and clang.
-          # nix-shell --arg docTools false --pure --run \
-          #   './tests/run-start-script.sh --use-nix'
+          [[ ${{ matrix.os }} != macos ]] &&
+            nix-shell --arg docTools false --pure --run \
+              './tests/run-start-script.sh --use-nix'
```
aherrmann, line 38:
```suggestion
          [[ ${{ matrix.os }} != macos ]] &&
            nix-shell --pure --run \
              './tests/run-start-script.sh --use-nix'
```
Otherwise the case `docTools=true` is untested on CI.

.buildkite/pipeline.yml
```diff
@@ -1,45 +0,0 @@
-steps:
```
aherrmann, line 1: `tests/RunTests.hs` now contains dangling references to `.circleci` and `.buildkite` in comments that need to be fixed.

aherrmann (commented):
Awesome, thank you! A couple small comments.

.github/workflows/workflow.yaml
mboes: In CircleCI this was also `false`, and the `docTools` argument only exists to avoid downloading many Nixpkgs derivations that are not required when not building the documentation. We could remove the `docTools` argument altogether, at the cost of slower CI. But that's orthogonal to moving existing CI jobs to GitHub Actions, so I think we'd want a separate PR.

.bazelrc
mboes: Bazel doesn't appear to support empty configurations (would make a good feature request upstream). We could hack around that, but I opted instead to remove this configuration from the README.

aherrmann: I've opened such a feature request: https://github.com/bazelbuild/bazel/issues/12844.
I don't think that [change](https://github.com/tweag/rules_haskell/pull/1471/commits/6a7d835b5d2fc7c0baea05780654b79562f4a088) is correct. On Linux/Nixpkgs we would need `--config=nixpkgs` but the README claims `(unneeded)`. On macOS we would now need both `--config=nixpkgs --config=macos-nixpkgs` but the README claims `--config=macos-nixpkgs`. The aim of https://github.com/tweag/rules_haskell/pull/1301 was to require only one top-level `--config` flag for developers.
I've pushed a commit that loads `--config=nixpkgs` from both `linux-nixpkgs` and `macos-nixpkgs`. With this developers only need a single top-level config and both linux-nixpkgs and macos-nixpkgs unfold the nixpkgs config as well.

mboes: I'm not sure I follow. `--config=nixpkgs` doesn't exist in this branch.

mboes: Sorry, spurious comment. I had the wrong branch checked out.

mboes:
@aherrmann just realized that yesterday when I did the force push, this somehow went to `master` instead of this branch. So this PR is effectively already merged, although in the local state when I did the push, i.e. without your latest commits. So I rebased this branch on top of the new `master`, taking into account your commits.

This is probably why `master` is red currently. :-/

I use Magit, which has a concept of "upstream" and "push" branch: https://magit.vc/manual/magit/The-Two-Remotes.html. This is likely culprit. I noticed before that it interacts badly with using Git CLI.

aherrmann:
> @aherrmann just realized that yesterday when I did the force push, this somehow went to `master` instead of this branch. So this PR is effectively already merged, although in the local state when I did the push, i.e. without your latest commits. So I rebased this branch on top of the new `master`, taking into account your commits.
> 
> This is probably why `master` is red currently. :-/
> 
> I use Magit, which has a concept of "upstream" and "push" branch: https://magit.vc/manual/magit/The-Two-Remotes.html. This is likely culprit. I noticed before that it interacts badly with using Git CLI.

Hmm, I thought branch protection on `master` would prevent pushing to master directly. Checking the settings, branch protection is enabled on `master`.

aherrmann (approved):

mboes:
AFAIU branch protection only protects against force pushes from admins.
Though we can enable stronger protection.

aherrmann:
> AFAIU branch protection only protects against force pushes from admins. Though we can enable stronger protection.

I see, if I understand [these docs](https://docs.github.com/en/github/administering-a-repository/about-protected-branches#restrict-who-can-push-to-matching-branches) correctly then admins can always push.

> People and apps with admin permissions to a repository are always able to push to a protected branch.

aherrmann:
This was failing on buildifier warnings

aherrmann:
Ecountered an error of the following form [here](https://github.com/tweag/rules_haskell/pull/1471/checks?check_run_id=1776055682#step:7:721)
```
ERROR: The Build Event Protocol upload failed: All retry attempts failed. DEADLINE_EXCEEDED: DEADLINE_EXCEEDED: deadline exceeded after 14999960100ns DEADLINE_EXCEEDED: DEADLINE_EXCEEDED: deadline exceeded after 14999960100ns
```
cc @siggisim

siggisim:
That looks like some potential network causing a timeout to be hit - the invocation uploaded just fine on our end, and I can't find any errors in the logs related to that invocation:
https://app.buildbuddy.io/invocation/d607713d-0b2b-47c0-8f3f-ef81dce03b56#

Are you seeing this consistently?

aherrmann:
Thanks for looking into this @siggisim .

> Are you seeing this consistently?

No, this was the first time I noticed this. Yes, looks like network flakiness. I'll let you know if this happens repeatedly.

aherrmann:
I've included an update to the mergify configuration and checked it in their config editor UI.

PR #1484, opened 2021-02-08
Run more test cases in bindist mode

aherrmann:
Some test cases were excluded from the bindist pipelines because they depend on targets imported from nixpkgs. However, this is not necessary in all cases. These targets can either be built by Bazel directly, e.g. `zlib`, or can be built by `stack_snapshot`, e.g. `doctest`.

This change moves nixpkgs dependencies to be built by Bazel instead where appropriate and removes the corresponding exclusion tags, e.g. `requires_doctest`.


tests/library-with-sysincludes/BUILD.bazel
```diff
@@ -21,21 +21,21 @@ cc_library(
     name = "zlib",
     hdrs = ["@zlib.dev//:include"],
     strip_include_prefix = "/external/zlib.dev/include",
-    tags = ["requires_zlib"],
+    tags = ["requires_nix"],
```
aherrmann, line 24: These tests don't require any zlib, but specifically the one provided by nixpkgs. So, the tag is named `requires_nix` instead of `requires_zlib`.

tests/stackage_zlib_runpath/BUILD.bazel
```diff
@@ -13,20 +13,21 @@ dynamic_libraries(
     srcs = ["@zlib.dev//:zlib"],
     filter = "libz",
     solib_names = "libz_soname",
-    tags = ["requires_zlib"],
+    # Only the Nix provided zlib is dynamically linked.
+    tags = ["requires_nix"],
```
aherrmann, line 17: These tests don't require any `zlib`, but specifically the one provided by nixpkgs. So, the tag is named `requires_nix` instead of `requires_zlib`.

mboes (approved):

aherrmann:
`c2hs` was failing on macOS in bindist mode. I'm skipping the corresponding tests on that platform for now and opened https://github.com/tweag/rules_haskell/issues/1486 to track the issue.

PR #1488, opened 2021-02-10
Use BuildBuddy API key

aherrmann:


guibou (approved):

guibou:
Approved because that's simple.

Just a question. This PR does just add an header in the HTTP request that `bazel` will do to the remote builder. So right now, except if I missed something, the only thing that this PR is doing is leaking the token to our current remote builder, if it exists.

aherrmann:
> Just a question. This PR does just add an header in the HTTP request that `bazel` will do to the remote builder. So right now, except if I missed something, the only thing that this PR is doing is leaking the token to our current remote builder, if it exists.

Thank you. Yes, this PR just sets the API key header as documented in BuildBuddy's setup guide. BuildBuddy had been configured before in https://github.com/tweag/rules_haskell/pull/1460. We're not using the remote builder, only the build event log.

PR #1487, opened 2021-02-09
Bazel 3.7.2

aherrmann:
Update Bazel from version 3.3.1 to 3.7.2.
Updates the nixpkgs revision accordingly.
Updates rules_haskell dependencies to latest released versions.

PR #1490, opened 2021-02-11
Update rules nixpkgs

aherrmann:
Depends on #1487 
(Not that rules_nixpkgs 0.8.0 needs Bazel 3.7.2 but there are changes in the Bazel update that would otherwise cause merge conflicts)

Update to latest 0.8.0 release of rules_nixpkgs.
Use the new hermetic nixpkgs cc toolchain and enable cc toolchain resolution to that end.


facundominguez (approved):
LGTM

PR #1491, opened 2021-02-11
Move cabal_wrapper arguments to json

facundominguez:
This is just so the cabal_wrapper doesn't need to parse the arguments on its own.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -274,7 +274,7 @@ def _prepare_cabal_inputs(
         for d in direct_include_dirs.to_list()
     ] + _uniquify(["--extra-lib-dirs=" + d for d in direct_lib_dirs])
     if with_profiling:
-        path_args.append("--enable-profiling")
+        extra_args.append("--enable-profiling")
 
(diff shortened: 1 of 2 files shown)


haskell/cabal.bzl
```diff
@@ -256,26 +255,40 @@ def _prepare_cabal_inputs(
...
+    path_args = [
+        "--package-db=" + _dirname(db)
+        for db in package_databases.to_list()
+    ] + [
+        "--extra-include-dirs=" + d
+        for d in direct_include_dirs.to_list()
+    ] + _uniquify(["--extra-lib-dirs=" + d for d in direct_lib_dirs])
     if with_profiling:
-        args.add("--enable-profiling")
+        path_args.append("--enable-profiling")
```
aherrmann, line 277:
```suggestion
        extra_args.append("--enable-profiling")
```
This doesn't contain a path, so doesn't need the execroot prepended.

aherrmann (approved):
Thank you! That's much nicer to work with.

The previous approach forced us to bundle arguments by whether they needed an execroot prefix or not. This approach is much more flexible.

PR #1505, opened 2021-02-25
Pass the path to the C compiler with envvars to cc_wrapper

facundominguez:
Bazel considers the cc toolchain a build dependency of the cc_wrapper, but we only need the cc toolchain at runtime. Without relying on environment variables, this is practically the only way we have to get our hands on the path to the C compiler other than using environment variables.

Using environment variables would be inconvenient to use with hie-bios because of technical reasons.

Specifying the dependency like this, however, interferes with cross-compilation, since the provided toolchain will target the execution platform. In this case, we use environment variables in the cc_wrapper scripts to indicate the path to the cc compiler, since we are not trying to cross-compile when using hie-bios.

### DIFF
--- a/haskell/private/cc_wrapper.bzl
+++ b/haskell/private/cc_wrapper.bzl
@@ -41,6 +41,22 @@ _cc_wrapper = rule(
             allow_single_file = True,
         ),
         "platform": attr.string(),
+        # Bazel considers the cc toolchain a build dependency of the cc_wrapper,
+        # but we only need the cc toolchain at runtime. Without relying on
+        # environment variables, this is practically the only way we have to get
+        # our hands on the path to the C compiler other than using environment
+        # variables.
+        #
+        # Using environment variables wouldn't work when using hie-bios.
+        #
+        # Specifying the dependency like this, however, interferes with
+        # cross-compilation, since the provided toolchain will target
+        # the execution platform. In this case, we use environment variables
+        # in the cc_wrapper scripts to indicate the path to the cc compiler.
+        # We are not trying to cross-compile when using hie-bios.
+        #
+        # TODO: Consider using execution groups to transition the toolchain
+        # to the target platform.
         "_cc_toolchain": attr.label(
             default = Label("@bazel_tools//tools/cpp:current_cc_toolchain"),
         ),
--- a/haskell/private/cc_wrapper.bzl
+++ b/haskell/private/cc_wrapper.bzl
@@ -2,6 +2,38 @@ load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
 load("@bazel_tools//tools/build_defs/cc:action_names.bzl", "ACTION_NAMES")
 load("@rules_python//python:defs.bzl", "py_binary")
 
+# Note [On configuring the cc_wrapper]
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# The cc_wrapper provides a thin compatibility layer between GHC and
+# the cc-toolchain that Bazel provides. It is a script that GHC will
+# invoke as the C compiler (pgmc) and that itself will invoke the C
+# compiler provided by Bazel's cc-toolchain. Therefore cc_wrapper only
+# needs the cc-toolchain at runtime, i.e. when building Haskell
+# binaries. More importantly, in case of cross-compilation, cc_wrapper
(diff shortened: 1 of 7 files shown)


haskell/private/cc_wrapper.py.tpl
```diff
@@ -80,8 +80,9 @@ import sys
 import tempfile
 
 WORKSPACE = "{:workspace:}"
-CC = "{:cc:}"
-PLATFORM = "{:platform:}"
+CC = os.environ.get("CC_WRAPPER_CC_PATH", "{:cc:}")
+PLATFORM = os.environ.get("CC_WRAPPER_PLATFORM", "{:platform:}")
```
facundominguez, line 84: I'll have to split this PLATFORM envvar in BUILD_PLATFORM and TARGET_PLATFORM. The script is using it to address concerns of both.

haskell/private/cc_wrapper.py.tpl
```diff
@@ -80,8 +80,9 @@ import sys
...
+CPU = os.environ.get("CC_WRAPPER_CPU", "{:cpu:}")
 INSTALL_NAME_TOOL = "/usr/bin/install_name_tool"
 OTOOL = "/usr/bin/otool"
```
facundominguez, line 87: These settings of INSTALL_NAME_TOOl and OTOOL are not going to work if we ever try to cross-compile from linux to osx. But I guess it is out of scope here.

haskell/private/cc_wrapper.bzl
```diff
@@ -41,6 +41,22 @@ _cc_wrapper = rule(
...
+        # Using environment variables wouldn't work when using hie-bios.
+        #
+        # Specifying the dependency like this, however, interferes with
+        # cross-compilation, since the provided toolchain will target
+        # the execution platform. In this case, we use environment variables
+        # in the cc_wrapper scripts to indicate the path to the cc compiler.
+        # We are not trying to cross-compile when using hie-bios.
+        #
+        # TODO: Consider using execution groups to transition the toolchain
+        # to the target platform.
```
aherrmann, line 59: I think this may be a bit hard to understand for someone who has less context about this. I gave it a stab to rephrase this to give a bit more context below.

---

The cc_wrapper provides a thin compatibility layer between GHC and the cc-toolchain that Bazel provides. It is a script that GHC will invoke as the C compiler (`pgmc`) and that itself will invoke the C compiler provided by Bazel's cc-toolchain. Therefore cc_wrapper only needs the cc-toolchain at runtime, i.e. when building Haskell binaries. More importantly, in case of cross-compilation, cc_wrapper needs a cc-toolchain that targets the platform that we are building the Haskell binary for.

However, we do not have control over the arguments that GHC passes to the cc_wrapper in all cases. Meaning, we cannot use command line arguments to configure cc_wrapper to use a particular cc-toolchain. Instead, we use environment variables to pass all relevant information about the cc-toolchain to cc_wrapper.

Unfortunately, environment variables are not available in the IDE use-case, because we cannot specify environment variables at the hie-bios interface. To work around this limitation, we configure a fall-back cc-toolchain as a build-time dependency to cc_wrapper.

Note, Bazel will not resolve the desired cc-toolchain at this build-time dependency in case of cross-compilation. The cc_wrapper is a dependency to the hs-toolchain and Bazel will build it to target the build platform (execute) instead of the ultimate target platform (target). Therefore, the fall-back cc-toolchain will not target the cross-platform and should not be used for any build actions.

TODO: Consider using execution groups to transition the toolchain to the target platform.

haskell/private/cc_wrapper.py.tpl
aherrmann: That is true. Yes, I'd consider that out of scope for now.

aherrmann (approved):
Thank you, looks good!

The comment could maybe provide a bit more context.

haskell/private/cc_wrapper.bzl
facundominguez: I didn't expected less of your review. :)

facundominguez: Adopted.

aherrmann: Thank you! Yes, using the note format is a good idea for this.

mboes:
So I take it Bazel configuration transitions are not sufficiently implemented yet?

facundominguez:
Indeed. Not yet. I created https://github.com/tweag/rules_haskell/issues/1507 to follow on the matter.

PR #1510, opened 2021-03-15
Invoke a runghc for the execution platform in the cabal_wrapper

facundominguez:
In this PR cabal_wrapper always uses a runghc binary from a toolchain that targets the execution platform. Otherwise, there would be no way to execute Setup.hs when cross-compiling.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -153,7 +153,7 @@ def _concat(sequences):
 def _uniquify(xs):
     return depset(xs).to_list()
 
-def _cabal_toolchain_info(hs, cc, workspace_name):
+def _cabal_toolchain_info(hs, cc, workspace_name, runghc):
     """Yields a struct containing the toolchain information needed by the cabal wrapper"""
 
     # If running on darwin but XCode is not installed (i.e., only the Command
@@ -168,7 +168,7 @@ def _cabal_toolchain_info(hs, cc, workspace_name):
     return struct(
         ghc = hs.tools.ghc.path,
         ghc_pkg = hs.tools.ghc_pkg.path,
-        runghc = hs.tools.runghc.path,
+        runghc = runghc.path,
         ar = ar,
         cc = cc.tools.cc,
         strip = cc.tools.strip,
@@ -198,6 +198,7 @@ def _prepare_cabal_inputs(
         flags,
         generate_haddock,
         cabal_wrapper,
+        runghc,
         package_database,
         verbose,
         transitive_haddocks,
@@ -345,11 +346,11 @@ def _prepare_cabal_inputs(
         runghc_args = runghc_args,
         extra_args = extra_args,
         path_args = path_args,
-        toolchain_info = _cabal_toolchain_info(hs, cc, workspace_name),
+        toolchain_info = _cabal_toolchain_info(hs, cc, workspace_name, runghc),
     )
 
     inputs = depset(
-        [setup, hs.tools.ghc, hs.tools.ghc_pkg, hs.tools.runghc],
+        [setup, hs.tools.ghc, hs.tools.ghc_pkg, runghc],
         transitive = [
             depset(srcs),
             depset(cc.files),
@@ -503,6 +504,7 @@ def _haskell_cabal_library_impl(ctx):
         flags = ctx.attr.flags,
         generate_haddock = ctx.attr.haddock,
         cabal_wrapper = ctx.executable._cabal_wrapper,
+        runghc = ctx.executable._runghc,
         package_database = package_database,
         verbose = ctx.attr.verbose,
         is_library = True,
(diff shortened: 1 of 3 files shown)


haskell/private/runghc.bzl
```diff
@@ -0,0 +1,38 @@
...
+# Producing runghc with a custom rule, as defined here, allows to use
+# it in rules which are using a Haskell toolchain for the target
+# platform.
+#
+
+def _runghc_impl(ctx):
+    hs_toolchain = ctx.toolchains["@rules_haskell//haskell:toolchain"]
+    (_, extension) = paths.split_extension(hs_toolchain.tools.runghc.path)
+
+    runghc_file = ctx.actions.declare_file(ctx.label.name + "2" + extension)
```
aherrmann, line 19: Why the `+ "2"`?

haskell/private/runghc.bzl
```diff
@@ -0,0 +1,38 @@
...
+    ctx.actions.symlink(
+        output = runghc_file,
+        target_file = hs_toolchain.tools.runghc,
+        is_executable = True,
+    )
+
+    return [DefaultInfo(
+        executable = runghc_file,
+        runfiles = hs_toolchain.cc_wrapper.runfiles.merge(
+            ctx.runfiles(files = [hs_toolchain.tools.runghc]),
```
aherrmann, line 29:
```suggestion
            ctx.runfiles(files = [runghc_file, hs_toolchain.tools.runghc]),
```
See https://groups.google.com/g/bazel-discuss/c/oEyEIr7vaF4/m/0E5QaK46AQAJ

aherrmann (commented):
On Windows this fails with 
```
runghc2.exe: C:\users\runneradmin\_bazel_runneradmin\z5des2pa\execroot\YOUR_PROJECT_NAME_HERE\bazel-out\host\bin\external\rules_haskell\haskell\ghc-stage2.exe: rawSystem: does not exist (No such file or directory)
```
I suspect the issue is that `ctx.actions.symlink` creates a copy instead of a symlink on Windows. (Symlinks not being available on Windows by default). `runghc` then tries to find GHC relative to itself and can't. One way to work around this would be to not create a symlink but a wrapper script that calls `runghc`. That will require two rules: One to generate the source file for the script and one `sh_binary` that makes the script executable.

haskell/private/runghc.bzl
facundominguez: I'm a tad confused. Why does CI work without adding `runghc_file` to the runfiles?

facundominguez: Ah, it is probably the way in which the output of the rule is depended upon.

aherrmann: `runghc` is added to the action inputs explicitly, https://github.com/tweag/rules_haskell/pull/1510/files#diff-908241729f5583f805b295fedcf5a975ae9f8099ee83fae2a08496f9ba78b847R353.
This change is only relevant if you have another executable target that adds this runghc target to its `data` dependencies.

haskell/private/runghc.bzl
facundominguez: Debugging artifact.
```suggestion
    runghc_file = ctx.actions.declare_file(ctx.label.name + extension)
```

haskell/private/runghc.bzl
```diff
@@ -0,0 +1,64 @@
...
+        content = """\
+#!/usr/bin/env python3
+
+import subprocess
+import sys
+from rules_python.python.runfiles import runfiles
+
+r = runfiles.Create()
+
+subprocess.run([r.Rlocation("{runghc}")] + sys.argv[1:], check=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
facundominguez, line 33: Letting the subprocess use stdout and stderr from the caller:
```suggestion
subprocess.run([r.Rlocation("{runghc}")] + sys.argv[1:], check=True)
```

haskell/cabal.bzl
```diff
@@ -345,11 +346,11 @@ def _prepare_cabal_inputs(
         runghc_args = runghc_args,
         extra_args = extra_args,
         path_args = path_args,
-        toolchain_info = _cabal_toolchain_info(hs, cc, workspace_name),
+        toolchain_info = _cabal_toolchain_info(hs, cc, workspace_name, runghc),
     )
 
     inputs = depset(
-        [setup, hs.tools.ghc, hs.tools.ghc_pkg, hs.tools.runghc],
+        [setup, hs.tools.ghc, hs.tools.ghc_pkg, runghc],
```
aherrmann, line 353: Does `runghc` still need to be added to `inputs`, if it's added to `tools` and `input_manifests` later?

aherrmann (approved):
Looks good, thank you!

haskell/cabal.bzl
facundominguez: I've tried and it seems to work.
```suggestion
        [setup, hs.tools.ghc, hs.tools.ghc_pkg],
```

PR #1513, opened 2021-03-19
Use different repo names when registering haskell toolchains

facundominguez:
When cross-compiling we need the ability to register more than one haskell toolchain.


aherrmann (approved):
LGTM

PR #1515, opened 2021-03-24
Set cc wrapper env vars when running ghc

facundominguez:
This is necessary to cross-compile with haskell_binary and haskell_library.

### DIFF
--- a/haskell/toolchain.bzl
+++ b/haskell/toolchain.bzl
@@ -1,5 +1,6 @@
 """Rules for defining toolchains"""
 
+load("@bazel_skylib//lib:dicts.bzl", "dicts")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load(":ghc_bindist.bzl", "haskell_register_ghc_bindists")
 load(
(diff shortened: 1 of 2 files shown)


haskell/toolchain.bzl
```diff
@@ -82,7 +83,7 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, params_file = None, e
         executable = hs.ghc_wrapper,
         mnemonic = mnemonic,
         progress_message = progress_message,
-        env = env,
+        env = dicts.add(env, cc.env),
```
aherrmann, line 86: I think this goes against the spirit of how the `env` argument has been used so far: If `env` is not specified explicitly or empty, then `_run_ghc` will automatically fall back to using `hs.env`. Otherwise, it will only use the `env` argument as given.

`run_ghc` is only used in `compile.bzl` and `link.bzl` and the `cc.env` was already passed explicitly in [`compile.bzl`](https://github.com/tweag/rules_haskell/blob/183f7f3e9889f2108984ad27efea1cbd51657d42/haskell/private/actions/compile.bzl#L366). 
So, it's only `link.bzl` that needs to have `cc.env` added. Perhaps this is a good opportunity to get rid of the implicit default env and set it explicitly in `link.bzl` as well.

aherrmann (approved):
LGTM
Thank you!

PR #1516, opened 2021-03-25
Allow to specify platform constraints for ghcs coming from nixpkgs

facundominguez:


haskell/nixpkgs.bzl
```diff
@@ -130,12 +144,18 @@ def _ghc_nixpkgs_toolchain_impl(repository_ctx):
...
+    if repository_ctx.attr.target_constraints == None:
+        target_constraints = ["@platforms//cpu:x86_64"]
+        if repository_ctx.os.name == "linux":
+            target_constraints.append("@platforms//os:linux")
+        elif repository_ctx.os.name == "mac os x":
+            target_constraints.append("@platforms//os:osx")
+    else:
+        target_constraints = repository_ctx.attr.target_constraints
+    if repository_ctx.attr.exec_constraints == None:
+        exec_constraints = list(target_constraints)
```
aherrmann, line 156: I think the defaulting behavior is a bit surprising here.
If the user only sets `target_constraint`, then `exec_constraint` is going to match `target_constraint`. However, if the user only sets `exec_constraint` then `target_constraint` is going to depend on `repository_ctx.os.name`.

I think it would be better to only provide a default if neither `target_constraint` nor `exec_constraint` are set by the user and otherwise require both to be set explicitly.

aherrmann (commented):
I'd prefer to limit this PR to specifying platform constraints on the nixpkgs ghc toolchain and add the cross compilation example in a separate PR.

facundominguez:
Right, I wasn't intending to submit the example just yet.

aherrmann (approved):
Looks good, thank you!

PR #1517, opened 2021-03-26
Add an example of cross-compiling to arm

facundominguez:
See `example/arm/README.md` for details.


examples/arm/arm-cross.nix
```diff
@@ -0,0 +1,96 @@
...
+    remote-iserv = crossPkgs.ghc-extra-packages.ghc8102.remote-iserv.components.exes.remote-iserv;
+    llvm = pkgs.llvm_9;
+    crossNumactl = crossPkgs.numactl;
+    qemu = pkgs.buildPackages.qemu;
+    qemuIservWrapper = pkgs.writeScriptBin "iserv-wrapper" ''
+      #!${pkgs.stdenv.shell}
+      set -euo pipefail
+      # Unset configure flags as configure should have run already
+      unset configureFlags
+      PORT=$((5000 + $RANDOM % 5000))
```
aherrmann, line 14: This looks like it's prone to flaky build errors if this happens to land on an occupied port. Does iserv perhaps support port=0 and a port file or something of the sort to safely choose a free port?

examples/arm/zlib.BUILD.bazel
```diff
@@ -0,0 +1,25 @@
...
+filegroup(
+    name = "include",
+    srcs = glob(["include/*.h"]),
+    visibility = ["//visibility:public"],
+)
+cc_library(
+    name = "zlib",
+	srcs = select({
+	    "//:is_linux_aarch64": ["@nixpkgs_zlib_arm//:lib"],
+        "//conditions:default": ["@nixpkgs_zlib_x86//:lib"],
```
aherrmann, line 20:
```suggestion
        "//:is_linux_aarch64": ["@nixpkgs_zlib_arm//:lib"],
        "//conditions:default": ["@nixpkgs_zlib_x86//:lib"],
```
Running `buildifer -r` over this directory is an easy way to fix all formatting issues.

examples/arm/README.md
```diff
@@ -0,0 +1,26 @@
+This directory contains an example of how to cross-compile to arm
+with `rules_haskell`.
+
+Build with
+```
+nix-shell --pure --run "cd examples/arm; bazel build //... --platforms=//:linux_aarch64"
```
aherrmann, line 6: Perhaps add a symlink to the toplevel `shell.nix` to `examples/arm`. `examples` does the same thing.

aherrmann (commented):
To run this on CI we could use the [`haskell.nix` binary-cache](https://input-output-hk.github.io/haskell.nix/tutorials/getting-started/#setting-up-the-binary-cache). The GH action to install nix has a flag for extra configuration that we can probably use for this https://github.com/cachix/install-nix-action

examples/arm/zlib.BUILD.bazel
facundominguez: Buildifier is not fixing this one. I'm running both
```
nix-shell --pure --run "bazel run --disk_cache=~/bazel-disk-cache //:buildifier -- -r examples/arm"
nix-shell --pure --run "bazel run --disk_cache=~/bazel-disk-cache //:buildifier-fix"
```
Both commands succeed without changing or reporting any files.

facundominguez: They do catch formatting issues on the WORKSPACE file (if I introduce them). But they seem to ignore `examples/arm/zlib.BUILD.bazel `

examples/arm/.bazelrc
```diff
@@ -0,0 +1,2 @@
+build --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host
+build --incompatible_enable_cc_toolchain_resolution
```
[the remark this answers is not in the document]
facundominguez: Done

examples/arm/README.md
facundominguez: Done.

examples/arm/WORKSPACE
```diff
@@ -0,0 +1,150 @@
+workspace(name = "arm_cross_example")
```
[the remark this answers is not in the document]
facundominguez: Done.

examples/arm/zlib.BUILD.bazel
aherrmann: Oh, I see. Sorry, in this case, it needs to be called directly on the file, given that it's not a standard BUILD file name.

examples/arm/arm-cross.nix
facundominguez: Here is where the binding happens,
https://gitlab.haskell.org/ghc/ghc/-/blob/master/libraries/libiserv/src/Remote/Slave.hs#L64
We can pass 0 to remote iserv, and it will bind to an available port, but it won't report back which port it was.

We are forced then to do something hacky. One easy way would be to try to start remote-iserv a few times before giving up.

examples/arm/zlib.BUILD.bazel
facundominguez: I renamed the file to `BUILD.zlib.bazel`, buildifier finds it now.

examples/arm/arm-cross.nix
aherrmann: >  We can pass 0 to remote iserv, and it will bind to an available port, but it won't report back which port it was.

I see, that's unfortunate. Perhaps worth a ticket on iserv.

> We are forced then to do something hacky. One easy way would be to try to start remote-iserv a few times on different random ports before giving up.

:+1:

facundominguez: I made a try. Instead of starting remote-iserv multiple times, I'm using `nc` to find an available port. It turns out to be simpler if it works.

facundominguez: One more try, this time without using nc. I think the approach using ss is less invasive as it doesn't need to interact with services running on occupied ports.

facundominguez:
```
for((PORT=5000;PORT<5000;PORT++)) looks like a typo, the start port and upper bound are the same.

nc -z PORT doesn't work for me, I have to specify a host as well, e.g. nc -z localhost 5000.

I think the random approach is preferable here. Always starting at 5000 and incrementing from there increases the risk of a collision. There is also a window for a race between the check with nc and iserv actually occupying the port. It doesn't seem all too unlikely to hit that window, e.g. I would imagine that bazel build //:a //:b could hit it if both targets a and b are scheduled to be built at the same time by Bazel and both require iserv.
```

aherrmann (commented):
Thank you! The iserv stuff looks good. Could you also add a line to the README how to run the resulting binary in qemu? I tried testing it locally and am getting
```
$ qemu-aarch64 bazel-out/k8-fastbuild/bin/example
bazel-out/k8-fastbuild/bin/example: error while loading shared libraries: libgcc_s.so.1: cannot open shared object file: No such file or directory
```
Side note, it's a bit odd that Bazel uses `k8-fastbuild` for the arm outputs. I would have expected it to put the artifacts into a separate output along the lines of `aarch64-fastbuild`. Then pattern here is `<cpu>-<mode>`. Perhaps the platform/toolchain is not setting the cpu correctly. But, that falls outside the scope of this PR.

examples/arm/qemu-shell.nix
```diff
@@ -0,0 +1,11 @@
+let
+  pkgs = (import ./arm-cross.nix).pkgs;
+  crossPkgs = pkgs.pkgsCross.aarch64-multiplatform;
+in
+pkgs.mkShell {
+  LANG="C.UTF-8";
+  LD_LIBRARY_PATH = "${crossPkgs.buildPackages.gcc-unwrapped}/aarch64-unknown-linux-gnu/lib64";
```
aherrmann, line 7: I'm wondering why this is needed and there's no corresponding `RUNPATH` entry generated. Something worth looking into outside of this PR.

aherrmann (approved):
Great work, thank you!  :tada: 

Thank you for the added instructions, the example works for me that way.

examples/arm/qemu-shell.nix
facundominguez: Created an issue here: https://github.com/tweag/rules_haskell/issues/1521

PR #1519, opened 2021-03-30
ylecornec/move attr doc to macros

ylecornec:
Some macros currently don't have API documentation on haskell.build, while the underlying rules contain doc attributes.

The first commit moves the documentation from the rule to the macro in the case the rule is private.

The second commit adds some fields in macro documentation,
with links to the documentation of the underlying rules when they are publics.

### DIFF
--- a/haskell/ghc_bindist.bzl
+++ b/haskell/ghc_bindist.bzl
@@ -604,6 +604,14 @@ def haskell_register_ghc_bindists(
 
     [toolchain-resolution]: https://docs.bazel.build/versions/master/toolchains.html#toolchain-resolution
 
+    Args:
+      version: [see ghc_bindist](#ghc_bindist-name)
+      compiler_flags: [see ghc_bindist](#ghc_bindist-compiler_flags)
+      haddock_flags: [see ghc_bindist](#ghc_bindist-haddock_flags)
+      repl_ghci_args: [see ghc_bindist](#ghc_bindist-repl_ghci_args)
+      cabalopts: [see ghc_bindist](#ghc_bindist-cabalopts)
+      locale: [see ghc_bindist](#ghc_bindist-local)
+
     """
     version = version or _GHC_DEFAULT_VERSION
     if not GHC_BINDIST.get(version):
--- a/haskell/ghc_bindist.bzl
+++ b/haskell/ghc_bindist.bzl
@@ -539,9 +539,13 @@ def ghc_bindist(
        ```
 
     Args:
-      version: The desired GHC version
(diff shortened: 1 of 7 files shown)


haskell/ghc_bindist.bzl
```diff
@@ -604,6 +604,14 @@ def haskell_register_ghc_bindists(
 
     [toolchain-resolution]: https://docs.bazel.build/versions/master/toolchains.html#toolchain-resolution
 
+    Args:
```
aherrmann, line 607: `rules_haskell_toolchains` defined in `haskell/toolchains.bzl` is the entry point that is documented in all the examples and use-case docs. It itself then calls `haskell_register_ghc_bindists`. I think in this case it's better to use this opportunity and move this docstring and the argument docs from `ghc_bindist` to `rules_haskell_toolchains` and have those reference `rules_haskell_toolchains`.

aherrmann (commented):
Thank you! Looks good!

One comment about `rules_haskell_toolchain`.

haskell/ghc_bindist.bzl
ylecornec: I figured that if users are loading the rule like this
`load("@rules_haskell//haskell:toolchain.bzl", "rules_haskell_toolchains")`
then I should expose the `toolchain.bzl` file in the documentation.

So now `haskell_toolchain` appears twice in the menu (since it is also reexported in `def.bzl`).

They are now three names for `rules_haskell_toolchains`.
 - `rules_haskell_toolchains` itself
 - `haskell_register_toolchains`
 - `haskell_register_ghc_bindists`
   
Are the last two depreciated ?

aherrmann: Good catch. `haskell_register_toolchains` is deprecated, see [here in the changelog](https://github.com/tweag/rules_haskell/blob/377db34fec4342ee0c317725669334479b6ecc10/CHANGELOG.md#changed-3). So, that should be marked accordingly.

Re `haskell_register_ghc_bindists`, that's a good question. There's ongoing work on more toolchains, in particular a [host toolchain](https://github.com/tweag/rules_haskell/pull/1466/files). I'm not sure yet how this will affect `rules_haskell_toolchains`. I'd say for the purpose of this PR the link from `haskell_register_ghc_bindists` to `rules_haskell_toolchains` is good the way it is.

ylecornec: Ok. I marked `haskell_register_toolchains` as deprecated.

aherrmann (approved):
Thank you, looks good!

PR #1523, opened 2021-04-02
Build and run the arm example in CI

facundominguez:


examples/arm/qemu-shell.nix
```diff
@@ -4,8 +4,9 @@ let
 in
 pkgs.mkShell {
   LANG="C.UTF-8";
-  LD_LIBRARY_PATH = "${crossPkgs.buildPackages.gcc-unwrapped}/aarch64-unknown-linux-gnu/lib64";
+  LD_LIBRARY_PATH = "${crossPkgs.buildPackages.gcc-unwrapped.out.lib}/aarch64-unknown-linux-gnu/lib";
   buildInputs = [
     pkgs.qemu
+    pkgs.strace
```
aherrmann, line 10: Is this still needed or was it only for debugging?

.github/workflows/workflow.yaml
```diff
@@ -51,6 +54,48 @@ jobs:
...
+          nix_path: nixpkgs=./nixpkgs/default.nix
+          extra_nix_config: |
+            trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=
+            extra-substituters = https://hydra.iohk.io
+      - name: Configure
+        env:
+          BUILDBUDDY_API_KEY: ${{ secrets.BUILDBUDDY_API_KEY }}
+        run: |
+          cat >.bazelrc.local <<EOF
+          # Enabling --config=ci causes cross-compilation to crawl
```
aherrmann, line 79: For reference, this is tracked in https://github.com/tweag/rules_haskell/issues/1524

.github/workflows/workflow.yaml
```diff
@@ -51,6 +54,48 @@ jobs:
             ./bazel-ci-bin/tests/run-tests
             bazel coverage //tests/...
             '
+
+  test-nixpkgs-cross:
+    name: Build & Test - Cross
```
aherrmann, line 59: Could you add this to the [mergify config](https://github.com/tweag/rules_haskell/blob/master/.mergify.yml#L3). You can use the [mergify dashboard](https://dashboard.mergify.io/github/tweag/queues) to make sure the config is valid.

aherrmann (approved):
Great work, thank you!

examples/arm/qemu-shell.nix
facundominguez: Only for debugging.
```suggestion
```

facundominguez:
Have I screwed the mergify config? It is not merging this, and the mergify dashboards appears empty.

PR #1526, opened 2021-04-05
Write a section on cross-compilation in the documentation for use cases

facundominguez:


docs/haskell-use-cases.rst
```diff
@@ -884,3 +884,127 @@ not possible to build docker images for haskell binaries as above using rules_do
 
 Following these steps you should end up with a fairly lightweight docker image, bringing the flexibility of nix
 as a docker base image manager and the power of ``rules_haskell`` for your haskell build together.
+
+Cross-compilation
+-----------------
+
+Currently, ``rules_haskell`` only supports cross-compiling to ``arm`` on linux.
```
mboes, line 891:
```suggestion
Currently, ``rules_haskell`` only supports cross-compiling to ``arm`` on Linux.
```

docs/haskell-use-cases.rst
```diff
@@ -884,3 +884,127 @@ not possible to build docker images for haskell binaries as above using rules_do
...
+Cross-compiling requires providing a cross-compiler, telling ``rules_haskell``
+about it, and then requesting Bazel to build for the target platform.
+
+Ideally, providing a cross-compiler would only require the advice in
+`Picking a compiler`_. However, the case of ``arm`` requires to configure
+a few aspects at this time. One has to make available the llvm tools
+to the compiler, emulation support needs to be set to enable
+compilation of Template Haskell splices via an external interpreter,
```
mboes, line 899: Can this be made optional, say if the user does not use TH?

docs/haskell-use-cases.rst
```diff
@@ -884,3 +884,127 @@ not possible to build docker images for haskell binaries as above using rules_do
...
+          "@platforms//os:linux",
+      ],
+      target_constraints = [
+          "@platforms//cpu:aarch64",
+          "@platforms//os:linux",
+      ],
+      repository = "@nixpkgs",
+  )
+
+This instantiation indicates the Nix file and the Nix attribute path
```
mboes, line 932:
```suggestion
This rule indicates the Nix file and the Nix attribute path
```

Bazel documentation confusingly uses rule and rule instance interchangeably.

docs/haskell-use-cases.rst
```diff
@@ -884,3 +884,127 @@ not possible to build docker images for haskell binaries as above using rules_do
...
+to reach the cross-compiler. It indicates to link a static
```
mboes, line 933:
```suggestion
to reach the cross-compiler. It says to link a static
```

(in particular to avoid repetition.)

mboes (approved):
Reads very well.

docs/haskell-use-cases.rst
facundominguez: It isn't needed if TH is not used, indeed.

Right now, the user can remove the `-fexternal-interpreter` flag in `arm-cross.nix` if she doesn't want TH. GHC doesn't start the interpreter unless some module uses TH, so it doesn't harm to leave the nix configuration as is if some targets use TH and others don't.

aherrmann (approved):
Very well written, thank you!

Issue #1059, opened 2019-08-27
ghcopts and cabalopts attributes


aherrmann:
**Describe the solution you'd like**
rules_haskell already allows passing additional options to GHC using the `compiler_flags` attribute. However, currently there is no mechanism to pass additional options to Cabal (`Setup.hs` really) in `haskell_cabal_library|binary`. Add an attribute `cabalopts` to allow users to define such options. For consistency `compiler_flags` should be renamed to `ghcopts`.

**Additional context**
`cabalopts` and `ghcopts` was suggested in https://github.com/tweag/rules_haskell/pull/1057#pullrequestreview-280142763. The PR #1057 allows to specify Cabal flags, i.e. arguments to `configure --flags`.

aherrmann:
This was implemented in #1394 and #1527.

PR #1527, opened 2021-04-06
rename `compiler_flags` options to `ghcopts`

ylecornec:
As mentioned in issue (#1059),
the parameters named `compiler_flags` of various macros and rules have been renamed to `ghcopts`. 

This is to keep naming consistent with the `cabalopts` parameter.

For now the old name still works but a deprecation message is printed,
and the build fails if both new and old named are used at the same time.

I modified some of the tests to use the new name, but not all of them yet
(since both names are supposed to work).

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -160,7 +160,7 @@ nixpkgs_local_repository(
     nix_file = "//nixpkgs:default.nix",
 )
 
-test_compiler_flags = [
+test_ghcopts = [
     "-XStandaloneDeriving",  # Flag used at compile time
     "-threaded",  # Flag used at link time
 
@@ -194,7 +194,7 @@ load(
 haskell_register_ghc_nixpkgs(
     attribute_path = "",
     cabalopts = test_cabalopts,
-    compiler_flags = test_compiler_flags,
+    ghcopts = test_ghcopts,
     haddock_flags = test_haddock_flags,
     locale_archive = "@glibc_locales//:locale-archive",
     nix_file_content = """with import <nixpkgs> {}; haskell.packages.ghc883.ghc""",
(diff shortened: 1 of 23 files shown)


mboes (commented):
Nice!

`compiler_flags` are used often enough in a large repo that printing a warning every time it is used is effectively forcing a pretty quick migration to `ghcopts`, lest users get spammed with _a lot_ of text on the terminal. And since `compiler_flags` instead of `ghcopts` is a benign enough name, we could support it in perpetuity, or at least on a very long deprecation cycle, with the only the documentation changing to say that the field is deprecated.

I believe this way we wouldn't need the extra helper functions in `workspace_utils.bzl`.

ylecornec:
Ok, I removed the warning when `compiler_flags` is used alone.

I think we must still fail if both the new and old name are used at the same time,
so I kept one helper function to check that.

But maybe `workspace_utils.bzl` is not the best place for it ?

aherrmann (commented):
Looks great, thank you!

> But maybe `workspace_utils.bzl` is not the best place for it ?

Yes, `workspace_util.bzl` is for repository rule helpers. Typically things that require a `repository_ctx`. That's not the case here.

The closest file that comes to mind is [`typing.bzl`](https://github.com/tweag/rules_haskell/blob/e39b87bcad0e8b720d88bea08c7ece9289857db7/haskell/private/typing.bzl) in the sense that it checks attributes for validity. The name `typing` is specific to type checking, but perhaps it could be renamed to `validate_attrs` or something like that.

ylecornec:
Alright, I renamed `typing.bzl` to `validate_attrs.bzl` and moved the `check_deprecated_attribute_usage` function there.

aherrmann (approved):
Thank you!

aherrmann:
@Mergifyio rebase

Issue #1528, opened 2021-04-07
Start script instructions ask for sh when it needs bash


tanyabouman:
**Describe the bug**
The [README](https://github.com/tweag/rules_haskell/blob/master/README.md) instructions for running the start script use `sh`.
```console
$ curl https://haskell.build/start | sh
```
or
```console
$ sh <(curl https://haskell.build/start) --use-nix
```

It returns an error:
```
sh: 146: declare: not found
```

**Expected behavior**
Start script runs without errors.  An easy solution would be to call the script with `bash` instead of `sh`, but maybe there's some reason to support `sh`.  That would mean replacing the parts of the start script which require `bash`.

**Environment**
 - OS name + version: Ubuntu 20.04.2 LTS
 - Bazel version: 3.3.1 and 4.0.0

**Additional context**
Running the script with bash works for me.
```console
$ curl https://haskell.build/start | bash
```
or
```console
$ bash <(curl https://haskell.build/start) --use-nix
```

mboes:
I believe your assessment is correct. Shellcheck concurs:

```
$ shellcheck -s sh start 

In start line 179:
declare -r ZLIB_BUILD_FILE="zlib.BUILD.bazel"
^------------------------^ SC2039: In POSIX sh, 'declare' is undefined.

For more information:
  https://www.shellcheck.net/wiki/SC2039 -- In POSIX sh, 'declare' is undefin...
```

We can either change the README or remove the use of `declare`. I vote for the latter. It doesn't add anything, and if don't use `declare`, Shellcheck flags no other issues.

mboes:
Interestingly, we already have a rule that lints the `start` script, but it states `bash` as the flavour: https://github.com/tweag/rules_haskell/blob/master/tests/shellcheck/BUILD.bazel#L67. If we keep the README as-is, then this line should be changed to say `sh_flavor = "sh"`.

facundominguez:
Fixed in #1530

PR #1530, opened 2021-04-07
make start script work with sh

tanyabouman:
This addresses issue #1528 by replacing `declare -r` with `readonly` in the script and updating lint rule to require sh rather than bash.


facundominguez (approved):
LGTM!

Issue #1432, opened 2020-09-24
have haskell_repl output absolute paths or paths relative to a different folder than the workspace one


teto:
**Is your feature request related to a problem? Please describe.**
This may be more of a bazel issue: I would like bazel to write paths relative to a different folder than the workspace root.

Specifically, calling 
```
	bazel build "//hie_bios" \
	--output_groups=hie_bios \
	--experimental_show_artifacts \
```
defined by
```
load("@rules_haskell//haskell:defs.bzl", "haskell_repl")

# See https://api.haskell.build/haskell/defs.html#haskell_repl
haskell_repl(
    name = "hie_bios",
    collect_data = False,
    deps = {deps},
    experimental_from_source = {deps},
    # Required because we might depend on tests, and bazel only allows
    # 'testonly' targets to depend on tests
    testonly = True,
)
```
returns 
```
-package-db
bazel-out/k8-fastbuild/bin/external/hackage/lrucaching-0.3.3/_install/lrucaching-0.3.3.conf.d
-package-db
bazel-out/k8-fastbuild/bin/external/hackage/haskell-lexer-1.1/_install/haskell-lexer-1.1.conf.d
-package-db
bazel-out/k8-fastbuild/bin/external/hackage/pretty-show-1.10/_install/pretty-show-1.10.conf.d
-package-db
bazel-out/k8-fastbuild/bin/external/hackage/unbounded-delays-0.1.1.0/_install/unbounded-delays-0.1.1.0.conf.d
-package-db
bazel-out/k8-fastbuild/bin/external/hackage/wcwidth-0.0.2/_install/wcwidth-0.0.2.conf.d
-package-db
bazel-out/k8-fastbuild/bin/external/hackage/tasty-1.2.3/_install/tasty-1.2.3.conf.d
-package-db
bazel-out/k8-fastbuild/bin/external/hackage/tasty-hunit-0.10.0.2/_install/tasty-hunit-0.10.0.2.conf.d
-pgma
bazel-out/k8-fastbuild/bin/external/rules_haskell/haskell/cc_wrapper-python
-pgmc
bazel-out/k8-fastbuild/bin/external/rules_haskell/haskell/cc_wrapper-python
-pgml
bazel-out/k8-fastbuild/bin/external/rules_haskell/haskell/cc_wrapper-python
-pgmP
bazel-out/k8-fastbuild/bin/external/rules_haskell/haskell/cc_wrapper-python -E -undef -traditional
[...]

PR #1531, opened 2021-04-08
added hie_bios_path_prefix attribute to haskell_repl

ylecornec:
I added an attribute named `hie_bios_path_prefix` to the `haskell_repl` rule to address issue #1432.

This makes it possible for hie-bios to output paths relative to a different folder than the workspace root.

In order to use absolute paths, it is possible to use a placeholder in the prefix,
and then replace it with a call to `pwd` in the user defined [.hie-bios script](https://rules-haskell.readthedocs.io/en/latest/haskell-use-cases.html#configuring-ide-integration-with-ghcide).

### DIFF
--- a/tests/hie-bios/BUILD.bazel
+++ b/tests/hie-bios/BUILD.bazel
@@ -1,0 +1,54 @@
+load(
+    "@rules_haskell//haskell:defs.bzl",
+    "haskell_repl",
+    "haskell_test",
+)
+
+haskell_test(
+    name = "binary-hie-bios-test",
+    srcs = ["Main.hs"],
+    deps = ["//tests/hackage:base"],
+)
+
+haskell_repl(
+    name = "hie-bios",
+    testonly = 1,
+    collect_data = True,
+    hie_bios_path_prefix = ["$magic_string"],
+    deps = [":binary-hie-bios-test"],
+)
+
+haskell_repl(
+    name = "hie-bios-no-prefix",
+    testonly = 1,
+    collect_data = True,
+    deps = [":binary-hie-bios-test"],
+)
+
+filegroup(
+    name = "hie-bios-file",
+    testonly = 1,
+    srcs = [":hie-bios"],
+    output_group = "hie_bios",
+)
+
+filegroup(
+    name = "hie-bios-file-no-prefix",
+    testonly = 1,
+    srcs = [":hie-bios-no-prefix"],
+    output_group = "hie_bios",
+)
+
+sh_test(
+    name = "test-hie-bios",
+    size = "small",
+    srcs = ["test_hie_bios.sh"],
+    args = [
+        "$(location :hie-bios-file)",
+        "$(location :hie-bios-file-no-prefix)",
+    ],
+    data = [
+        ":hie-bios-file",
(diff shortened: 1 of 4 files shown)


aherrmann (commented):
That looks good, thank you!

Could you add a test-case as well? We should test both cases: No prefix and some prefix. 

Perhaps an `sh_test` that has a `data` dependency on a `haskell_repl`'s hie-bios output would work.

ylecornec:
I added a test case which contains two `hs_repl` rules, one with the `hie_bios_path_prefix` option and one without which generate hie-bios files.

Then a `sh_test` rule check that the prefix appears in the first file, and that if we remove this prefix using sed, then the two files are equals.

Some notes : 
 - Since the test uses a shell script, I added the `dont_test_on_windows` tag.
 - I also added a newline at the end of the hie-bios files since sed on mac complained about it.
 - I left some debug print which can be useful in case the test fails later, is this good practice ?

tests/hie-bios/BUILD.bazel
```diff
@@ -0,0 +1,55 @@
...
+    srcs = ["test_hie_bios.sh"],
+    args = [
+        "$(rootpath :hie-bios-file)",
+        "$(rootpath :hie-bios-file-no-prefix)",
+    ],
+    data = [
+        ":hie-bios-file",
+        ":hie-bios-file-no-prefix",
+    ],
+    tags = ["dont_test_on_windows"],
```
aherrmann, line 54:
```suggestion
```
This shouldn't be necessary. rules_haskell requires msys2 on Windows anyway, so a bash is availabl. And Bazel can generate executables from `.sh` scripts on Windows if a bash is a available.

aherrmann (commented):
> I left some debug print which can be useful in case the test fails later, is this good practice ?

I think it's better to not print this output in the success case and only print it in the failure case.

---

It looks like there's a conflict with `master` in `haskell/repl.bzl`. Could you rebase on `master`?

tests/hie-bios/BUILD.bazel
aherrmann: Just saw the [Windows CI failure](https://github.com/tweag/rules_haskell/runs/2325545499#step:6:102):
```
grep: tests/hie-bios/hie-bios@hie-bios: No such file or directory
```
I think the issue has to do with runfiles. On Unix Bazel will create a tree of symlinks in the tests working directory pointing to `data` dependencies. However, on Windows symlinks are not generally available, so instead Bazel generates a runfiles manifest file and code needs to use a runfiles library to query the manifest file for the location of `data` dependencies. The bash runfiles library is documented [here](https://github.com/bazelbuild/bazel/blob/master/tools/bash/runfiles/runfiles.bash#L39-L68). Example uses can be found in the repository. In tests you can use the `TEST_WORKSPACE` env-var to find the workspace name, e.g.
```
file1="$(rlocation "$TEST_WORKSPACE/$1")"
```

ylecornec: Ah I see, thank you.

So I activated the windows test back on and it works with the bash runfiles library.
And now, debug information is only printed in case the test fails.

aherrmann (approved):
Thank you, looks good!

Issue #1419, opened 2020-09-13
Examples do not build


TomMD:
**Describe the bug**
The examples in rules_haskell do not build.

**To Reproduce**
On my OS-X machine

```
git clone https://github.com/tweag/rules_haskell
cd rules_haskell/examples
bazel build //cat_hs/exec/cat_hs
```

And observe

```
ERROR: /private/tmp/rules_haskell/examples/cat_hs/exec/cat_hs/BUILD.bazel:6:15 HaskellCabalLibrary @stackage//:unliftio-core failed (Exit 1) cabal_wrapper failed: error executing command bazel-out/host/bin/external/rules_haskell/haskell/cabal_wrapper lib:unliftio-core unliftio-core-0.1.2.0 true external/stackage/unliftio-core-0.1.2.0/Setup.hs external/stackage/unliftio-core-0.1.2.0 ... (remaining 7 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
INFO: Elapsed time: 3.009s, Critical Path: 2.85s
INFO: 20 processes: 20 darwin-sandbox.
FAILED: Build did NOT complete successfully
```

**Expected behavior**
I thought this would build a binary that could be run.

**Environment**
 - OS name + version: OS-X with bazel 3.14 from brew.
 - Bazel version: 3.14
 - Version of the rules: This was on git rev aabeedc18f5e5db030ca1aa0c10a7dc14e4a4a55

**Additional context**

TomMD:
I also tried on linux:

```
docker run --rm -it ubuntu:20.04 bash
# follow bazel instructions here https://docs.bazel.build/versions/master/install-ubuntu.html
apt update && apt install -y git
git clone https://github.com/tweag/rules_haskell
cd rules_haskell/examples
bazel build //cat_hs/exec/cat_hs
```

Yielding:

```
# bazel build //cat_hs/exec/cat_hs
Starting local Bazel server and connecting to it...
DEBUG: /root/.cache/bazel/_bazel_root/0d15a96fd12ed9dca003553c71536f42/external/rules_haskell/haskell/private/versions.bzl:60:10: WARNING: bazel version is too recent. Supported versions range from 2.1.0 to 3.3.1, but found: 3.5.0
INFO: Repository rules_haskell_python_local instantiated at:
  no stack (--record_rule_instantiation_callstack not enabled)
Repository rule _config_python3_toolchain defined at:
[...]

PR #1533, opened 2021-04-09
hopefully a more helpful error for when make is missing

tanyabouman:
The `build-essential` package is a required dependency, and when it is missing, `make` fails with an error that's not really friendly, as evidenced in #1419.
The new error also links to the other required dependencies, because there's a good chance that some other things are missing too.


aherrmann (approved):
Thank you!

aherrmann:
@Mergifyio rebase

PR #1534, opened 2021-04-12
Add boot_files field to HaskellInfo. Export boot files to hie_bios.

### LINKED ISSUE #1532: export hs-boot files to repl
When running haskell-language-server, it chokes on `SOURCE` imports
`import {-# SOURCE #-}           My.Module ()     ■ Could not find module`

Turns out this is supported in hls and if I add the `*.hs-boot` files to the flags loaded by the hie.yaml file it works fine (I am not a 100% sure yet but in the 2 cases I added the flags manually, the error went away).

I would like `bazel build "//hie_bios"   --output_groups=hie_bios   --experimental_show_artifacts ` to output the hs-boot files ?

I tried to hack my way around in `haskell/repl.bzl` where the sources are listed, seems like .hs-boot files end up in extra_haskell_sources (cf  `haskell/private/actions/compile.bzl`) but then retreieving the info required to have one more level of indirection so I wanted to confirm first what approach to adopt.

ak3n:
This PR should fix #1532. I haven't checked if haskell-language-server works, but `*.hs-boot` files are now listed in `.%s.hie-bios`.


haskell/private/haskell_impl.bzl
```diff
@@ -233,10 +233,17 @@ def _haskell_binary_common_impl(ctx, is_test):
         version = ctx.attr.version,
     )
 
+    boot_files = depset([
+        s
+        for s in c.extra_source_files.to_list()
+        if s.extension in ["hs-boot", "lhs-boot"]
+    ])
```
aherrmann, line 240: Can we push the separation all the way through? I.e. already have a separate `boot_files` returned by `compile_binary` and not include the boot files in `c.extra_source_files` in the first place? That would seem like a more consistent implementation.

ak3n: Please, take a look at my attempt. I tried not to forget every `extra_source_files` usage and provide the boot files.

aherrmann (approved):
Thank you, looks good to me!

aherrmann:
@Mergifyio rebase

teto:
haskell-language-server 1.0 suffers from a bug such that I can't confirm it worked for the whole project but in my manual testing it worked great. thank you

PR #1539, opened 2021-04-13
Fix links in the cross-compilation section

facundominguez:


facundominguez:
I surrender. I couldn't find a way to have named external links with a custom text. So I renamed some links and inlined others.

facundominguez:
It would be helpful to have de `deploy/netlify` job render the use-cases documentation too.

aherrmann (approved):

PR #1542, opened 2021-04-14
fix absolute paths in api doc

ylecornec:
When the documentation is served locally with `./serve-docs.sh`, the api documentation is under the `api/` folder.
But it is not the case on https://api.haskell.build, so some absolute links were broken.

I added a new build_setting for the documentation to add this `/api` prefix for absolute links if needed.
(when serving the doc locally, or later to be able to preview the guide next to the api doc on netlify).

So the `./serve-docs.sh` now builds the local documentation this way:

    bazel build //docs:api_html --//docs:render_dev_website

(Using relative links did not work because the same table of content is used at the root and in the `api` directory)


aherrmann (commented):
Nice! And a nice use of build settings and custom make variables.

Loading the style sheet also needs that path prefix. I think the following patch should fix it
``` diff
diff --git a/docs/BUILD.bazel b/docs/BUILD.bazel
index 5adb65c2..531447aa 100644
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -190,6 +190,7 @@ $(location @bazel_tools//tools/zip:zipper) c $@ %s
             ":index_html",
         ] + [":%s_html" % lib for lib in libraries]
     ]),
+    toolchains = [":set_site_end_location"],
     tools = ["@bazel_tools//tools/zip:zipper"],
 )
 
diff --git a/docs/templates/main.html b/docs/templates/main.html
index 57282dbb..02088354 100644
--- a/docs/templates/main.html
+++ b/docs/templates/main.html
@@ -30,7 +30,7 @@ Documentation generated by Stardoc
     <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
     <link rel="stylesheet" href="https://code.getmdl.io/1.1.1/material.green-light_blue.min.css">
     <script defer src="https://code.getmdl.io/1.1.1/material.min.js"></script>
-    <link rel="stylesheet" href="/main.css">
+    <link rel="stylesheet" href="$path_prefix$/main.css">
 $for(header-includes)$
 $header-includes$
 $endfor$
```

ylecornec:
Ah yes, thank you.

I did not include the following line yet:

     toolchains = [":set_site_end_location"],

I think it is only needed if we use the make variable `$(SITE_END_LOCATION)` in the rule ?
But here it is not the case.

aherrmann:
> I did not include the following line yet:
> 
> ```
>  toolchains = [":set_site_end_location"],
> ```
> 
> I think it is only needed if we use the make variable `$(SITE_END_LOCATION)` in the rule ?
> But here it is not the case.

Indeed, you're right. I had an outdated version in the browser cache and thought this change had fixed a missing style sheet when in fact it was just due to clearing the browser cache.

aherrmann (approved):

Issue #1525, opened 2021-04-02
use local packages in snapshot.yml


teto:
**Is your feature request related to a problem? Please describe.**
I am trying to update our ghc version to 8.10.4. As usual some packages dont support the platform yet so I am patching them.
To speed up the testing workflow, I tried to use local packages in my snapshot.yaml instead of github repositories to avoid having to update the `commit` bit (i.e., git revision).
```
resolver: lts-17.8

# - local-package
# - location: vendor/binary
packages:
  - /home/teto/streaming-conduit
```

I suppose it fails for good reasons (not reproducible) but I wonder if I could hack my way around ?
```
bazel build //... ; noti
INFO: Invocation ID: f98a3dfb-de70-47cc-9bd6-986f5be239b7
INFO: Repository hackage instantiated at:
  /home/teto/nova/jinko3/WORKSPACE:55:20: in <toplevel>
  /home/teto/.cache/bazel/_bazel_teto/b25a84786e6d9b137eae45b9dec297ea/external/buildlib/haskell.bzl:117:19: in jinko_haskell_setup
  /home/teto/.cache/bazel/_bazel_teto/b25a84786e6d9b137eae45b9dec297ea/external/rules_haskell/haskell/cabal.bzl:2374:20: in stack_snapshot
Repository rule _stack_snapshot defined at:
  /home/teto/.cache/bazel/_bazel_teto/b25a84786e6d9b137eae45b9dec297ea/external/rules_haskell/haskell/cabal.bzl:2027:34: in <toplevel>
ERROR: An error occurred during the fetch of repository 'hackage':
   Traceback (most recent call last):
	File "/home/teto/.cache/bazel/_bazel_teto/b25a84786e6d9b137eae45b9dec297ea/external/rules_haskell/haskell/cabal.bzl", line 1829, column 37, in _stack_snapshot_impl
		resolved = _resolve_packages(
	File "/home/teto/.cache/bazel/_bazel_teto/b25a84786e6d9b137eae45b9dec297ea/external/rules_haskell/haskell/cabal.bzl", line 1256, column 42, in _resolve_packages
		exec_result = _execute_or_fail_loudly(
	File "/home/teto/.cache/bazel/_bazel_teto/b25a84786e6d9b137eae45b9dec297ea/external/rules_haskell/haskell/private/workspace_utils.bzl", line 27, column 13, in execute_or_fail_loudly
[...]

PR #1544, opened 2021-04-16
Expand stack_snapshot use-case docs

aherrmann:
This expands the use-case documentation for `stack_snapshot`. It covers
- Simple use case of `stack_snapshot`
- Pinning with `stack_snapshot_json`
- Version overrides and Hackage dependencies
- C-library dependencies, e.g. `libz`
- Vendored packages
- Patching packages as a special case of vendored packages.

This was prompted by https://github.com/tweag/rules_haskell/issues/1525#issuecomment-818289717.


docs/haskell-use-cases.rst
```diff
@@ -308,20 +308,195 @@ However, you seldom want to use them directly. Cabal packages
...
+      visibility = ["//visibility:public"],
+  )
+  cc_library(name = "z", srcs = glob(["*.c"]), hdrs = glob(["*.h"]))
+  """,
+      sha256 = "c3e5e9fdd5004dcb542feda5ee4f0ff0744628baf8ed2dd5d66f8ca1197cb1a1",
+      strip_prefix = "zlib-1.2.11",
+      urls = ["http://zlib.net/zlib-1.2.11.tar.gz"],
+  )
+
+Vendoring packages
```
guibou, line 421: The difference between `vendored_packages` and `extra_deps` is unclear to me.

guibou (approved):

docs/haskell-use-cases.rst
```diff
@@ -308,20 +308,195 @@ However, you seldom want to use them directly. Cabal packages
...
+attribute of your Haskell targets. Note that ``base`` is a core package and its
+version is determined by the GHC toolchain and not the Stackage snapshot.
+
+Use the ``local_snapshot`` attribute to refer to a `custom Stack snapshot`_.
+
+Pinning
+^^^^^^^
+
+The ``stack_snapshot`` rule invokes ``stack`` for version and dependency
+resolution. By default this will happen on every fetch of the repository. You
```
mboes, line 349: Calling `stack` doesn't sound too onerous in of itself, and it's unclear how often or when "fetching" would happen. Perhaps we can tighten this point by making it more clear what the advantages of pinning are to the user? Like, pinning means that users don't even need Stack installed on their system first, right? Also, running Stack means requiring arbitrary network access and that's what makes it slow, right? Also, "repository" here is in the sense of "external repository", but that won't necessarily be all that obvious except for Bazel cognoscenti. Could say "external repository" explicitly, and make it a link to the Bazel doc.

docs/haskell-use-cases.rst
```diff
@@ -308,20 +308,195 @@ However, you seldom want to use them directly. Cabal packages
...
+
+  stack_snapshot(
+      ...
+      packages = [
+          ...
+          "optparse-helper-0.2.1.1",
+      ],
+  )
+
+C-library dependencies
```
mboes, line 379:
```suggestion
Non-Haskell dependencies
```

docs/haskell-use-cases.rst
mboes: The two attributes don't have the same type. `extra_deps` serves to state additional dependencies to Haskell packages. `vendored_packages` serves to replace Haskell packages with targets defined elsewhere. According to the respective sections. What could be said to make the distinction clearer?

mboes (approved):
Very nice! Just one section that could be clearer I think.

docs/haskell-use-cases.rst
guibou: Haa, got it. Perhaps not using the same `zlib` (the C library and the Haskell package) in the example may be more clear?

docs/haskell-use-cases.rst
aherrmann: I've expanded this to more clearly explain the issues and how pinning resolves them.

docs/haskell-use-cases.rst
aherrmann: I've changed it to use a vendored version of `hashable` and to illustrate on `unordered-containers` that the vendored version is injected into the dependency graph.

docs/haskell-use-cases.rst
aherrmann: > Like, pinning means that users don't even need Stack installed on their system first, right?

True, though this one is perhaps a bit less visible to users since rules_haskell will fetch a version of stack if it doesn't find one installed already.

teto:
cool PR. Thanks

Issue #1549, opened 2021-05-12
broken `@haddock` when using `ghc_plugin`


sellout:
**Describe the bug**

If a target depends on a `ghc_plugin`, then the `@haddock` for that target fails with "haddock: can't find a package database at bazel-out/k8-fastbuild/bin/{{{plugin library}}}" unless the target depends on the plugin's underlying library.

**To Reproduce**

```bazel
haskell_library(
    name = "dummy_plugin",
    srcs = ["Dummy.hs"],
    deps = [...],
)

ghc_plugin(
    name = "dummy",
    module = "Dummy",
    deps = [":dummy_plugin"],
)

haskell_library(
    name = "test_dummy",
    srcs = ["Dummy/Test.hs"],
    plugins = [":dummy"],
)
```
then `bazel build :test_dummy@haddock` will fail as described.

Building and using `:test_dummy` otherwise behaves correctly.

**Expected behavior**

Should build the haddock successfully.

**Environment**
 - OS name + version: NixOS (kernel 5.4)
 - Bazel version: 3.3.1
 - Version of the rules: From 5 Jan 2021, "d58e2318e84f0280002ae95569fa09f593c0482f"

**Additional context**

This has a workaround, which is to add the plugin library as an explicit dependency on the downstream library:

```bazel
haskell_library(
    name = "test_dummy",
    srcs = ["Dummy/Test.hs"],
    plugins = [":dummy"],
    deps = [":dummy_plugin"], # redundant, to make Haddock happy
)
```

We benefit here, I think, from the weakness of GHC's `-Wunused-packages` -- it doesn't always warn on transitive dependencies, and that seems to be the case here, as this builds cleanly.

aherrmann:
@sellout Thanks for reporting this and the repro instructions. I was able to reproduce this in https://github.com/tweag/rules_haskell/pull/1550.

I tried a quick fix by just adding plugin dependencies to the list of dependencies tracked in the [haddock aspect](https://github.com/tweag/rules_haskell/blob/86959463147dee44dbd0b962deddfe472ebb8bbf/haskell/haddock.bzl#L73). However, that was not sufficient. So, this needs a little more investigation.

Thanks for providing a work-around. I confirmed that it also works on the repro in #1550. Looking at the arguments passed to haddock the only difference are additional `--read-interface` flags:
```
[...]

PR #1550, opened 2021-05-18
Add failure test for haskell_doc with plugin

aherrmann:
Adds a regression test for https://github.com/tweag/rules_haskell/issues/1549.
This does not fix the issue, but adds a test-case that is expected to fail. It is tagged as `manual` and added to the failure tests in `//tests:run-tests` accordingly.


guibou (approved):
LGTM.

Issue #1552, opened 2021-06-03
Support fine grained module level build definitions


aherrmann:
# Goal

Add a rule `haskell_module` to support fine grained build definitions at the level of individual Haskell modules.

# Motivation

The existing `haskell_library|binary|test` rules operate on the level of [Haskell packages](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/packages.html#packages). In these rules building consist of two main steps with corresponding Bazel build actions: Compile all modules of a package in one build action and link all modules together into a library in a separate action (one for static and one for dynamic linking). Additionally, there is an action to construct the package's package database and several supporting actions.

This means that all modules of such a package need to be recompiled even if only a single module in a package changed compared to a previous cached build.

With a `haskell_module` rule we could instead issue one compile action per Haskell module. This would allow us to benefit from Bazel's caching and parallel builds on the level of individual modules.

# Outline

Not all the details are worked out at this stage, however, the idea is that the `haskell_module` rule may be used like this:

```
haskell_module(
    name = "ModuleA",
    src = "ModuleA.hs",
    deps = ["@stackage//:base"],
)
haskell_module(
    name = "ModuleB",
    src = "ModuleB",
    deps = [
        "@stackage//:base",
        ":ModuleA",
    ],
)
haskell_library(
    name = "my-library",
    modules = [
        ":ModuleA",
        ":ModuleB",
    ],
    deps = ["@stackage//:base"],
)
```
In this example the `haskell_module` rules perform the compilation of modules to object and interface files, and the `haskell_library` rule performs the linking of these compiled modules into a static archive or shared object and the generation of a package database for the package.

# Alternatives considered

The goal is to provide what we called "sub-target incremental builds" in the context of the current `haskell_library|binary|test` rules. I.e. avoid recompiling modules that have not changed between builds even if another module that is part of the same library or executabe target did change. An alternative approach that we considered is to use [Bazel's persistent worker](https://www.tweag.io/blog/2019-09-25-bazel-ghc-persistent-worker-internship/) mechanism to provide a persistent GHC session that caches previous compilation outputs (object and interface files). Downsides of this approach are that:
- It is a more complex solution, since it requires maintaining our own cache and maintaining state across worker invocations.
- It introduces potential inhermeticity, since the worker would persist state across build actions outside of Bazel's control.
- The benefits are limited to a worker instance, i.e. modules cannot be shared across build nodes through a remote cache.

# Anticipated problems

## Slower uncached builds

[...]

PR #1553, opened 2021-06-03
Provide an experimental haskell_module rule

aherrmann:
This is the first step towards https://github.com/tweag/rules_haskell/issues/1552.

- Adds a first implementation of `haskell_module` as an experimental rule.
  - Can compile a module to an object and interface file.
  - Can depend on other `haskell_module` targets.
  - Can depend on other `haskell_library` targets.
  - Does not yet integrate with `haskell_library|binary|test`. I.e. these rules cannot yet depend on `haskell_module` targets. This is left open for future PRs.
- Adds test cases for the currently supported use-cases of the `haskell_module` rule.

This rule is still in an early experimental stage and not yet usable in practice. However, I'm proposing to merge it in its current state marked as an experimental rule and defined separately from the other Haskell rules in order to avoid a long standing feature branch that regularly needs to rebased on `master` and enable incremental development on this rule with regular review on small PRs.

### DIFF
--- a/haskell/experimental/defs.bzl
+++ b/haskell/experimental/defs.bzl
@@ -20,6 +20,7 @@ _haskell_module = rule(
             allow_single_file = [".hs", ".lhs"],  #, ".hs-boot", ".lhs-boot", ".hsc", ".h"],
             mandatory = True,
         ),
+        "src_strip_prefix": attr.string(),
         "extra_srcs": attr.label_list(
             allow_files = True,
         ),
--- a/haskell/experimental/defs.bzl
+++ b/haskell/experimental/defs.bzl
@@ -1,0 +1,95 @@
+"""Experimental Haskell rules"""
+
+load(
+    "//haskell/experimental/private:module.bzl",
+    _haskell_module_impl = "haskell_module_impl",
+)
+load("//haskell:private/cc_libraries.bzl", "haskell_cc_libraries_aspect")
+
+_haskell_module = rule(
+    _haskell_module_impl,
+    # NOTE: Documentation needs to be added to the wrapper macros below.
+    #   Currently it is not possible to automatically inherit rule documentation
+    #   in wrapping macros. See https://github.com/bazelbuild/stardoc/issues/27
+    attrs = {
+        # TODO[AH] Merge with _haskell_common_attrs in //haskell:defs.bzl
+        "src": attr.label(
+            # TODO[AH] How to handle boot files?
+            # TODO[AH] How to handle .hsc files?
+            # TODO[AH] Do we need .h files in here?
+            allow_single_file = [".hs", ".lhs"],  #, ".hs-boot", ".lhs-boot", ".hsc", ".h"],
+            mandatory = True,
+        ),
+        "extra_srcs": attr.label_list(
+            allow_files = True,
+        ),
+        "deps": attr.label_list(
+            aspects = [haskell_cc_libraries_aspect],
+        ),
+        "ghcopts": attr.string_list(),
+        #"repl_ghci_args": attr.string_list(),
+        "plugins": attr.label_list(
+            aspects = [haskell_cc_libraries_aspect],
+        ),
+        "tools": attr.label_list(
+            cfg = "host",
+            allow_files = True,
+        ),
+        "_cc_toolchain": attr.label(
+            default = Label("@bazel_tools//tools/cpp:current_cc_toolchain"),
+        ),
+        "_ghc_wrapper": attr.label(
+            executable = True,
+            cfg = "host",
+            default = Label("@rules_haskell//haskell:ghc_wrapper"),
+        ),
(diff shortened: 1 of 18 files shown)


haskell/experimental/providers.bzl
```diff
@@ -0,0 +1,8 @@
+HaskellModuleInfo = provider(
+    doc = "Information about a single compiled Haskell module.",
+    fields = {
+        "object_file": "The compiled `.o` file.",
+        "interface_dir": "The import search directory for the interface file.",
```
Radvendii, line 5: For a while I mistakenly thought that this was just the directory the `interface_file` is in, and this kind of confused me.

After working with it for a bit I realized that the directory structure between `interface_dir` and `interface_file` is relevant to the module import path.

I don't know if this would be obvious to most people working with `rules_haskell` or if it's worth putting some kind of explanation.

haskell/experimental/defs.bzl
```diff
@@ -0,0 +1,127 @@
...
+    #    for attr in ["testonly", "tags"]
+    #    if attr in kwargs
+    #}
+    #haskell_repl(
+    #    name = "%s@repl" % name,
+    #    deps = [name],
+    #    experimental_from_source = [":%s" % name],
+    #    repl_ghci_args = [],
+    #    **repl_kwargs
+    #)
```
Radvendii, line 127: Why are these here but commented? Presumably they don't work yet but should be added later?

tests/haskell_module/nested/Branch/Left/BUILD.bazel
```diff
@@ -0,0 +1,12 @@
+load("@rules_haskell//haskell/experimental:defs.bzl", "haskell_module")
+
+haskell_module(
+    name = "Module",
+    src = "Module.hs",
+    src_strip_prefix = "/tests/haskell_module/nested",
```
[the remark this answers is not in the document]
aherrmann: I'm not sure I understand the question. GHC needs an import path here, otherwise, it won't find the interface file when another module imports `Branch.Left.Module`. I.e. if you remove this `src_strip_prefix`, then 
```
$ bazel build //tests/haskell_module/nested:LeafModule
```
fails with
```
tests/haskell_module/nested/LeafModule.hs:3:1: error:
    Could not find module ‘Branch.Left.Module’
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
  |
3 | import Branch.Left.Module
  | ^^^^^^^^^^^^^^^^^^^^^^^^^
Target //tests/haskell_module/nested:LeafModule failed to build
```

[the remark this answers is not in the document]
Radvendii: Oh, I understand. And since we're in a subdirectory of that directory here, it has to be workspace-relative. My mistake.

haskell/experimental/providers.bzl
aherrmann: Ah, yes, that is a misleading name. I've renamed it to `import_dir` to be consistent with the package configuration's `import-dirs` field and expanded the docstring a bit.

haskell/experimental/defs.bzl
aherrmann: Yes, I followed the pattern of `haskell_library` and left this part commented since it doesn't work, yet. However, there's no good reason to keep it around. I've removed it.

Radvendii: Oh, I didn't realize this part of the code was also taken from there. I should have checked.

Radvendii (approved):

Radvendii:
Two potential issues:
1. CI is failing for bindist / macos for some reason. It seemed unrelated to the changes though, so I'm trying just rerunning it.
1. Except for `single`, the tests included here don't actually define any bazel tests, so they don't get run by CI. Is this intentional?

aherrmann:
>  2\. Except for `single`, the tests included here don't actually define any bazel tests, so they don't get run (or built) by CI. Is this intentional?

@Radvendii Good catch, this is an inconsistency between `run-tests` and `bazel test` (one more reason to get rid of `run-tests`). `bazel test //...` [also builds targets that are not tests](https://github.com/bazelbuild/bazel/issues/4257#issuecomment-396268175), i.e. apart from scheduling it's equivalent to `bazel build //... && bazel test //...`. I've confirmed that the modules are built [here](https://github.com/tweag/rules_haskell/runs/3058263166#step:6:178) by introducing a syntax error in one of the modules. The bindist pipeline fails, however, the nixpkgs pipeline does not. The reason is that run-tests uses `bazel test //... --build_tests_only` while the bindist one uses `bazel test //tests/...`. This also explains why profiling related issues didn't show up on CI: Profiling mode is only tested by `run-tests` right now.

I think the best solution, short of getting rid of run-tests, is to align both to run `bazel test //...` without `--build_tests_only`. However, this is unrelated to this PR and best treated in a separate PR.

PR #1554, opened 2021-06-04
Remove .bak files when installing ghc from bindist

facundominguez:
.bak files don't affect builds, but they are declared as inputs, causing remote cache misses.

This commit just arranges to have .bak files removed after sed produces them.

### DIFF
--- a/haskell/ghc_bindist.bzl
+++ b/haskell/ghc_bindist.bzl
@@ -373,7 +373,7 @@ def _ghc_bindist_impl(ctx):
         # IMPORTANT: all these scripts have to be compatible with BSD
         # tools! This means that sed -i always takes an argument.
         execute_or_fail_loudly(ctx, ["sed", "-e", "s/RelocatableBuild = NO/RelocatableBuild = YES/", "-i.bak", "mk/config.mk.in"], working_directory = unpack_dir)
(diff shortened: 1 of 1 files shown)


haskell/ghc_bindist.bzl
```diff
@@ -373,6 +373,7 @@ def _ghc_bindist_impl(ctx):
         # IMPORTANT: all these scripts have to be compatible with BSD
         # tools! This means that sed -i always takes an argument.
         execute_or_fail_loudly(ctx, ["sed", "-e", "s/RelocatableBuild = NO/RelocatableBuild = YES/", "-i.bak", "mk/config.mk.in"], working_directory = unpack_dir)
+        execute_or_fail_loudly(ctx, ["rm", "-f", "config.mk.in.bak"], working_directory = unpack_dir)
```
aherrmann, line 376: Should that be `mk/config.mk.in.bak`? Or does `sed` produce the file in the current working directory?

facundominguez: Good catch
```suggestion
        execute_or_fail_loudly(ctx, ["rm", "-f", "mk/config.mk.in.bak"], working_directory = unpack_dir)
```

aherrmann (approved):
Thanks for finding and fixing this!

Issue #1506, opened 2021-02-28
Support Cabal named libraries


mboes:
**Is your feature request related to a problem? Please describe.**
Recent versions of Cabal support defining multiple libraries in a package. These additional libraries must be named. Compiling packages with named libraries fails. See #1439 for an example.

**Describe the solution you'd like**
Make it possible to build packages that define multiple libraries. Unclear how to do this without substantial changes to [`haskell/cabal.bzl`](https://github.com/tweag/rules_haskell/blob/master/haskell/cabal.bzl). If Stack can give us a list of *components*, rather than a list of packages when constructing a build plan, then we could add a `component` attribute to `haskell_cabal_library` and make each `haskell_cabal_library` (each with a separate value for `component`) depend on each other.

**Additional context**
See #1439.

aherrmann:
> If Stack can give us a list of _components_, rather than a list of packages when constructing a build plan, then we could add a `component` attribute to `haskell_cabal_library` and make each `haskell_cabal_library` (each with a separate value for `component`) depend on each other.

Relevant upstream feature request from a while ago: https://github.com/commercialhaskell/stack/issues/5275

ylecornec:
Yes I think. Choosing the component we want to build with `haskell_cabal_library` is possible with the `sublibrary_name` attribute.
And if using `stack_snapshot`, the components and their dependencies can be manually specified via the `components` and `components_dependencies` attributes.

Issue #1439, opened 2020-11-09
`dns-internal -any` internal library not visible from rules_haskell


teto:
**Describe the bug**

I am trying to build https://hackage.haskell.org/package/dns via a stack_snapshot but it fails with 
```
A clear and concise description of what the bug is.
bazel build @hackage//:dns
INFO: Invocation ID: 0bc6895d-90fe-4520-9ced-ec6fe4302334
INFO: Analyzed target @hackage//:dns (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: /home/teto/.cache/bazel/_bazel_teto/f3097620197e6c203645e1230d1ae88c/external/hackage/BUILD.bazel:1696:22: HaskellCabalLibrary @hackage//:dns failed (Exit 1) cabal_wrapper failed: error executing command bazel-out/k8-fastbuild/bin/external/rules_haskell/haskell/cabal_wrapper lib:dns dns-4.0.1 true external/hackage/dns-4.0.1/Setup.hs external/hackage/dns-4.0.1 ... (remaining 27 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
Setup.hs: Encountered missing or private dependencies:
dns-internal -any
```

dns-internal is defined within the dns package https://github.com/kazu-yamamoto/dns/blob/master/dns.cabal#L23 : what's the best way to deal with this with rules_haskell ?

**Environment**
 - OS name + version: nixos-unstable
 - Bazel version: >3
 - Version of the rules: current master

**Additional context**
This happens while trying to build https://hackage.haskell.org/package/mongoDB with ghc 8.8.4.

Profpatsch:
Named/internal libraries were only added to cabal a short while ago iirc; do we know what changes we need to do to the `Setup.hs` compilation/call to also support these?

aherrmann:
The Cabal rules `haskell_cabal_library` and `haskell_cabal_binary` each build only a [specific cabal component](https://github.com/tweag/rules_haskell/blob/6ad9e2b336d24e311137234f2315d593ca4a1c77/haskell/cabal.bzl#L416-L418). For a library that is `lib:<package-name>`. I assume if we pointed that to `lib:dns-internal` then it would build the internal library that's missing in this case.

[...]

PR #1555, opened 2021-06-08
cabal sublibraries

ylecornec:
This branch adds a way to expose a sub-library of a cabal package by adding an attribute named `sublibrary_name` to the  `haskell_cabal_library` rule.

As mentioned in #1506 something like this is needed to compile a cabal package that depends on its own sub-libraries (such as in #1439).

In this case, it is now possible to make one `haskell_cabal_library` call for each library and make them depend on each other.


aherrmann (approved):
Thank you! That looks good to me.

PR #1558, opened 2021-06-18
integrate haskell_module() dependencies into haskell_library() rule

Radvendii:
Depends on #1553 
^^Not sure if I'm doing this right, since it's still showing all of the changes in the diff, not just mine.

Allows `haskell_library()` rules to depend on `haskell_module()`s through a new `modules` attribute.

### DIFF
--- a/haskell/experimental/transitions.bzl
+++ b/haskell/experimental/transitions.bzl
@@ -1,0 +1,17 @@
+load("//haskell:private/pkg_id.bzl", "pkg_id")
+
+def _package_name_out_transition_impl(settings, attr):
+    package_name = getattr(attr, "package_name", None)
+    version = getattr(attr, "version", None)
+
+    # we don't have direct access to ctx.label here, so we've recreated it in
+    # "attr.label". See //haskell:defs.bzl
+    label = Label(attr.label_string)
+    my_pkg_id = pkg_id.new(label, package_name, version)
+    return {"//haskell/experimental:package_name_setting": pkg_id.to_string(my_pkg_id)}
+
+package_name_out_transition = transition(
+    implementation = _package_name_out_transition_impl,
+    inputs = [],
+    outputs = ["//haskell/experimental:package_name_setting"],
+)
--- a/haskell/private/actions/link.bzl
+++ b/haskell/private/actions/link.bzl
@@ -35,7 +35,7 @@ def merge_parameter_files(hs, file1, file2):
     )
     return params_file
 
-def _create_objects_dir_manifest(hs, posix, objects_dir, dynamic, with_profiling):
+def _create_objects_dir_manifest(hs, posix, objects_dir, extra_objects, dynamic, with_profiling):
     suffix = ".dynamic.manifest" if dynamic else ".static.manifest"
     objects_dir_manifest = hs.actions.declare_file(
         objects_dir.basename + suffix,
@@ -58,12 +58,14 @@ def _create_objects_dir_manifest(hs, posix, objects_dir, dynamic, with_profiling
         # for efficient caching. See
         # https://github.com/tweag/rules_haskell/issues/1126.
         command = """
-        "{find}" {dir} -name '*.{ext}' | "{sort}" > {out}
+        "{cat}" <("{find}" {dir} -name '*.{ext}') <(echo -n "{extra}") | "{sort}" > {out}
         """.format(
             find = posix.commands["find"],
             sort = posix.commands["sort"],
+            cat = posix.commands["cat"],
             dir = objects_dir.path,
             ext = ext,
+            extra = "\n".join([e.path for e in extra_objects]),
             out = objects_dir_manifest.path,
         ),
     )
@@ -158,6 +160,7 @@ def link_binary(
         hs,
         posix,
         objects_dir,
+        [],
         dynamic = dynamic,
         with_profiling = with_profiling,
     )
@@ -246,7 +249,7 @@ def _so_extension(hs):
     """
     return "dylib" if hs.toolchain.is_darwin else "so"
 
-def link_library_static(hs, cc, posix, dep_info, objects_dir, my_pkg_id, with_profiling):
+def link_library_static(hs, cc, posix, dep_info, objects_dir, extra_objects, my_pkg_id, with_profiling):
     """Link a static library for the package using given object files.
 
     Returns:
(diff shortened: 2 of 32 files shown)


aherrmann:
>  ^^Not sure if I'm doing this right, since it's still showing all of the changes in the diff, not just mine.

The "depends on" is correct, *dpulls* shows up in the list of checks. However, you also need to set the PR's base branch to that of #1553 so that the diff only shows the relative changes. (You can do that with the *edit* button next to the PR's title or when creating the PR.)

tests/haskell_module/library/BUILD.bazel
```diff
@@ -0,0 +1,37 @@
...
+package(default_testonly = 1)
+
+haskell_library(
+    name = "TestLib",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//tests/hackage:base",
+    ],
+    modules = [
+        "//tests/haskell_module/library:TestLibModule"
```
aherrmann, line 18: Why not just `:TestLibModule`?

haskell/private/haskell_impl.bzl
```diff
@@ -401,33 +408,36 @@ def haskell_library_impl(ctx):
...
         ls_modules = ctx.executable._ls_modules,
+        modules = modules,
         other_modules = other_modules,
         exposed_modules_reexports = exposed_modules_reexports,
         interfaces_dir = c.interfaces_dir,
         with_profiling = with_profiling,
     )
 
-    if srcs_files:
+    if srcs_files or modules:
```
aherrmann, line 418: Perhaps this could be factored out into a variable, `non_empty` or something of that sort.

haskell/private/actions/link.bzl
```diff
@@ -319,7 +323,7 @@ def link_library_dynamic(hs, cc, posix, dep_info, extra_srcs, objects_dir, my_pk
 
     args = hs.actions.args()
     args.add_all(["-optl" + f for f in cc.linker_flags])
-    args.add_all(["-shared", "-dynamic"])
+    args.add_all(["-shared", "-dynamic", "-fPIC"])
```
aherrmann, line 326: Is `-fPIC` needed on linking? IIRC GHC has some handling of these flags built-in, see [this note](https://github.com/tweag/rules_haskell/blob/5acb7432a55ff5c203043b9acdc7ab228cacebcf/haskell/private/actions/link.bzl#L195-L236).

haskell/experimental/transitions.bzl
```diff
@@ -0,0 +1,16 @@
+load("//haskell:private/pkg_id.bzl", "pkg_id")
+
+def _package_name_out_transition_impl(settings, attr):
+    package_name = getattr(attr, "package_name", None)
+    version = getattr(attr, "version", None)
+    # we don't have direct access to ctx.label here, so we've recreated it in
+    # "attr.label". See //haskell:defs.bzl
```
aherrmann, line 7: Ah, yes, that's tricky. Sadly, I don't have a better idea how to deal with this either. Nice workaround!

aherrmann (commented):
Nice work, thank you!

Hmm, GH actions is not picking this up. I think we need to add `pull_request` to the triggers in the workflow config.
@Radvendii you can also push to a branch of rules_haskell under the Tweag org. Then GH actions should pick it up.

tests/haskell_module/library/BUILD.bazel
Radvendii: Good question. XD

haskell/private/haskell_impl.bzl
Radvendii: Yeah, I was debating about this and don't remember why I didn't end up doing it. Given that I can't think of an objection now, I'll assume there wasn't one before either.

haskell/private/actions/link.bzl
Radvendii: I could have *sworn* it didn't compile without that, but now it does. Probably something I changed after that removed the need for it.

Radvendii:
I've pushed my changes to the `tweag/haskell_module_library_support` branch, but I don't see a way to change where the PR is coming from. Should I open a new PR from there and close this one?

aherrmann:
@Radvendii I believe Github picks up the checks based on git commit. At least the checks show up on this PR now. So, as long as you push to both `tweag/haskell_module_library_support` and your fork this PR should be up-to-date. So, should be fine to keep this PR.

CI shows some errors, though.

aherrmann:
The bindist failure occurs on a proto target:
```
ERROR: /home/runner/.cache/bazel/_bazel_runner/c91220d87f26118b19246641d5266d21/external/com_google_protobuf/BUILD:356:15: in //haskell:protobuf.bzl%_haskell_proto_aspect aspect on proto_library rule @com_google_protobuf//:descriptor_proto: 
	File "/home/runner/work/rules_haskell/rules_haskell/haskell/protobuf.bzl", line 216, column 105, in _haskell_proto_aspect_impl
		[hs_info, cc_info, coverage_info, default_info, library_info, output_groups] = _haskell_library_impl(patched_ctx)
	File "/home/runner/work/rules_haskell/rules_haskell/haskell/private/haskell_impl.bzl", line 357, column 23, in haskell_library_impl
		modules = ctx.attr.modules
Error: 'struct' value has no field or method 'modules'
```
I suspect the issue is that the [patched ctx](https://github.com/tweag/rules_haskell/blob/66f31670c6abc316ef047784f3ceaaa3e513bc8b/haskell/protobuf.bzl#L165-L186) needs to be extended to also include `attr.modules`.

About the nixpkgs failure I'm pretty puzzled. Some more debug output would be helpful. You could add `-v` to `ghcopts` and perhaps extend the GHC wrapper to print all args to `stderr`.

Radvendii:
Okay, I squashed all the commits into one and rebased onto the new version of the branch with `import_dir` isntead of `interface_dir`. 

The same code is pushed to `tweag/haskell_module_library_support`, so that's getting checked by CI.

It's worth looking through the new changes to make sure I did everything the right way.

Also, in the past I've definitely messed things up monkeying with git and rebasing and branches. I think everything is fine for this one, but I'm never sure.

aherrmann (approved):
:tada: Good work, thank you!

If you're happy with #1553 then you can approve that one and set the merge-queue label.

Afterwards you can merge this PR (#1558) by also setting the merge-queue label.
You may need to push once more to the tweag org branch after dpulls rebased this PR.

Radvendii:
I'm no longer able to reopen the PR or change what branch it's based at. I guess I'll open a new PR

aherrmann:
> I'm no longer able to reopen the PR or change what branch it's based at. I guess I'll open a new PR

Weird, not sure what went wrong with Mergify there.

PR #1564, opened 2021-07-06
add profiling support to haskell_module()

Radvendii:


Radvendii:
Note: This should not be passing. There are several tests under `//tests/haskell_module` that don't technically define any tests, so they're not run by CI

haskell/experimental/private/module.bzl
```diff
@@ -63,6 +69,15 @@ def haskell_module_impl(ctx):
...
         "-Wmissing-home-modules",
     ])
+    if with_profiling:
+        args.add_all([
+            "-prof",
+            "-fexternal-interpreter",
+            "-hisuf",
+            "p_hi",
+            "-osuf",
+            "p_o",
```
aherrmann, line 79: Are those still required when `-o` and `-ohi` are being set explicitly?

tests/haskell_module/single/test.bzl
```diff
@@ -6,24 +6,21 @@ def _default_info_test_impl(ctx):
...
-    asserts.equals(env, "Single.hi", hi_file.basename)
+    if with_profiling:
+        expected_files = ["Single.p_o", "Single.p_hi"]
+    else:
+        expected_files = ["Single.o", "Single.hi"]
+
+    for f in expected_files:
+        asserts.true(env, f in actual_files, "Expected file %s not produced" % f)
+    for f in actual_files:
+        asserts.true(env, f in expected_files, "Found unexpected file %s" % f)
```
aherrmann, line 23: Perhaps this could be simplified using [`asserts.set_equals`](https://github.com/bazelbuild/bazel-skylib/blob/fd75066f159234265efb8f838b056be5a2e00a59/lib/unittest.bzl#L502)?

aherrmann (approved):
Thank you, looks good!

haskell/experimental/private/module.bzl
Radvendii: Yes! This was exactly the headache I caused myself, because I thought the same thing, removed this, and the more complicated tests started failing.

It will generate the correct files even without this, but it won't *look* for the right files when you pass in other profiling-enabled modules as dependencies. It will look for `.hi` files instead of `.p_hi` files.

tests/haskell_module/single/test.bzl
Radvendii: It could, but I thought the more precise error message this gives (exactly which files are / are not present) would be more useful.

I suppose I could rephrase the error message to "Exepected files: blah. Got files: blah instead"

I'm happy to change it if you think that's better.

haskell/experimental/private/module.bzl
Radvendii: Also worth noting that `haskell_library()` does both set `-o`, `-ohi`, and `-osuf`, `-hisuf`

aherrmann: > It will generate the correct files even without this, but it won't _look_ for the right files when you pass in other profiling-enabled modules as dependencies. It will look for `.hi` files instead of `.p_hi` files.

That makes sense, thanks for clarifying!

tests/haskell_module/single/test.bzl
aherrmann: > It could, but I thought the more precise error message this gives (exactly which files are / are not present) would be more useful.

That is useful, indeed. [Looking at the code](https://github.com/bazelbuild/bazel-skylib/blob/fd75066f159234265efb8f838b056be5a2e00a59/lib/unittest.bzl#L423-L426), though, `asserts.set_equals` should produce that kind of message.

Radvendii: Good catch. switched it to using `asserts.set_equals`

aherrmann: Thank you! In that case the sets need to be first constructed with [`sets.make`](https://github.com/bazelbuild/bazel-skylib/blob/fd75066f159234265efb8f838b056be5a2e00a59/lib/sets.bzl)

Radvendii: Ah, thank you. I saw that error on CI. For some reason it's not happening when I run the test locally :eyes:

aherrmann:
Just saw this wasn't merged yet. @Radvendii when it's approved you can add the `merge-queue` label to get it merged by mergify.

PR #1565, opened 2021-07-07
add hs-boot support to haskell_module()

Radvendii:
depends on: #1564 

I'm not sure if this is the right way to do hs-boot support in haskell_module(). In particular, it would be nice if the `.hs` and `.hs-boot` files could be passed in to the same `haskell_module()` command. To use this, however, would create a cyclic dependency in bazel, which I'm pretty sure it won't be happy about.

Instead the way I've implemented it you have to create a separate module for the `.hs-boot` file by itself.

### DIFF
--- a/haskell/experimental/defs.bzl
+++ b/haskell/experimental/defs.bzl
@@ -15,10 +15,9 @@ _haskell_module = rule(
         # TODO[AH] Merge with _haskell_common_attrs in //haskell:defs.bzl
         "package_name": attr.string(),
         "src": attr.label(
-            # TODO[AH] How to handle boot files?
             # TODO[AH] How to handle .hsc files?
             # TODO[AH] Do we need .h files in here?
-            allow_single_file = [".hs", ".lhs"],  #, ".hs-boot", ".lhs-boot", ".hsc", ".h"],
+            allow_single_file = [".hs", ".lhs", ".hs-boot"],  #, ".hs-boot", ".lhs-boot", ".hsc", ".h"],
             mandatory = True,
         ),
         "src_strip_prefix": attr.string(),
--- a/haskell/experimental/defs.bzl
+++ b/haskell/experimental/defs.bzl
@@ -17,7 +17,7 @@ _haskell_module = rule(
         "src": attr.label(
             # TODO[AH] How to handle .hsc files?
             # TODO[AH] Do we need .h files in here?
-            allow_single_file = [".hs", ".lhs", ".hs-boot"],  #, ".hs-boot", ".lhs-boot", ".hsc", ".h"],
+            allow_single_file = [".hs", ".lhs", ".hs-boot", ".lhs-boot"],  #, ".hsc", ".h"],
             mandatory = True,
         ),
         "src_strip_prefix": attr.string(),
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -46,7 +46,7 @@ def haskell_module_impl(ctx):
 
     # Determine outputs
     with_profiling = is_profiling_enabled(hs)
-    hs_boot = paths.split_extension(src.path)[1] == ".hs-boot"
+    hs_boot = paths.split_extension(src.path)[1] in [".hs-boot", ".lhs-boot"]
     extension_template = "%s"
     if hs_boot:
         extension_template = extension_template + "-boot"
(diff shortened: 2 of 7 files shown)


Radvendii:
Two further things to investigate before this is ready:
1. The test is failing with profiling on.
1. The [GHC guide on this](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/separate_compilation.html#how-to-compile-mutually-recursive-modules) specifically says not to link the `.o-boot` file that's produced. I haven't put in any specific code to exclude that, and it doesn't seem to be breaking. I don't know if that's because (a) it's not actually a problem to link it, (b) our code excludes it for other reasons, or (c) I just got lucky with this test.

Radvendii:
I fixed the profiling issue. The way I fixed it though, produces a `.p_hi-boot` file if profiling is enabled and we have a `.hs-boot` file as input. I don't know if this is the right behaviour, but it's all internal so it shouldn't matter?

Radvendii:
Regarding linking the `.o-boot` file, we should be all set as long as the `haskell_module()` that contains the `.o-boot` is not passed in as a `module` to a `haskell_library()` rule, which it has no reason to be anyways. Transitive dependencies are not included.

haskell/experimental/defs.bzl
```diff
@@ -15,10 +15,9 @@ _haskell_module = rule(
         # TODO[AH] Merge with _haskell_common_attrs in //haskell:defs.bzl
         "package_name": attr.string(),
         "src": attr.label(
-            # TODO[AH] How to handle boot files?
             # TODO[AH] How to handle .hsc files?
             # TODO[AH] Do we need .h files in here?
-            allow_single_file = [".hs", ".lhs"],  #, ".hs-boot", ".lhs-boot", ".hsc", ".h"],
+            allow_single_file = [".hs", ".lhs", ".hs-boot"],  #, ".hs-boot", ".lhs-boot", ".hsc", ".h"],
```
aherrmann, line 20:
```suggestion
            allow_single_file = [".hs", ".lhs", ".hs-boot", ".lhs-boot"],  #, ".hsc", ".h"],
```
I think `.lhs-boot` files should work exactly the same.

haskell/experimental/private/module.bzl
```diff
@@ -46,12 +46,20 @@ def haskell_module_impl(ctx):
 
     # Determine outputs
     with_profiling = is_profiling_enabled(hs)
+    hs_boot = paths.split_extension(src.path)[1] == ".hs-boot"
```
aherrmann, line 49:
```suggestion
    hs_boot = paths.split_extension(src.path)[1] in [".hs-boot", ".lhs-boot"]
```
Following the above.

aherrmann (commented):
> I'm not sure if this is the right way to do hs-boot support in haskell_module(). In particular, it would be nice if the `.hs` and `.hs-boot` files could be passed in to the same `haskell_module()` command. To use this, however, would create a cyclic dependency in bazel, which I'm pretty sure it won't be happy about.

I think requiring a separate target for the boot file is the right approach. As you say, trying to place both into the same target would lead to a cycle in Bazel's dependency graph, which is forbidden.

> The way I fixed it though, produces a `.p_hi-boot` file if profiling is enabled and we have a `.hs-boot` file as input. I don't know if this is the right behaviour, but it's all internal so it shouldn't matter?

That sounds correct. The other rules (`haskell_library`, etc.) have the same behavior. E.g. `bazel build //tests/hs-boot/... -c dbg` produces such `.p_hi-boot` and `.p_o-boot` files.

> Regarding linking the `.o-boot` file, we should be all set as long as the `haskell_module()` that contains the `.o-boot` is not passed in as a `module` to a `haskell_library()` rule, which it has no reason to be anyways. Transitive module dependencies are not linked in.

I think it would be good to raise an error to the user via `fail` if they get this wrong.

haskell/experimental/defs.bzl
Radvendii: Yep, good call. Should be fixed

haskell/experimental/private/module.bzl
Radvendii: Fixed this too

aherrmann (approved):
Looks good, thank you!

aherrmann:
@Radvendii There seems to be a merge conflict. Looking at the list of commits perhaps because one of the profiling commits is contained in this PR?

Radvendii:
Good catch. I rebased and I think it fixed it

PR #1566, opened 2021-07-08
Don't set -fplugin by default if plugin.module is empty

facundominguez:
This is necessary when the plugin fails if run on some modules of a package.

### DIFF
--- a/haskell/plugins.bzl
+++ b/haskell/plugins.bzl
@@ -19,7 +19,7 @@ ghc_plugin = rule(
     ghc_plugin_impl,
     attrs = {
         "module": attr.string(
-            doc = "Plugin entrypoint.",
+            doc = "Plugin entrypoint. Omit it to prevent -fplugin= from being set by default.",
         ),
         "deps": attr.label_list(
             doc = "Plugin dependencies. These are compile-time dependencies only.",
--- a/haskell/private/actions/compile.bzl
+++ b/haskell/private/actions/compile.bzl
@@ -303,9 +303,10 @@ def _compilation_defaults(hs, cc, java, posix, dep_info, plugin_dep_info, srcs, 
 
     # Plugins
     for plugin in plugins:
-        args.add("-fplugin={}".format(plugin.module))
-        for opt in plugin.args:
-            args.add_all(["-fplugin-opt", "{}:{}".format(plugin.module, opt)])
+        if plugin.module != "":
+            args.add("-fplugin={}".format(plugin.module))
+            for opt in plugin.args:
+                args.add_all(["-fplugin-opt", "{}:{}".format(plugin.module, opt)])
 
     plugin_tool_inputs = depset(transitive = [plugin.tool_inputs for plugin in plugins])
     plugin_tool_input_manifests = [


haskell/plugins.bzl
```diff
@@ -19,7 +19,7 @@ ghc_plugin = rule(
     ghc_plugin_impl,
     attrs = {
         "module": attr.string(
-            doc = "Plugin entrypoint.",
+            doc = "Plugin entrypoint. Omit it to prevent -fplugin= from being set by default.",
         ),
```
aherrmann, line 23: What is the meaning of a `ghc_plugin` with an empty `module`? My understanding is that without `-fplugin=<Module>` no plugin will be run and the `ghc_plugin` could be omitted entirely. What am I missing?

facundominguez: `ghc_plugin` can't be omitted if some modules use the pragma `{-# GHC_OPTIONS -fplugin=<module> #-}`

facundominguez: The difference with the `module` attribute is that the user chooses which modules use the plugin.

aherrmann: Oh, I see, thanks for clarifying!

That makes me wonder what the interface to this should look like. Should this really be configured on the plugin target or should this be configured at the use site on the `haskell_*` target that uses the plugin? Any thoughts on this?

I suspect that configuring this on the plugin could lead to confusion. E.g. the following library target definition would make me expect that only `PluginA` is applied to all modules of the target.
```
haskell_library(
    name = "lib",
    plugins = [":plugin-a", "plugin-b"],
    ghcopts = ["-fplugin=PluginA"],
    ...
)
```
But, that would not be the case if the plugins are defined as follows.
```
haskell_plugin(name = "plugin-a", ...)
haskell_plugin(name = "plugin-b", module = "PluginB", ...)
```

facundominguez: It is a matter of settling on a spec. I don't use plugins a lot. My preference would be for `ghc_plugin` to never set `-fplugin`. In which case the `module` attribute can be eliminated.

Another view at it, is that this PR is fixing `ghc_plugin` which misbehaves if the optional `module` attribute isn't specified, and the documentation does say that the `module` attribute is optional.

I'm fine going either way here.

aherrmann: As an alternative approach: Looking at the GHC user-guide I see that GHC has a flag to disable previously enabled plugins: [`-fclear-plugins`](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/extending_ghc.html#ghc-flag--fclear-plugins). Would this be a viable option for the issues that you encountered? I.e. leave `ghc_plugin` as is (even make `module` mandatory) and add a GHC options pragma with `-fclear-plugins` on the failing modules.

---

> It is a matter of settling on a spec. I don't use plugins a lot. My preference would be for `ghc_plugin` to never set `-fplugin`. In which case the `module` attribute can be eliminated.

Never setting `-fplugin` is indeed a simple solution. Having to repeat `-fplugin=...` everywhere seems like an annoyance, though. It's also a pretty subtle breaking change. There may be cases where a plugin not being enabled goes unnoticed.

> Another view at it, is that this PR is fixing `ghc_plugin` which misbehaves if the optional `module` attribute isn't specified, and the documentation does say that the `module` attribute is optional.

The part of the documentation that says that the `module` attribute is optional is auto-generated by stardoc based on the fact that `mandatory = True` is not set on the attribute definition and Bazel defaults to `mandatory = False`. Given that the current implementation emits invalid GHC flags and fails the build if `module` is empty, I would assume that this was not intentional and is an oversight from when `ghc_plugin` was added.

facundominguez: > As an alternative approach: Looking at the GHC user-guide I see that GHC has a flag to disable previously enabled plugins: -fclear-plugins.

This is fixing a problem that rules_haskell is introducing itself, so it looks convoluted. Also it affects all plugins not just the one that is making compilation fail.

> I suspect that configuring this on the plugin could lead to confusion. E.g. the following library target definition would make me expect that only PluginA is applied to all modules of the target.

This is a confusion that I had until I "discovered" how `ghc_plugin` works. Specifying a dependency on a plugin is not quite the same as running it on every module!

> Having to repeat -fplugin=... everywhere seems like an annoyance, though.

It is how I do it with plugins in my experience. "Everywhere" means the `compiler_flags` attribute of the rules that use the plugin, so it is not as bad as every module. I've also seen some code bases use a global list of flags that are used, which also reduces the repetition. 

> There may be cases where a plugin not being enabled goes unnoticed.

Removing a required attribute is hardly a subtle breaking change. It certainly won't go unnoticed. You can't use `ghc_plugin` without `module` today.

aherrmann: > > There may be cases where a plugin not being enabled goes unnoticed.
> 
> Removing a required attribute is hardly a subtle breaking change. It certainly won't go unnoticed. You can't use `ghc_plugin` without `module` today.

By subtle I mean the use-site at a `haskell_*` target. Removing the `module` attribute will only point at the `ghc_plugin` target, not at any of its use-sites.

---

Another question that this raises: How to deal with the `args` attribute of `ghc_plugin`. With this PR in it's current form it is silently ignored if `module` is left empty. That seems problematic as it means that `args` is silently ignored in the following kind of use-case:

```
ghc_plugin(name = "plugin-foo", args = ["foo", "bar"], ...)
haskell_library(name = "lib", plugins = [":plugin-foo"], ghcopts = ["-fplugin=PluginFoo"], ...)
```

Notably, `args` cannot function without knowing the module name. So, if we remove `module` we'd also have to remove `args`.

How does GHC handle `-fplugin-opt` flags without a corresonding `-fplugin` flag? Are those considered an error, or does GHC simply ignore them?

facundominguez: > How does GHC handle -fplugin-opt flags without a corresonding -fplugin flag? Are those considered an error, or does GHC simply ignore them?

Yes, ghc does ignore the arguments when -fplugin is omitted.

aherrmann: > Update: Or maybe add instead a field `non_default_plugins`. And keep `plugins` with the meaning it has today, which would be equivalent to `default_plugins` as defined above.

I like that option. Considering an example where `Plugin.A` should always be enabled and `Plugin.B` only on demand it would look like this:
```
haskell_library(
    name = "lib",
    plugins = [":plugin-a"],
    non_default_plugins = [":plugin-b"],
)
```
whereas the `default_plugins` approach (or perhaps it could be `enable_plugins` defaulting to `True`) would look something like this:
```
haskell_library(
    name = "lib",
    plugins = [":plugin-a", ":plugin-b"],
    enable_plugins = False,
    ghcopts = ["-fplugin=Plugin.A"],
)
```
The first option seems preferable to me: It requires fewer attributes and plugin A only needs to be mentioned once.

This approach is also good for consistency with `haskell_module` going forward, where we'd just not have the `non_default_plugins` attribute on `haskell_module`.

> > How does GHC handle -fplugin-opt flags without a corresonding -fplugin flag? Are those considered an error, or does GHC simply ignore them?
> 
> Yes, ghc does ignore the arguments when -fplugin is omitted.

That's good to know. I think in that case it would make sense to always pass the `ghc_plugin`'s `args` via `-fplugin-opt` even if the plugin is part of `non_default_plugins`. So that the user only needs to enable the plugin in the appropriate modules via a `-fplugin` pragma without having to repeat common `-fplugin-opt` flags.

facundominguez:
Superseded by #1578

PR #1568, opened 2021-07-13
integrate haskell_module() dependencies into haskell_library() rule

Radvendii:
Repeat of #1558 since that one got stuck on closed


aherrmann (approved):
Thank you!

PR #1569, opened 2021-07-16
Test `bazel test //...` in all configurations

aherrmann:
As discussed in https://github.com/tweag/rules_haskell/pull/1553#issuecomment-876459314 and https://github.com/tweag/rules_haskell/pull/1553#issuecomment-879215867 `bazel test //...` was previously not fully tested in all configurations. The nixpkgs pipelines only tested `bazel test //... --build_tests_only` and the bindist configurations only tested `bazel test //tests/...`. This PR changes this so that all configurations and the coverage tests use `bazel test //...`. This required adding exclusion tags to some targets that don't work in all configurations.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -52,7 +52,7 @@ jobs:
             ./tests/run-start-script.sh --use-nix
             bazel build //tests:run-tests
             ./bazel-ci-bin/tests/run-tests
-            bazel coverage //tests/...
+            bazel coverage //...
             bazel build //docs:api_html
             bazel build //docs:guide_html
             '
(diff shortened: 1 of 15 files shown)


Radvendii:
Why was it like this before? Are we assuming it's because of the tests that fail in some configurations?

aherrmann:
> Why was it like this before? Are we assuming it's because of the tests that fail in some configurations?

Historical reasons. The nixpkgs pipelines existed first and the bindist pipelines were only introduced later and only incrementally. Some parts still require Nix provided tools that have no replacement in the bindist toolchains, e.g. sphinx in `//docs/...`.

Radvendii (approved):
Okay, seems good to me.

aherrmann:
> GitHub App like Mergify are not allowed to merge pull request where .github/workflows is changed.

Requires manual merge.

PR #1572, opened 2021-07-27
Set the CLASSPATH for repl targets

facundominguez:


facundominguez:
cc @buonuomo

haskell/repl.bzl
```diff
@@ -463,7 +482,7 @@ def _create_repl(hs, posix, ctx, repl_info, output):
                 hs.tools.ghci,
                 ghci_repl_script,
             ],
-            transitive_files = inputs,
+            transitive_files = depset(transitive = [inputs, repl_info.java_deps]),
```
facundominguez, line 485: I would have thought the runfiles of the haskell library should contain already the jars, but in my tests it doesn't. So I added them here. What am I missing?

facundominguez:
If you have any suggestions on how to add a test involving the repl, I could give it a try.

facundominguez:
> If you have any suggestions on how to add a test involving the repl, I could give it a try.

hm, I think I could try something similar to:
https://github.com/tweag/rules_haskell/blob/master/tests/RunTests.hs#L67-L69

haskell/repl.bzl
```diff
@@ -74,6 +76,7 @@ HaskellReplCollectInfo = provider(
     fields = {
         "load_infos": "Dictionary from labels to HaskellReplLoadInfo.",
         "dep_infos": "Dictionary from labels to HaskellReplDepInfo.",
+        "java_deps": "depset of Files to jars needed for building.",
```
aherrmann, line 79: Could this be carried in `HaskellReplLoadInfo` instead?

The idea here is that `HaskellReplCollectInfo` collects "load from source" and "load from binary" information separately for each target associated to the target's label. This information is then combined into `HaskellReplInfo` later on based on the `from_source` and `from_binary` patterns. Adding `java_deps` directly here changes that design. Is there a good reason to do so?

---

If I understand correctly then JARs are only compile time dependencies for targets loaded from source to interpret inline-java TH splices. With targets loaded from binary they are only runtime dependencies and so should only be made available if `collect_data = True` to be consistent. Does that make sense?

haskell/private/haskell_impl.bzl
```diff
@@ -320,11 +320,12 @@ def _haskell_binary_common_impl(ctx, is_test):
             ctx.file._bash_runfiles,
             hs.toolchain.tools.hpc,
             binary,
-        ] + mix_runfiles + srcs_runfiles
+        ] + mix_runfiles + srcs_runfiles + java.inputs.to_list()
 
     return [
         hs_info,
         cc_info,
+        java,
```
aherrmann, line 328: This makes `JavaInteropInfo` part of the public rule interface. I don't think that's necessary to accomplish this. Instead, you could change `java_interop_info` so that it only depends on `deps` and then call it inside `_create_HaskellReplCollectInfo`.

facundominguez: Done.

haskell/repl.bzl
```diff
@@ -432,6 +444,13 @@ def _create_repl(hs, posix, ctx, repl_info, output):
 
     env = dict(hs.env)
 
+    classpath_inputs = collections.uniq([
+        paths.join("$RULES_HASKELL_EXEC_ROOT", f.path)
+        for f in repl_info.java_deps.to_list()
+    ])
```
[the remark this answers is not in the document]
facundominguez: I copied it from the implementation of `java_interop_info`, I removed it for now.

haskell/repl.bzl
facundominguez: I thought that was the case. But after the latest edits it seems I don't need it anymore. I might have tested it wrong before. I just removed it.

haskell/repl.bzl
facundominguez: > If I understand correctly then JARs are only compile time dependencies for targets loaded from source to interpret inline-java TH splices. 

JARs are both compile and runtime dependencies in this case.

> With targets loaded from binary they are only runtime dependencies and so should only be made available if collect_data = True to be consistent.

This makes sense.

facundominguez: I still don't understand the propose rearrangement.

If I move `java_deps` to `HaskellReplLoadInfo`, how do I keep track of the jars that need to be added to the classpath when the target is loaded from binary?

facundominguez: hm, strictly speaking, the CLASSPATH doesn't need to be set when running. But it would be a convenient way to signal to the program the list of jar paths to feed to the JVM when initializing it.

[This issue](https://github.com/tweag/inline-java/issues/179) is related. Maybe we should handle this concern separately.

facundominguez: I moved `java_deps` to `HaskellReplLoadInfo`.

aherrmann: > If I move `java_deps` to `HaskellReplLoadInfo`, how do I keep track of the jars that need to be added to the classpath when the target is loaded from binary?

Right, so they would also be needed in `HaskellReplDepInfo` for the runtime deps case.

aherrmann (approved):
Thank you, looks good. Thank you for adding the test case!

haskell/repl.bzl
aherrmann: > hm, strictly speaking, the CLASSPATH doesn't need to be set when running. But it would be a convenient way to signal to the program the list of jar paths to feed to the JVM when initializing it.
> 
> [This issue](https://github.com/tweag/inline-java/issues/179) is related. Maybe we should handle this concern separately.

Oh, I remember. Yes, happy to treat this separately.

facundominguez:
Thanks for reviewing!

Issue #1573, opened 2021-07-28
a cabal_haskell_library that depends on a haskell_library fails to build with profiling.


ylecornec:
When building a `haskell_cabal_library` rule with profiling, Cabal seems to be looking for a non profiling version of its `haskell_library` dependencies, which do not exist.

We wonder if cabal tries to do two versions of the build as described [here](https://downloads.haskell.org/ghc/8.10.1/docs/html/users_guide/glasgow_exts.html#using-template-haskell-with-profiling).

PR #1574, opened 2021-07-28
test for a cabal_library that depends on a haskell_library

ylecornec:
This adds a test where a `haskell_cabal_library` target depends on a regular `haskell_library`.

This target fails to build in the profiling case as described in issue (#1573).


aherrmann (approved):
Thank you!

Issue #1247, opened 2020-02-18
data-files are not propagated from `haskell_cabal_library` to `haskell_binary`


cocreature:
**Describe the bug**

If you have a `haskell_cabal_library` that uses `data-files` and you depend on it from a `haskell_binary`, the `data-files` will not be found. This is only an issue for `haskell_binary`. `haskell_cabal_binary` works fine.

**To Reproduce**

I’ve added a (failing) testcase to `rules_haskell` here https://github.com/cocreature/rules_haskell/commit/dcc6d9116cf2a892f25d21bba33e52b31ad428fd

**Expected behavior**

The `data-files` should be located properly.

**Environment**
 - Version of the rules: 1fedad7e64ea00ecf1bcb4be70ca51a11dc48312

No clue what exactly is going wrong or how to fix this. Presumably cabal does some magic to make the `pkgroot`-relative `data-dir` work that is missing from `rules_haskell`.

Profpatsch:
Thanks for the test case!

aherrmann:
This is resolved by using a Bazel generated `Paths_` module. See https://github.com/tweag/rules_haskell/pull/1575.
The following patch on-top of https://github.com/cocreature/rules_haskell/commit/dcc6d9116cf2a892f25d21bba33e52b31ad428fd achieves as much:
```patch
diff --git a/tests/haskell_cabal_library/BUILD.bazel b/tests/haskell_cabal_library/BUILD.bazel
index f32547ff..81bd2ba5 100644
--- a/tests/haskell_cabal_library/BUILD.bazel
+++ b/tests/haskell_cabal_library/BUILD.bazel
@@ -23,6 +23,7 @@ haskell_cabal_library(
         "lib.cabal",
         "foobar",
     ],
+    generate_paths_module = True,
     cabalopts = [
         "--ghc-option=-DONE=1",
         "--haddock-option=--optghc=-DONE=1",
@@ -40,7 +41,7 @@ haskell_cabal_library(
         "dont_test_on_windows",
     ],
     version = "0.1.0.0",
-    deps = [":c-lib"],
+    deps = [":c-lib", "//tools/runfiles"],
 )
 
 haskell_cabal_library(
diff --git a/tests/haskell_cabal_library/lib.cabal b/tests/haskell_cabal_library/lib.cabal
index fb1f7c27..355f3176 100644
--- a/tests/haskell_cabal_library/lib.cabal
+++ b/tests/haskell_cabal_library/lib.cabal
[...]

PR #1575, opened 2021-07-28
generate_paths_module attribute to haskell_cabal_library rule.

ylecornec:
This PR adds the ability to override the `getDatadir` and `getDataFileName` functions [provided by cabal](https://cabal.readthedocs.io/en/3.4/cabal-package.html#accessing-data-files-from-package-code) to access runfiles, in order to use Bazel's [runfile library](https://hackage.haskell.org/package/bazel-runfiles-0.12/docs/Bazel-Runfiles.html).

This makes the data-files of `haskell_cabal_libraries` rules accessible to `haskell_binary` rules, which has been a problem before (#1247).

This works by generating a `Paths_{libname}.hs` file, which provides the functions of interest, before cabal can do so.

This option is off by default  and activated by setting the `generates_paths_module` attribute of `haskell_cabal_library` to `True`.

Some other functions are defined by this module (`getBinDir`, `getLibDir`, `getDynLibDir`, `getLibexecDir` and `getSysconfDir`):
for these we imitate the previous behavior of cabal.

## Tests

 - Three of the tests check that the datafiles are accessible from binaries in various locations.
 - `//tests/haskell_cabal_datafiles/compare_other_cabal_functions:same_other_functions`  compiles two copies of a library (with and without the option) and  checks that the functions we do not intend to change yield the same result.

## Profiling

Profiling is not compatible with this `generate_paths_module` option for now as it runs into the following issue: #1573.

### DIFF
--- a/haskell/private/generate_cabal_paths_module.py
+++ b/haskell/private/generate_cabal_paths_module.py
@@ -41,7 +41,7 @@ def normalise_arch(arch):
     if arch == "arm64": return "aarch64"
     return arch
  
-def generate_cabal_paths_module(component_name, mangled_component_name, ghc_version, is_windows, cabal_basename, cabal_dirname,
+def generate_cabal_paths_module(component_name, ghc_version, is_windows, cabal_basename, cabal_dirname,
                                 ghc, libdir, dynlibdir, bindir, datadir, pkgroot, workspace):
 
     # cabal calls ghc --info to recover the target arch and os, and uses these in path names.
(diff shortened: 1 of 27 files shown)


haskell/private/generate_cabal_paths_module.py
```diff
@@ -0,0 +1,243 @@
+""" Generates a Paths_libname module instead of cabal.
+
+We modify the getDataFileName and getDataDir functions to use bazel's runfile library.
+And we try to keep the same behavior as before for the others functions
+(getBinDir getLibDir getDynLibDir getLibexecDir getSysconfDir).
+
```
aherrmann, line 6: Could you add a link to the relevant code in the Cabal code-base for reference?

aherrmann (approved):
Thank you, good work!

Just a few minor comments

ylecornec:
Thanks for the review,

I took the comments into account.

For the `haskell_cabal_binary` case we only made use of the component name until now,
and here the name of the cabal package is also needed (to name the generated file).

So, as for the package version, it is now recovered from the cabal file.

aherrmann:
In the context of #1247 I just noticed that this also requires adding `runfiles` to the `.cabal` file. The docs don't currently mention this, they only say to add the runfiles library to the `deps` attribute. Was this a case of me holding it wrong, or are the docs incomplete?

ylecornec:
The docs are incomplete, I opened PR #1703 to make this explicit.

It is a bit unfortunate that we have to modify the `.cabal` though, because this dependency is only needed with Bazel.
Maybe there is another way to make the `runfiles` library available to cabal? Or we could patch the `.cabal` file in cabal rules.

PR #1576, opened 2021-07-28
Fix com_google_protobuf URL

aherrmann:
The previous URL now produces 404. Perhaps the tag was removed. Note that the sha256 remains the same with this change.
https://developers.google.com/ also links to https://github.com/protocolbuffers/protobuf/releases. 

Thanks @lunaris for finding the replacement URL.


facundominguez (approved):
Works for inline-java. Thank you!

PR #1577, opened 2021-07-28
Fix protobuf url

cocreature:
It looks like google/protobuf got moved to protocolbuffers/protobuf
and then they renamed google/protobuf-dart to google/protobuf so now
the url returns a 404.

changelog_begin
changelog_end


cocreature:
Duplicate of #1576

PR #1578, opened 2021-07-29
Implement non_default_plugins

facundominguez:
With this new attribute, `haskell_library` and `haskell_binary` can avoid setting `-fplugin` for every module, which is handy if the plugin is slow or fails to process some of the modules.

### DIFF
--- a/tests/binary-with-plugin/BUILD.bazel
+++ b/tests/binary-with-plugin/BUILD.bazel
@@ -19,6 +19,7 @@ haskell_toolchain_library(name = "process")
 haskell_library(
     name = "plugin-lib",
     srcs = ["Plugin.hs"],
+    version = "0",
     deps = [
         ":base",
         ":bytestring",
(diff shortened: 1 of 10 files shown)


facundominguez:
This was discussed in https://github.com/tweag/rules_haskell/pull/1566#discussion_r672327179

tests/binary-with-plugin/BUILD.bazel
```diff
@@ -26,27 +28,59 @@ haskell_library(
...
+    name = "binary-with-plugin2",
+    srcs = ["main2.hs"],
+    compiler_flags = ["-fplugin=Plugin"],
+    non_default_plugins = select({
+        ":debug_build": [],
+        "//conditions:default": [":plugin2"],
+    }),
+    visibility = ["//visibility:public"],
+    deps = [":base"],
+)
```
facundominguez, line 76: I think these tests fail in CI because the main program is now checking if the plugin is enabled or not. Is there some way to skip these tests in the profiling configuration?

aherrmann: Yes, you can add `tags = ["skip_profiling"]`.

aherrmann (approved):
Thank you!

PR #1579, opened 2021-08-04
haskell_module(): add plugin support

Radvendii:
depends on #1565

I modified the test from `tests/binary-with-plugin`, but I'm not sure it's doing what we want it to do. The test succeeds even if the plugin is not loaded at all.

### DIFF
--- a/tests/haskell_module/plugin/BUILD.bazel
+++ b/tests/haskell_module/plugin/BUILD.bazel
@@ -1,0 +1,78 @@
+load(
+    "@rules_haskell//haskell:defs.bzl",
+    "ghc_plugin",
+    "haskell_library",
+    "haskell_test",
+    "haskell_toolchain_library",
+)
+load("@rules_haskell//haskell/experimental:defs.bzl", "haskell_module")
+
+package(default_testonly = 1)
+
+haskell_toolchain_library(name = "base")
+
+haskell_toolchain_library(name = "bytestring")
+
+haskell_toolchain_library(name = "ghc")
+
+haskell_toolchain_library(name = "process")
+
+haskell_library(
+    name = "plugin-lib",
+    srcs = ["Plugin.hs"],
+    version = "0",
+    deps = [
+        ":base",
+        ":bytestring",
+        ":ghc",
+        ":process",
+    ],
+)
+
+ghc_plugin(
+    name = "plugin",
+    args = [
+        "$(location //tests/binary-simple)",
+        "active",
+    ],
+    module = "Plugin",
+    tools = ["//tests/binary-simple"],
+    deps = [":plugin-lib"],
+)
+
+haskell_module(
+    name = "module-with-plugin",
+    src = "ModuleWith.hs",
+    plugins = [":plugin"],
+    tags = ["skip_profiling"],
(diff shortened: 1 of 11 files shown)


aherrmann (commented):
> I modified the test from `tests/binary-with-plugin`, but I'm not sure it's doing what we want it to do. The test succeeds even if the plugin is not loaded at all.

Indeed, those have been improved very recently in https://github.com/tweag/rules_haskell/pull/1578.
- It's using tags as a simpler way to disable the test in profiling mode.
- It fails if the plugin is not enabled.

I think you can use this more recent version to improve the haskell_module version of this test. You can skip anything using `non_default_plugins` as that attribute doesn't make sense for `haskell_module` (see https://github.com/tweag/rules_haskell/pull/1566#discussion_r673414169).

tests/haskell_module/plugin/BUILD.bazel
```diff
@@ -0,0 +1,86 @@
...
+
+package(default_testonly = 1)
+
+config_setting(
+    name = "debug_build",
+    values = {
+        "compilation_mode": "dbg",
+    },
+)
+
```
aherrmann, line 18:
```suggestion
```

Remove unused code.

aherrmann (commented):
Thank you!

Windows CI is failing in a [strange way](https://github.com/tweag/rules_haskell/pull/1579/checks?check_run_id=3246987216#step:6:154) though:
```
ERROR: D:/a/rules_haskell/rules_haskell/tests/haskell_module/plugin/BUILD.bazel:65:16: HaskellRegisterPackage tests/haskell_module/plugin/testsZShaskellZUmoduleZSpluginZSlib/testsZShaskellZUmoduleZSpluginZSlib.conf failed: (Exit 1): ghc-pkg.exe failed: error executing command 
  cd C:/users/runneradmin/_bazel_runneradmin/minshlu6/execroot/rules_haskell
  SET PATH=C:/Program Files/Git/usr/bin
  external/rules_haskell_ghc_windows_amd64/bin/ghc-pkg.exe recache --package-db=bazel-out/x64_windows-fastbuild/bin/tests/haskell_module/plugin/testsZShaskellZUmoduleZSpluginZSlib -v0 --no-expand-pkgroot
Execution platform: @local_config_platform//:host
ghc-pkg.exe: .:10:45: 
unexpected "t"
expecting space or "\""
```

That's the second `HaskellRegisterPackage` action of the `:lib` target (see `bazel aquery //tests/haskell_module/plugin:lib`). The `.conf` file is only generated when building something that depends on `:lib`, e.g. `:bin`.

Radvendii:
Update: it's failing because bazel doesn't do any sandboxing on windows. 

Our  `ls_modules.py` script searches the whole tree under the given search paths for interface files, and lists all of those as exposed modules. 

This is grabbing some extra built artifacts from a different target, that's not listed as a dependency.

Radvendii:
I realized that we have all the information to construct the module paths ourselves in the bazel code, so we can just pass those to `ls_modules.py` for it to append after

aherrmann (approved):
Good work! Thank you for debugging and fixing that Windows issue!

PR #1582, opened 2021-08-15
Support using TH and FFI with static C dependencies

jcpetruzza:
Consider the following scenario:

  - Package H1 uses FFI and needs to be linked against a pre-built static library X
  - Package H2 depends on H1 and uses TH (but TH code doesn't depend on X)
  - The ghc in the toolchain doesn't have a static runtime

In this case, as far as I can see, it is not currently possible to build H2 using `haskell_library`: during the compile step, `-lX` is passed to ghc, and the moment TH is used, ghc will try to load `libX.a` and fail. For example, on the added test-case, I currently see:

```
$ bazel build //tests/library-with-static-cc-dep:mypkg
INFO: Analyzed target //tests/library-with-static-cc-dep:mypkg (10 packages loaded, 372 targets configured).
INFO: Found 1 target...
ERROR: /home/daniel/repos/rules_haskell/tests/library-with-static-cc-dep/BUILD.bazel:10:16: HaskellBuildLibrary //tests/library-with-static-cc-dep:mypkg failed: (Exit 1): ghc_wrapper failed: error executing command bazel-out/host/bin/haskell/ghc_wrapper bazel-out/k8-fastbuild/bin/tests/library-with-static-cc-dep/compile_flags_mypkg_HaskellBuildLibrary ... (remaining 1 argument(s) skipped)

<command line>: User-specified static library could not be loaded (bazel-out/k8-fastbuild/bin/tests/library-with-static-cc-dep/libcbits-static.a)
Loading static libraries is not supported in this configuration.
Try using a dynamic library instead.
...
```

Cabal on the other hand doesn't pass `-lX` to ghc in this case, so falling back to `haskell_cabal_library` is the only workaround I see.

To fix this in haskell_library, the proposed solution is to optimistically ignore the static dependencies during the compile
action whenever the ghc runtime will not be able to load them. The only visible difference to the user should be when their TH
code in H2 depends in fact on X (a case that could never work under this scenario anyway): now they will get a symbol-not-found error, instead of a failure to load the static lib.

### DIFF
--- a/haskell/private/cc_libraries.bzl
+++ b/haskell/private/cc_libraries.bzl
@@ -49,7 +49,7 @@ def get_cc_libraries(cc_libraries_info, libraries_to_link):
         if not cc_libraries_info.libraries[cc_library_key(lib_to_link)].is_haskell
     ]
 
-def get_ghci_library_files(hs, cc_libraries_info, libraries_to_link, *, include_real_paths = False, for_th_only=False):
+def get_ghci_library_files(hs, cc_libraries_info, libraries_to_link, *, include_real_paths = False, for_th_only = False):
     """Get libraries appropriate for loading with GHCi.
 
     See get_library_files for further information.
--- a/haskell/private/cc_libraries.bzl
+++ b/haskell/private/cc_libraries.bzl
@@ -49,7 +49,7 @@ def get_cc_libraries(cc_libraries_info, libraries_to_link):
         if not cc_libraries_info.libraries[cc_library_key(lib_to_link)].is_haskell
     ]
 
-def get_ghci_library_files(hs, cc_libraries_info, libraries_to_link, include_real_paths = False):
+def get_ghci_library_files(hs, cc_libraries_info, libraries_to_link, *, include_real_paths = False, for_th_only=False):
     """Get libraries appropriate for loading with GHCi.
 
     See get_library_files for further information.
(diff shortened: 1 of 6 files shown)


haskell/private/cc_libraries.bzl
```diff
@@ -49,7 +49,7 @@ def get_cc_libraries(cc_libraries_info, libraries_to_link):
         if not cc_libraries_info.libraries[cc_library_key(lib_to_link)].is_haskell
     ]
 
-def get_ghci_library_files(hs, cc_libraries_info, libraries_to_link, include_real_paths = False):
+def get_ghci_library_files(hs, cc_libraries_info, libraries_to_link, *, include_real_paths = False, for_th_only=False):
```
aherrmann, line 52:
```suggestion
def get_ghci_library_files(hs, cc_libraries_info, libraries_to_link, include_real_paths = False, for_th_only=False):
```
Is this needed anywhere?

aherrmann (commented):
Thank you! Yes, this looks like a reasonable solution to me. Thank you for including a test as well!

haskell/private/cc_libraries.bzl
jcpetruzza: The `*` was there to disallow calls like `get_ghci_library(hs, cc_libs, True)`; I guess I find keyword args more clear when dealing with boolean flags? (and maybe safer when refactoring). Anyway, happy to remove it if you rather

aherrmann: Thanks for clarifying, makes sense to enforce this.

aherrmann (approved):
Thank you! Looks good!

buildifier was complaining, I pushed the result of `bazel run //:buildifier-fix`.

aherrmann:
@Mergifyio refresh

PR #1585, opened 2021-08-21
Sort exposed-modules in package conf

jcpetruzza:
The order of the `exposed-modules` in the generated package conf was non-deterministic, making builds non-reproducible.

This would happen since python's `os.walk()` will return files in arbitrary order. So now we sort them before writing to the file.


aherrmann (approved):
Good catch! Thank you!

PR #1586, opened 2021-08-23
Fix locale setting in GHC bindist on Linux

aherrmann:
The if expression was not parenthesized leading to the locale always being set to `C.UTF-8` on Linux.


ylecornec (approved):
Looks good to me.

Issue #951, opened 2019-06-17
Move as much of tests/RunTests.hs into bazel, remove if possible


Profpatsch:
For adding our repo to the upstream integration CI (https://github.com/tweag/rules_haskell/issues/754, https://github.com/bazelbuild/continuous-integration/issues/235) we need to pull many/all of our tests into bazel rules.

I’m pretty certain many of these can be implemented as simple `sh_test`s (or `genrule`s) instead.

guibou:
What's provided as a build context ? Is the building blindly running `bazel test //...` and can it be configured to call multiples bazel commands?

Profpatsch:
> What's provided as a build context ?

I don’t understand that question :)



> Is the builder blindly runs `bazel test //...` or can it be configured to call multiples bazel commands?

Ideally, `bazel test //...` should do what it says on the tin, run every test. If we cannot represent some tests *within* bazel (or compile in different modes), that is a bug in bazel in my book.

guibou:
@Profpatsch 

I was wondering how our code will be tested on the upstream integration CI? What control do we have on how it will be tested? What interface should we provide.

As you correctly said, most of our tests can be rewrited as `genrule` or `sh_test`, but a few of them needs more control on the options we pass to bazel. If I remember well, some of them need the result of `bazel query` before calling a few bazel command. Some others needs some custom bazel flags (such as `-c dbg`).

To understand how / what we must rewrite for the upstream CI system, we need to know how it behaves.

> Ideally, bazel test //... should do what it says on the tin, run every test. If we cannot represent some tests within bazel (or compile in different modes), that is a bug in bazel in my book.

There is a lot of limitations in bazel which are not bug, but just limitations due to the fact that this project is not finished.

regnat:
IIRC a relatively common practice is to run bazel whithin bazel test (see for example https://github.com/bazelbuild/bazel-integration-testing)

guibou:
Actually, just having `bazel test //...` part of bazel CI is a great improvement. We can iterate later on the removal of the test runner, if necessary.

Profpatsch:
> Actually, just having `bazel test //...` part of bazel CI is a great improvement. We can iterate later on the removal of the test runner, if necessary.

You should add that comment to the upstream issue.

guibou:
Which one?

Profpatsch:
https://github.com/bazelbuild/continuous-integration/issues/235

aherrmann:
[...]

PR #1596, opened 2021-09-20
Create "all_files" filegroup contains all rules_haskell sources

k1nkreet:
#951: To be able to use ```buildifier_test``` target and integration test targets like  ```go_bazel_build``` from https://github.com/bazelbuild/rules_go, this PR adds testonly filegroup contains all source files.


BUILD.bazel
```diff
@@ -15,6 +15,25 @@ buildifier_exclude_patterns = [
...
+        "BUILD.bazel",
+        "WORKSPACE",
+        "constants.bzl",
+        "//debug/linking_utils:all_files",
+        "//docs:all_files",
+        "//haskell:all_files",
+        "//nixpkgs:all_files",
+        "//protobuf:all_files",
+        "//rule_info:all_files",
+        "//tests:all_files",
```
aherrmann, line 31: For the buildifer use-case we'll want to include all files under `tests` to make sure that they are format tested as well. But, for the `go_bazel_test` use-case perhaps not. Including them means that `go_bazel_test` type integration tests are re-run not only when rules_haskell's implementation changes, but also when any of its tests changes.
AFAIK rules_go does not include the tests tree in the `all_files` filegroups.
A simple option might be to omit `//tests:all_files` here and then add it to the buildifier test explicitly. What do you think?

aherrmann: A related instance is `//vendor`. It's needed for functionality, so will be needed for `go_bazel_test`, but it would be preferable to skip it for buildifier tests.

k1nkreet: Probably it  makes sense to enumerate all necessary filegroups in ```buildifier_test``` srcs explicitly, because otherwise we'll inevitably end up with some kind of "sources for buildifer_test" filegroup. In this case name ```all_files``` is a bit odd  for filegroup dedicated for integration_tests. ```rules_go``` for example doesn't include tests in ```all_files``` but use it both for integration tests and buildifier which is a bit kludgy imho. Some bazelbuild projects like ```bazel-skylib``` or ```rules_python``` uses name ```distribution``` for such filegroup, probably it makes sense to follow

aherrmann: > Some bazelbuild projects like `bazel-skylib` or `rules_python` uses name `distribution` for such filegroup, probably it makes sense to follow

Nice find! Yes, that sounds like a good precedent to follow.

> Probably it makes sense to enumerate all necessary filegroups in `buildifier_test` srcs explicitly, [...]

You mean at the top-level something like this?
```
buildifier_test(
    srcs = [
        "//haskell:all_files",
        "//nixpkgs:all_files",
        "//tests:all_files",
        ...
    ],
```
Yes, that sounds like a good solution to the problem.

k1nkreet: > You mean at the top-level something like this?

Yes, that's exactly what I meant :)

k1nkreet: > Some bazelbuild projects like bazel-skylib or rules_python uses name distribution for such filegroup, probably it makes sense to follow

aherrmann (approved):
Thank you, looks good!

PR #1597, opened 2021-09-21
Add buildifier_test target

k1nkreet:
#951: Create ```buildifier_test``` target target instead of running ```build run //:buildifer``` from RunTests.hs

Depends on https://github.com/tweag/rules_haskell/pull/1596

### DIFF
--- a/tests/RunTests.hs
+++ b/tests/RunTests.hs
@@ -18,9 +18,6 @@ import Test.Hspec (context, hspec, it, describe, runIO, shouldSatisfy, expectati
 
 main :: IO ()
 main = hspec $ do
-  it "bazel lint" $ do
-    assertSuccess (bazel ["run", "//:buildifier"])
-
   it "bazel test" $ do
     assertSuccess (bazel ["test", "//..."])
 
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -66,4 +66,5 @@ buildifier_test(
         "@examples//:all_files",
         "@tutorial//:all_files",
     ],
+    tags = ["dont_test_on_windows"],
 )
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -66,5 +66,6 @@ buildifier_test(
         "@examples//:all_files",
         "@tutorial//:all_files",
     ],
+    mode = "diff",
     tags = ["dont_test_on_windows"],
 )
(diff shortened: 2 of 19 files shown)


tests/RunTests.hs
```diff
@@ -18,9 +18,6 @@ import Test.Hspec (context, hspec, it, describe, runIO, shouldSatisfy, expectati
 
 main :: IO ()
 main = hspec $ do
-  it "bazel lint" $ do
-    assertSuccess (bazel ["run", "//:buildifier"])
```
aherrmann, line 22: I think this also checked files under examples and tutorial which are not covered by the new `buidifier_test` with `//:all_files`. Any way to keep those checked as well?

k1nkreet: I would suggest to add them as a ```local_repositories``` providing same ```all_files``` targets and use them as srcs of ```buildifier_test```

aherrmann: :+1: Sounds like a good approach.

aherrmann:
I see that CI is failing on Windows with
```
ERROR(tools/test/windows/tw.cc:1288) ERROR: src/main/native/windows/process.cc(202): CreateProcessW("C:\users\runneradmin\_bazel_runneradmin\minshlu6\execroot\rules_haskell\bazel-out\x64_windows-fastbuild\bin\tests\buildifier_test.bash"): %1 is not a valid Win32 application.
```
This happens when a Bazel rule emits a script like a bash script as its executable. This works on Unix, but not on Windows where such files are not valid executables. I suspect this may be the issue here.

One way to fix this is to split the rule into two parts, one that fills in the script template and a separate `native.sh_binary`. We do this e.g. in [`cc_wrapper`](https://github.com/tweag/rules_haskell/blob/ea0e70ace2432a490d4ab4c4e54617612466e584/haskell/private/cc_wrapper.bzl#L122).

That said, I don't think we need to block this PR on that. Only checking formatting on Unix is sufficient and we can skip this test on Windows.

k1nkreet:
I've missed buildtools readme mentions explicitly Bazel target doen't supported on Windows. This is actually true for run target used in RunTests.hs as well :)

aherrmann:
> I've missed buildtools readme mentions explicitly Bazel target is not supported on Windows. This is actually true for run target used in RunTests.hs as well :)

Ah, good to know. Thanks for clarifying!

k1nkreet:
I've tagged this test as "dont_test_on_windows" so it should not be failing now.

tests/RunTests.hs
k1nkreet: Done

aherrmann:
Rebase on master should fix the CI error. #1606 is merged.

Issue #1598, opened 2021-09-30
Rules haskell generates too much symlink


guibou:
**Describe the bug**

`bazel build //...` generates a lot of files in bazel-bin.

This starts to be a problem when the number of files goes above a few dozen millions. It saturates all of our CI disk (too much inodes used). It is also a huge performance problem, because creating so much files put a lot of pressure on the filesystem. I do have a CI build here which took 4 hours, of which 97% where CPU time used by the system to manage the filesystem.

**To Reproduce**

`bazel build //...` on a medium project. To give you an idea, I observed the problem on two codebase with ~1000 haskell modules and ~300 targets (library, binary, tests, ...).

A few numbers.

`bazel build //...` generated 24.7 GiB in `bazel-bin`, for a total numbor of 12 millions files.

This number can be reduced to 18 GiB, for 6 million files using bazel `--nolegacy_external_runfiles`. However this may not work depending on the other ruleset used. (I got it to work on a codebase without rules_python, but not on a codebase with rules_python).

Amongs theses 18 Gib (and 6 million files), we have duplicated files in the `runfiles` directory of each rules:
- 10 GiB of `hackage` data.
- 6.4 GiB of `repl` runfiles

The rest, ~1.8 Gib are the artifacts of my build, for 120K files. That's still a lot (there are still a lot of duplication and symlink), but it is acceptable now.

I was able to remove these by disabling `-repl` build in CI and I stop using rules_haskell `stack_snapshot` for hackage dependencies.

Theses changes "fixed" our CI (disk space usage and number of files was reduced by more than 10x).

**Expected behavior**

Well, not creating millions of files.

**Environment**
 - OS name + version:
 - Bazel version:
 - Version of the rules:

**Additional context**

This is a known issue on bazel design: https://github.com/bazelbuild/bazel/issues/2951

However this is also something rules_haskell may improve by pushing less things in the runfiles directories by default.

aherrmann:
> I was able to remove these by disabling `-repl` build in CI

This could be addressed by tagging the auto-generated `haskell_repl` targets with `tags = ["manual"]`. What do you think?

> and I stop using rules_haskell `stack_snapshot` for hackage dependencies.

Given that haddock files seem to be culprit here, would it be sufficient to set [`haddock = False`](https://api.haskell.build/haskell/cabal.html#stack_snapshot-haddock) on `stack_snapshot`?

---

[...]

PR #1599, opened 2021-10-01
Set all repl as "manual"

guibou:
This is part of an attempt to improve situation in #1598.

When calling `bazel build //...`, repl runfiles directory were created,
generating a lot of files on the disk. This was responsible for
important slowdown as well as disk usage.

This change set the `"manual"` tag to all implicit repl targets, so they
are only built (and run) when using explicit bazel call.


haskell/defs.bzl
```diff
@@ -356,6 +352,20 @@ def haskell_binary(
         **repl_kwargs
     )
 
+def make_repl_kwargs(args_list, kwargs):
```
aherrmann, line 355:
```suggestion
def make_repl_kwargs(args_list, kwargs):
    """Create extra attributes for the auto-generated haskell_repl target.
    
    Copies the extra attributes specified in `args_list` from the extra
    `haskell_library|binary|test` attributes listed in `kwargs`.
    
    Adds a `manual` tag so that the REPL is not built by default.
    """
```
Adding a doc-string.

aherrmann (approved):
Thank you, looks good to me.

guibou:
@aherrmann Thank you. I've "accepted" your suggestion. Could you merge? (i.e. I unfortunately do not have the "add label" right anymore ;)

PR #1614, opened 2021-10-19
External dependencies for asterius

ylecornec:
This PR sets up dependencies needed for asterius support.
    
  - Updates the dependency on rules_sh to a version without target constraint for the toolchain.

  - Install a recent version of rules_nodejs if bazel supports it.
    Executing asterius tests requires a recent version of node, and recent enough rules_nodejs only support bazel >= 4.

  - The `asterius_repositories.bzl` files contains functions setting up the node toolchain that need to be called before using asterius. There is one macro for bindists, one for nix, and one way to supports both. That last case is a bit more complicated because we use one workspace rule to determine whether nix is supported and depending on the result we call the `node_repository` workspace rule from rules_nodejs differently.

We use it this way in the `WORKSPACE` file:
```
load("//haskell/asterius:asterius_repositories.bzl", "prepare_asterius_dependencies")

prepare_asterius_dependencies(
    name = "prepare_asterius_dependencies",
    nix_repository = "@nixpkgs_default",
    nixpkgs_package_rule = nixpkgs_package,
)

load("@prepare_asterius_dependencies//:deps.bzl", "asterius_node_dependencies")

asterius_node_dependencies()
```
  
### Note on Stardoc

Stardoc documentation for the `asterius_repositories.bzl` file is not set up at the moment because 
of the following  [issue of rules_nodejs](https://github.com/bazelbuild/rules_nodejs/issues/2874).


ylecornec:
As discussed, I removed the code that supports nix and bindists at the same time.

It is not worth the complexity since it is not a common problem and in our case it is simpler to rely on the `is_nix_shell` variable.

PR #1624, opened 2021-11-10
Add support for dynamic builds in haskell_module

facundominguez:
This is a preparation step towards Template Haskell support for `haskell_module`.


aherrmann (approved):
Thank you, looks great!

PR #1625, opened 2021-11-12
Add template haskell support to haskell_module

facundominguez:
This PR exposes object and .dyn_o files to the build action of haskell_module.

It also adds a test of building a module which uses Template Haskell.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 /bazel-*
 .bazelrc.local
+*.swp
+*.swo
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -31,7 +31,20 @@ load(
     "HaskellModuleInfo",
 )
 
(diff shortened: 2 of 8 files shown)


facundominguez:
I see some `haskell_module` tests missing in the console output. Are they being run at all?
For instance in `Build & Test - Nixpkgs (ubuntu-latest)`, searching for `haskell_module` in the console output yields no result.

facundominguez:
Hm, some tests only build but have no test executables. I see in the Build Buddy interface that the th test is built in at least one job, but I don't understand this interface enough to find if all the haskell_module tests are built in all jobs.

aherrmann:
Yes, the list of tests only lists proper test targets, e.g. `haskell_test`. However, `bazel test //...` will also build every target, not just test targets. You can see that the `th:lib` target was built [here](https://app.buildbuddy.io/invocation/acf4e505-8c66-4247-b58e-e93846df4917?target=%2F%2Ftests%2Fhaskell_module%2Fth%3Alib).

aherrmann (approved):
Thank you, looks good!

PR #1628, opened 2021-11-15
Add a module_name attribute to Haskell module.

### LINKED ISSUE #1627: Allow to turn off haskell_module inference of module names
When building a haskell binary, the main module sometimes lives in the same folder as other main modules for other binaries. This makes inconvenient to require all the main modules to be called `Main.hs` only so `haskell_module` can infer the module name correctly.

It would work better, if we could turn off the module name inference by specifying it in an attribute. e.g.
```python
haskell_module(
    name = "test-main",
    src = "exes/Test.hs",
    module_name = "Main",
)
```

facundominguez:
Closes #1627

### DIFF
--- a/haskell/experimental/defs.bzl
+++ b/haskell/experimental/defs.bzl
@@ -19,6 +19,7 @@ _haskell_module = rule(
             mandatory = True,
         ),
         "src_strip_prefix": attr.string(),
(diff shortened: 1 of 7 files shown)


aherrmann (approved):
Thanks, looks good.

PR #1632, opened 2021-11-19
Update Asterius bundle

ylecornec:
This PR updates the Asterius bundle:

- It is now in a Tweag owned repository (until it is built in the Asterius CI).

- This new bundle supports literate Haskell (the previous one was missing the `unlit` binary provided by Asterius) so there is now a test for this.


aherrmann (approved):
Thank you!

PR #1641, opened 2021-11-24
Cabal archive for testing sublibraries through stack_snapshot

ylecornec:
This PR adds an archive based on the `haskell_cabal_library_sublibrary_name` test, that can be referenced by a custom stack snapshot, in order to test support for cabal sublibraries.

Directly referencing the repository as such does not work:

```
- github: tweag/rules_haskell
  commit: "eec33928e7ee2337a8d01ab964cb556353345696"
  sha256: "2ebc619bcd80b1106c48cc6306d06d523b86b69b31fc4b51f05fc9f363f6d581"
  subdirs:
    - tests/haskell_cabal_library_sublibrary_name
```
 because of dangling symlink `examples/.bazelrc.local` which makes stack fail with this error.

```
Symbolic link dest not found from examples/.bazelrc.local
...
This may indicate that the source is a git archive which uses git-annex.
```

But referencing it as such works:

```
  - archive: https://github.com/tweag/rules_haskell/raw/4f17633f3b35d16265470db1bb8b622c29739f55/tests/haskell_cabal_library_sublibrary_name/package1.tar
    sha256: "302d8ddda8330c825da61fe0a2315c899ab083e641c7716ebdacb5c951682445"
```

(Except that the commit number will have to be changed after this branch is rebased).


aherrmann (approved):
It's a bit awkward to have to do it this way, but, I don't see a better solution at the moment.
Could you add a README next to that tarball explaining why this tarball is there?

ylecornec:
I added the `README.md` file next to the tarball.

aherrmann (approved):
Thank you!

PR #1642, opened 2021-11-24
explicitly pass hsc2hs binary to cabal_wrapper.py.

ylecornec:
This PR explicitly pass the `hsc2hs` binary from the haskell toolchain to cabal, which is needed for asterius to build the `@stackage-zlib//:zlib` target, and make the new test pass.


aherrmann (approved):
Thank you!

PR #1644, opened 2021-11-29
Disallow non-module dependencies in the deps attribute of haskell_module

facundominguez:
Since non-haskell_module deps are ignored in the deps attribute of `haskell_module`, this PR has bazel produce an error when non-haskell_module dependencies are placed in the modules attribute.

### DIFF
--- a/haskell/experimental/defs.bzl
+++ b/haskell/experimental/defs.bzl
@@ -23,9 +23,7 @@ _haskell_module = rule(
         "extra_srcs": attr.label_list(
             allow_files = True,
         ),
-        "deps": attr.label_list(
-            aspects = [haskell_cc_libraries_aspect],
-        ),
+        "modules": attr.label_list(),
         "ghcopts": attr.string_list(),
         "plugins": attr.label_list(
             aspects = [haskell_cc_libraries_aspect],
@@ -57,13 +55,16 @@ def haskell_module(
         src = None,
         extra_srcs = [],
         module_name = "",
-        deps = [],
+        modules = [],
         ghcopts = [],
         plugins = [],
         tools = [],
         worker = None,
         **kwargs):
-    """Compile a module from Haskell source.
+    """Declare a dependency between modules.
+
+    This allows library, binary, and test rules to do incremental builds when only a
+    few modules are affected by a change.
 
     Note: This rule is experimental and not ready for production, yet.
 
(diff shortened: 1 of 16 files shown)


haskell/experimental/defs.bzl
```diff
@@ -57,13 +55,16 @@ def haskell_module(
...
         module_name = "",
-        deps = [],
+        modules = [],
         ghcopts = [],
         plugins = [],
         tools = [],
         worker = None,
         **kwargs):
-    """Compile a module from Haskell source.
+    """Declare a dependency between modules.
```
facundominguez, line 64: I don't think I know how to explain this concisely.

haskell/experimental/private/module.bzl
```diff
@@ -390,7 +390,13 @@ def build_haskell_modules(ctx, hs, cc, posix, package_name, with_shared, hidir,
     )
 
 def haskell_module_impl(ctx):
-    module_deps = [dep for dep in ctx.attr.deps if HaskellModuleInfo in dep]
+    module_deps = [dep for dep in ctx.attr.modules if HaskellModuleInfo in dep]
+    if len(module_deps) < len(ctx.attr.modules):
+        fail("""
+{rule_name} contains non-haskell_module labels in the modules attribute: {unexpected_labels}
+Maybe put them in the deps attributes of the corresponding library, binary, or test rule.
+""".format(rule_name = ctx.label, unexpected_labels = ", ".join([str(dep.label) for dep in ctx.attr.modules if not HaskellModuleInfo in dep])))
```
aherrmann, line 398: Bazel has builtin support to check for providers in attributes with the `providers` parameter of [`attr.label_list`](https://docs.bazel.build/versions/main/skylark/lib/attr.html#label_list). If you set that, then I think this check shouldn't be necessary.

aherrmann (commented):
Do we need to rename the attribute? They are still dependencies, module dependencies of another module, so, the name `deps` seems fitting.

facundominguez:
> Do we need to rename the attribute? They are still dependencies, module dependencies of another module, so, the name deps seems fitting.

The name `deps` is fitting, but it doesn't hint which dependencies are expected. Isn't `modules` more precise?

aherrmann:
> > Do we need to rename the attribute? They are still dependencies, module dependencies of another module, so, the name deps seems fitting.
> 
> The name `deps` is fitting, but it doesn't hint which dependencies are expected. Isn't `modules` more precise?

To me it seems confusing, because it adds a different meaning to this attribute name compared to `haskell_library`. In `haskell_library` `modules` means the modules that are *contained in* this library, i.e. it's not a dependency in the sense of a library depending on another library, but it's the modules that make up that library. With `haskell_module` on the other hand it is a regular dependency, a module depending on another module. `deps` is the standard attribute name in Bazel for that kind of relationship.

facundominguez:
> To me it seems confusing, because it adds a different meaning to this attribute name compared to haskell_library.

I can live with it. I don't feel strong about it. Let me undo the renaming then.

haskell/experimental/private/module.bzl
facundominguez: Done.

aherrmann (approved):
Thank you!

PR #1645, opened 2021-11-30
Implement hs-bin@repl test as go bazel test

k1nkreet:
According to #951 this PR makes a first step to implement integration tests as ```go_bazel_test``` from ```rules_go```


tests/integration_tests.bzl
```diff
@@ -0,0 +1,22 @@
+load("@io_bazel_rules_go//go/tools/bazel_testing:def.bzl", "go_bazel_test")
+
+def integration_test(name, **kwargs):
+    test_src = name + ".go"
+    size = kwargs.pop("size", "medium")
+
+    native.config_setting(
+        name = "nixpkgs",
+        values = {"define": "nixpkgs=true"},
+    )
```
aherrmann, line 10: I think you could reuse [this `config_setting`](https://github.com/tweag/rules_haskell/blob/7675f021fb2bb64d21f4b1c6bb3f77e02ceb8c83/tests/BUILD.bazel#L39-L44) based on the `support_nix` platform constraint. Then you also don't need to introduce the `--define=nixpkgs` above.

k1nkreet: Thanks! Fixed

k1nkreet:
According to [last ci task on Windows](https://github.com/tweag/rules_haskell/runs/4580603106?check_suite_focus=true) there is no bazel binary in PATH occurred inside bazel test, so it probably makes sense to implement an external dependency of ```integration_test``` rule on bazel binary within this PR

k1nkreet:
All CI tests are finally passed but I have a strong feeling that using tmp filesystem for testcases bazel cache is not a good idea. Cache size will constantly grow and tmpfs on average machine is pretty limited. I've already experienced freezing of my laptop sometimes when I'm running these tests. So I will try to find another place writable by bazel tests and invariant between tests

k1nkreet:
For basic rules_haskell's integration Bazel test creates about 800MB of cache. It nearly exceeds size of $TMPDIR folder created by Nix on my machine for example but fits /tmp without any problem. So I see two ways to store cache of internal bazel:

- Use golang os.TempDir() to pass it as $HOME for test: it uses $TMPDIR on Unix system and %TEMP% on Windows. In this case we'll need to add ```TMPDIR=/tmp``` as shellHook in shell.nix for nixpkgs configurations. Also Bazel cache will stuck in /tmp after tests run which concerns me a bit.
- Pass $HOME through the --action_env: it will share cache between external and internal Bazel which probably could even speed up things sometimes but also breaks sandboxing of tests a bit more. Also it could affect sandboxing of other tests since they all will have non-sandboxed $HOME. Also it worth mentioning ```go_bazel_tests``` in rules_go just use common $HOME but would break with strict action_env.

I'm leaning towards the first option, @aherrmann do you have an opinion?

aherrmann:
> I'm leaning towards the first option, @aherrmann do you have an opinion?

Yes, the first option seems preferable. I'd prefer to avoid the need for `--action_env` because it would affect all targets and impact their cache keys.

k1nkreet:
@aherrmann It's finally ready for being reviewed :)

WORKSPACE
```diff
@@ -533,6 +533,8 @@ register_toolchains(
 # starting from 0.29, rules_go requires bazel >= 4.2.0
 http_archive(
     name = "io_bazel_rules_go",
+    patch_args = ["-p1"],
+    patches = ["//:rules_go_integration_testing.patch"],
```
aherrmann, line 537: Please add a comment why this is needed and also pointing out that it is not needed for downstream users of rules_haskell.

WORKSPACE
```diff
@@ -581,9 +583,12 @@ bind(
 
 # For persistent worker (tools/worker)
 load("//tools:repositories.bzl", "rules_haskell_worker_dependencies")
+load("//tools:repositories.bzl", "bazel_binaries_for_integraion_testing")
 
 rules_haskell_worker_dependencies()
 
+bazel_binaries_for_integraion_testing()
```
aherrmann, line 590:
```suggestion
bazel_binaries_for_integration_testing()
```

tests/integration_testing.go
```diff
@@ -0,0 +1,157 @@
...
+`
+        if bazel_testing.OutputUserRoot != "" {
+                bazelrc += fmt.Sprintf("startup: --output_user_root=%s\n", bazel_testing.OutputUserRoot)
+        }
+        return bazelrc
+}
+
+func BazelEnv() []string {
+        env := []string{}
+	// It's important value of $HOME to be invariant between different integration test runs
```
aherrmann, line 71: Indentation looks off here. Perhaps best to run `gofmt` on the go sources.

tests/integration_testing.go
```diff
@@ -0,0 +1,157 @@
...
+        // and to be writable directory for bazel test.
```
aherrmann, line 72:
```suggestion
	// It's important that the value of $HOME is invariant between different integration test runs
        // and that the directory is writable for bazel test.
```
Maybe a bit easier to read?

tests/integration_tests.bzl
```diff
@@ -0,0 +1,31 @@
...
+        name = name,
+        srcs = [test_src],
+        size = size,
+        rule_files = ["//:distribution"],
+        args = select({
+            "//tests:nix": ["nixpkgs=true"],
+            "//conditions:default": ["nixpkgs=false"],
+        }) + select({
+            "@platforms//os:osx": ["bazel_bin=$(location @bazel_bin_darwin//file)"],
+            "@platforms//os:linux": ["bazel_bin=$(location @bazel_bin_linux//file)"],
```
aherrmann, line 22: Shouldn't it use a Nix provided Bazel in the Nix case?

aherrmann (commented):
:tada: Good work, thank you!
A couple comments/questions.

tests/integration_testing.go
k1nkreet: It sounds reasonable to include the same rule as we use for ```buildifier``` and ```buildifier-fix``` for ```gofmt``` but apparently there is no such thing :(

k1nkreet: Fixed

tests/integration_testing.go
k1nkreet: Fixed, sorry for my runglish :)

tests/integration_tests.bzl
k1nkreet: Done

WORKSPACE
k1nkreet: Done

aherrmann (approved):
This is great, thank you!
I've tested it locally and tests work as expected with and without Nix.

Issue #1646, opened 2021-12-01
Support bazel 4.2.0 without warnings


nishanthkarthik:

aherrmann:
fixed in #1647

nishanthkarthik:
Thanks for the fix!

Issue #1622, opened 2021-11-02
Should I run `HEAD`?


ocharles:
I'm evaluating `rules_haskell`, and I see a lot of active development but the last release was over a year ago. Should I run `HEAD`, or should I still use the latest release?

aherrmann:
Thanks for reaching out. I would recommend using `HEAD` at this point.

ocharles:
Ok, thanks!

avdv:
Hi. :smiley: 

Just a quick follow-up question: what is the current minimum bazel version supported by HEAD?

Asking because I was trying to use bazel 3.7.2 with commit d3d8bede5597c4ffc373f9742a6a733c728505b2 and got this:
```
ERROR: /home/claudio/.cache/bazel/_bazel_claudio/a0b0c5e02d9eb86a96afefb39001e2b9/external/rules_haskell/haskell/cabal.bzl:1955:9: name 'json' is not defined
ERROR: /home/claudio/.cache/bazel/_bazel_claudio/a0b0c5e02d9eb86a96afefb39001e2b9/external/rules_haskell/haskell/cabal.bzl:2116:22: name 'json' is not defined
ERROR: error loading package '': Extension 'haskell/cabal.bzl' has errors
```
AFAICS, the `json` module is only available with bazel +4.0.0. Or is it possible to add this module for older bazel versions somehow?

aherrmann:
Hi @avdv, good catch. Yes, `json` is only available in Bazel +4.0.0. So, the bounds should be bumped accordingly. Do you need to use rules_haskell with an older Bazel version?

avdv:
Thanks for confirming, @aherrmann . I have no need to stay at bazel 3.x, and I just upgraded to the bazel_4 package without much ado. :+1:

PR #1647, opened 2021-12-02
Update bazel version and bounds.

ylecornec:
This PR updates the Bazel version to 4.2.1.

- The supported versions are now 4.0.0 to 4.2.1.

    - This should remove warnings such as the one in: #1646
    - We make use of the json module which requires at least bazel `4.0.0` as raised up in [this comment](https://github.com/tweag/rules_haskell/issues/1622#issuecomment-978917512).
    - Asterius supports depends on a recent `rules_nodejs` which also requires bazel `>= 4`.
 
- The nixpkgs version of the reposiory was bumped to 21.11 which provides bazel 4.2.1.
    ghc 8.10.4 is not in this revision so I updated it to 8.10.7, which is the latest one with a stackage lts.

- Since the minimal supported Bazel version is now 4.0.0, I removed the `supports_rules_nodejs_ge_4` function, which was testing this to decide which `rules_nodejs` installation to depend on, and whether to enable asterius support.

##### Checklist
- [x] Use `.ci/update-bazel-version` to update `.ci/bazelversion` and      `.ci/bazel-*.sha256`.
- [x] Update all bazel rules dependencies in `WORKSPACE` (e.g.`io_bazel_stardoc`)
- [x] Update bazel in nixpkgs and bump `nixpkgs/default.nix`
- [x] Bump `MAX_BAZEL_*` in `start` and `haskell/private/versions.bzl`
- If we are updating to a new LTS:
  - [x] Bump `MIN_BAZEL_*` in `start` and `haskell/private/versions.bzl`
- [x] Add update notice to `CHANGELOG`


aherrmann (approved):
Awesome, thank you!

Issue #1600, opened 2021-10-05
Rules haskell remote caching


guibou:
A question about remote caching.

In my latest project, I was using bazel remote caching to cache our build (including rules_haskell) and it was ok. Bazel caching is slow (because it needs to fetch all the intermediate artifacts), but with a fast internet connection, it was generally a bit faster than building everything locally. We also had a lot of C++ and AFAIK `gcc` produces reproducible binaries. Our hackage dependencies where built using nix, so this part of the caching was not handled by bazel.

I recently moved to a new codebase which is 100% haskell and uses `stack_snapshot` to build hackage dependencies.

And I observed that remote caching is totally unreliable. Sometimes it works, fetching our hackage dependencies in less than an hour, sometimes it fails, and we have to rebuild everything, which can take a few hours.

I traced different build using `--execution_log_json_file` and all the Haskell `.hi` files generated are different.

GHC status about it is a bit unclear: https://gitlab.haskell.org/ghc/ghc/-/issues/4012, it says that it won't be ready for 9.2 (btw, we are using 8.10), but there is also mentions that interfaces files are reproducible, but that's not what I'm observing.

On rules_haskell master:

```console
$ bazel clean
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.

[nix-shell:~/the_company/rules_haskell]$ bazel build @stackage//:primitive --execution_log_json_file=exec_log.json
INFO: Analyzed target @stackage//:primitive (54 packages loaded, 1245 targets configured).
INFO: Found 1 target...
Target @stackage//:primitive up-to-date:
  bazel-bin/external/stackage/primitive-0.7.1.0/_install/lib/libHSprimitive-0.7.1.0.a
  bazel-bin/external/stackage/primitive-0.7.1.0/_install/lib/libHSprimitive-0.7.1.0-ghc8.10.4.so
INFO: Elapsed time: 60.916s, Critical Path: 59.62s
INFO: 25 processes: 23 internal, 2 linux-sandbox.
INFO: Build completed successfully, 25 total actions

[nix-shell:~/the_company/rules_haskell]$ bazel clean
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.

[nix-shell:~/the_company/rules_haskell]$ bazel build @stackage//:primitive --execution_log_json_file=exec_log2.json
INFO: Analyzed target @stackage//:primitive (54 packages loaded, 1245 targets configured).
INFO: Found 1 target...
[...]

PR #1648, opened 2021-12-02
Use more deterministic distdir path in haskell_cabal_*

aherrmann:
Addresses indeterministic interface generation as reported in #1600.

Use a path for distdir in the `haskell_cabal_*` rules that is based on the package name and deterministic unless there is a collision with an already existing file or directory.

The distdir path enters flags that are passed to GHC and thereby enters the 'flag hash' field of the generated interface files. If the distdir path is indeterministic, then the interface file will also be indeterministic.

With this PR a build of the `hspec` Cabal library and its dependencies has been found to be bit-reproducible by building it twice without caching and comparing the outputs.

Notably, there is still an issue with non-deterministic builds if a `haskell_cabal_*` target has a C library dependency that was built by Bazel. In that case the absolute path to the sandbox execroot, which can change between builds, enters the [include and library search path flags](https://github.com/tweag/rules_haskell/blob/a7241fa64c7cd36462a1f6ac4c660d1247d5e07b/haskell/cabal.bzl#L343-L345). Cabal forces us to provide these paths as absolute paths, see https://github.com/haskell/cabal/issues/1317 and https://github.com/haskell/cabal/issues/694. Meaning, the generated interface files of `haskell_cabal_library` targets with Bazel built C library dependencies remain indeterministic.

### DIFF
--- a/haskell/private/cabal_wrapper.py
+++ b/haskell/private/cabal_wrapper.py
@@ -37,7 +37,9 @@ from __future__ import print_function
 
 from contextlib import contextmanager
 from glob import glob
+import itertools
 import json
+import hashlib
 import os
 import os.path
 import re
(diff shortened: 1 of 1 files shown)


ylecornec (approved):
It seems good to me :+1:

regnat (approved):
The motivation for the change made me scream a bit, but the change itself looks good :+1:

aherrmann:
@Mergifyio refresh

PR #1649, opened 2021-12-03
Tweaks to introductory tutorial from a fresh pair of eyes

Xophmeister:
These are just the tweaks to the "Introduction to Bazel: Building a Haskell project". I may make further tweaks to subsequent pages: I can ball them up into a single PR if it's more convenient (rather than piecemeal).


aherrmann (approved):
Thank you!

> I can ball them up into a single PR if it's more convenient (rather than piecemeal).

I'm happy with small PRs.

Issue #1631, opened 2021-11-18
haskell_module rebuilds more than necessary


facundominguez:
When a module M depends on a library L, it is made to depend on all the modules of the library, even if it doesn't import all of them.

Two problems arise from this situation:

1. Unnecessary rebuilds: Module M might be rebuilt when no module in its transitive imports changes. This can happen when library L needs to be rebuilt, but the modules from library L that changed are not in the transitive imports of module M.

2. Unnecessary sequentialization: Building module M might have to wait until modules of library L not in its transitive imports are built. This can happen when a change in a module in library L causes multiple modules in L to rebuild, but not all of these modules are in the transitive imports of module M.

### Fixing it

Bazel offers a solution for (1) with a file attribute of actions called [unused_input_list](https://docs.bazel.build/versions/main/skylark/lib/actions.html#run.unused_inputs_list). This file must contain a list of all the unused inputs. In our case, the action to build module M would list all the module interface files not in the transitive imports.

Avoiding (2), however, would require narrowing the inputs. Instead of depending on a library, the action that builds module M would need to depend only on the modules in the transitive imports. This also happens to be a solution for (1). And there are two cases to consider:

A. Module M doesn't use Template Haskell. In this case we only need the interface files of transitive imports.

B. Module M does use Template Haskell, and we need to supply the object files of transitive imports as well.

In case (A) we can feed ghc the package database of library L to expose the interface files. And we can calculate the interface files that are in the transitive imports by asking the user or `gazelle_haskell_modules` to make the imports explicit in rule attributes.

Case (B) is a bit trickier. We can calculate which object files are needed as inputs in the same way we do for interface files. However, ghc will not take on its own any object files in the package database. We need to explicitly pass the object files to load with TH in the command line. e. g.
```
$ ghc -c M.hs -package-db libraryL/pkgdb libraryL/M_1.o ... libraryL/M_n.o
Warning: the following files would be used as linker inputs, but linking is not being done: libraryL/M_1.o ... libraryL/M_n.o
```
Additionally, `ghc` will insist in loading some shared library file in `libraryL/libHSlibraryL-ghc8.10.7.so`. We can create an empty library with the expected name to satisfy GHC. But long term, we would need to engage with ghc devs to get an official way to feed object files instead of the shared library for single-module compilation.

facundominguez:
One thing we need to come up here is with a way to specify dependencies on modules coming from other libraries.

[...]

PR #1654, opened 2021-12-07
Narrowing of dependencies when not using template haskell

facundominguez:
Partial implementation of #1631 

This implements the narrowing that supplies the needed interface files from dependencies as inputs to the action of `haskell_module`.

### DIFF
--- a/haskell/defs.bzl
+++ b/haskell/defs.bzl
@@ -35,6 +35,10 @@ load(
     "check_deprecated_attribute_usage",
 )
 load(
+    "//haskell:providers.bzl",
+    "HaskellLibraryInfo",
+)
+load(
     "//haskell/experimental:providers.bzl",
     "HaskellModuleInfo",
 )
@@ -53,6 +57,7 @@ _haskell_common_attrs = {
     "deps": attr.label_list(
         aspects = [haskell_cc_libraries_aspect],
     ),
+    "narrowed_deps": attr.label_list(providers = [HaskellLibraryInfo]),
     "modules": attr.label_list(
         providers = [HaskellModuleInfo],
     ),
@@ -248,6 +253,7 @@ def haskell_binary(
         srcs = [],
         extra_srcs = [],
         deps = [],
+        narrowed_deps = [],
         data = [],
         compiler_flags = [],
         ghcopts = [],
@@ -309,6 +315,10 @@ def haskell_binary(
       srcs: Haskell source files. File names must match module names, see above.
       extra_srcs: Extra (non-Haskell) source files that will be needed at compile time (e.g. by Template Haskell).
       deps: List of other Haskell libraries to be linked to this target.
+      narrowed_deps: Like deps, but only for dependencies using the modules
+          attribute. These dependencies are only used if this library uses
+          the modules attribute and the haskell_module rules depend on modules
+          provided by these dependencies.
       modules: List of extra haskell_module() dependencies to be linked into this binary.
       data: See [Bazel documentation](https://docs.bazel.build/versions/master/be/common-definitions.html#common.data).,
       compiler_flags: DEPRECATED. Use new name ghcopts.
@@ -332,6 +342,7 @@ def haskell_binary(
         srcs = srcs,
         extra_srcs = extra_srcs,
         deps = deps,
+        narrowed_deps = narrowed_deps,
         data = data,
         compiler_flags = compiler_flags,
         ghcopts = ghcopts,
@@ -391,6 +402,7 @@ def haskell_test(
         srcs = [],
         extra_srcs = [],
         deps = [],
+        narrowed_deps = [],
         data = [],
         compiler_flags = [],
         ghcopts = [],
@@ -442,6 +454,10 @@ def haskell_test(
       srcs: Haskell source files. File names must match module names, see above.
       extra_srcs: Extra (non-Haskell) source files that will be needed at compile time (e.g. by Template Haskell).
       deps: List of other Haskell libraries to be linked to this target.
+      narrowed_deps: Like deps, but only for dependencies using the modules
+          attribute. These dependencies are only used if this library uses
+          the modules attribute and the haskell_module rules depend on modules
+          provided by these dependencies.
       modules: List of extra haskell_module() dependencies to be linked into this test.
       data: See [Bazel documentation](https://docs.bazel.build/versions/master/be/common-definitions.html#common.data).,
       compiler_flags: DEPRECATED. Use new name ghcopts.
@@ -478,6 +494,7 @@ def haskell_test(
         srcs = srcs,
         extra_srcs = extra_srcs,
         deps = deps,
+        narrowed_deps = narrowed_deps,
         data = data,
         compiler_flags = compiler_flags,
         ghcopts = ghcopts,
@@ -523,6 +540,7 @@ def haskell_library(
         srcs = [],
         extra_srcs = [],
         deps = [],
+        narrowed_deps = [],
         modules = [],
         data = [],
         compiler_flags = [],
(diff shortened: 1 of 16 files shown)


facundominguez:
Currently, narrowing will fail on any module that uses Template Haskell, but the PR is otherwise self-contained and can give me early feedback for the rest of the implementation.

facundominguez:
cc @Radvendii

haskell/defs.bzl
```diff
@@ -53,6 +57,7 @@ _haskell_common_attrs = {
     "deps": attr.label_list(
         aspects = [haskell_cc_libraries_aspect],
     ),
+    "narrowed_deps": attr.label_list(providers = [HaskellLibraryInfo]),
```
aherrmann, line 60:
```suggestion
    "narrowed_deps": attr.label_list(providers = [HaskellInfo, HaskellLibraryInfo]),
```
IIUC this needs to provide `HaskellInfo` for the modules information, and `HaskellLibraryInfo`, because it should be a `haskell_library`, correct?

haskell/experimental/private/module.bzl
```diff
@@ -361,29 +418,36 @@ def build_haskell_modules(ctx, hs, cc, posix, package_name, with_shared, hidir,
...
         object_inputs = _collect_module_inputs(module_objects, os, dep)
+        narrowed_interfaces = _collect_narrowed_deps_module_files(per_module_transitive_interfaces, dep)
 
         _build_haskell_module(
             ctx,
             hs,
             cc,
             posix,
             dep_info,
+            narrowed_deps_info,
```
aherrmann, line 443: Are they used separately, or could they be passed in merged form same as `interface_inputs`?

haskell/private/haskell_impl.bzl
```diff
@@ -511,6 +511,7 @@ def haskell_library_impl(ctx):
         compile_flags = c.compile_flags,
         user_compile_flags = user_compile_flags,
         user_repl_flags = _expand_make_variables("repl_ghci_args", ctx, ctx.attr.repl_ghci_args),
+        per_module_transitive_interfaces = module_outputs.per_module_transitive_interfaces,
```
aherrmann, line 514: Wouldn't it be enough for this dictionary to only capture the modules that are in the `haskell_library`'s `modules` attribute? A `haskell_module` can only depend on other `haskell_module`s that are either part of the enclosing `haskell_library|...`'s `modules` attribute or contained in a direct `narrowed_deps` dependency, correct?

IIUC this could currently capture more modules than just that, namely all transitive module dependencies of all the modules in the `modules` attribute. I worry that in a project where every `haskell_library` uses `narrowed_deps` this could grow very large.

haskell/defs.bzl
facundominguez: That is correct. The reason I didn't add `HaskellInfo` is that I expect it to be present whenever `HaskellLibraryInfo` is.

Also, isn't the quality of the error message better with a fail call? Is the user supposed to know which rules provide which providers?

haskell/experimental/private/module.bzl
facundominguez: They are used separately. All of the interface files in `dep_info` are passed as inputs, while only a few from `narrowed_deps_info` are passed.

facundominguez:
I still need to test binaries that use `narrowed_deps`. I also found linking errors when trying to use a library that uses `narrowed_deps`, looks like I need to somehow propagate the fact that the libraries in `narrowed_deps` need to be linked as well.

haskell/private/haskell_impl.bzl
aherrmann: Ah, I think I missed the "excluding modules that come from libraries in narrowed_deps" in https://github.com/tweag/rules_haskell/blob/321129c1e669adbb3b342093948c1690aa53812d/haskell/experimental/private/module.bzl#L339-L341
Makes sense then. Thank you for clarifying.

haskell/experimental/private/module.bzl
aherrmann: Makes sense, thanks for clarifying.

haskell/defs.bzl
aherrmann: > Also, isn't the quality of the error message better with a fail call? Is the user supposed to know which rules provide which providers?

Fair enough, that is an issue. The error message takes the form
> in deps attribute of haskell_library rule //some:target: '//some:dep' does not have mandatory providers: ['HaskellInfo', 'HaskellLibraryInfo']

which is indeed not terribly informative. Happy to go with a `fail` approach instead.

facundominguez:
When testing a binary with `narrowed_deps`, a flaw in the design became evident. It is related to our discussion on computing `per_module_transitive_interfaces`.

`HaskellModuleInfo.transitive_module_dep_labels` is a `depset` that contains the labels of transitive dependencies in the graph of `haskell_module` rules. This depset is used to compute the depsets of inputs to each build action.

Since we are allowing cross-library dependencies between modules, now `transitive_module_dep_labels` contains modules from many libraries down the dependency hierarchy. If we are to narrow inputs for all of these libraries, I will need to collect the interface files for all of them. Right now, I was collecting only interfaces from libraries pointed to with `narrowed_deps` and excluding the libraries that they in turn point in their `narrowed_deps`.

This I can probably arrange, but then the depset flattening we are doing per module in `_reorder_module_deps_to_postorder` is going to become expensive since we will start doing it on larger depsets.

I think that in order to fix this, I will need to compute separately the depset with the old meaning of transitive_module_dep_labels. That is, labels of transitive module dependencies coming from the same enclosing library. However, I don't think I can do this without recursion or emulating it with unbounded loops, unless we split the `deps` attribute of `haskell_module` as well.

Thus, we would have `deps` to refer to modules in the same enclosing library, and `cross_library_deps` to refer to modules in other libraries.

What do you think?

aherrmann:
> Is the depset flattening so bad to deserve a new attribute in `haskell_module`?

As I understand it would imply an `O(N*M)` cost with `N` being the total number of `haskell_library` rules in the project and `M` the total number of `haskell_module` rules in the project (assuming everything uses narrowed deps). Compared to `O(M)` if we separate "internal" and "external" modules for the purposes of `depset` flattening.

> Thus, we would have `deps` to refer to modules in the same enclosing library, and `cross_library_deps` to refer to modules in other libraries.

Right, similar to the `extra_imports` in https://github.com/tweag/rules_haskell/issues/1631#issuecomment-984451514, but with labels.

Yes, that may be necessary. I suspect the distinction might be relevant in other places as well, e.g. the REPL in future.

facundominguez:
Ladys and gentlemen :loudspeaker:, you claimed loud and enthusiastically for splitting `deps` into `deps` and `cross_library_deps`:
Split-ting! :fist_raised: Split-ting! :fist_raised:  Split-ting! :fist_raised:
...
And here it is! :fireworks: :fireworks: :fireworks:

haskell/defs.bzl
facundominguez: I tried `fail` in 1095d23.

haskell/defs.bzl
```diff
@@ -309,6 +315,10 @@ def haskell_binary(
       srcs: Haskell source files. File names must match module names, see above.
       extra_srcs: Extra (non-Haskell) source files that will be needed at compile time (e.g. by Template Haskell).
       deps: List of other Haskell libraries to be linked to this target.
+      narrowed_deps: Like deps, but only for dependencies using the modules
+          attribute. These dependencies are only used if this library uses
+          the modules attribute and the haskell_module rules depend on modules
+          provided by these dependencies.
       modules: List of extra haskell_module() dependencies to be linked into this binary.
```
aherrmann, line 322: `haskell_module` is still marked as experimental. Let's also mark `narrowed_deps` and `modules` as experimental while that is the case.

haskell/experimental/private/module.bzl
```diff
@@ -304,28 +314,95 @@ def _collect_module_inputs(module_input_map, directs, dep):
...
+
+    Args:
+      rule_label: The label of the rule with the modules attribute
+      narrowed_deps: The contents of the narrowed_deps attribute
+
+    Returns:
+      struct(per_module_transitive_interfaces):
+        per_module_transitive_interfaces: dict of module labels to their
+            interfaces and the interfaces of their transitive module
+            dependencies
```
aherrmann, line 383: Nit: Why the surrounding `struct`?

aherrmann (approved):
Nice work, thank you for pushing this through!

Only minor comments. Feel free to merge after addressing.

haskell/defs.bzl
facundominguez: I added notes to both attributes saying that they are experimental.

aherrmann: Could you also add such notes to the other instances at `haskell_library` and `haskell_test`?

facundominguez: Done. I forgot that this piece of documentation was duplicated.

PR #1653, opened 2021-12-07
Fix Bazel CI

aherrmann:
Closes #1650

- Fixes [failure due to `CcToolchainType` mismatch](https://buildkite.com/bazel/rules-haskell-haskell/builds/964)
    ```
    (00:15:43) ERROR: /workdir/tests/binary-with-lib-dynamic/BUILD.bazel:13:16: in _haskell_library rule //tests/binary-with-lib-dynamic:lib:
    Traceback (most recent call last):
    	File "/workdir/haskell/private/haskell_impl.bzl", line 378, column 25, in haskell_library_impl
    		cc = cc_interop_info(
    	File "/workdir/haskell/cc.bzl", line 86, column 57, in cc_interop_info
    		feature_configuration = cc_common.configure_features(
    Error in configure_features: in call to configure_features(), parameter 'cc_toolchain' got value of type 'ToolchainInfo', want 'CcToolchainInfo'
    ```
    The issue was that the asterius toolchain setup was not calling `finc_cc_toolchain` but instead directly referencing the cc toolchain. This can cause mismatching result types, either `CcToolchainInfo` or `ToolchainInfo` depending on the Bazel configuration.
- Changes rules_haskell to consistently use `rules_cc` for the CC toolchain.
- ~Fixes [failure due to missing `libgmp`](https://buildkite.com/bazel/rules-haskell-haskell/builds/966#0085fa7a-c25c-4dcf-8862-a26f59d82dde/229-290)~
    ~Adds commands to `presubmit.yml` to install `libgmp-dev` and `libgmp10`.~
    Fixed by https://github.com/tweag/rules_haskell/pull/1655

Tested [successfully on Bazel CI](https://buildkite.com/bazel/rules-haskell-haskell/builds/971#9c534d1f-3d3c-4c71-bbc2-5279d6de3a1d)

cc @meteorcloudy

PR #1655, opened 2021-12-07
Fix "missing libgmp.so" error

philwo:
Bazel CI's base image no longer contains a lot of potential dependencies used by projects, but instead now it's possible to install dependencies using `sudo apt install` easily. This should probably fix the reported error.


aherrmann (approved):
Thank you!

aherrmann:
I merged this one in favor of the corresponding fix in https://github.com/tweag/rules_haskell/pull/1653. Once the rest of https://github.com/tweag/rules_haskell/pull/1653 is merged Bazel CI should be green again.

Issue #1657, opened 2021-12-08
Bazel CI: rules_haskell failing with Bazel@HEAD in downstream


meteorcloudy:
After the fixes of https://github.com/tweag/rules_haskell/pull/1653 and https://github.com/tweag/rules_haskell/pull/1655, there are new failures with Bazel@HEAD

https://buildkite.com/bazel/bazel-at-head-plus-disabled/builds/1135#bd2706c7-71d5-4ff3-ba18-0db702a7456a
```
(10:47:02) ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/1a898b3d70076f020535b91eba1e94e2/external/linux_amd64_asterius/BUILD:793:18: While resolving toolchains for target @linux_amd64_asterius//:toolchain-impl: no matching toolchains found for types @build_bazel_rules_nodejs//toolchains/node:toolchain_type
```

This is reproducible with Bazel 5.0.0rc2:
```
export USE_BAZEL_VERSION=5.0.0rc2
bazelisk build --build_tag_filters=-requires_nix,-requires_lz4,-requires_shellcheck,-requires_threaded_rts,-dont_test_with_bindist,-dont_test_on_bazelci //tests/...
```

aherrmann:
@meteorcloudy Thank you for notifying!

I can reproduce this locally:
```
$ USE_BAZEL_VERSION=5.0.0rc2 bazel test //... --nobuild --toolchain_resolution_debug=nodejs.*
...
... Rejected toolchain @nodejs_linux_amd64_config//:toolchain; mismatching values: linux, x86_64
...
ERROR: ... While resolving toolchains for target @linux_amd64_asterius//:toolchain-impl: no matching toolchains found for types @build_bazel_rules_nodejs//toolchains/node:toolchain_type
$ USE_BAZEL_VERSION=4.2.2 bazel test //... --nobuild --toolchain_resolution_debug
...
... Selected execution platform @local_config_platform//:host, type @build_bazel_rules_nodejs//toolchains/node:toolchain_type -> toolchain @nodejs_linux_amd64_config//:toolchain, ...
...
```

I think this has to do with the Asterius toolchain depending on the NodeJS toolchain and the change of the default of [`incompatible_override_toolchain_transition`](https://github.com/bazelbuild/bazel/issues/11584) in Bazel 5.
In particular, I can get Bazel 5 to pass with
```
[...]

PR #1659, opened 2021-12-09
override_toolchain_transition support

ylecornec:
This PR intends to fix issue #1657 by supporting the `--incompatible_override_toolchain_transition` flag, which becomes enable by default on bazel 5 (I did not however test it on bazel 5).

The asterius haskell toolchain now depends on others toolchains through an attribute with cfg = "exec".
With the exception of rules_sh which declares no target platform constraints, but I now wonder if it would be more consistent to do the same thing for it.

Some new utility targets are declared like `@rules_haskell//haskell:nodejs_toolchain`.
I also tried to declare them in the same external repository as the `haskell_toolchain` but I think this would require adding a parameter to the `haskell_toolchain` rule in order to know the correct workspace name to use [in the default labels](https://github.com/tweag/rules_haskell/compare/ylecornec/override_toolchain_transition?expand=1#diff-f83278238101cead941a012a809724f4fcac6997d0106d70678372f5c533ca84R336-R339).

The second commit adds the `--incompatible_override_toolchain_transition` flag to `.bazelrc` but the ci was also run without it on the first commit.

### DIFF
--- a/haskell/BUILD.bazel
+++ b/haskell/BUILD.bazel
@@ -20,6 +20,13 @@ load(
     "@rules_haskell//haskell:toolchain_info.bzl",
     "haskell_toolchain_info",
 )
+load(
+    "@rules_haskell//haskell:toolchain.bzl",
+    "get_cc_toolchain",
+    "get_haskell_toolchain",
+    "get_nodejs_toolchain",
+    "get_posix_toolchain",
+)
 
 exports_files(
     glob(["*.bzl"]) + [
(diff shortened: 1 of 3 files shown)


haskell/BUILD.bazel
```diff
@@ -173,6 +179,21 @@ haskell_toolchain_info(
...
+
+get_haskell_toolchain(
+    name = "haskell_toolchain",
+    visibility = ["//visibility:public"],
+)
+
+get_cc_toolchain(
+    name = "cc_toolchain",
+    visibility = ["//visibility:public"],
+)
```
aherrmann, line 195: A common convention is to call these `current_XYZ_toolchain`. Might as well follow it here.

haskell/toolchain.bzl
```diff
@@ -212,9 +213,9 @@ def _haskell_toolchain_impl(ctx):
 
     if ctx.attr.asterius_binaries:
         tools_config = asterius_tools_config(
-            exec_cc_toolchain = find_cc_toolchain(ctx),
+            exec_cc_toolchain = ctx.attr._exec_cc_toolchain[platform_common.ToolchainInfo],
```
aherrmann, line 216:
```suggestion
            exec_cc_toolchain = ctx.attr._exec_cc_toolchain[cc_common.CcToolchainInfo],
```
Otherwise this fails with Bazel 5.

You can relatively easily test against Bazel 5 in bindist mode by using bazelisk with the env-var `USE_BAZEL_VERSION=5.0.0rc2`.

aherrmann (approved):
Thank you!

Issue #1635, opened 2021-11-19
rules_haskell demands additional dependencies when using bazelisk


facundominguez:
When using `bazelisk`, `rules_haskell` requires the user to manually define `@build_bazel_rules_nodejs`.

```
# bazelisk build //...
2021/11/19 19:55:34 Using unreleased version at commit a6e97a7e59e80bbe5dd4c63cf1c4f4802daa94ae
Extracting Bazel installation...
Starting local Bazel server and connecting to it...
INFO: Invocation ID: ac1ef87c-65e6-4288-af84-26b0b823e21e
ERROR: /mnt/gazelle_haskell_modules/example/package-a/BUILD.bazel:108:15: Target '//package-a:package-a.PackageA.Other.B' depends on toolchain '@rules_haskell_ghc_linux_amd64//:toolchain-impl', which cannot be found: error loading package '@rules_haskell_ghc_linux_amd64//': at /root/.cache/bazel/_bazel_root/3799ddd95a8b72a9bb7ce448d71f5b59/external/rules_haskell/haskell/asterius/defs.bzl:8:6: Unable to find package for @build_bazel_rules_nodejs//:index.bzl: The repository '@build_bazel_rules_nodejs' could not be resolved: Repository '@build_bazel_rules_nodejs' is not defined.'
ERROR: Analysis of target '//package-a:package-a.PackageA.Other.B' failed; build aborted: Analysis failed
INFO: Elapsed time: 247.747s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (97 packages loaded, 7860 targets configured)
    currently loading: @rules_haskell_ghc_linux_amd64//
```

I'm conjecturing that [this version check](https://github.com/tweag/rules_haskell/blob/master/haskell/repositories.bzl#L104) is failing and leaves `build_bazel_rules_node_js` undefined. Adding the following to the workspace file fixes it.

```pyhton
load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")

maybe(
    http_archive,
    name = "build_bazel_rules_nodejs",
    sha256 = "0fa2d443571c9e02fcb7363a74ae591bdcce2dd76af8677a95965edf329d778a",
    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/3.6.0/rules_nodejs-3.6.0.tar.gz"],
)
```

I'm not using Asterious in this project.

facundominguez:
cc @ylecornec

[...]

PR #1660, opened 2021-12-09
separate ahc.BUILD.tpl file for asterius.

ylecornec:
Before this PR, the same `ghc.BUILD.tpl` file was used for both asterius and regular haskell toolchain, but now the asterius one uses a new `ahc.BUILD.tpl` file.

Most notably, this prevents loading the `@rules_haskell//haskell/asterius:defs.bzl` file when it is not needed, which is what seems to be happening in issue #1635.


aherrmann (approved):
Thank you!

PR #1662, opened 2021-12-10
Add CC toolchain and wrapper inputs to REPL

### LINKED ISSUE #1661: cc_wrapper-python dependency is not built by haskell-language-server flags
**Describe the bug**

When building `--output_groups=hie_bios` for some `haskell_library` or `haskell_binary`, the output contains references to the path to `cc_wrapper-python` which may not exists.

It actually depends on the type of `haskell_library` built. If that ones contains `chs` files, the file will exists, if not, it won't.

This breaks haskell-language-server in a surprising way here, because the file may be missing or not depending on the history of build on this repository.

**To Reproduce**

In a codebase using rules_haskell, after a `bazel clean`:

```
# Build the HLS flags
$ bazel build //path/to/a/repl --output_groups=hie_bios
[...]
  bazel-bin/path/to/a/repl@hie-bios
[...]

# Look at the flag file
$ cat bazel-bin/path/to/a/repl@hie-bios | grep cc_wrapper
bazel-out/k8-fastbuild/bin/external/rules_haskell/haskell/cc_wrapper-python
[...]

# Test if the dependency exists
$ ls bazel-out/k8-fastbuild/bin/external/rules_haskell/haskell/cc_wrapper-python
ls: cannot access 'bazel-out/k8-fastbuild/bin/external/rules_haskell/haskell/cc_wrapper-python': No such file or directory

# Build another dependency
$ bazel build //path/to/another/repl --output_groups=hie_bios
[...]

# Check for the file
$ ls bazel-out/k8-fastbuild/bin/external/rules_haskell/haskell/cc_wrapper-python
bazel-out/k8-fastbuild/bin/external/rules_haskell/haskell/cc_wrapper-python
```

The difference between the first and second repl is that the first does not need to run convert `chs` files to `hs` files in order to build the flag list, hence it does not need the `cc_wrapper-python`.

**Expected behavior**

All the files listed in the content of the `hie_bios` file must exists.

**Environment**
 - OS name + version:
 - Bazel version:
 - Version of the rules:

**Additional context**

A workaround is to prebuild the dependency using `bazel build @rules_haskell//haskell:cc_wrapper-python`

aherrmann:
Closes #1661

The CC wrapper and other CC inputs were not declared as required inputs to the generated hie-bios file.

cc @guibou

### DIFF
--- a/haskell/repl.bzl
+++ b/haskell/repl.bzl
@@ -347,7 +347,7 @@ def _compiler_flags_and_inputs(hs, cc, repl_info, static = False, path_prefix = 
         depset(all_library_files),
         depset([hs.toolchain.locale_archive] if hs.toolchain.locale_archive else []),
         cc.all_files,
-        hs.toolchain.cc_wrapper.inputs,
+        hs.toolchain.cc_wrapper.runfiles.files,
         repl_info.dep_info.interface_dirs,
     ])
     return (args, inputs)
--- a/haskell/repl.bzl
+++ b/haskell/repl.bzl
@@ -3,6 +3,7 @@
 load("@bazel_skylib//lib:dicts.bzl", "dicts")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@bazel_skylib//lib:shell.bzl", "shell")
+load("@rules_cc//cc:find_cc_toolchain.bzl", "find_cc_toolchain")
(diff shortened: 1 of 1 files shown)


guibou:
Thank you!

ylecornec (approved):
Looks good to me :+1:

PR #1663, opened 2021-12-14
Narrowing of module dependencies when using TemplateHaskell

facundominguez:
This PR implements the part of #1631 needed for Template Haskell.


facundominguez:
This PR has warnings printed in the console as follows:
```
$ nix-shell --pure --run "bazel test //tests/haskell_module/dep-narrowing-th/..."
INFO: Invocation ID: b7aead2d-321f-45b9-8bf2-3324990e5b0c
INFO: Analyzed 10 targets (0 packages loaded, 0 targets configured).
INFO: Found 9 targets and 1 test target...
INFO: From HaskellBuildObject //tests/haskell_module/dep-narrowing-th:lib //tests/haskell_module/dep-narrowing-th:TestModule:
Warning: the following files would be used as linker inputs, but linking is not being done: bazel-out/k8-fastbuild/bin/tests/haskell_module/dep-narrowing-th/_obj/TestLib2/SimpleFoo.o bazel-out/k8-fastbuild/bin/tests/haskell_module/dep-narrowing-th/_obj/TestLib2/SimpleFoo.dyn_o bazel-out/k8-fastbuild/bin/tests/haskell_module/dep-narrowing-th/_obj/TestLib2/TestLibModule2.o
INFO: From HaskellBuildObject //tests/haskell_module/dep-narrowing-th:lib //tests/haskell_module/dep-narrowing-th:TestModule2:
Warning: the following files would be used as linker inputs, but linking is not being done: bazel-out/k8-fastbuild/bin/tests/haskell_module/dep-narrowing-th/_obj/TestLib2/SimpleFoo.o
INFO: Elapsed time: 1.169s, Critical Path: 1.08s
INFO: 10 processes: 9 remote cache hit, 1 linux-sandbox.
INFO: Build completed successfully, 10 total actions
//tests/haskell_module/dep-narrowing-th:Test                    (cached) PASSED in 0.0s

Executed 0 out of 1 test: 1 test passes.
INFO: Build completed successfully, 10 total actions
```
I still need to fix that somehow.

haskell/private/haskell_impl.bzl
```diff
@@ -53,6 +53,20 @@ load("@rules_cc//cc:find_cc_toolchain.bzl", "find_cc_toolchain")
...
+# for running TemplateHaskell splices. It wants to do this even when all the
+# necessary object files are passed in the command line.
+#
+# In order to satisfy GHC, and yet avoid passing the linked library as input, we
+# create a ficticious package which points to an empty shared library. The
+# ficticious and the real package share the same interface files.
+#
+# Avoiding to pass the real shared library as input is necessary when building
+# individual modules with haskell_module, otherwise building the module would
+# need to wait until all of the modules of library dependencies have been built.
```
aherrmann, line 68: Thank you for adding this!

Do we also document the motivation in a note somewhere? If not, I think that would be good to document as well[ either here, or in a separate note.

facundominguez: > Do we also document the motivation in a note somewhere?

I'm sorry, I think I need this spelled more explicitly. Which motivation would you like to see in a note?

haskell/experimental/private/module.bzl
```diff
@@ -465,6 +469,7 @@ def build_haskell_modules(ctx, hs, cc, posix, package_name, with_shared, hidir,
             module_outputs[dep.label],
             depset(transitive = [interface_inputs, narrowed_interfaces, narrowed_objects]),
             object_inputs,
+            narrowed_objects,
```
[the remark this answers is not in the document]
facundominguez: I'd expect `interface_inputs` to only contain interface files, while `narrowed_objects` would only contain objects. Is there any overlap still possible?

haskell/private/haskell_impl.bzl
aherrmann: The narrowed inputs. Essentially, why is the following a problem that we want to solve?

> GHC 8.10.x wants to load the shared libraries corresponding to packages needed for running TemplateHaskell splices

facundominguez: Does the last paragraph in the note address that question?
```
# Avoiding to pass the real shared library as input is necessary when building
# individual modules with haskell_module, otherwise building the module would
# need to wait until all of the modules of library dependencies have been built.
```
The other way to go about it would be to write a grand story for what narrowing does and how it works. This would amount roughly to putting the description of #1631 in a comment and providing links to the intervening parts of the implementation. GHC has some notes like that.

facundominguez:
Passing object files on the command line doesn't convince `ghc` to load them in the external interpreter, they are only loaded in the internal interpreter. Thus, profiling is beyond reach for now.

One thing I could do is to try to disable narrowing when profiling is enabled and treat `narrowed_deps` as `deps` in that case. What do you think?

facundominguez:
> One thing I could do is to try to disable narrowing when profiling is enabled and treat narrowed_deps as deps in that case. 

Ah, another thing we could do is to ask the user to say if the module uses TH, and when building with profiling we can have the modules thus marked handle `narrowed_deps` the same as `deps`. That is, they will depend on the real libraries of the dependencies in `narrowed_deps`.

The difference with my former proposal is that modules that don't use TH can still use narrowing.

haskell/private/haskell_impl.bzl
facundominguez: > The other way to go about it would be to write a grand story for what narrowing does and how it works.

I tried a note in 02778b9.

haskell/experimental/private/module.bzl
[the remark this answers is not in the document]
facundominguez: Ah, I got it. I removed `narrowed_objects` from the `interface_inputs` parameter of `_build_haskell_module` in f8681508.

haskell/private/haskell_impl.bzl
aherrmann: Thank you! Yes, I think that describes it well.

aherrmann:
> Ah, another thing we could do is to ask the user to say if the module uses TH, and when building with profiling we can have the modules thus marked handle `narrowed_deps` the same as `deps`. That is, they will depend on the real libraries of the dependencies in `narrowed_deps`.

Yes, that sounds like a good option. IIUC this would also allow us to just not pass objects in if the module does not use TH, thereby potentially further increasing incrementality.

> I would call the new attribute `dont_narrow_deps`, and the user could use `select` to disable narrowing on profiling builds.

Why not call it something like `enable_th` and automatically switch to the non-narrowing behavior on profiling builds without requiring the select?

facundominguez:
> Why not call it something like enable_th and automatically switch to the non-narrowing behavior on profiling builds without requiring the select?

That is an option. The problem with that name is that `enable_th` is only necessary if using `cross_library_deps` and only if doing a profiling build. TH is "enabled/supported" in all other cases.

aherrmann:
> The problem with that name is that `enable_th` is only necessary if using `cross_library_deps` and only if using the external interpreter.

Couldn't the other modes also benefit from it by avoid inputs that are unused anyway? That could prevent recompilation in cases where only the object changes, but not the interface.

> Also, profiling is only one such configuration that needs the external interpreter, cross-compilation is another one, do we need the external interpreter in other cases?

Those are the only two cases.

facundominguez:
> Couldn't the other modes also benefit from it by avoid inputs that are unused anyway? That could prevent recompilation in cases where only the object changes, but not the interface.

hm, is that possible at all?

aherrmann:
> > Couldn't the other modes also benefit from it by avoid inputs that are unused anyway? That could prevent recompilation in cases where only the object changes, but not the interface.
> 
> hm, is that possible at all?

Without TH we don't need any dependency objects or shared or static libraries in the inputs, correct? So, we could just omit them if `enable_th = False`. Perhaps I'm missing something?

facundominguez:
> Without TH we don't need any dependency objects or shared or static libraries in the inputs, correct?

This is correct. What I am pondering is how likely it is that the interface file doesn't change while the object file does. It is looking like it could be frequent in unoptimized builds for changes that don't affect the ABI of a module (types and type-signatures).

I am lazy to do this, because then I will have to make `gazelle_haskell_modules` detect if a module needs TH. But it is looking like it is worth to consider.

aherrmann:
> I was lazy to do this because then I will have to make `gazelle_haskell_modules` detect if a module needs TH. But it is worth considering it.

I'd imagine it comes down to whether the module has `{-# LANGUAGE TemplateHaskell #-}` or the `ghcopts` include `-XTemplateHaskell`. Is there more to it?



> I renamed and repurposed `dont_narrow_deps` to `enable_th`. The new `enable_th` attribute controls whether object files and libraries are passed as inputs to the build action. When doing profiling builds, it passes libraries of `narrowed_deps` as inputs instead of the object files.

Thanks!

aherrmann (approved):
Thank you! Good work!

haskell/private/ghc_wrapper.sh
```diff
@@ -14,5 +14,5 @@ if [ "$2" == "--persistent_worker" ]; then
 else
     while IFS= read -r line; do extra_args+=("$line"); done < "$2"
     "${compile_flags[@]}" "${extra_args[@]}" 2>&1 \
-      | while IFS= read -r line; do [[ $line =~ ^Loaded ]] || echo "$line"; done >&2
+      | while IFS= read -r line; do [[ $line =~ ^(Loaded|Warning: the following files would be used as linker inputs, but linking is not being done:) ]] || echo "$line"; done >&2
```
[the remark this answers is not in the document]
facundominguez: Done in e78e162.

facundominguez:
> I'd imagine it comes down to whether the module has {-# LANGUAGE TemplateHaskell #-} or the ghcopts include -XTemplateHaskell. Is there more to it?

The interpreter is used when using `ANN` pragmas, `QuasiQuotes` implies `TemplateHaskell`, and `-XTemplateHaskell` could appear in `OPTIONS_GHC` pragmas. Maybe we could get away in the beginning with just looking for `{-# LANGUAGE ..., TemplateHaskell, ... #-}`.

aherrmann:
> > I'd imagine it comes down to whether the module has {-# LANGUAGE TemplateHaskell #-} or the ghcopts include -XTemplateHaskell. Is there more to it?
> 
> The interpreter is used when using `ANN` pragmas, `QuasiQuotes` implies `TemplateHaskell`, and `-XTemplateHaskell` could appear in `OPTIONS_GHC` pragmas. Maybe we could get away in the beginning with just looking for `{-# LANGUAGE ..., TemplateHaskell, ... #-}` and `-XTemplateHaskell` in `ghcopts`.

Thanks for clarifying, that makes sense. Does GHC's parser API expose these pragmas in some form? If so it seems reasonably straight forward to check for them in `gazelle_haskell_module`. At least in most common cases. We don't have to cover every possible case...

facundominguez:
> Does GHC's parser API expose these pragmas in some form? 

We are using the lexer, and the lexer exposes those comments as a token
```
ITblockComment "{-# LANGUAGE DefaultSignatures, TemplateHaskell #-}"
```
We can parse the contents of the comments easily enough. Finding ANN pragmas is more expensive since we must lex the whole file to determine if they contain any ANN pragmas, we'll have to see if it is noticeable to the user.

PR #1669, opened 2021-12-21
Fix computation of transitive dependencies

facundominguez:
The most important fix here is exposing interface files coming from transitive cross-library dependencies. Otherwise, the strengthened tests would fail.

### DIFF
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -376,14 +376,14 @@ def _collect_module_inputs(module_input_map, directs, dep):
     return all_inputs
 
 def _collect_narrowed_deps_module_files(ctx_label, per_module_transitive_files, dep):
-    transitive_cross_library_dep_labels = dep[HaskellModuleInfo].transitive_cross_library_dep_labels.to_list()
+    direct_cross_library_deps = dep[HaskellModuleInfo].direct_cross_library_deps
     transitives = [
-        per_module_transitive_files[m]
-        for m in transitive_cross_library_dep_labels
-        if m in per_module_transitive_files
+        per_module_transitive_files[m.label]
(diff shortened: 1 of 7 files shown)


haskell/experimental/private/module.bzl
```diff
@@ -376,14 +376,14 @@ def _collect_module_inputs(module_input_map, directs, dep):
     return all_inputs
 
 def _collect_narrowed_deps_module_files(ctx_label, per_module_transitive_files, dep):
-    direct_cross_library_deps = dep[HaskellModuleInfo].direct_cross_library_deps
+    transitive_cross_library_dep_labels = dep[HaskellModuleInfo].transitive_cross_library_dep_labels.to_list()
```
facundominguez, line 379: This flattening could be problematic because the depset could be large. I'm still thinking how to avoid it.

facundominguez: I reverted this and created a better solution in af97995.

haskell/private/haskell_impl.bzl
```diff
@@ -177,6 +177,9 @@ def _haskell_binary_common_impl(ctx, is_test):
     if modules and ctx.files.srcs:
         fail("""Only one of "srcs" or "modules" attributes must be specified in {}""".format(ctx.label))
 
+    if not modules and ctx.attr.narrowed_deps:
+        fail("""The attribute "narrowed_deps" is enabled only if "modules" is specified in {}""".format(ctx.label))
```
aherrmann, line 181:
```suggestion
        fail("""The attribute "narrowed_deps" can only be used if "modules" is specified in {}""".format(ctx.label))
```
Perhaps a bit clearer this way?

aherrmann (approved):
Looks good, thank you!

One small comment.

PR #1671, opened 2021-12-27
Fix title ;)

mobileink:
Truth in advertising. ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <p align="left"><img src="logo/horizontal.png" alt="rules_haskell" height="100px"></p>
 
-# Haskell rules for [Bazel][bazel]
+# [Bazel][bazel] rules for Haskell
 
 [![Continuous integration](https://github.com/tweag/rules_haskell/workflows/Continuous%20integration/badge.svg)](https://github.com/tweag/rules_haskell/actions?query=branch%3Amaster)
 


aherrmann:
Hmm, to me this seems like one of those cases where the English language is ambiguous and the meaning is provided by context:
I think it could either be "Haskell rules for Bazel" as in "rules to build Haskell with Bazel", or it could be "Bazel rules for Haskell" as in "Bazel extension to build Haskell". The context being that Bazel is the build system and Haskell is the language. Though, I'm not a native speaker, so I might be wrong.

For what it's worth, all the following rule sets use the same phrasing: https://github.com/bazelbuild/rules_go, https://github.com/bazelbuild/rules_cc, and https://github.com/bazelbuild/rules_python. Are they all wrong, or is this indeed ambiguous?

AlexeiDrake:
As a fun example of this ambiguity, Microsoft is notorious for using backwards terminology:
 - [Windows Subsystem for Linux](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux)
 - [VFS for Git](https://github.com/microsoft/VFSForGit)
 - Configuring your computer for Windows

aherrmann:
I'll close as it doesn't seem clear that the current form is actually wrong and at least it's in-line with what others do.

Issue #1673, opened 2022-01-10
all_files captures files under the bazel-out symlink


aherrmann:
**Describe the bug**
A build on CI recently failed with Bazel complaining about missing source files in the `all_files` file group:
- https://github.com/tweag/rules_haskell/runs/4754510983?check_suite_focus=true#step:7:195
- https://github.com/tweag/rules_haskell/runs/4754511059?check_suite_focus=true#step:7:200
This occurred only on the nixpkgs pipeline, but on both, Linux and MacOS.

The [corresponding filegroup](https://github.com/tweag/rules_haskell/blob/3a133d3953ee4cbb8f043f70332c6f9f306d7ec3/tutorial/BUILD.bazel#L4) includes `glob(["**"])`.

**To Reproduce**
The PR on which this occurred does not change anything seemingly related. So, this may be a flaky issue.

**Expected behavior**
Bazel should not include files under `bazel-out` or `bazel-ci-out` in the `all_files` filegroup.

**Environment**
 - OS name + version:
   - Mac OS X 10.15.7
   - Ubuntu 20.04.3 (nixpkgs)
 - Bazel version: 4.2.1
 - Version of the rules: https://github.com/tweag/rules_haskell/pull/1672 on top of `master`.

PR #1674, opened 2022-01-10
Fix all_files captures "bazel-*" directories

k1nkreet:
Fix for #1673

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,3 +1,14 @@
+queue_rules:
+  - name: default
+    conditions:
+      - 'status-success~=Build & Test - Nixpkgs \(ubuntu-.*\).*'
+      - 'status-success~=Build & Test - Nixpkgs \(macos-.*\).*'
+      - 'status-success~=Build & Test - bindist \(ubuntu-.*\).*'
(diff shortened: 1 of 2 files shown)


aherrmann (approved):
Thank you for fixing this so quickly!

k1nkreet:
There is a problem with Mergify configuration got deprecated today afaiu: https://blog.mergify.com/strict-mode-deprecation/

aherrmann:
> There is a problem with Mergify configuration got deprecated today afaiu: https://blog.mergify.com/strict-mode-deprecation/

Indeed, looks like it. I think the Mergify config on rules_sh uses the new format and could probably be adapted for rules_sh as well. See https://github.com/tweag/rules_sh/blob/master/.mergify.yml.

aherrmann:
Looks like mergify still uses the configuration from master. But, I've checked the new config in the [dashboard](https://dashboard.mergify.com/github/tweag/repo/rules_haskell/config-editor?repositoryName=rules_haskell&pullRequestNumber=1674) and it looks good.
So, merging manually...
@k1nkreet thank you!

PR #1680, opened 2022-01-13
Expose object files of transitive dependencies when using TH

facundominguez:
Turns out that we only collected object files of cross-library dependencies if TH was enabled for a given module.

Unfortunately, if a module M doesn't use TH, those dependencies might still be necessary if some other module using TH imports module M.

This PR adds a test to rehearse this case and a fix.


aherrmann (approved):
Thank you, makes sense.

PR #1682, opened 2022-01-18
Suppress bogus warnings when using plugins

### LINKED ISSUE #1681: haskell_module emits bogus warnings when using plugins
When a `haskell_module` rule uses plugins, and the rule doesn't enable TH, warnings of the following form are printed in the console:
```
INFO: From HaskellBuildObject //tests/haskell_module/plugin:lib //tests/haskell_module/plugin:module-with-plugin:

<no location info>: warning: [-Wmissed-extra-shared-lib]
    libHSexceptions-0.10.4.so: cannot open shared object file: No such file or directory
    It's OK if you don't want to use symbols from it directly.
    (the package DLL is loaded by the system linker
     which manages dependencies by itself).

<no location info>: warning: [-Wmissed-extra-shared-lib]
    libHSsplitmix-0.1.0.4.so: cannot open shared object file: No such file or directory
    It's OK if you don't want to use symbols from it directly.
    (the package DLL is loaded by the system linker
     which manages dependencies by itself).

<no location info>: warning: [-Wmissed-extra-shared-lib]
    libHSrandom-1.2.0.so: cannot open shared object file: No such file or directory
    It's OK if you don't want to use symbols from it directly.
    (the package DLL is loaded by the system linker
     which manages dependencies by itself).

<no location info>: warning: [-Wmissed-extra-shared-lib]
    libHStemporary-1.3.so: cannot open shared object file: No such file or directory
    It's OK if you don't want to use symbols from it directly.
    (the package DLL is loaded by the system linker
     which manages dependencies by itself).
```

The cause of these messages is that loading plugins requires the internal interpreter, which in turn wants to load the shared libraries of the packages in the dependencies. However, when TH is not enabled, those shared libraries are not provided as inputs to the build action, thus producing the above warnings.

This state of affairs should not cause build failures, because when the plugin needs some shared library, it will be listed in the dependencies of the plugin, and these plugin dependecies are fed as inputs even with TH disabled.

This can be reproduced with [these changes](https://github.com/tweag/rules_haskell/compare/fd/plugins-warnings) and then running
```
nix-shell --pure --run "bazel build //tests/haskell_module/plugin:lib"
```
### Expected behavior

I can think of the following fixes
1. Have `rules_haskell` add `-Wno-missed-extra-shared-lib` to the compiler flags when using plugins and TH is disabled.
2. Ask the user to suppress the warnings with `-Wno-missed-extra-shared-lib` when using plugins

Perhaps (1) is the way to go. I don't know in which cases the user would want to have the warning enabled other than debugging `rules_haskell` itself.

facundominguez:
When a `haskell_module` rule doesn't enable TH, but uses plugins, we pass `-Wno-missed-extra-shared-lib` to the build action in order to suppress bogus warnings.

Fixes #1681

### DIFF
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -225,8 +225,6 @@ def _build_haskell_module(
         for manifest in plugin.tool_input_manifests
     ]
 
-    # TODO[AH] Support package id - see `-this-unit-id` flag.
-
     args.add_all(hs.toolchain.ghcopts)
 
     args.add_all(expand_make_variables("ghcopts", ctx, ctx.attr.ghcopts, haskell_library_extra_label_attrs(ctx.attr)))
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -81,6 +81,7 @@ def _build_haskell_module(
         package_name,
         with_profiling,
         with_shared,
+        enable_th,
         hidir,
         odir,
         module_outputs,
@@ -100,6 +101,7 @@ def _build_haskell_module(
       package_name: name of this package, or empty if building a binary
       with_profiling: Whether to build profiling object files
       with_shared: Whether to build dynamic object files
+      enable_th: Whether object files and shared libraries need to be exposed to the build action
       hidir: The directory in which to output interface files
       odir: The directory in which to output object files
       module_outputs: A struct containing the interfaces and object files produced for a haskell_module.
@@ -235,7 +237,7 @@ def _build_haskell_module(
         moduleAttr.tools,
     ]
     args.add_all(expand_make_variables("ghcopts", ctx, moduleAttr.ghcopts, module_extra_attrs))
-    if module[HaskellModuleInfo].attr.enable_th:
+    if enable_th:
         args.add_all(narrowed_objects)
 
     outputs = [module_outputs.hi]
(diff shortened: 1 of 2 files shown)


haskell/experimental/private/module.bzl
```diff
@@ -236,8 +236,12 @@ def _build_haskell_module(
         moduleAttr.plugins,
         moduleAttr.tools,
     ]
+    if plugins and not enable_th:
+        # For #1681. These suppresses bogus warnings about missing libraries which
+        # aren't really needed.
```
aherrmann, line 241:
```suggestion
        # For #1681. This suppresses bogus warnings about missing libraries which
        # aren't really needed.
```
Is there a GHC ticket about these? Bogus warnings seem like a compiler issue.

aherrmann (approved):
Thank you!

haskell/experimental/private/module.bzl
facundominguez: They are bogus in the context of `rules_haskell`. But if the user asks to have these missing libraries reported with `-Wmissing-extra-shared-lib`, I couldn't say the messages are bogus in general.

We could try to argue perhaps that the interpreter shouldn't try to load these libraries automatically. But it isn't clear to me either that a plugin can't request the interpreter to evaluate code that depends on these libraries. It would count pretty much as using TH, I guess.

PR #1684, opened 2022-01-18
Avoid passing dyn_o files in the command line of ghc

### LINKED ISSUE #1683: Narrowing passes unneeded .dyn_o files in the command line
The following [test](https://github.com/tweag/rules_haskell/commit/cd77cdec42218aab64f032a6b55c111a6bea9f86) shows that dyn_o files are passed in the command line, which in turn are ignored by ghc.

```
$ nix-shell --pure --run "bazel build //tests/haskell_module/plugin:lib"
...
Warning: ignoring unrecognised input `bazel-out/k8-fastbuild/bin/tests/haskell_module/dep-narrowing-th/_obj/TestLib2/SimpleFoo.dyn_o'
...
INFO: Build completed successfully, 3 total actions
```

These dyn_o files shouldn't be passed in the command line.

### Fixing it

The dyn_o files are necessary to create shared libraries and to load home package modules when compiling modules with TH. To load modules coming from other packages we don't need `dyn_o` files, apparently.

The dyn_o files are collected together with the .o files in order to expose them to build actions. Two solutions:
1. scrap them from the command line, or
2. avoid putting them together with `.o` files.

Both solutions can be compared in the upcoming PR.

facundominguez:
This PR first scraps dyn_o files from the command line, and then undoes that to collect `.dyn_o` files and `.o` files in separate depsets, which is likely the best to maintain (but I'm only going on an intuition here).

Fixes #1683

### DIFF
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -347,14 +347,16 @@ def _collect_module_outputs_of_direct_deps(with_shared, module_outputs, dep):
             for m in dep[HaskellModuleInfo].direct_module_deps
             if m.label in module_outputs
         ]
-        os += [
+        dyn_os = [
             dyn_o
             for m in dep[HaskellModuleInfo].direct_module_deps
             if m.label in module_outputs
             for dyn_o in [module_outputs[m.label].dyn_o]
             if dyn_o  # boot module files produce no useful object files
         ]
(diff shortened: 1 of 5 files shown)


aherrmann (approved):
I agree that separating the objects seems preferable.

PR #1685, opened 2022-01-19
Handling of toolchain libraries in stack_snapshot

ylecornec:
This PR removes the hard coded list of toolchain libraries from the `bazel.bzl` file.

If the stack_snapshot is to be used by ghc, we will rely on stack to decide the toolchain libraries.
With a different compiler such as asterius, this list of toolchain libraries is configurable through the new `toolchain_libraries` attribute.

### About the call to stack

We consider libraries without a `location` field in the result of `stack ls dependencies json --global-hint --external` to be part of the toolchain.

This happens in two cases:
  - If it is present in the generated `stack.yaml` (considered as local)
  - if it is known to stack (through [this file](https://github.com/commercialhaskell/stackage-content/blob/master/stack/global-hints.yaml)) and not present in the snapshot.

So there are two kinds of mismatch to fix: 
1. toolchain libraries that stack does not know about should be added to the `stack.yaml` file.
2. libraries that stack wrongly considers part of the toolchain should be added to a custom stack snapshot.

### The `toolchain_libraries` attribute
If this attribute is set, it must contain all of the toolchain libraries.
So that we can know that a library is not in the toolchain and have an error message telling the user it should be added a custom stack snapshot.

### The WORKSPACE file

To avoid writing the list of toolchain libraries explicitely, the asterius toolchain writes this list to the `toolchain_libraries.bzl` file so that it can be imported in the WORKSPACE file.

Since this is linux only and we cannot conditionally `load` this file, the `toolchain_libraries` workspace rule will create a file with a default `None` value.


WORKSPACE
```diff
@@ -584,6 +584,29 @@ load("//tools:repositories.bzl", "rules_haskell_worker_dependencies")
...
+    stack_snapshot(
+        name = "stackage_asterius",
+        local_snapshot = "@rules_haskell//tests/asterius/stack_toolchain_libraries:snapshot.yaml",
+        packages = [
+            "xhtml",
+        ],
+        stack_snapshot_json = "@rules_haskell//tests/asterius/stack_toolchain_libraries:snapshot.json",
+        toolchain_libraries = toolchain_libraries,
+    ) if is_linux else None
+)
```
aherrmann, line 608: I think the outer parenthesis are not strictly necessary.

ghcide-stack-snapshot.yaml
```diff
@@ -10,3 +10,6 @@ packages:
 - implicit-hie-cradle-0.3.0.2
 - lsp-test-0.11.0.7
 - retrie-0.1.1.1
+
+drop-packages:
+- Win32
```
aherrmann, line 15: Maybe worth a comment as to why this is being dropped.

stackage-pinning-test.yaml
```diff
@@ -12,3 +12,6 @@ packages:
 
   - archive: https://github.com/tweag/rules_haskell/raw/e4e74f17f743488f564bd0d69c580106d5b910a5/tests/haskell_cabal_library_sublibrary_name/package1.tar
     sha256: "302d8ddda8330c825da61fe0a2315c899ab083e641c7716ebdacb5c951682445"
+
+drop-packages:
+  - Win32
```
aherrmann, line 17: Same here

stackage_snapshot.yaml
```diff
@@ -0,0 +1,3 @@
+resolver: lts-18.18
+drop-packages:
+  - Win32
```
aherrmann, line 3: Same here

aherrmann (approved):
Good work, thank you!

Thank you for the test and the changelog entry!

PR #1686, opened 2022-01-25
haskell_module: Pass cc toolchain linker flags when using the interpreter

facundominguez:
This gets rid of the "warning: type and size of dynamic symbol ... are not defined"

### DIFF
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -188,6 +188,11 @@ def _build_haskell_module(
 
     args.add_all(cc.include_args)
 
+    if plugins or enable_th:
+        # cc toolchain linker flags would be necessary when the interpreter wants to
+        # load any libraries
+        args.add_all(["-optl" + f for f in cc.linker_flags])
+
     # Collect library dependency arguments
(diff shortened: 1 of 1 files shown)


haskell/experimental/private/module.bzl
```diff
@@ -188,6 +188,11 @@ def _build_haskell_module(
 
     args.add_all(cc.include_args)
 
+    if plugins or enable_th:
+        # cc toolchain linker flags would be necessary when the interpreter wants to
+        # load any libraries
+        args.add_all(["-optl" + f for f in cc.linker_flags])
```
aherrmann, line 194:
```suggestion
        args.add_all(cc.linker_flags, format_each = "-optl%s")
```
See [optimization recommendations](https://docs.bazel.build/versions/main/skylark/performance.html#use-ctxactionsargs-for-command-lines).

aherrmann (approved):
Thank you, looks good!
Only one minor comment.

haskell/experimental/private/module.bzl
facundominguez: Done.

PR #1689, opened 2022-02-07
Reproducible interface files for Cabal packages with C dependencies

aherrmann:
This is the continuation of https://github.com/tweag/rules_haskell/pull/1648 addressing the case where a Cabal target has C dependencies.

Addresses the non-reproducible interface files aspect of https://github.com/tweag/rules_haskell/issues/1600.

As described in https://github.com/haskell/cabal/issues/1317#issuecomment-1025942396 we can avoid passing absolute C header and library search paths to Cabal builds if we do not call `Setup.hs install`. [`Setup.hs install` fails](https://github.com/haskell/cabal/issues/1317#issuecomment-985537653) if the configuration contains relative paths. Instead, we call `Setup.hs copy` to copy the generated files and then generate a package configuration file with `Setup.hs register --gen-pkg-config`, patch it to replace relative paths by paths starting with `${pkgroot}`, and then invoke `ghc-pkg` directly to cache the package-db.

This adds a test-case that be used to confirm that the build is now reproducible.
```
$ n=1; bazel clean; bazel build //tests/haskell_cabal_reproducibility/pkg-b --execution_log_json_file=execlog$n.json; rsync -aL bazel-bin output$n
$ n=2; bazel clean; bazel build //tests/haskell_cabal_reproducibility/pkg-b --execution_log_json_file=execlog$n.json; rsync -aL bazel-bin output$n
$ diff -u execlog1.json execlog2.json
$ diff -ur output1 output2
$ diff -u <(ghc --show-iface output1/bazel-bin/tests/haskell_cabal_reproducibility/pkg-a/_install/pkg-a-0.1.0.0_iface/LibA.dyn_hi) <(ghc --show-iface output2/bazel-bin/tests/haskell_cabal_reproducibility/pkg-a/_install/pkg-a-0.1.0.0_iface/LibA.dyn_hi)
```

When https://github.com/tweag/rules_haskell/pull/1645 is merged this could be turned into an integration test.

### DIFF
--- a/haskell/private/cabal_wrapper.py
+++ b/haskell/private/cabal_wrapper.py
@@ -107,7 +107,8 @@ execroot = os.getcwd()
 setup = os.path.join(execroot, json_args["setup_path"])
 srcdir = os.path.join(execroot, json_args["pkg_dir"])
 # By definition (see ghc-pkg source code).
-pkgroot = os.path.realpath(os.path.join(execroot, os.path.dirname(json_args["package_db_path"])))
+rel_pkgroot = os.path.dirname(json_args["package_db_path"])
+pkgroot = os.path.realpath(os.path.join(execroot, rel_pkgroot))
 libdir = os.path.join(pkgroot, "{}_iface".format(name))
 dynlibdir = os.path.join(pkgroot, "lib")
 bindir = os.path.join(pkgroot, "bin")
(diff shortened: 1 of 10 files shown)


ylecornec (approved):
Looks good to me :+1:

cocreature:
Thanks a lot!

PR #1690, opened 2022-02-09
Add th_deps attribute for compile-time dependencies

jcpetruzza:
Implements the proposal in #1382. 

A new `th_deps` attribute is added to `haskell_(library|binary|test)` that is similar to `deps` but makes only the libraries in the formaer available at compile-time. Dependencies in `th_deps` are also made available during linking: FFI declarations also need to be resolved at link time, even if they were used only for the sake of TH.

As discussed in the issue, this reduces the number of dependencies for the build action, creating more opportunities for parallel work.

In addition, adding all cc deps at compile time was preventing some configurations from working: ghc's dynamic linker
apparently tries to load the .so files in the order in which they are specified so if a.so declares symbols referenced by b.so,
a.so needs to be given first. We don't seem to control the order in which these are passed from the rules, though. E.g., without `th_deps` defaulting to `[]` I'm unable to package llvm-hs to use a bazel-built llvm, since the former uses TH and FFI (but no FFI during TH) and the latter splits things in several several packages.

### DIFF
--- a/haskell/private/haskell_impl.bzl
+++ b/haskell/private/haskell_impl.bzl
@@ -173,6 +173,12 @@ def _haskell_binary_common_impl(ctx, is_test):
     dep_info = gather_dep_info(ctx.attr.name, ctx.attr.deps)
     all_deps_info = gather_dep_info(ctx.attr.name, deps)
 
+    th_deps = getattr(ctx.attr, "th_deps", [])
+    if not th_deps and ctx.attr.use_deps_on_empty_th_deps:
+        th_deps = deps
+
+    th_dep_info = gather_dep_info(ctx.attr.name, th_deps)
+
     modules = ctx.attr.modules
     if modules and ctx.files.srcs:
         fail("""Only one of "srcs" or "modules" attributes must be specified in {}""".format(ctx.label))
@@ -231,6 +237,7 @@ def _haskell_binary_common_impl(ctx, is_test):
         java,
         posix,
         dep_info,
+        th_dep_info,
         plugin_dep_info,
         srcs = srcs_files,
         module_map = module_map,
@@ -423,6 +430,12 @@ def haskell_library_impl(ctx):
     dep_info = gather_dep_info(ctx.attr.name, ctx.attr.deps + ctx.attr.exports)
     narrowed_deps_info = gather_dep_info(ctx.attr.name, ctx.attr.narrowed_deps)
     all_deps_info = gather_dep_info(ctx.attr.name, deps)
+
+    th_deps = getattr(ctx.attr, "th_deps", [])
+    if not th_deps and ctx.attr.use_deps_on_empty_th_deps:
+        th_deps = deps
+    th_dep_info = gather_dep_info(ctx.attr.name, th_deps)
+
     all_plugins = ctx.attr.plugins + ctx.attr.non_default_plugins
     plugin_dep_info = gather_dep_info(
         ctx.attr.name,
@@ -482,6 +495,7 @@ def haskell_library_impl(ctx):
         java,
         posix,
         dep_info,
+        th_dep_info,
         plugin_dep_info,
         srcs = srcs_files,
         module_map = module_map,
--- a/haskell/BUILD.bazel
+++ b/haskell/BUILD.bazel
@@ -176,6 +176,13 @@ config_setting(
     },
 )
 
+config_setting(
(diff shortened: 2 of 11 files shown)


ylecornec:
Thank you for the PR.

I will let others take a look but I have some comments:

- For backward compatibility, I think the old behavior should be kept for now if the `th_deps` field is not present.
    Eventually though we do not want to write `th_deps=[]` everywhere to activate this feature, so there could be a deprecation period with a warning in the documentation of the `deps` attribute.
- I also have one question: I do not find where the `th_package_ids` field of `HaskellContext` is used.

haskell/private/haskell_impl.bzl
```diff
@@ -419,7 +419,8 @@ def _create_empty_library(hs, cc, posix, my_pkg_id, with_shared, with_profiling,
 
 def haskell_library_impl(ctx):
     hs = haskell_context(ctx)
-    deps = ctx.attr.deps + ctx.attr.exports + ctx.attr.narrowed_deps
+    th_deps = getattr(ctx.attr, "th_deps", [])
```
aherrmann, line 422: This looks unused.

aherrmann (commented):
Thank you @jcpetruzza for working on this!

Thanks for the review @ylecornec, I have the same questions.

Another question, how does this interact with the [narrowing in haskell_module](https://github.com/tweag/rules_haskell/pull/1663)? CC @facundominguez

haskell/private/haskell_impl.bzl
facundominguez: Isn't it used in the following line to build `deps`?

facundominguez (commented):
Thanks @jcpetruzza.

I still have some questions about how `th_deps` should work. But here goes a first analysis of interactions with `haskell_module`.

If `th_deps` are only needed at compile time, then they are necessary to build individual modules with `haskell_module` only if the `enable_th` attribute is true.

Regarding narrowing, `th_deps` (same as `deps`), are necessary to find cross-library dependencies with `gazelle_haskell_modules`. But in the interface of `haskell_library/binary/test`, I'd expect libraries in `th_deps` to be moved by the user to `narrowed_deps` if their modules need to be cross-referenced, and nothing would need to change in the implementation of the rules.

jcpetruzza:
Hi all, thanks for the comments! Digging a bit more, I don't think this is doing what's advertised for Haskell `deps` just yet: running an `aquery` shows that the static/dynamic libraries of the Haskell `deps` are still inputs of the HaskellBuildXXX action.  Just to confirm, the expected behaviour here would be to include as inputs only the package.conf and hi files of each dep, correct? I'll try to fix this and add some tests.

Regarding the questions:

- I agree that for backwards compatibility it may be preferable not to default to [] at this point.
- `th_deps` shouldn't be included at link-time. I had added them since `//tests/template-haskell-with-cbits` was failing with a linker error otherwise, even though the FFI import was being seemingly used only for a TH quotation... but I failed to notice that the module had no export list. I'll correct that.

jcpetruzza:
I think I'm ultimately a bit lukewarm on this proposal:

  - It is not the right solution to C libs making a build fail due to TH (it mitigates it, but haskel_repl is still unusable; I think the correct solution is to use "topological" order for the libraries depset?),
  - Afaics, `th_deps` needs to be always a subset of `deps`: TH splices need to refer to Q-expressions defined in other modules, so Q-expressions need to be exported, so code is always generated for them in object files.
  - From the user's pov, then, the distinction between the two become unclear: adding `th_deps` could make the build faster in some cases, but it may not be easy to measure and things can easily drift.

Anyway, in case someone wants to look into it further, I've pushed a new version that handles:
  - Making default behaviour of th_deps configurable, for backwards compatibility.
  - Removing hs libraries from the inputs of the build step for non th_deps.
  - Ignores th_deps in the link step.

aherrmann:
@jcpetruzza Thank you for further digging into this!

> * It is not the right solution to C libs making a build fail due to TH (it mitigates it, but haskel_repl is still unusable; I think the correct solution is to use "topological" order for the libraries depset?),

Do you have a repro for that issue? 
I thought we do use topological order for C library dependencies. At least, I'd expect mostly-static linking of binaries to fail if that wasn't the case. (mostly-static linking is the default)

jcpetruzza:
@aherrmann Yeah, it's strange, it doesn't seem to be related to the order in which the libraries are given... I've filed #1696 with instructions to reproduce

PR #1691, opened 2022-02-10
Fix haskell_module when dependencies appear both as narrowed and not-narrowed

facundominguez:
Without this fix, when a module uses TH, ghc would find missing symbols when loading normal dependencies.
This would happen when this dependencies need in turn dependencies that are listed as narrowed. Narrowed dependencies point to empty shared libraries that don't provide the necessary symbols to regular deps. However they had precedence over packages pointing to the real libraries which also appear in the environment file.

This PR changes the order of packages so those with a real library take precedence over those with an empty library.

### DIFF
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -71,6 +71,20 @@ load("//haskell:providers.bzl", "HaskellInfo", "HaskellLibraryInfo")
 # profiling builds use the libraries of narrowed_deps instead of the
 # their object files.
 
+# Note [Deps as both narrowed and not narrowed]
+#
+# Package databases are searched in reversed order with respect to
+# how they appear in the command line of GHC or in environment files.
+#
+# If a package appears both in narrowed_deps and in normal deps, different
+# versions of the package will appear in different package databases.
+# One of the versions points to an empty shared library, and the other
+# version points to the real shared library.
+#
+# When both versions are needed, we want package database pointing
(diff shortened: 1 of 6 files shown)


haskell/experimental/private/module.bzl
```diff
@@ -200,8 +214,10 @@ def _build_haskell_module(
             package_ids = hs.package_ids,
             package_databases = depset(
                 transitive = [
-                    dep_info.package_databases,
+                    # Mind the order in which databases are specified here.
+                    # See Note [Deps as both narrowed and not narrowed].
                     narrowed_deps_info.empty_lib_package_databases,
+                    dep_info.package_databases,
                 ],
```
facundominguez, line 221: hm, in some tests this is still not producing the environment file with the flags in the order given here.
Can I set `order = "postorder"` here, regardless of the order of the transitive depsets?

facundominguez: I had to use preorder here and at another place. It seem to work now.

aherrmann: The [documentation](https://docs.bazel.build/versions/main/skylark/lib/depset.html) says the following about the `order` parameter:

> Two depsets may only be merged if either both depsets have the same order, or one of them has "default" order. In the latter case the resulting depset's order will be the same as the other's order.

aherrmann (approved):
Thank you!

PR #1692, opened 2022-02-10
Implement hs-lib@repl test as integration_test

k1nkreet:
Since #1645  is merged we can gradually implement integration tests from ```RunTests.hs``` as ```integration_test``` rules. Here we have hs-lib@repl test scenario reimplemented this way


aherrmann:
@k1nkreet It looks like CI is failing with a couple instances of 
```
       ERROR: The Build Event Protocol upload failed: All retry attempts failed. DEADLINE_EXCEEDED: DEADLINE_EXCEEDED: deadline exceeded after 14.999828601s. [remote_addr=cloud.buildbuddy.io/34.82.173.239:443] DEADLINE_EXCEEDED: DEADLINE_EXCEEDED: deadline exceeded after 14.999828601s. [remote_addr=cloud.buildbuddy.io/34.82.173.239:443]`
```
Maybe worth asking on BuildBuddy Slack for some help.

k1nkreet:
There was a discussion on the BuildBuddy slack about month ago about exactly the same issue occurred when "running multiple bazel invocations one after the other in github actions". Here is the fixes proposed in the thread:

```
If so running either bazel shutdown before the command - or setting the flag --keep_backend_build_event_connections_alive=false  would likely help

Another thing that might help (that I've seen fix other github actions related networking issues) is running this before the invocation in your github action:
sudo sysctl -w net.ipv4.tcp_keepalive_time=60

Either that or setting the --keep_backend_build_event_connections_alive=false Bazel flag are two options I'd try out.
```

aherrmann:
@k1nkreet Thank you for digging into this. Yes, either option is worth a shot.

k1nkreet:
https://github.com/tweag/rules_haskell/blob/master/.bazelrc#L67

It looks like we've already tried to fix exactly same problem with the second solution :)

.github/workflows/workflow.yaml
```diff
@@ -50,6 +50,7 @@ jobs:
           nix-shell --arg docTools false --pure --run '
             set -euo pipefail
             ./tests/run-start-script.sh --use-nix
+            bazel shutdown
             bazel build //tests:run-tests
             ./bazel-ci-bin/tests/run-tests
```
aherrmann, line 55: One issue might be that `tests/run-tests` runs a couple Bazel commands as well.

k1nkreet:
I've found out this test wasn't supposed to be passing on macos-bindist because ```hs-lib``` depend on c2hs target and tagged with ```dont_test_on_darwin_with_bindist```. So I've used this tag for ```hs_lib_repl_test``` integration test as well.

aherrmann (approved):
Good work, thank you!
A couple minor comments. Feel free to merge once addressed.

Issue #936, opened 2019-06-11
Haskell_doctest does not includes -package-db


guibou:
**Describe the bug**

This is #367 reopened.

I'm using `haskell_doctest`. However it fails for libraries with dependencies

```
doctest: <command line>: cannot satisfy -package-id QuickCheck-2.12.6.1-2neFyIv0PQcKVFUR4jOgvi
    (use -v for more information)
```

(This is not specific to `QuickCheck`, it depens on the order of the dependencies in target `deps` attribute).

**To Reproduce**

I cannot reproduce it in `rules_haskell`. I'm using a `nixpkgs` registered toolchain which contains all packages in a global database.

I can workaround the issue by just specifying the `-package-db` in my `haskell_doctest` rule call:

```
        haskell_doctest(
            name = "foo"
            deps = [target_label],
            tags = ["doctest"],

	    doctest_flags = [
                # TODO(guillaume.bouchard)
                # For some reasons, rules_haskell does not see the package DB for doctests
                #
                # Luckyly, we have a global package DB in
                # @io_tweag_rules_haskell_ghc_nixpkgs//:lib/ghc-8.6.3/package.conf.d/package.cache
                #               
                # A clean temporary solution involves too much hack,
                # so I'm hacking this uglyly here until a more robust
                # solution is implemented in rules_haskell
                #
                # Unluckly, bazel does not do expansion for label in
                # list of flags, neither it accept a label to a
                # directory, neither it accepts a label to a file
                # which may be inside the directory if the file is not
                # declared with the right visibility. So the 
                #
                #
                # This path must be upgraded each time your ghc from
                # nixpkgs changes.  have a look inside
                # /bazel-heavisoft/external/io_tweag_rules_haskell_ghc_nixpkgs/
                # until you find the /nix/store path which works
	    	"-package-db",
	    	"/nix/store/q5xl9hv2zy6cwqzjsg56bb4yy99r3dr2-ghc-8.6.3-with-packages/lib/ghc-8.6.3/package.conf.d",
	    ]
        )
```

**Environment**
 - OS name + version: nixos
 - Bazel version: 0.26
 - Version of the rules: 98e3bbd6fa5f77273be5cc5b003e8344bb06d069

**Additional context**

I'm able to fix this issue, but I need some discussion first on the design.

[...]

PR #1697, opened 2022-02-16
Add a regression test for haskell_doctest

aherrmann:
Regression test for https://github.com/tweag/rules_haskell/issues/936.
That issue seems to have been resolved in the meantime. However, that case wasn't covered by the test-suite, yet.


Xophmeister (approved):
Looks good to me :+1:

PR #1700, opened 2022-02-18
Mark fetch_stack as configure and dependent on PATH

aherrmann:
To ensure that Bazel re-runs the rule if PATH changes or a reconfiguration is requested.

Closes https://github.com/tweag/rules_haskell/issues/1079


ylecornec (approved):
Looks good to me :+1:

Issue #1173, opened 2019-12-17
rules_haskell requires python2 on Ubuntu when not using Nix


aherrmann:
**Describe the bug**
rules_haskell in GHC bindist mode (not Nix shell) requires the `python` package on Ubuntu which provides Python 2.

**To Reproduce**
Try to run rules_haskell test suite on Ubuntu image without Python 2 installed. See https://buildkite.com/tweag-1/rules-haskell/builds/667#a3798e34-7be3-433f-bfa3-0aca1b0a303f/36-2757.
```
<no location info>: error:
--
  | Warning: Couldn't figure out C compiler information!
  | Make sure you're using GNU gcc, or clang
  | /usr/bin/env: ‘python’: No such file or directory
  | `cc_wrapper-python' failed in phase `Assembler'. (Exit code: 127)
```

**Expected behavior**
Python 3 alone (`python3` on Ubuntu) should be sufficient.

**Environment**
 - OS name + version: Ubuntu 19.04
 - Bazel version: 1.1
 - Version of the rules: 36593735

**Additional context**
The `cc_wrapper` script that raises the error is a `py_binary` target. The source file specifies `#!/usr/bin/env python3`, however, it seems that Bazel generates a Python script requiring `/usr/bin/env python` anyway.

cc @flokli

smelc:
I have what looks like a similar issue using Ubuntu18.04, **with** python2 installed:

```
> python --version
Python 2.7.17
> python3 --version
Python 3.6.9
> bazel test //...
...
ERROR: /home/churlin/.cache/bazel/_bazel_churlin/a2ae29489331b4803c5f3e551b1f4565/external/stackage-zlib/BUILD.bazel:14:1: HaskellCabalLibrary @stackage-zlib//:zlib failed (Exit 127) cabal_wrapper failed: error executing command bazel-out/host/bin/haskell/cabal_wrapper lib:zlib zlib-0.6.2 true external/stackage-zlib/zlib-0.6.2/Setup.hs external/stackage-zlib/zlib-0.6.2 ... (remaining 10 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
/usr/bin/env: ‘python’: No such file or directory
```

Version of the rules: 8f7d75961a1f459b4edb5faeb6e75dd251a2147f, bazel version: 2.0.0

smelc:
Here are more details (version of the rules: 9a52a45410e0aa0e7005a572b40bf1320fd65220).

[...]

PR #1702, opened 2022-02-21
Set stub_shebang in bindist Python toolchain

aherrmann:
Closes https://github.com/tweag/rules_haskell/issues/1173

This is the bindist analog of the rules_nixpkgs solution here: https://github.com/tweag/rules_nixpkgs/pull/174

### DIFF
--- a/haskell/ghc_bindist.bzl
+++ b/haskell/ghc_bindist.bzl
@@ -1,6 +1,7 @@
 """Workspace rules (GHC binary distributions)"""
 
 load("@bazel_skylib//lib:paths.bzl", "paths")
+load("@bazel_skylib//lib:versions.bzl", "versions")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "patch")
 load("@bazel_tools//tools/cpp:lib_cc_configure.bzl", "get_cpu_value")
 load("@rules_sh//sh:posix.bzl", "sh_posix_configure")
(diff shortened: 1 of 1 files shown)


Xophmeister (approved):

meteorcloudy:
This change is causing rules_haskell to break with Bazel@HEAD, we'll lose test coverage for you :(
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2375#73315dac-a9aa-49b4-bee1-38507d2f86ab
```
(04:04:06) ERROR: An error occurred during the fetch of repository 'rules_haskell_python_local':
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/afd0e027d4dc894108ae8f85e604d52d/external/rules_haskell/haskell/ghc_bindist.bzl", line 774, column 28, in _configure_python3_toolchain_impl
		if versions.is_at_least("4.2.0", versions.get()):
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/afd0e027d4dc894108ae8f85e604d52d/external/bazel_skylib/lib/versions.bzl", line 80, column 32, in _is_at_least
		return _parse_bazel_version(version) >= _parse_bazel_version(threshold)
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/afd0e027d4dc894108ae8f85e604d52d/external/bazel_skylib/lib/versions.bzl", line 55, column 22, in _parse_bazel_version
		return tuple([int(n) for n in version.split(".")])
Error in int: empty string
```

meteorcloudy:
Looks like [_is_at_least](https://github.com/bazelbuild/bazel-skylib/blob/main/lib/versions.bzl#L69) in bazel-skylib isn't compatible with unreleased Bazel version, so you can probably check `versions.get()` before passing to it.

aherrmann:
@meteorcloudy Thank you for reporting and looking into the issue. You're spot-on, I was able to reproduce the failure locally with `USE_BAZEL_VERSION=last_green` and fixing that line removes that issue.
It turns out that Skylib has [`versions.check`](https://github.com/bazelbuild/bazel-skylib/blob/e30197f3799eb038fbed424e365573f493d52fa5/lib/versions.bzl#L82) which handles this case already.l

meteorcloudy:
But `_check_bazel_version` will fail if the version isn't in range, but I guess you already noticed..

aherrmann:
> But `_check_bazel_version` will fail if the version isn't in range, but I guess you already noticed..

Indeed, I shot too quickly there. :man_facepalming: 

https://github.com/tweag/rules_haskell/pull/1707 should hopefully fix it.

PR #1703, opened 2022-02-22
add documentation for the generate_paths_module option of cabal rules.

ylecornec:
As noted in https://github.com/tweag/rules_haskell/pull/1575#issuecomment-1046778016, the `runfiles` dependency needs to be added to the `.cabal` file when the `generate_paths_module` option is used. 
This PR adds this fact to the documentation.


aherrmann (approved):
Thank you!

PR #1705, opened 2022-02-23
add `use_stack()` to override `stack` from anywhere

### LINKED ISSUE #1080: Globally set the `stack` executable for `stack_snapshot`

`stack_snapshot` have a `stack` attribute to select the `stack` binary we want to use. However if the `stack_snapshot` call is hidden in another repository, we cannot override it.

In #1078 and #1079 I observed that a workspace using rules_haskell may implicitly depends on a `stack_snapshot` call, without any control on the `stack` argument, leading to the following cases:

- if `stack_snapshot` provides a `stack` argument, no way to change it
- if `stack_snapshot` does not provide a `stack` argument, `stack_snapshot` will try the following:
   - downloading a `stack` prebuilt binaries. It may not be a working solution (see #1079).
   - use the environment (i.e. `PATH`) `stack`. That's not a satisfying solution for hermetic build if we want to provide our `stack` binary as a bazel label.

A way to register a `stack` binary as a bazel toolchain.

fricklerhandwerk:
closes #1080


haskell/cabal.bzl
```diff
@@ -2561,6 +2563,12 @@ def stack_snapshot(
 
     """
     typecheck_stackage_extradeps(extra_deps)
+
+    # Allow overriding stack binary by external repositories.
+    # Otherwise this is a no-op.
+    if existing_rule("rules_haskell_stack"):
```
aherrmann, line 2569: IIUC this would also override a user provided `stack` attribute. Checking `if not stack and existing_rule(...)` should fix that.

haskell/cabal.bzl
```diff
@@ -2626,3 +2634,54 @@ def _expand_make_variables(name, ctx, strings):
...
+    local_repository(
+        name = "my_stack",
+        path = "../some_custom_stack",
+    )
+    use_stack("@my_stack//:stack")
+    local_repository(
+        name = "A",
+        path = "../A",
+    )
+    # using `@A//:a` as dependency will build on `my_stack`
```
aherrmann, line 2680: That's not how Bazel workspaces work. Bazel will only consider *one* `WORKSPACE` file, the one in the main repository, all other `WORKSPACE` files are ignored.

So, the correct usage is to place `use_stack` into the `WORKSPACE` before any `stack_snapshot`.

haskell/cabal.bzl
```diff
@@ -2626,3 +2634,54 @@ def _expand_make_variables(name, ctx, strings):
...
+        "stack": attr.label(),
+    },
+)
+
+def use_stack(stack):
+    """
+    use given `stack` binary in `stack_snapshot`.
+    allows overriding `stack` installation independent of call site,
+    specifically from repositories outside of where a `stack_snapshot`
+    has been declared.
```
aherrmann, line 2658:
```suggestion
```
See comment about `WORKSPACE` files below.

aherrmann (commented):
Thank you! That's already looking good.
A few comments.

Re testing. Testing repository rules is a bit tricky. One way is to factor out the business logic into pure code and test it in a [unittest](https://bazel.build/rules/testing#testing-starlark-utilities). An example exists in rules_nixpkgs [here](https://github.com/tweag/rules_nixpkgs/blob/bf6da6adf97c2dae6b477a0f8266c56025fc8cec/tests/location_expansion_unit_test.bzl) and [here](https://github.com/tweag/rules_nixpkgs/blob/bf6da6adf97c2dae6b477a0f8266c56025fc8cec/nixpkgs/private/location_expansion.bzl#L16-L17). However, in this case it seems like an integration test would be more appropriate. To that end we have [`integration_test`](https://github.com/tweag/rules_haskell/blob/e8550a3aa2195813bdb5e8ce53b49c0f0bfad860/tests/integration_tests.bzl), see [here](https://github.com/tweag/rules_haskell/blob/f53c6f0f44a4bcc968bc51b349484f54fef749b6/tests/repl-targets/BUILD.bazel#L117) and [here](https://github.com/tweag/rules_haskell/blob/f53c6f0f44a4bcc968bc51b349484f54fef749b6/tests/repl-targets/hs_bin_repl_test.go).

haskell/cabal.bzl
fricklerhandwerk: I thought this was the point of #1080 - allowing an override?

aherrmann: Good catch, I overlooked
> if stack_snapshot provides a stack argument, no way to change it

Let's keep it this way then.

haskell/cabal.bzl
```diff
@@ -2653,34 +2653,35 @@ _use_stack = repository_rule(
...
-    allows global override of `stack` executable, independent of call site,
+    Allows global override of `stack` executable, independent of call site,
     specifically from dependants of a `stack_snapshot`.
+    Call this in `WORKSPACE` before any use of `stack_snapshot`.
 
-    call this in `WORKSPACE` before any use of `stack_snapshot`.
+    WARNING: Use it only as a stop-gap measure if tracing discrepancies in
+    `stack` versions becomes overwhelming. Try to avoid its spooky action at
+    distance, and instead explicitly pass `stack` to macros which end up
+    calling `stack_snapshot`. It should always be possible and is a bug otherwise.
```
aherrmann, line 2665: I understand the sentiment, but this seems a bit out of place. `WORKSPACE` files are not declarative, instead they are sequential and imperative. This kind of behavior is very common in `WORKSPACE` files, in fact the common approach for declaring external dependencies is to use [`maybe`](https://bazel.build/rules/lib/repo/utils#maybe), which has a similar kind of behavior: The first declaration of a dependency is effectively an override of all later ones. So, this approach fits naturally into that model.
(For a more declarative model we'll have to wait for `bzlmod`.)

I'm not aware of a need to use different stack versions at different `stack_snapshot` instances. So, I'm not sure when using the `stack` attribute would be preferable.

That said, what does seem appropriate as a warning is that this should only be used in the top-level repository. Dependencies should not implicitly impose global overrides as they could conflict with other dependencies or the main project.
Perhaps it could be phrased like this: "WARNING: This will globally override the stack version used by all instances of stack_snapshot. This should only be used in the top-level repository."

tests/stack-snapshot-deps/hs_override_stack_test.go
```diff
@@ -0,0 +1,107 @@
...
+main :: IO ()
+main = putStrLn "Hello GHCi!"
+`)
+}
+
+func TestHsBinRepl(t *testing.T) {
+    // FIXME: This test is only partial due to limitations of `bazel_testing`.
+    // It is not possible to define a file as executable in `txtar`, and no
+    // obvious way to inject a pre-made executable into the build.
+    // To do this properly, the whole testing method needs to be reworked.
```
aherrmann, line 96: I'm happy to leave this out of this PR and address in a follow-up to reduce scope, but, a good way to achieve this may be to define the binary in the top-level repository, i.e. somewhere under `@rules_haskell//tests/...`, and then make it a data dependency of the integration test. To that end you'll need to expose the [data attribute](https://github.com/tweag/rules_haskell/blob/102cf5ff1e2e0982c34c888cfd0555b2d9dd252b/tests/integration_tests.bzl#L21). Then it would be a pre-built file from the perspective of the integration test and could be used in `use_stack`.

haskell/cabal.bzl
```diff
@@ -1017,11 +1020,11 @@ _STACK_BINDISTS = \
...
-    exec_result = _execute_or_fail_loudly(repository_ctx, [stack_cmd, "--numeric-version"])
-
-    stack_major_version = int(exec_result.stdout.split(".")[0])
-    stack_minor_version = int(exec_result.stdout.split(".")[1])
-    return stack_major_version >= 2 and stack_minor_version >= 3
+    exec_result = repository_ctx.execute([stack_cmd, "--numeric-version"])
+    if exec_result.return_code != 0:
+        return False, "not"  # HACK: use as "Stack {} found".format(version)
+    version = tuple([int(x) for x in exec_result.stdout.split(".")[:2]])
+    return version >= _STACK_MIN_VERSION, exec_result.stdout.strip()
```
aherrmann, line 1026: IIUC all use-sites construct the same message based on this output. So, maybe an easy way to avoid the hack and also avoid some duplication is to construct the full message here and return it.

aherrmann (commented):
Thank you! I took another look at it and left some comments.
I'm happy to leave the fix and extension of the integration testing for a follow-up to reduce the scope of this PR.

aherrmann (approved):
That looks good to me! Thank you, good work!

@k1nkreet could you take a look at the integration-testing changes?

haskell/cabal.bzl
```diff
@@ -1018,12 +1018,16 @@ _STACK_BINDISTS = \
...
 def _stack_version_check(repository_ctx, stack_cmd):
-    """Returns False if version not recent enough."""
+    """Returns error string if version is not recent enough"""
     exec_result = repository_ctx.execute([stack_cmd, "--numeric-version"])
     if exec_result.return_code != 0:
-        return False, "not"  # HACK: use as "Stack {} found".format(version)
-    version = tuple([int(x) for x in exec_result.stdout.split(".")[:2]])
-    return version >= _STACK_MIN_VERSION, exec_result.stdout.strip()
+        print(exec_result.stderr)
+        print(exec_result.stdout)
```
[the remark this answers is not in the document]
fricklerhandwerk: No, deliberately placed for debugging purposes such that it is immediately clear from output what happened.

k1nkreet (commented):
Looks good to me also!

tests/stack-snapshot-deps/hs_override_stack_test.go
```diff
@@ -0,0 +1,122 @@
...
+				return
+			}
+			return os.Chmod(stack, 0700|info.Mode())
+		},
+	})
+}
+
+func TestHsBinRepl(t *testing.T) {
+	_, err := it.BazelOutput(it.Context.BazelBinary, "run", "//:hs-bin@repl", "--", "-ignore-dot-ghci", "-e", ":main")
+	out := string(err.(*bt.StderrExitError).Err.Stderr)
```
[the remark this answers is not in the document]
fricklerhandwerk: Just have to use Go's `error` interface correctly. Thanks for the reminder.

PR #1706, opened 2022-02-23
Use Skylib versions.check

aherrmann:
`versions.get` can be empty on pre-release versions which breaks `versions.is_at_least`, the Skylib function [`versions.check`](https://github.com/bazelbuild/bazel-skylib/blob/main/lib/versions.bzl#L82) already handles this case.

Fixes the issue reported in https://github.com/tweag/rules_haskell/pull/1702#issuecomment-1048888698


aherrmann:
`versions.check` is not appropriate after all, it `fail`s if the version doesn't match.

PR #1707, opened 2022-02-23
Handle Bazel pre-release in version check

aherrmann:
Fixes the issue reported in https://github.com/tweag/rules_haskell/pull/1702#issuecomment-1048888698
Second attempt after https://github.com/tweag/rules_haskell/pull/1706

`versions.get` can be empty on pre-release versions which breaks [`versions.is_at_least`](https://github.com/bazelbuild/bazel-skylib/blob/e30197f3799eb038fbed424e365573f493d52fa5/lib/versions.bzl#L69).
This PR adds a variant of [`versions.check_version`](https://github.com/bazelbuild/bazel-skylib/blob/e30197f3799eb038fbed424e365573f493d52fa5/lib/versions.bzl#L82) that doesn't fail on version mismatch, but instead returns a bool indicating whether the version matches or not.

`rules_haskell` already had a very similar function `check_version` to generate a warning if the Bazel version is outside of a known compatibility range. This PR changes that function to re-use the newly introduced `check_bazel_version` based on Skylib to avoid duplication.

### DIFF
--- a/haskell/private/versions.bzl
+++ b/haskell/private/versions.bzl
@@ -1,4 +1,4 @@
-# check_version below cannot be called from anywhere
+# check_bazel_version_compatible below cannot be called from anywhere
 # because native.bazel_version's availability is restricted:
 #   https://github.com/bazelbuild/bazel/issues/8305
 # An alternative hacky way is as follows:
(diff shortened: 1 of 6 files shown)


haskell/private/versions.bzl
```diff
@@ -38,23 +38,81 @@ def _parse_bazel_version(bazel_version):
...
-        return  # Unexpected format
+def _is_at_least(threshold, version):
+    """Check that a version is higher or equals to a threshold.
+    Args:
+      threshold: the minimum version string
+      version: the version string to be compared to the threshold
+    Returns:
+      True if version >= threshold.
+    """
+    # Vendored from https://github.com/bazelbuild/bazel-skylib/blob/e30197f3799eb038fbed424e365573f493d52fa5/lib/versions.bzl
```
aherrmann, line 49: These are vendored because Bazel warning check occurs in `haskell/repositories.bzl` which is the place that imports Skylib in the first place. So, we can't depend on it here.

Xophmeister:
...oh, but the CI doesn't like it. Let me have a quick look

Xophmeister (approved):

PR #1718, opened 2022-03-16
Don't discard dead_strip_dylibs in macos

### LINKED ISSUE #1714: Shared libraries have too many dependencies on Mac
One difference between `stack` and `rules_haskell` in Mac is that the former uses the linker flag `-dead_strip_dylibs` to keep the amount of load commands low. To fix #170, `rules_haskell` started discarding `-dead_strip_dylibs` in #627.

This however is causing some libraries to exceed a limit of 254 load commands, which is imposed by how MachO file format deals with two-level namespaces and library ordinals. The description of the `n_desc` field in [this source](https://github.com/aidansteele/osx-abi-macho-file-format-reference/blob/master/README.md#nlist_64) indicates the limit.

Looks like we would need either to recover our ability to use `-dead_strip_dylibs` or to start splitting libraries in an attempt to keep the count of load commands low. 

**Environment**
 - OS name + version: Mac
 - Bazel version: 5
 - Version of the rules: 0.14

## Why do we need to discard `-dead_strip_libraries`?

cc libraries underlink in `bazel`, which means that they don't report all of their dynamic dependencies, and `bazel` keeps track of the transitive dependencies to use them when linking a cc binary.

When cc libraries are used in a haskell library, `rules_haskell` links all of the transitive dependencies of the cc libraries, so they don't need to be tracked separately. However, when `-dead_strip_dylibs` is enabled, it has the effect of discarding all of the indirect dependencies, including indirect cc dependencies which aren't otherwise declared to the dynamic loader. This would cause missing undefined symbols when trying to run binaries which use the haskell libraries, which use cc libraries, which depend on other cc libraries that have been stripped.

A possible fix would be to stop linking indirect cc dependencies in Haskell libraries, and collect them to link them in haskell binaries and tests instead. This is implementing in `rules_haskell` the same linking strategy for cc dependencies as cc rules use.

facundominguez:
Fixes #1714

### DIFF
--- a/haskell/private/actions/compile.bzl
+++ b/haskell/private/actions/compile.bzl
@@ -104,7 +104,28 @@ def _process_hsc_file(hs, cc, hsc_flags, hsc_inputs, hsc_file):
 
     return hs_out, idir
 
-def _compilation_defaults(hs, cc, java, posix, dep_info, plugin_dep_info, srcs, module_map, import_dir_map, extra_srcs, user_compile_flags, output_mode, with_profiling, interfaces_dir, objects_dir, my_pkg_id, version, extra_ldflags_file, plugins, non_default_plugins, preprocessors):
+def _compilation_defaults(
+        hs,
+        cc,
+        java,
+        posix,
+        dep_info,
+        plugin_dep_info,
+        srcs,
+        module_map,
+        import_dir_map,
+        extra_srcs,
+        user_compile_flags,
+        output_mode,
+        with_profiling,
+        interfaces_dir,
+        objects_dir,
+        my_pkg_id,
+        version,
+        extra_ldflags_file,
+        plugins,
+        non_default_plugins,
+        preprocessors):
     """Compute variables common to all compilation targets (binary and library).
 
     Returns:
@@ -437,7 +458,29 @@ def compile_binary(
         source_files: set of Haskell source files
         boot_files: set of Haskell boot files
     """
-    c = _compilation_defaults(hs, cc, java, posix, dep_info, plugin_dep_info, srcs, module_map, import_dir_map, extra_srcs, user_compile_flags, "dynamic" if dynamic else "static", with_profiling, interfaces_dir, objects_dir, my_pkg_id = None, version = version, extra_ldflags_file = extra_ldflags_file, plugins = plugins, non_default_plugins = non_default_plugins, preprocessors = preprocessors)
+    c = _compilation_defaults(
+        hs,
+        cc,
+        java,
+        posix,
+        dep_info,
+        plugin_dep_info,
+        srcs,
+        module_map,
+        import_dir_map,
+        extra_srcs,
+        user_compile_flags,
+        "dynamic" if dynamic else "static",
+        with_profiling,
+        interfaces_dir,
+        objects_dir,
+        my_pkg_id = None,
+        version = version,
+        extra_ldflags_file = extra_ldflags_file,
+        plugins = plugins,
+        non_default_plugins = non_default_plugins,
+        preprocessors = preprocessors,
+    )
     c.args.add_all(["-main-is", main_function])
     if dynamic:
         # For binaries, GHC creates .o files even for code to be
@@ -512,7 +555,29 @@ def compile_library(
         boot_files: set of Haskell boot files
         import_dirs: import directories that should make all modules visible (for GHCi)
     """
-    c = _compilation_defaults(hs, cc, java, posix, dep_info, plugin_dep_info, srcs, module_map, import_dir_map, extra_srcs, user_compile_flags, "dynamic-too" if with_shared else "static", with_profiling, interfaces_dir, objects_dir, my_pkg_id = my_pkg_id, version = my_pkg_id.version, extra_ldflags_file = extra_ldflags_file, plugins = plugins, non_default_plugins = non_default_plugins, preprocessors = preprocessors)
+    c = _compilation_defaults(
+        hs,
+        cc,
+        java,
+        posix,
+        dep_info,
+        plugin_dep_info,
+        srcs,
+        module_map,
+        import_dir_map,
+        extra_srcs,
+        user_compile_flags,
+        "dynamic-too" if with_shared else "static",
+        with_profiling,
+        interfaces_dir,
+        objects_dir,
+        my_pkg_id = my_pkg_id,
+        version = my_pkg_id.version,
+        extra_ldflags_file = extra_ldflags_file,
+        plugins = plugins,
+        non_default_plugins = non_default_plugins,
+        preprocessors = preprocessors,
+    )
     if with_shared:
         c.args.add("-dynamic-too")
 
--- a/haskell/private/actions/compile.bzl
+++ b/haskell/private/actions/compile.bzl
@@ -376,7 +376,7 @@ def _compilation_defaults(
 
     input_files = []
     if extra_ldflags_file:
-        args.add_all(["-optl", "@{}".format(extra_ldflags_file.path)])
+        args.add("-optl@{}".format(extra_ldflags_file.path))
         input_files.append(extra_ldflags_file)
 
     return struct(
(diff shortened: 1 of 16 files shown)


facundominguez:
I need some advice to complete this. Probably this PR needs to add some provisions to have `-u` flags produced when building with cabal rules.

haskell/private/actions/link.bzl
```diff
@@ -36,6 +36,63 @@ def merge_parameter_files(hs, file1, file2):
...
+        dynamic,
+    )
+
+    hs.actions.run_shell(
+        inputs = cc_dynamic_libs,
+        outputs = [linker_flags_file],
+        command = """
+        touch {out}
+        for lib in {solibs}; do
+            {nm} -Ujp "$lib" | head -1 | sed 's/^/-u /' >> {out}
```
aherrmann, line 86: What do these `nm` flags mean? The man page on my system only knows `-u` (lower-case) and `-p`, but no `-U` or `-j`. Are those flags portable, or only available on some specific system?

aherrmann (commented):
Thank you!

> Probably this PR needs to add some provisions to have -u flags produced when building with cabal rules. I'm rather unfamiliar with the cabal rules, so I'd be helpful for some pointers on what to change.

Yes, I think so, too. Also a test case that covers the Cabal case would be good.

I think this can be achieved by adding the same linker response file to `--ghc-option=-optl` to [`Setup.hs configure`](https://github.com/tweag/rules_haskell/blob/4c8e6d3b83122ee43949994f4343697d98dbe281/haskell/private/cabal_wrapper.py#L255).

Note that the Cabal rules change directory during the build, so you'll have to prefix the path to the response file with `cfg_execroot`.

haskell/private/actions/link.bzl
facundominguez: They are macos specific: http://www.manpagez.com/man/1/nm/osx-10.12.6.php

```
-p     Don't sort; display in symbol-table order.
-j     Just display the symbol names (no value or type).
-U     Don't display undefined symbols.
```

aherrmann: Thanks for clarifying! Makes sense. :)

haskell/private/actions/compile.bzl
```diff
@@ -104,7 +104,7 @@ def _process_hsc_file(hs, cc, hsc_flags, hsc_inputs, hsc_file):
 
     return hs_out, idir
 
-def _compilation_defaults(hs, cc, java, posix, dep_info, plugin_dep_info, srcs, module_map, import_dir_map, extra_srcs, user_compile_flags, output_mode, with_profiling, interfaces_dir, objects_dir, my_pkg_id, version, plugins, non_default_plugins, preprocessors):
+def _compilation_defaults(hs, cc, java, posix, dep_info, plugin_dep_info, srcs, module_map, import_dir_map, extra_srcs, user_compile_flags, output_mode, with_profiling, interfaces_dir, objects_dir, my_pkg_id, version, extra_ldflags_file, plugins, non_default_plugins, preprocessors):
```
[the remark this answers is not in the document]
facundominguez: `compile_binary` has 22 parameters. `compile_library` has 20 parameters. `_compilation_defaults` has 21.

I propose to list the parameters one per line in `_compilation_defaults` :)

Probably we can make the function more readable not by reducing the parameters, but by refactoring it so it calls smaller functions instead of having a body of 280 lines as it is today.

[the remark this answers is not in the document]
aherrmann: > I propose to list the parameters one per line in _compilation_defaults :)

It's a start :)

> Probably we can make the function more readable not by reducing the parameters, but by refactoring it so it calls smaller functions instead of having a body of 280 lines as it is today.

Yes, that sounds like a good approach. Though better kept in a separate PR.

haskell/private/actions/link.bzl
facundominguez: I added those bits of documentation to a comment in the code.

haskell/private/actions/compile.bzl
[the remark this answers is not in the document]
facundominguez: I aligned the parameters vertically.

> Yes, that sounds like a good approach. Though better kept in a separate PR.

I created [this issue](https://github.com/tweag/rules_haskell/issues/1719) to track it.

facundominguez:
All green!

> Also a test case that covers the Cabal case would be good.

This is giving me some trouble. cabal rules don't link C dependencies by default, and my attempts to use `-dynamic` are giving me errors about missing shared libraries. I will try it some more.

haskell/private/actions/link.bzl
```diff
@@ -36,6 +36,74 @@ def merge_parameter_files(hs, file1, file2):
...
+        touch {out}
+        for lib in {solibs}; do
+            {nm} -Ujpg "$lib" | head -1 | sed 's/^/-u /' >> {out}
+        done
+        """.format(
+            nm = cc.tools.nm,
+            solibs = " ".join(["\"" + l.path + "\"" for l in cc_dynamic_libs]),
+            out = linker_flags_file.path,
+        ),
+    )
```
facundominguez, line 104: This is going to rerun whenever the PATH variable changes. Right?
And I guess the user needs to have `nm`, `head` and `sed` in the PATH.

Should we do anything about it?

aherrmann: `nm` is provided by the toolchain.
You're correct on `head` and `sed`. You can pull those tools in from rules_sh's POSIX toolchain to avoid that. E.g. 
```head = posix.commands["head"]```

Comment (aherrmann) [reply haskell/private/actions/link.bzl:104]: > This is going to rerun whenever the PATH variable changes. Right?

No, Bazel controls `PATH` for this action.

File: tests/haskell_cabal_indirect_link/BUILD.bazel [hunk 4]
```diff
@@ -0,0 +1,71 @@
...
+# haskell_cabal_binary handle indirect C dependencies
+# correctly
+
+cc_library(
+    name = "c-lib2",
+    srcs = ["c-lib2.c"],
+)
+
+# Embedding libraries like this causes libc-lib2.so
+# to be created for the extra-libraries field of cabal
```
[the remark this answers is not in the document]
aherrmann: Looks good to me. For context, the shared libraries generated by `cc_library` have mangled names by default. However, Cabal looks for libraries with unmangled names. Wrapping the `cc_library` in another `cc_library` like here prevents this name mangling.

tests/haskell_cabal_indirect_link/lib.cabal
```diff
@@ -0,0 +1,18 @@
+cabal-version: >=1.10
+name: lib
+version: 0.1.0.0
+build-type: Simple
+
+library
+  exposed-modules: Lib
+  build-depends: base
+  extra-libraries: c-lib c-lib2
```
aherrmann, line 9: Shouldn't this only list `c-lib`? Otherwise, `c-lib2` is also a direct dependency.
(Same below).

aherrmann (commented):
Thank you! Looks good!
A few small comments.

tests/haskell_cabal_indirect_link/lib.cabal
facundominguez: Let's see: `c-lib` is underlinked. If we only specify `c-lib`, Cabal is going to specify `-lc-lib` when linking. Therefore, `c-lib2` would never be considered to be linked. By specifying both `c-lib` and `c-lib2`, the linker is made aware of both libraries.

I actually ran this test without the commit that introduces `-u` and verified that it fails with errors about missing symbols as we expect when using `-dead_strip_dyld`.

aherrmann: Ah, yes, you're right. Forgot that we wanted to list the transitive C libs directly because of underlinking. Sorry for the noise.

haskell/private/actions/link.bzl
facundominguez: I've changed this to use the posix toolchain.

aherrmann (approved):
Thanks!

PR #1725, opened 2022-03-30
Add missing input_manifests to hsc2hs action

jcpetruzza:
Without this, running hs2chs actions via rbe fails with cc_wrapper-python complaining that it can't find the .runfiles directory


ylecornec (approved):
Thank you, it looks good to me.

PR #1729, opened 2022-04-18
Support some includes in hsc files

### LINKED ISSUE #1728: Support includes in hsc files
Very often when people have `.hsc` files they will also have some headers which they're including. We currently don't seem to have support for `#include`, as we don't do any work in the hsc invocations to make sure files that are being included are actually there during a build.

Some way to have headers available for hsc to process. My current idea (which I've implemented) is to allow users to specify headers in `extra_srcs`, and then try to mirror the directory structure that's in the project in our build environment.

I'm not sure if reusing `extra_srcs` is a good idea - it does the job, but if someone is using `extra_srcs` already.

I'm also not sure about the "mirror the directory structure", as it seems a bit fragile (e.g. it doesn't work if someone includes a file from outside the project structure). We could use `hsc2hs`'s ability to include files via the CLI, however that would require either
* we tell the user to adapt their project and manually remove `#include`s
* we manipulate source files to remove the `#include`s from there
both of which are rather unsatisfying. (and way more involved than the current solution that I'm proposing).

googleson78:
Somewhat improves #1728

### DIFF
--- a/haskell/private/actions/compile.bzl
+++ b/haskell/private/actions/compile.bzl
@@ -10,6 +10,7 @@ load("@bazel_skylib//lib:paths.bzl", "paths")
 load(
     ":private/path_utils.bzl",
     "declare_compiled",
+    "rel_path_to_module",
     "target_unique_name",
 )
 load(":private/pkg_id.bzl", "pkg_id")
@@ -27,7 +28,7 @@ load(
 load(":private/set.bzl", "set")
 load("//haskell/experimental:providers.bzl", "HaskellModuleInfo")
 
-def _process_hsc_file(hs, cc, hsc_flags, hsc_inputs, hsc_file):
+def _process_hsc_file(hs, cc, hsc_flags, hsc_inputs, hsc_file, extra_srcs):
     """Process a single hsc file.
 
     Args:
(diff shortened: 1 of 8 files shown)


googleson78:
Unresolved questions:

* Is this approach ok?
* How should I document this?
* How should I extend the tests - is it enough to just have the new target I've added be running, or should there be some other checks too?

haskell/private/actions/compile.bzl
```diff
@@ -105,6 +113,30 @@ def _process_hsc_file(hs, cc, hsc_flags, hsc_inputs, hsc_file):
...
+    we need to have the same directory structure.
+    Args:
+      hs: Haskell context.
+      header: The header we want to copy. Note that this is not strictly required to be a header.
+      hsc_dir_raw: The directory where we will be outputting a result with hsc2hs.
+        This is where we'll be mirroring the module structure.
+    Returns:
+      File: The generated file which is now a symlink to header.
+    """
+    # TODO[GL] fine to use rel_path_to_module here?
```
googleson78, line 132: Kind of seems like "abusing" the function. Maybe if we rename it, it'll sound better :sweat_smile:

haskell/private/actions/compile.bzl
```diff
@@ -105,6 +113,30 @@ def _process_hsc_file(hs, cc, hsc_flags, hsc_inputs, hsc_file):
 
     return hs_out, idir
 
+# Not strictly for headers but that's my motivating example..
+# TODO[GL} this will fail if the header is outside of our projec root
```
googleson78, line 117: Not sure what to do about this - my judgement is to just "leave it as future work", unless there's some straightforward solution I'm missing..

haskell/private/actions/compile.bzl
```diff
@@ -82,12 +83,19 @@ def _process_hsc_file(hs, cc, hsc_flags, hsc_inputs, hsc_file):
     if hs.env.get("PATH") == None and hs.toolchain.is_windows:
         hs.env["PATH"] = ""
 
+    # TODO[GL] should we be using extra_srcs for this?
```
googleson78, line 86: Opinions?

haskell/private/path_utils.bzl
```diff
@@ -465,11 +465,11 @@ def create_rpath_entry(
     else:
         return prefix + "/" + path
 
-def _rel_path_to_module(hs, f):
+def rel_path_to_module(hs, f):
     """Make given file name relative to the directory where the module hierarchy
     starts.
 
-    _rel_path_to_module(
+    rel_path_to_module(
```
googleson78, line 472: Is this fine to expose?

googleson78:
Closing as this seems too hacky and niche, for what could be accomplished without it (see https://github.com/tweag/rules_haskell/issues/1728#issuecomment-1106369758)

PR #1730, opened 2022-04-19
Ignore the pycache folder

googleson78:
I noticed this being untracked, and it seems like something that belongs in `.gitignore`


ylecornec (approved):
It looks good to me :+1:

PR #1731, opened 2022-04-19
Support hsc in haskell_module rules

### LINKED ISSUE #1727: Support for hsc in haskell_modules
`haskell_module` currently [doesn't support](https://github.com/tweag/rules_haskell/blob/a20de0b8437448a86c784b64332a49e9c1a0cd90/haskell/experimental/defs.bzl#L19) hsc files. It would be nice to have support for that.

Ideally, we should have it work with `.hsc` files as if they were normal `.hs` files.

I don't think there are any alternatives. One could manually or automatically generate the `.hs` files and commit them into their repo, but that seems like a very unsatisfactory solution.

googleson78:
Fixes #1727

### DIFF
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -6,7 +6,7 @@ load(
 load(
     "//haskell:private/expansions.bzl",
     "expand_make_variables",
-    "haskell_library_extra_label_attrs",
+    "haskell_library_expand_make_variables",
 )
 load(
     "//haskell:private/mode.bzl",
@@ -255,7 +255,7 @@ def _build_haskell_module(
 
     args.add_all(hs.toolchain.ghcopts)
 
-    args.add_all(expand_make_variables("ghcopts", ctx, ctx.attr.ghcopts, haskell_library_extra_label_attrs(ctx.attr)))
+    args.add_all(haskell_library_expand_make_variables("ghcopts", ctx, ctx.attr.ghcopts))
     module_extra_attrs = [
         [moduleAttr.src],
         moduleAttr.extra_srcs,
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -32,6 +32,11 @@ load(
     "HaskellModuleInfo",
 )
 load("//haskell:providers.bzl", "HaskellInfo", "HaskellLibraryInfo")
+load(
+    "//haskell:private/actions/process_hsc_file.bzl",
+    "preprocess_hsc_flags_and_inputs",
+    "process_hsc_file",
+)
 
 # Note [Narrowed Dependencies]
 #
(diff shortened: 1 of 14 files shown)


haskell/private/expansions.bzl
```diff
@@ -19,11 +19,12 @@ def expand_make_variables(name, ctx, strings, extra_label_attrs):
...
     return strings
 
-def haskell_library_extra_label_attrs(attr):
-    return [
-        attr.srcs,
-        attr.extra_srcs,
-        attr.data,
-        attr.plugins,
-        attr.tools,
+def haskell_library_expand_make_variables(name, ctx, strings):
```
googleson78, line 22: Unrelated change but I did this to help me understand what exactly the difference was between the two invocations of `expand_make_variables` in `private/module.bzl`. I think it's an improvement overall so I think we should keep it.

haskell/experimental/private/module.bzl
```diff
@@ -133,6 +138,26 @@ def _build_haskell_module(
...
+        moduleAttr.tools,
+    ]
+
+    user_ghcopts += expand_make_variables("ghcopts", ctx, moduleAttr.ghcopts, module_extra_attrs)
+
+    if src.extension == "hsc":
+        # TODO[AH] Support version macros
+        hsc_flags, hsc_inputs = preprocess_hsc_flags_and_inputs(dep_info, user_ghcopts, None)
+
+        # TODO[GL] Use idir when providing HaskellReplLoadInfo
```
googleson78, line 157: Not sure what to do here. I'd need to construct a repl provider to return from here, but we aren't currently doing so (and I'm not sure we can construct one from a single module?)

aherrmann: The REPL providers are generated in the REPL aspect: https://github.com/tweag/rules_haskell/blob/f2a0c08425e49150063e9d65310d6766c1b0953e/haskell/repl.bzl#L547
We'll need to extend that aspect to support module targets.

googleson78: Sounds like something for another issue to me. I didn't find one open though, but maybe I'm just not recognising that some issue directly relates to this.

aherrmann: Yes, definitely worth a ticket. I think we don't have a dedicated one, yet. There is a list of similar topics [here](https://github.com/tweag/rules_haskell/issues/1552#issuecomment-1000147400). But, I agree that they should be dedicated tickets.

googleson78: Created #1732

googleson78, line 157: Actually the discussion we had about repls is still valid, but this is also present in `HaskellInfo`, so I should probably propagate it from here in this PR.

googleson78: Tried to address this.

googleson78: I think what I've implemented is broken, but it only shows up when trying to use a `haskell_repl` on a `haskell_library` that has `modules`. I should add a test case for this and fix it.

googleson78: In order to test this with a repl, we need to support repls for targets with `haskell_module`s in them. I'm doing this in a separate PR. So when I have a draft for that I'll probably drop this commit and implement it later.

haskell/experimental/private/module.bzl
```diff
@@ -133,6 +138,26 @@ def _build_haskell_module(
     src = moduleAttr.src.files.to_list()[0]
     extra_srcs = [f for t in moduleAttr.extra_srcs + ctx.attr.extra_srcs for f in t.files.to_list()]
 
+    user_ghcopts = []
+    user_ghcopts += haskell_library_expand_make_variables("ghcopts", ctx, ctx.attr.ghcopts)
```
facundominguez, line 142: Same as
```python
user_ghcopts = haskell_library_expand_make_variables("ghcopts", ctx, ctx.attr.ghcopts)
```
?

facundominguez (approved):
LGTM!

PR #1733, opened 2022-04-28
Implement haskell_repl support for targets that have modules in them

### LINKED ISSUE #1732: Extend repl aspect to support haskell_module targets
The current way to provide repl info (an aspect) doesn't support `haskell_module` targets.

We should not forget to also take into account `haskell_module`s with `hsc` files in them, as they potentially generate additional import directories.

googleson78:
Fixes #1732


googleson78:
TODO: tests

haskell/experimental/private/module.bzl
```diff
@@ -548,6 +569,7 @@ def build_haskell_modules(
             object files and the object files of their transitive module
             dependencies. See Note [Narrowed Dependencies].
         per_module_transitive_dyn_objects: like per_module_transitive_objects but for dyn_o files
+        repl_info: struct(source_files, boot_files, import_dirs, user_compile_flags)
```
googleson78, line 572: Is this enough documentation? If not, how should I document the fields of the returned struct?

haskell/experimental/defs.bzl
```diff
@@ -7,6 +7,7 @@ load(
 )
 load("//haskell:private/cc_libraries.bzl", "haskell_cc_libraries_aspect")
 
+# TODO[GL] should we have repl_ghci_args here?
```
googleson78, line 10: Not sure about this one.

haskell/experimental/private/module.bzl
ylecornec: I would add some, to match the style of the doc for the other arguments. It looks you can indent it as it is done [here](https://github.com/tweag/rules_haskell/blob/master/haskell/private/path_utils.bzl#L101-L106).

googleson78: Added some docs here.

k1nkreet (commented):
Overall looks good to me

tests/haskell_module/repl/haskell_module_repl_test.go
```diff
@@ -0,0 +1,171 @@
+package haskell_module_repl_test
```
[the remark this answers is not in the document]
k1nkreet: > Perhaps this should get automated to read from the fs somehow

We are going there slowly :). The current workspace representation inherits the one used by `rules_go`. The `rules_bazel_integration_test` do this, but has other issues which is incompatible with `rules_haskell` needs.  Anyway it worth to either adopt `rules_bazel_integration_test`, either implement the same approach on top of `rules_go` rules we are using now

> however writing test cases inline in a go file feels a bit weird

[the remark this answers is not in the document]
googleson78: I don't mind go (but of course Haskell would be more convenient :P). I was thinking more along the lines of having the file structure of the test on the file system, and then instead of having a inline string, reading them from the fs. However, this will
* definitely slow down tests
* make file management more complex?

On the other hand,
* it makes the structure easier to observe and inspect in my opinion - people are used to browsing a file system, not so much a custom (albeit extremely lightweight) format
* error reports are better - e.g. when bazel fails to parse a test file now we get a line in that file, but that doesn't correspond to the source line in the .go file.

Anyways, I'm fine with following the status quo for now - it's definitely workable. I was just raising these things to potentially start a discussion. I think a definite improvement would be somehow exposing a common `WORKSPACE` for all these tests, or at least the ones that don't do anything fancy with the `WORKSPACE`.

tests/haskell_module/repl/haskell_module_repl_test.go
```diff
@@ -0,0 +1,171 @@
...
+    name = "lib",
+    modules = [
+        ":root",
+        ":branch_left",
+        ":branch_right",
+        ":leaf",
+    ],
+    deps = [":base"],
+)
+
```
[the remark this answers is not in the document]
googleson78: Thanks! Deleted. (copy paste..)

googleson78:
Now that https://github.com/tweag/rules_haskell/pull/1731 is merged, I should take care of the import_dirs situation here (after rebasing).

pranaysashank:
I just tested this PR on a local workspace with multiple packages, the repl target doesn't seem to build / include its dependencies.
Here is the workspace to test with https://github.com/pranaysashank/nd/tree/test-modules-repl 
```
$ bazel run //package-b:package-b@repl
GHCi, version 8.10.7: https://www.haskell.org/ghc/  :? for help
[1 of 2] Compiling PackageB.Mod2    ( package-b/src/PackageB/Mod2.hs, interpreted )

package-b/src/PackageB/Mod2.hs:4:1: error:
    Could not find module ‘PackageA.Mod2’
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
  |
4 | import qualified PackageA.Mod2
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Failed, no modules loaded.
[1 of 2] Compiling PackageB.Mod2    ( package-b/src/PackageB/Mod2.hs, interpreted )

package-b/src/PackageB/Mod2.hs:4:1: error:
    Could not find module ‘PackageA.Mod2’
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
  |
4 | import qualified PackageA.Mod2
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Failed, no modules loaded.
```

googleson78:
Thanks a lot @pranaysashank. I'll look into it.

googleson78:
I just tried out your example repo and the repl starts up fine for me and loads modules, at least for `package-a`.

How are you running your repl?

Both `bazel run //package-a:package-a@repl` and defining an explicit `haskell_repl` target and doing `bazel run //package-a:repl` work for me.

googleson78:
Derp, I see. I'll look into why it doesn't work cross package.

googleson78:
Including your cross-library dependencies from `package-a` in the `modules` of the `package-b` `haskell_library` makes the repl work fine. 

So this is either
* an issue where we're not taking into account `cross_library_deps` somewhere for the `repl` target
* a missing check to yell "things are missing from the `modules` attribute of your `haskell_library`"

Seeing as how the library builds without them included in there I assume it's the first one.

pranaysashank:
For the repl target, is it better to have behaviour similar to the normal build without modules attribute i.e. we build the cross package libraries that we depend on, instead of creating a repl with all the modules across packages. 

Perhaps you're missing building `narrowed_deps` in the repl target.

googleson78:
After a lot of struggling with the "empty libs" mechanism, I managed to get your example working in my latest commit, 346cdd75e936c9f815fb9edf61a6e656807c4963, by propagating the repl aspect along `narrowed_deps` too. 

I don't think this is the proper solution however, since this causes the repl target to rebuild all the modules in libraries specified in `narrowed_deps`, regardless if they are used.

@facundominguez can you offer some help here?

googleson78:
TODO: I should also add your example as a test to the repo

googleson78:
After a short discussion with @facundominguez, it turns out it might not be possible or it might be rather complicated. I think it would be better to leave it as an open question for another issues, as something working slowly is better than not working at all.

haskell/repl.bzl
```diff
@@ -182,6 +182,7 @@ def _create_HaskellReplCollectInfo(target, ctx):
 
     hs_info = target[HaskellInfo]
 
+    # TODO[GL]: do we need to also take into account narrowed_deps here, everywhere "deps" is checked?
```
googleson78, line 185: ?

googleson78:
Tests pass locally - one of the runs seems to be a [flaky error](https://github.com/tweag/rules_haskell/runs/6623005428?check_suite_focus=true) the [other one](https://github.com/tweag/rules_haskell/runs/6623005153?check_suite_focus=true) I'm not sure about - it passes locally :crying_cat_face:

haskell/repl.bzl
facundominguez: I would imagine it is necessary. To rehearse that we would need to add java dependencies to dependencies of Haskell libraries using `modules`.

haskell/experimental/defs.bzl
googleson78: Probably not for now. Seems like something we can try to add later if a user wants it.

facundominguez:
The changes look good to me in intent, and given that the tests that have been added pass, the PR looks like a net improvement. I'm not versed in the aspects that get the repl working to foresee unknown issues.

The code that manages java and CC dependencies in libraries of narrowed_deps doesn't have a test yet. If you still have steam, it would be necessary to add such tests. Otherwise, just make a new issue to add them later.

googleson78:
Apart from the CC and java dependencies, it would also be nice to have tests for hie bios output. I'll try to add all three next week.

googleson78:
Actually, I can't seem to find tests for the normal repl usage with cc/java deps :sweat_smile: If they exist, can you point me to them? If not, I think we should leave this as a separate issue (add tests for both "normal" and module usage).

facundominguez:
@aherrmann is this good to merge?

googleson78:
Opened #1756 and #1757

PR #1735, opened 2022-05-04
Remove unused `unfold_fields` function

avdv:


aherrmann (approved):
Good catch!

avdv:
@Mergifyio requeue

PR #1740, opened 2022-05-10
Add -main-is argument to _build_haskell_module.

pranaysashank:
when we have `modules` attr in `haskell_binary`, `main_function` is not respected which results in linking error when we try to build

### DIFF
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -183,6 +183,10 @@ def _build_haskell_module(
     # Construct compiler arguments
 
     args = ctx.actions.args()
+
+    if (moduleAttr.src == ctx.attr.main_file):
+        args.add_all(["-main-is", ctx.attr.main_function])
+
     args.add_all([
         "-c",
         "-odir",
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -190,7 +190,12 @@ def _build_haskell_module(
     if main_function:
         main_file = getattr(ctx.attr, "main_file", None)
         main_function_module = infer_main_module(main_function)
-        if (moduleAttr.src == main_file or main_function_module == moduleAttr.module_name):
+        if moduleAttr.module_name:
+            guess_module_name = moduleAttr.module_name
+        else:
+            guess_module_name = paths.split_extension(get_module_path_from_target(module))[0].replace("/", ".")
+
+        if (moduleAttr.src == main_file or main_function_module == guess_module_name):
             args.add_all(["-main-is", ctx.attr.main_function])
 
     args.add_all([
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -184,7 +184,8 @@ def _build_haskell_module(
 
     args = ctx.actions.args()
 
-    if (moduleAttr.src == ctx.attr.main_file):
+    if (moduleAttr.src == getattr(ctx.attr, "main_file", None)
+            and getattr(ctx.attr, "main_function", None)):
         args.add_all(["-main-is", ctx.attr.main_function])
 
     args.add_all([
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -188,13 +188,13 @@ def _build_haskell_module(
     main_function = getattr(ctx.attr, "main_function", None)
 
     if main_function:
-        main_file = getattr(ctx.attr, "main_file", None)
-        main_function_module = infer_main_module(main_function)
         if moduleAttr.module_name:
             guess_module_name = moduleAttr.module_name
         else:
             guess_module_name = paths.split_extension(get_module_path_from_target(module))[0].replace("/", ".")
 
+        main_file = getattr(ctx.attr, "main_file", None)
+        main_function_module = infer_main_module(main_function)
         if (moduleAttr.src == main_file or main_function_module == guess_module_name):
             args.add_all(["-main-is", ctx.attr.main_function])
 
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -1,4 +1,5 @@
 load("@bazel_skylib//lib:paths.bzl", "paths")
+load("//haskell:private/path_utils.bzl", "infer_main_module")
 load(
     "//haskell:private/dependencies.bzl",
     "gather_dep_info",
@@ -184,9 +185,13 @@ def _build_haskell_module(
 
     args = ctx.actions.args()
 
-    if (moduleAttr.src == getattr(ctx.attr, "main_file", None)
-            and getattr(ctx.attr, "main_function", None)):
-        args.add_all(["-main-is", ctx.attr.main_function])
+    main_function = getattr(ctx.attr, "main_function", None)
+
+    if main_function:
+        main_file = getattr(ctx.attr, "main_file", None)
+        main_function_module = infer_main_module(main_function)
+        if (moduleAttr.src == main_file or main_function_module == moduleAttr.module_name):
+            args.add_all(["-main-is", ctx.attr.main_function])
 
     args.add_all([
         "-c",
--- a/haskell/experimental/private/module.bzl
+++ b/haskell/experimental/private/module.bzl
@@ -191,7 +191,7 @@ def _build_haskell_module(
         if moduleAttr.module_name:
             guess_module_name = moduleAttr.module_name
         else:
-            guess_module_name = paths.split_extension(get_module_path_from_target(module))[0].replace("/", ".")
+            guess_module_name = get_module_path_from_target(module).replace("/", ".")
 
         main_file = getattr(ctx.attr, "main_file", None)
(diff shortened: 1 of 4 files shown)


ylecornec:
Thanks for the PR, 

One issue (that should not be too hard to fix) is that the `ctx.attr.main_file` attribute only exists for `haskell_binary` rules (and not `haskell_library`) but the `build_haskell_modules` function is called in both cases (this makes the CI fail).

Other than that, the fix works on this [example](https://github.com/tweag/rules_haskell/commit/e5e29e021a1520265029b570a56dd83eb0279e75) that uses the `main_file`,`main_function` and `modules` attributes of `haskell_test`.

However I am not familiar with the module system. @facundominguez, is there a better way to address this maybe ?

pranaysashank:
Pushed a fix for the failure.

haskell/experimental/private/module.bzl
```diff
@@ -147,6 +147,11 @@ def _build_haskell_module(
     # Construct compiler arguments
 
     args = ctx.actions.args()
+
+    if (moduleAttr.src == getattr(ctx.attr, "main_file", None)
+            and getattr(ctx.attr, "main_function", None)):
+        args.add_all(["-main-is", ctx.attr.main_function])
```
facundominguez, line 153: Reading the documentation of [main_function](https://api.haskell.build/haskell/defs.html#haskell_binary-main_function) looks like `-main-is` should be also passed when `main_file` is unset and the module name matches the name qualifier in `main_function` or `Main` if there is no qualifier.

Otherwise, the documentation would need to be updated.

facundominguez (commented):
Thanks @pranaysashank! Looks like a reasonable way to fix it. There is still some disagreement to fix with the documentation. Also, could we have a test added for this?

haskell/experimental/private/module.bzl
pranaysashank: Should I just do `args.add_all(["-main-is", ctx.attr.main_function])` without checking that the `moduleAttr.src` matches the main_file (so its added to all the modules while compiling for a binary), because I don't know how to construct an input file target similar to `moduleAttr.src` from the module name.

facundominguez: Maybe
```python
main_file = getattr(ctx.attr, "main_file", None)
main_function = getattr(ctx.attr, "main_function", None)
module_path = paths.split_extension(get_module_path_from_target(module))[0]
if main_function and (moduleAttr.src == main_file or main_function.replace(".", "/") == module_name):
    args.add_all(["-main-is", ctx.attr.main_function])
```
`get_module_path_from_target` is supposed to return "A/B/C.hs" for a module named "A.B.C".

I think this implements what the documentation says, but I'd feel better if someone double-checked it.

pranaysashank: Looks like this should work, however in my test setup `module_name` is an empty string for some modules but not others need to investigate what's happening.

pranaysashank:
I added a test for this custom main_function usage, and I am getting strange errors. This is how I am running the test
```
 bazel test //tests/haskell_module/binary-custom-main:Test
 ```
 and I'm getting the error
 ```
ERROR: /home/pranaysashank/src/tweag/rules_haskell/tests/haskell_module/binary-custom-main/BUILD.bazel:10:13: in _haskell_test rule //tests/haskell_module/binary-custom-main:Test: 
        File "/home/pranaysashank/src/tweag/rules_haskell/haskell/private/haskell_impl.bzl", line 97, column 39, in haskell_test_impl
                return _haskell_binary_common_impl(ctx, is_test = True)
        File "/home/pranaysashank/src/tweag/rules_haskell/haskell/private/haskell_impl.bzl", line 212, column 40, in _haskell_binary_common_impl
                module_map = determine_module_names(srcs_files, not main_as_haskell_module, ctx.attr.main_function, ctx.file.main_file)
        File "/home/pranaysashank/src/tweag/rules_haskell/haskell/private/path_utils.bzl", line 126, column 21, in determine_module_names
                fail("""\
Error in fail: main_file = "tests/haskell_module/binary-custom-main/TestBin.hs" is not listed in the "srcs" attribute.
ERROR: Analysis of target '//tests/haskell_module/binary-custom-main:Test' failed; build aborted: Analysis of target '//tests/haskell_module/binary-custom-main:Test' failed
INFO: Elapsed time: 0.045s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (1 packages loaded, 2 targets configured)
FAILED: Build did NOT complete successfully (1 packages loaded, 2 targets configured)
 ```
 
Am I missing something or is this another bug?

**Comment by facundominguez** [comment]
It looks to me like a bug, or a yet unsupported feature. No test in `tests/haskell_module` uses main_file so far.

**Comment by googleson78** [comment]
So there are at least two bugs here:
* `infer_main_module` is either "buggy" (it's getting passed `custom_main`, which is not a valid module name which gets checked, hence it returns `Main`) or getting called with the wrong arg
* `is_main_as_haskell_module` is returning `False`, even though (assuming from the name) it should return `True` in this situation

Still investigating

**Comment by googleson78** [comment]
Upon further inspection, the further assumption is currently made (and not documented as far as I can tell):

* if `main_file` is present, we're **not** building with `haskell_module`s

`determine_module_names` looks for the main module in `srcs` **only** when that happens.

**Comment by googleson78** [comment]
I think it would be fair to just document this and say "if you're building with modules, you **must** write the fully qualified name of the main function".

**Comment by pranaysashank** [comment]
If we change to always require a fully qualified main_function, do we try to get a module name for usages like this https://github.com/tweag/rules_haskell/pull/1740/files#diff-500f449c248c36e74ead8385c8dd2b98bb9244e21fafd0be214b0eb927c99abbR157 or is it okay to just pass `-main-is` argument to all modules. I thought it's not a good idea to always pass the option as the modules can be from a different package.

**Comment by pranaysashank** [comment]
>    if main_file is present, we're not building with haskell_modules

where do you see that assumption?

**Comment by pranaysashank** [comment]
BTW, in my project where I needed to use this, this fix does work. It's only in the test case above that it doesn't

**Comment by googleson78** [comment]
> where do you see that assumption?

`path_utils.bzl`, line 127 - https://github.com/pranaysashank/rules_haskell/blob/f0af91817ea66972c05b48b471fd58e5aae195f5/haskell/private/path_utils.bzl#L122

**Comment by pranaysashank** [comment]
> I think it would be fair to just document this and say "if you're building with modules, you must write the fully qualified name of the main function in main_function".

Is this the current behaviour for haskell_binary? Without this PR, fully qualifying main_function still doesn't work because we don't pass `-main-is` to ghc

**Comment by pranaysashank** [comment]
> But tbh I'm not seeing the benefit of using main_file/main_module when you can already fully qualify main_function, at least in the cases I'm thinking of. Perhaps there are some other considerations I'm missing?

`haskell_module` allows `module_name` to be empty, so we need `main_file` in those cases. I pushed a new commit that tries to guess `module_name` from `src` attribute of `haskell_module`, if that looks good enough, we can merge this PR

File: haskell/experimental/private/module.bzl [hunk 2]
```diff
@@ -183,6 +184,20 @@ def _build_haskell_module(
     # Construct compiler arguments
 
     args = ctx.actions.args()
+
+    main_function = getattr(ctx.attr, "main_function", None)
+
+    if main_function:
+        main_file = getattr(ctx.attr, "main_file", None)
```
googleson78, line 191: Can you move this below next to the second `if`, where it's actually used?

haskell/experimental/private/module.bzl
```diff
@@ -183,6 +184,20 @@ def _build_haskell_module(
...
+        main_function_module = infer_main_module(main_function)
```
googleson78, line 192: Can you move this below next to the second `if`, where it's actually used?

haskell/experimental/private/module.bzl
```diff
@@ -183,6 +184,20 @@ def _build_haskell_module(
...
+        if moduleAttr.module_name:
+            guess_module_name = moduleAttr.module_name
+        else:
+            guess_module_name = paths.split_extension(get_module_path_from_target(module))[0].replace("/", ".")
```
googleson78, line 196: Doesn't `get_module_path_from_target` already do the `.hs` removal internally? This seems redundant here.

googleson78:
The buildifier test has failed (`bazel test //:buildifier_test`). You should run `bazel run //:buildifier-fix` and commit the fix it generated.

haskell/experimental/private/module.bzl
googleson78: Fixed it

haskell/experimental/private/module.bzl
googleson78: Fixed it

haskell/experimental/private/module.bzl
googleson78: Fixed it

facundominguez (approved):
LGTM!

facundominguez (approved):
LGTM!

I created https://github.com/tweag/rules_haskell/issues/1806 to deal with `main_file`.

PR #1743, opened 2022-05-16
Configure the build to use the CC toolchain's linker for GHC >= 9.0.1

### LINKED ISSUE #1736: Configure the build to use the CC toolchain's linker explicitly
We configure the build with GHC or Cabal or similar to use the C compiler provided by Bazel's toolchain explicitly (through the CC wrapper), e.g. [here](https://github.com/tweag/rules_haskell/blob/df5aeb7169021d88ef7bdf2a1d7655838cd1b375/haskell/private/cabal_wrapper.py#L265). And we do the same for the linker, see [here](https://github.com/tweag/rules_haskell/blob/df5aeb7169021d88ef7bdf2a1d7655838cd1b375/haskell/cc.bzl#L204-L205).

However, recent versions of GHC (>9.0.1) expose a [new flag `-pgmlm`](https://downloads.haskell.org/~ghc/9.0.1/docs/html/users_guide/phases.html#ghc-flag--pgmlm%20%E2%9F%A8cmd%E2%9F%A9) to configure the linker used to merge objects separately. If we don't configure this one explicitly either then GHC will not use the linker from Bazel's CC toolchain, but instead the defined in its settings file. This can lead to inconsistent build results, in particular, it was found to lead to build errors on M1 MacOS in a Nix configuration.

We should also set `-pgmlm` in recent GHC versions. A solution has been implemented [here](https://github.com/digital-asset/daml/blob/81c7a8ae00b596ce4fcb34f2e8a58b9efa1003dd/bazel_tools/haskell-arm-m1.patch) and needs to be upstreamed into rules_haskell.

**Additional context**
This was discovered in [the Daml repository](https://github.com/digital-asset/daml/pull/13401/commits/81c7a8ae00b596ce4fcb34f2e8a58b9efa1003dd#).

avdv:
# Purpose

For newer GHC (>= 9.0.1) we have to pass on the correct linker of the CC toolchain.

Fixes #1736

Note, this PR is based upon #1739


aherrmann (approved):
Thank you, looks good!

PR #1755, opened 2022-06-08
Persistent worker prototype for haskell_module

facundominguez:
Not ready for general use. See the accompanying README for details.

### DIFF
--- a/haskell/toolchain.bzl
+++ b/haskell/toolchain.bzl
@@ -32,14 +32,12 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, env, params_file = No
     extra_inputs = []
 
     # Detect persistent worker support
-    flagsfile_prefix = ""
     execution_requirements = {}
     tools = []
     if hs.worker != None and worker == None:
         worker = hs.worker
 
     if worker != None:
-        flagsfile_prefix = "@"
         execution_requirements = {
             "supports-workers": "1",
             "requires-worker-protocol": "proto",
@@ -100,7 +98,7 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, env, params_file = No
         # for different actions. This allows persistent workers to be
         # reused for different actions. Otherwise, a different worker
         # would be spawned for each action.
-        arguments = [exe_path, flagsfile_prefix + flagsfile.path],
+        arguments = [exe_path, "@%s" % flagsfile.path],
         execution_requirements = execution_requirements,
     )
 
--- a/haskell/toolchain.bzl
+++ b/haskell/toolchain.bzl
@@ -44,10 +44,10 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, env, params_file = No
             "supports-workers": "1",
             "requires-worker-protocol": "proto",
         }
-        args.add(worker.path)
+        exe_path = worker.path
         tools = [worker]
     else:
-        args.add(hs.tools.ghc)
+        exe_path = hs.tools.ghc.path
         extra_inputs += [hs.tools.ghc]
 
     # XXX: We should also tether Bazel's CC toolchain to GHC's, so that we can properly mix Bazel-compiled
@@ -62,18 +62,17 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, env, params_file = No
     hs.actions.write(compile_flags_file, args)
     hs.actions.write(extra_args_file, arguments)
 
-    extra_inputs += [
-        compile_flags_file,
-        extra_args_file,
-    ] + cc.files + hs.toolchain.bindir + hs.toolchain.libdir
+    extra_inputs += cc.files + hs.toolchain.bindir + hs.toolchain.libdir
 
     if hs.toolchain.locale_archive != None:
         extra_inputs.append(hs.toolchain.locale_archive)
 
-    flagsfile = extra_args_file
     if params_file:
-        flagsfile = merge_parameter_files(hs, extra_args_file, params_file)
-        extra_inputs.append(flagsfile)
+        flagsfile0 = merge_parameter_files(hs, extra_args_file, params_file)
+    else:
+        flagsfile0 = extra_args_file
+    flagsfile = merge_parameter_files(hs, compile_flags_file, flagsfile0)
+    extra_inputs.append(flagsfile)
 
     if type(inputs) == type(depset()):
         inputs = depset(extra_inputs, transitive = [inputs])
@@ -97,7 +96,11 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, env, params_file = No
         mnemonic = mnemonic,
         progress_message = progress_message,
         env = env,
-        arguments = [compile_flags_file.path, flagsfile_prefix + flagsfile.path],
+        # Mind that the argument list here only differs on the last argument
+        # for different actions. This allows persistent workers to be
+        # reused for different actions. Otherwise, a different worker
+        # would be spawned for each action.
+        arguments = [exe_path, flagsfile_prefix + flagsfile.path],
         execution_requirements = execution_requirements,
     )
 
--- a/haskell/toolchain.bzl
+++ b/haskell/toolchain.bzl
@@ -27,7 +27,7 @@ load(
 
 _GHC_BINARIES = ["ghc", "ghc-pkg", "hsc2hs", "haddock", "ghci", "runghc", "hpc"]
 
-def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, env, params_file = None, progress_message = None, input_manifests = None, extra_name = ""):
+def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, env, params_file = None, progress_message = None, input_manifests = None, extra_name = "", worker = None):
     args = hs.actions.args()
     extra_inputs = []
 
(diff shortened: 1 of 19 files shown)


haskell/toolchain.bzl
```diff
@@ -27,21 +27,39 @@ load(
...
+    if hs.worker != None and worker == None:
+        worker = hs.worker
+
+    if worker != None:
+        execution_requirements = {
+            "supports-workers": "1",
+            "requires-worker-protocol": "proto",
+            "supports-multiplex-workers": "1",
+            "supports-multiplex-sandboxing": "1",
+        }
```
facundominguez, line 58: I think this change in `execution_requirements` is disabling the old persistent worker for `haskell_library`. Should I make efforts to keep it functioning? (if it is still functioning condition, I haven't tried it).

tools/haskell_module_worker/README.md
```diff
@@ -0,0 +1,60 @@
...
+constrain the total amount of workers. `bazel` can only constrain the total
+amount of workers assigned to a combination of mnemonic and startup parameters.
+Or said differently it can only constrain the amount of workers assigned to a
+particular library. This is why this persistent worker implements an LRU cache
+to keep the amount of workers limited, and thus ensures that the memory does
+not fill up with unused persistent workers.
+
+A few issues need to be investigated to make this implementation ready for
+general use. At the time of this writting they are:
+
```
facundominguez, line 52: This README needs to say something about how the keys for the LRU cache are computed. Specially, because the keys could be computed from something more meaningful than now.

tools/haskell_module_worker/README.md
```diff
@@ -0,0 +1,60 @@
...
+  --worker_max_instances=1 \
+  --worker_quit_after_build \
+  --define use_haskell_module_worker=True \
+  --worker_verbose \
+  --experimental_worker_max_multiplex_instances=2 \
+  --worker_extra_flag --num-workers=2 \
+  //tests/haskell_module/dep-narrowing-th/..
+```
+
+Bazel 6 is necessary to use the flag `--experimental_worker_multiplex_sandboxing`.
```
googleson78, line 17: I'm guessing adding tests will be difficult because of this restriction?

tools/haskell_module_worker/src/Compile.hs
```diff
@@ -0,0 +1,279 @@
...
+              return $
+                CompileErrors logs [show (e :: PlainGhcException)]
+            _ -> case fromException se of
+              Just e | not (isPanic e) ->
+                return $
+                  CompileErrors logs [show (e :: GhcException)]
+              _ ->
+                throwIO se
+        Right _ ->
+          return (Succeeded logs)
```
googleson78, line 183: Might be an improvement if you move this case first since it's much shorter?

tools/haskell_module_worker/src/Compile.hs
```diff
@@ -0,0 +1,279 @@
...
+-- | Creates a status from the exceptions thrown by the given action
+-- and the logs in the given 'IORef'.
+collectStatus :: IORef [String] -> IO () -> Ghc Status
+collectStatus logsRef action = do
+    hsc_env <- GHC.getSession
+    liftIO $ do
+      ee <- try action
+      logs <- reverse <$> readIORef logsRef
+      case ee of
+        Left se -> case fromException se of
```
googleson78, line 165: I think the `fromException` calls here would be clearer with a type application/annotation.

tools/haskell_module_worker/src/Compile.hs
```diff
@@ -0,0 +1,279 @@
...
+  :: (String -> IO a)
+  -> DynFlags
+  -> WarnReason
+  -> Severity
+  -> SrcSpan
+  -> PprStyle
+  -> MsgDoc
+  -> IO a
+renderLog k dflags reason severity srcSpan pprStyle msg
+    = (>>= k) $ case severity of
```
googleson78, line 226:
```suggestion
    = k =<< case severity of
```
Or is this part copied and perhaps not worth changing?

tools/haskell_module_worker/README.md
facundominguez: Yeah, or run without a sandbox, or run it in singleplex mode instead.

tools/haskell_module_worker/src/Compile.hs
```diff
@@ -0,0 +1,279 @@
...
+smallestGroups :: WarningFlag -> [String]
+smallestGroups flag = mapMaybe go warningHierarchies where
+    -- Because each hierarchy is arranged from smallest to largest,
+    -- the first group we find in a hierarchy which contains the flag
+    -- is the smallest.
+    go (group:rest) = fromMaybe (go rest) $ do
+        flags <- lookup group warningGroups
+        guard (flag `elem` flags)
+        pure (Just group)
+    go [] = Nothing
```
googleson78, line 279: It seems to me like you return a `Maybe (Maybe a)` here, but the value can never be `Just Nothing`. So this can be simplified to
```haskell
smallestGroups :: WarningFlag -> [String]
smallestGroups flag = mapMaybe go warningHierarchies where
    -- Because each hierarchy is arranged from smallest to largest,
    -- the first group we find in a hierarchy which contains the flag
    -- is the smallest.
    go :: [String] -> Maybe String
    go (group:rest) = do
        flags <- lookup group warningGroups
        if flag `elem` flags
        then Just group
        else go rest
    go [] = Nothing
```
Additionally, I also suspect you can slightly simplify this further by using `find` instead:
```haskell
-- Because each hierarchy is arranged from smallest to largest,
-- the first group we find in a hierarchy which contains the flag
-- is the smallest.
smallestGroups :: WarningFlag -> [String]
smallestGroups flag = mapMaybe (find go) warningHierarchies where
    go :: String -> Bool
    go group =
      case lookup group warningGroups of
        Nothing -> False
        Just flags -> flag `elem` flags
```

googleson78:
Do you think it would be worth it to ask the ghc people to eventually expose some of the things that are copied from ghc? I can also do the exposing myself in a MR to ghc, but I don't know which ones you need.

tools/haskell_module_worker/src/Main.hs
```diff
@@ -0,0 +1,226 @@
...
+  -> ProtoClient
+  -> IO ()
+serveRequests opts semWorkers outChan lruV pc = go
+  where
+    go = do
+      forwardRequest semWorkers outChan lruV pc
+      if optPersist opts then
+        go
+      else
+        return ()
```
googleson78, line 134:
```suggestion
      when (optPersist opts) go
```

tools/haskell_module_worker/src/ProtoClient.hs
```diff
@@ -0,0 +1,92 @@
...
+
+-- | Writes a response to whatever file given to 'createProtoClient',
+-- then flushes it.
+--
+-- Takes the exit code and the output.
+writeWorkResponse :: ProtoClient -> Int -> Int -> String -> IO ()
+writeWorkResponse (ProtoClient pc) requestId exitCode output = do
+    rc <- withCString utf8 output $
+            c_writeWorkResponse pc (fromIntegral requestId) (fromIntegral exitCode)
+    when (rc /= 0) $ error "writeWorkResponse failed"
```
googleson78, line 92: Should probably be `throwIO` as well?

tools/haskell_module_worker/src/Compile.hs
facundominguez: Thanks @googleson78. This code is copied from GHC though, so I rather leave it as is.

tools/haskell_module_worker/src/Main.hs
```diff
@@ -0,0 +1,226 @@
...
+      case Binary.decodeOrFail bs of
+        Right (rest, _, wr) -> do
+          liftIO $ forM_ (wrSandboxDir wr) (setCurrentDirectory . (cwd </>))
+          st <- compile (wrArgs wr) (wrVerbosity wr)
+          liftIO $ do
+            ByteString.Lazy.putStr (Binary.encode st)
+            hFlush stdout
+          go cwd rest
+        Left (_, _, err) ->
+          liftIO $ hPutStrLn stderr $ "Error decoding work request: " ++ err
```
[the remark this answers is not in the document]
facundominguez: The background worker terminates since it doesn't call recursively to `go`, this closes stdout, which lets the main worker know that something went wrong.

tools/haskell_module_worker/src/Worker.hs
```diff
@@ -0,0 +1,100 @@
...
+        outBS <- ByteString.Lazy.hGetContents pout
+        _ <- forkIO $ forever $
+          Text.hGetLine perr >>= handleStderrLine
+        return Info
+          { wiProcessHandle = ph
+          , wiIn = pin
+          , wiOut = pout
+          , wiOutBS = outBS
+          }
+      _ -> error "error starting worker"
```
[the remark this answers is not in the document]
facundominguez: `throwIO` would require writing an exception type to throw here. `error` also gives us the source code location in the error message which is nice.

While we don't have a need to catch and manage these exceptions in a more structured way, I would keep it simple as it is here.

tools/haskell_module_worker/src/Main.hs
facundominguez: Done.

tools/haskell_module_worker/src/Main.hs
```diff
@@ -0,0 +1,226 @@
...
+
+    statusExitCode = \case { Succeeded{} -> 0; _ -> 1 }
+
+    statusOutput = \case
+      Succeeded logs -> unlines $ intersperse "" logs
+      CompileErrors logs errs -> unlines $ intersperse "" $ logs ++ errs
+      NonHaskellInputs files ->
+        unlines $ "haskell_module_worker error: non-haskell inputs:" : files
+      NonOneShotCompilation ->
+        "haskell_module_worker error: only on-shot compilation (-c) is supported"
```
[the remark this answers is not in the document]
facundominguez: Done.

tools/haskell_module_worker/src/Compile.hs
facundominguez: Done.

tools/haskell_module_worker/src/Compile.hs
facundominguez: Done.

tools/haskell_module_worker/src/Compile.hs
googleson78: Ah, of course :sweat_smile:

tools/haskell_module_worker/README.md
facundominguez: Done.

PR #1761, opened 2022-06-14
Support version macros in the haskell_module rule

facundominguez:


tests/version-macros/BUILD.bazel
```diff
@@ -32,6 +33,40 @@ haskell_test(
...
 
+haskell_test(
+    name = "version_macros_haskell_module",
+    size = "small",
+    modules = [
+        ":HsLib",
+        ":HscLib",
+        ":Main",
+    ],
+    version = "3.4.5.6",
```
googleson78, line 44: The tests only seem to check if the define for `versioned_lib` exist, not the one for `version_macros_haskell_module`. This also seems to be the case in the previously existing test, there's no check for for whether `VERSION_version_macros` is defined. Perhaps we should add checks for those?

Am I misunderstanding something?

facundominguez: > not the one for version_macros_haskell_module

I just tried it. `VERSION_version_macros` isn't defined when building `version_macros`. If Cabal does define it, it could make sense to create a feature request.

facundominguez:
@googleson78, thanks for having a look. Is this good to merge?

googleson78:
Is the following test case supposed to pass? I just tried it and it doesn't seem to be passing, but maybe I am completely not understanding how `version` works.
```
# BUILD.bazel
haskell_test(
    name = "version_macros_haskell_module_test",
    size = "small",
    modules = [
        ":ModuleHsLib",
        ":Main",
    ],
    version = "1.2.3.4",
    deps = [
        "//tests/hackage:base",
    ],
)

haskell_module(
    name = "ModuleHsLib",
    src = "ModuleHsLib.hs",
)

haskell_module(
    name = "Main",
    src = "Main.hs",
    deps = [
        ":ModuleHsLib",
    ],
)

# Main.hs
module Main where

import qualified ModuleHsLib

main :: IO ()
main = do
  ModuleHsLib.check
# ModuleHsLib
{-# LANGUAGE CPP #-}

module ModuleHsLib where

import Control.Monad (unless)

check_version_version_macros_haskell_module_test :: IO ()
#ifndef VERSION_version_macros_haskell_module_test
check_version_version_macros_haskell_module_test = fail "hs: VERSION_version_macros_haskell_module_test missing"
#else
check_version_version_macros_haskell_module_test =
  unless (VERSION_version_macros_haskell_module_test == "1.2.3.4") $
    fail "hs: VERSION_version_macros_haskell_module_test invalid"
#endif

check_min_version_version_macros_haskell_module_test :: IO ()
#ifndef MIN_VERSION_version_macros_haskell_module_test
check_min_version_version_macros_haskell_module_test = fail "hs: MIN_VERSION_version_macros_haskell_module_test missing"
#elif !MIN_VERSION_version_macros_haskell_module_test(1,2,3)
check_min_version_version_macros_haskell_module_test = fail "hs: MIN_VERSION_version_macros_haskell_module_test invalid"
#else
check_min_version_version_macros_haskell_module_test = pure ()
#endif

check :: IO ()
check = do
  check_version_version_macros_haskell_module_test
  check_min_version_version_macros_haskell_module_test
```

facundominguez:
I don't know if it should pass. But I think that if it should pass, it should be a new feature request.

A similar test for `version_macros` (i.e. not using `haskell_module`) doesn't pass either. The current implementation relies on `ghc` to generate the version macros from the information in package databases. `ghc` cannot generate version macros for the binaries in the compile action because we don't give it any information about how they are called or which version they have. Thus, supporting this would require manually creating `-DVERSION_blah=3.4.5.6` flags. This is why I say it could make sense if Cabal makes something similar.

This PR would already improve the capabilities of `rules_haskell` by bringing `haskell_module` on par with `haskell_test`. Argueably, there is not much benefit in keeping this on hold until more is implemented.

googleson78:
Good point, thanks for clarifying! Sorry for dragging this on so long due to misunderstanding :sweat_smile: Feel free to merge :)

facundominguez:
How come? github won't let me approve my own PR :). And furthermore, it want's the PR approved so it can be merged :shrug:

googleson78:
Added an approval :sweat_smile:

PR #1763, opened 2022-06-16
Simplify iserv usage with the arm example

avdv:
I noticed that the setup for using an "external-interpreter" with ghc is a bit complicated, using two processes communicating with one another using TCP and one ghc communicating with the proxy using FIFOs, although every process is run on the same system.

The performance hit is probably negligible, but trying at most 4 times to find a port [5000, 5999] is probably also not very solid.

However, my main motivation was that I could not easily use a different compiler, or a ghc from nixpkgs with it because iserv-proxy and remote-iserv are not readily available. Trying to use ghc 9.0.2 from haskell.nix and the accompanying tools was taking a long time (probably that ghc variant is not in the cache?), so I figured that I could simply call out to `ghc-iserv` directly...


avdv:
@facundominguez was there a reason why you went with this approach in the first place, or do you think there could be any problems with just calling `ghc-iserv`?

facundominguez:
I have recollections of picking the `remote-iserv` use from another example. If `ghc-iserv` works for you, the change looks like a net win :+1:.

avdv:
I also asked in this GHC issue for the reason using proxy and remote: https://gitlab.haskell.org/ghc/ghc/-/issues/21767

The answer was as expected:
> For qemu-user mode (transparent execution) where file handles are preserved, yes, we don’t need remote and proxy. For others where the communication needs to be bridged over the network, we need proxy and remote.

So, indeed we can drop using proxy and remote and simply use ghc-iserv instead.

facundominguez (approved):

PR #1765, opened 2022-06-23
Change integration tests implementation to one based on `rules_bazel_integration_test`

k1nkreet:
This PR suggests alternative approach to integration testing, based on rules_bazel_integration_test ruleset. Unfortunately I could not come up with anything less radical than put all required changes in one PR, cause otherwise both implementation would stay for a while in repo which could be very unclear and misleading 

So here is a list of changes I've put here:

1. Create haskell_bazel_integration_test rule for integration test
      with test scenario written in haskell and similar to go_bazel_test setup procedure
2. Create rules_haskell_integration_test macro for integration test running
      haskell scenario with all bazel versions supported in rules_haskell
3. Remove rules_go-based integration_test implementation
4. Remove rules_go patching for the sake of this implementation
5. Create IntegrationTesting library with helper functions formerly
      located in RunTests.hs
6. Reimplement existing integration_test implementations with
      rules_haskell_integration_test

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -66,4 +66,5 @@ buildifier_test(
         "@examples//:all_files",
         "@tutorial//:all_files",
     ],
+    tags = ["dont_test_on_windows"],
 )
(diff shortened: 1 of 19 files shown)


k1nkreet:
Sorry, took a wrong branch for PR

PR #1766, opened 2022-06-23
Change integration tests implementation to one based on `rules_bazel_integration_test`

k1nkreet:
This PR suggests alternative approach to integration testing, based on `rules_bazel_integration_test ruleset`. Unfortunately I could not come up with anything less radical than put all required changes in one PR, cause otherwise both implementation would stay for a while in repo which could be very unclear and misleading 

So here is a list of changes I've put here:

1. Create haskell_bazel_integration_test rule for integration test
      with test scenario written in haskell and similar to `go_bazel_test` setup procedure
2. Create `rules_haskell_integration_test` macro for integration test running
      haskell scenario with all bazel versions supported in rules_haskell
3. Remove rules_go-based integration_test implementation
4. Remove rules_go patching for the sake of this implementation
5. Create IntegrationTesting library with helper functions formerly
      located in RunTests.hs
6. Reimplement existing integration_test implementations with
      `rules_haskell_integration_test`


googleson78:
Awesome!

Regarding 6., I added a couple more tests using the old framework in `tests/haskell_module/repl`

tests/integration_testing/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
...
+    deps = [
+        "//tests/hackage:base",
+        "//tests/hackage:directory",
+        "//tests/hackage:filepath",
+        "//tests/hackage:process",
+        "//tests/hackage:text",
+        "@stackage//:hspec",
+        "@stackage//:hspec-core",
+    ],
+    # visibility = ["//visibility:public"],
```
googleson78, line 16: Redundant comment?

googleson78:
Thanks! Sorry for creating additional work for the migration.

k1nkreet:
> 

> Thanks! Sorry for creating additional work for the migration.

There is plenty of it still in RunTests.hs, so don't be :)

tests/integration_testing/BUILD.bazel
k1nkreet: Yep, thanks for noticing

k1nkreet:
Tests on Windows are failing due to the [bug in `rules_bazel_integration_test`](https://github.com/bazel-contrib/rules_bazel_integration_test/issues/69). I'm going to help with debug and fixing it. In the meantime we can ignore these tests on windows until it fixed, or just wait for the fix

tests/integration_testing/IntegrationTesting.hs
```diff
@@ -0,0 +1,163 @@
...
+\ build:linux-nixpkgs --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host \n\
+\ build:linux-nixpkgs --incompatible_enable_cc_toolchain_resolution \n\
+\ build:macos-nixpkgs --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host \n\
+\ build:macos-nixpkgs --incompatible_enable_cc_toolchain_resolution \n\
+\ build:linux-bindist --incompatible_enable_cc_toolchain_resolution \n\
+\ build:macos-bindist --incompatible_enable_cc_toolchain_resolution \n\
+\ build:windows-bindist --crosstool_top=@rules_haskell_ghc_windows_amd64//:cc_toolchain \n\
+\ ")
+
+setupWorkspace :: IO (String, String)
```
aherrmann, line 71: I think a docstring that describes the workspace setup and how rules_haskell is important would be useful

shell.nix
```diff
@@ -35,6 +34,8 @@ mkShell {
     # check the start script for problems
     shellcheck
     file
+    # Needed for rules_bazel_integration_test
+    clang
```
aherrmann, line 38: Out of curiosity, why is that needed?

aherrmann (commented):
Thanks a lot for pushing this forward! This looks a lot more pleasant to use!

A couple comments/questions.

> I'm going to help with debug and fixing it. In the meantime we can ignore these tests on windows until it fixed, or just wait for the fix

Thanks! Happy to temporarily skip those on Windows. You could already line up a PR to re-enable them so that it's not forgotten.

tests/integration_testing/IntegrationTesting.hs
k1nkreet: I was thinking about put a README.md in tests/integration_tesinting/ forlder

.bazelrc
```diff
@@ -111,6 +111,12 @@ build: --incompatible_override_toolchain_transition
 
 coverage --build_tag_filters "coverage-compatible" --test_tag_filters "coverage-compatible" --test_output=all
 
+# To update these lines, execute
+# `bazel run @contrib_rules_bazel_integration_test//tools:update_deleted_packages`
```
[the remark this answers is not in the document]
k1nkreet: It will cause a test failure. But problem is it will have a bit cryptic error output

shell.nix
k1nkreet: This puzzled me also, ruleset implementation demanding this explicitly https://github.com/bazel-contrib/rules_bazel_integration_test/blob/main/bazel_integration_test/private/bazel_integration_test.bzl#L137-L141, but I still didn't figure out why

aherrmann: Probably worth an upstream ticket. The only occurrences of `clang` in that repo according to GH search are around the `CC` variable. Maybe it's a leftover from some previous context?

k1nkreet: https://github.com/bazel-contrib/rules_bazel_integration_test/issues/70

tests/integration_testing/IntegrationTesting.hs
```diff
@@ -0,0 +1,163 @@
...
+import System.Exit (ExitCode(..))
+import Control.Monad (when, unless, forM_)
+import Test.Hspec (shouldSatisfy, expectationFailure)
+
+bazelCmd :: String -> String -> IO ([String] -> Process.CreateProcess)
+bazelCmd workspaceDir outputUserRoot = do
+    bazelPath <- getEnv "BIT_BAZEL_BINARY"
+    config <- (fmap bazelConfig isNixpkgs)
+    return (\args -> case args of
+        subcommand:xs | elem subcommand ["build", "test", "run", "query"] -> (Process.proc bazelPath (["--output_user_root", outputUserRoot, subcommand, "--config", config] ++ xs)) { Process.cwd = Just workspaceDir }
```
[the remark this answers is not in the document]
aherrmann: Thanks!

.bazelrc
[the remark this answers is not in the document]
k1nkreet: I've drop a note about it in "adding a new integration test guide" in integration tests README

tests/integration_testing/IntegrationTesting.hs
k1nkreet: Done

aherrmann (approved):
Thank you, that looks great!

I fixed a couple typos in the README, please take another look in case I messed something up.

k1nkreet:
> Thank you, that looks great!
> 
> I fixed a couple typos in the README, please take another look in case I messed something up.

Thanks! Btw there is a problem with `Cross` ci job. It complains on "no space left on device" for some Nix operations. Do you have an idea how we could fix it?

aherrmann:
> Btw there is a problem with `Cross` ci job. It complains on "no space left on device" for some Nix operations. Do you have an idea how we could fix it?

Hmm, this is odd. It seems to be failing when fetching `haskell.nix` already.

> unpacking 'https://github.com/input-output-hk/hackage.nix/archive/3bce0587d9ffe56d11e83b5eeccd63e49b643bdd.tar.gz'...
tar: hackage/sloane-1.8.2-r0-26781c13cac7e1a8a70fb3928e39a4b6bac5b6ea742f83ddc174a3b915ba8207.nix: Cannot write: No space left on device

I'm not sure what's causing this. AFAIK nothing in this PR should affect that pipeline, right?

k1nkreet:
> I'm not sure what's causing this. AFAIK nothing in this PR should affect that pipeline, right?

Yes, and it was passing before README fixes, which I hardly believe could influence this

k1nkreet:
Since https://github.com/bazel-contrib/rules_bazel_integration_test/pull/71 is merged, there is an ability to specify the `env` for integration test. So I've changed a way how `nixpkgs` indication get passed into the test from command line to environment. Among other things, It allows to wrap integration tests with `hspec` again as it was done in `RunTests.hs` and which I removed because it was impossible to use command line with them. I was going to return it back, but I'm doubting if it's actually useful in this scenario (when every test is a separate executable running probably one function). @aherrmann Do you have an opinion?

aherrmann:
@k1nkreet I'd say, what's important is that test failures produce equally clear and understandable error messages. Otherwise, as you say, if each binary represents a single test anyway, then there isn't much benefit in wrapping it in `hspec`.

k1nkreet:
@aherrmann Can we add this to the merge queue, or there is something else to fix?

Issue #1760, opened 2022-06-13
Implement unused_inputs_list for haskell_module


facundominguez:
`rules_haskell` provides the interface file of a module M as input to the action that builds every module that imports it transitively. This is because the compiler might need to read the interface file for M, and the build system can't predict reliably whether it will be needed or not.

If the interface file of M changes, then Bazel will arrange to rerun all those actions whether the interface files of the modules along the path in the dependency graph are needed or not. A remedy for this could be to use [unused_input_list](https://bazel.build/rules/lib/actions#run) to report to bazel the interface files that don't affect the output. This might be possible to do if we can get the compiler to report which interface files have been actually read during compilation.

It needs to be investigated whether this is possible to implement with GHC as it is today, or perhaps with the GHC API, or perhaps by using `strace`.

Related https://github.com/tweag/rules_haskell/issues/1552

GuillaumeGen:
#1771 also introduces small example: `basic_modules`. It contains 3 files:

A.hs:
```haskell
module A (f, g, T, a) where

data T = T0 | T1

a :: T
a = T1

f :: a -> b -> a
f x y = x

g :: T -> T
g T0 = a
g T1 = T1
```

B.hs:
``` haskell
module B where

import qualified A (f, g, T)

f :: A.T -> A.T
f x = A.f (A.g x) x

g :: a -> a -> a -> a
g x y = A.f (A.f x) (A.f x)
```

and C.hs:
```haskell
module C where

import qualified B

data N = Z | S N

f :: a -> a -> a
f x = B.g x x
```

There are then quite a lot of changes one can make to `A.hs` or `B.hs` which changes the interface file `B.hi` but do not affect the `ABI` hash of `B` avoiding to trigger the recompilation of `C.hs`.

Changing the export list of `A`:
```diff
- module A (f, g, T, a) where 
+ module A (f, g, T) where 
```

Change the type of a function of `A`:
```diff
- f :: a -> b -> a
- f x y = x
+ f :: a -> b -> b
+ f x y = y
```

Inline a function of `A`:
```diff
+ {-# INLINE f #-}
f :: a -> b -> a
```

Modify the import list of `B`:
```diff
- import qualified A (f, g, T)
+ import qualified A (f, g, T, a)
```

All those modifications affects the part of the interface file of `B` regarding imports, hence it changes `B.hi` but not `B.abi`, so without #1771 it triggers a recompilation of `C`, it is not the case anymore.

[...]

PR #1771, opened 2022-07-08
Generate ABI files and use them to avoid recompilation

GuillaumeGen:
This tries to address #1760 by generating a file containing only the "ABI" hash in order to avoid recompilation when the interface file of a dependency is modified in a place which does not affect the ABI hash, meaning that the modification in the dependency is irrelevant to the module depending of it.


facundominguez (commented):
Thanks @GuillaumeGen! I think this PR is helpful ground for refining a solution.

There are some structural changes to do (mostly merging actions) and there are some missing abi files I think.

I'm deferring style comments and requests for documentation after the next round of changes.

haskell/toolchain.bzl
```diff
@@ -27,7 +27,58 @@ load(
 
 _GHC_BINARIES = ["ghc", "ghc-pkg", "hsc2hs", "haddock", "ghci", "runghc", "hpc"]
 
-def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, env, params_file = None, progress_message = None, input_manifests = None, extra_name = ""):
+def _create_abi(hs, inputs, outputs, progress_message = None):
+    hs.actions.run_shell(
+        inputs = inputs,
+        outputs = outputs,
+        command =
+            """sed -n "s/^\\s*ABI hash:\\s*\\(\\S*\\)$/\\1/p" {readable_hi_input} > {abi_output}""".format(
```
aherrmann: Please obtain `sed` from the POSIX toolchain. See [here](https://github.com/tweag/rules_haskell/blob/2b3c715d57b5f8a36a7b4d2e2f4729d6397e3a8a/haskell/private/actions/link.bzl#L103) for an example.

haskell/private/ghc_wrapper.sh
```diff
@@ -6,21 +6,33 @@ export PATH=${PATH:-} # otherwise GCC fails on Windows
 # support older bash versions.
 while IFS= read -r line; do compile_flags+=("$line"); done < "$1"
 
-# Detect if we are in the persistent worker mode
-if [ "$2" == "--persistent_worker" ]; then
```
aherrmann, line 10: Uhm, IIUC this removes support for the persistent worker. That doesn't look right...

GuillaumeGen: Indeed, I suppressed this line when I created 2 versions of the script, but I did not keep it, when going back to only one version. This is an error.

haskell/toolchain.bzl
GuillaumeGen: For some reason I do not understand why this comment is not tagged as "outdated". Does your comment still applies to the current version, where `sed` is called within the bash script `ghc-wrapper.sh`?

aherrmann: > For some reason I do not understand why this comment is not tagged as "outdated".

Good question, perhaps because I commented on the individual commit and not the overall diff?

> Does your comment still applies to the current version, where `sed` is called within the bash script `ghc-wrapper.sh`?

Good question. I think it would still be preferable to have this and potentially other tools provided by the POSIX toolchain. IIRC we already have such a mechanism in place with the [`path_for_run_ghc`](https://github.com/tweag/rules_haskell/blob/662f1423754a2dfd72f991d3346bdefe793bc696/haskell/toolchain.bzl#L83). @ylecornec AFAIK this is currenlty only used for the Asterius toolchian. Do you think we could also use it to provide POSIX tools for the regular toolchain?

haskell/toolchain.bzl
```diff
@@ -56,6 +71,49 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, env, params_file = No
...
+
+        extra_inputs += [compiler_file, flags_file, interface_with_abis_list]
+        outputs += [abi_file, readable_hi_file, unused_inputs_list]
+        env.update({"MUST_EXTRACT_ABI": "true"})
+
+        new_args = [compiler_file.path, flags_file.path, readable_hi_file.path, abi_file.path, interface_with_abis_list.path, unused_inputs_list.path]
+    else:
+        env.update({"MUST_EXTRACT_ABI": "false"})
+        new_args = []
+        unused_inputs_list = None
```
facundominguez, line 115: All of `compiler_file.path`, `readable_hi_file.path` and `abi_file.path` are inferrable from the flags files that are already given to the `ghc_wrapper`. Do we need to pass them again from bazel, or can the ghc_wrapper figure on its own ?

`readable_hi_file.path` in particular is an intermediate file only needed by the action to produce the abi file, does it need to trickle into all of the logic that handles action outputs in the `haskell_module` implementation?

haskell/toolchain.bzl
```diff
@@ -56,6 +71,49 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, env, params_file = No
...
+        hs.actions.write(interface_with_abis_list, interface_with_abis_args)
+
+        # The interface_with_abis_list should be used as unused_inputs_list.
+        # However, since it is more traditional to generate the unused_inputs_list within the action itself,
+        # and this proposition is already quite hackish, we wanted to stick to the usage,
+        # expecting it to offer us a stronger reliability, even with the future evolution of Bazel.
+        unused_inputs_list = hs.actions.declare_file("unused_%s_%s" % (hs.name, extra_name))
+
+        extra_inputs += [compiler_file, flags_file, interface_with_abis_list]
+        outputs += [abi_file, readable_hi_file, unused_inputs_list]
```
facundominguez, line 108: I would have the functions calling to `run_ghc` add the `abi_file` to the outputs if they want it.

Since `readable_hi_file` is a temporary file only needed by the action, I don't think it should be listed as an output.

haskell/toolchain.bzl
```diff
@@ -56,6 +71,49 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, env, params_file = No
...
+        ghc_args.set_param_file_format("multiline")
+        ghc_args.add(hs.tools.ghc.path)
+        hs.actions.write(compiler_file, ghc_args)
+
+        # The second one contains the "--show-iface" option and the name of the hi file.
+        flag_args = hs.actions.args()
+        flag_args.set_param_file_format("multiline")
+        flag_args.add("--show-iface ")
+        flag_args.add(hi_file.path)
+        hs.actions.write(flags_file, flag_args)
```
facundominguez, line 90: The `ghc_wrapper` could produce these flags itself when extracting the ABI is required. It doesn't seem necessary to assemble them in the caller side.

facundominguez (commented):
The logic in `build_haskell_modules` looks better. There are still some things to revise.

haskell/experimental/private/module.bzl
```diff
@@ -652,8 +658,53 @@ def build_haskell_modules(
...
+        # and the narrowed dependencies.
         interface_inputs = _collect_module_inputs(module_interfaces, narrowed_interfaces, his, dep)
-        abi_inputs = depset(direct = abis)
+        # Similarly abi_inputs contains the direct dependencies to a module from the library we are compiling,
+        # and the narrowed dependencies.
+        # One only wants the direct abi files, since if a modifiaction in a transitive dependency did not affect any abi of a direct dependency,
+        # it means that those changes do not impact the file we are considering, hence recompilation can be avoided.
+        abi_inputs = depset(direct = abis, transitive = [narrowed_abis])
+
+        ### On the ABI hash ###
```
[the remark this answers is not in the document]
GuillaumeGen: What is the benefit of formatting the comments title like this. Is there any tool which benefit of it or is it just for homogeneity?

haskell/experimental/private/module.bzl
```diff
@@ -703,17 +772,25 @@ def build_haskell_modules(
...
     _merge_depset_dicts(per_module_transitive_objects0, per_module_maps.transitive_objects)
     _merge_depset_dicts(per_module_transitive_dyn_objects0, per_module_maps.transitive_dyn_objects)
+    _merge_depset_dicts(per_module_abi0, per_module_maps.abis)
 
     return struct(
         his = hi_set,
         dyn_his = dyn_hi_set,
         os = o_set,
         dyn_os = dyn_o_set,
+        abis = abi_set,
```
facundominguez, line 789: Maybe this can be removed as we can't find where it is used.

haskell/experimental/private/module.bzl
```diff
@@ -619,6 +676,7 @@ def build_haskell_modules(
         # called in all cases to validate cross_library_deps, although the output
         # might be ignored when disabling narrowing
         narrowed_interfaces = _collect_narrowed_deps_module_files(ctx.label, per_module_maps.transitive_interfaces, dep)
+        narrowed_abis = [per_module_maps.abis[m] for m in per_module_maps.abis]
```
facundominguez, line 679: Revise this because it looks like it collects all the narrowed abis, while we only want the ones from modules that are directly imported.

haskell/toolchain.bzl
```diff
@@ -56,6 +70,41 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, env, params_file = No
...
 
+    if abi_file != None:
+        # We declare the file containing informations about the call to GHC with --show-iface.
+        show_iface_file = hs.actions.declare_file("show_iface_%s_%s" % (hs.name, extra_name))
+        show_iface_args = hs.actions.args()
+        show_iface_args.set_param_file_format("multiline")
+        show_iface_args.add(hs.tools.ghc.path)
+        show_iface_args.add("--show-iface ")
+        show_iface_args.add(hi_file.path)
+        hs.actions.write(show_iface_file, show_iface_args)
```
facundominguez, line 81: @aherrmann, we have some controversy on what the interface of the GHC wrapper should be.

@GuillaumeGen is arguing that it is easier to accomplish the task if we keep this `show_iface_file`, while I think it is preferable to have the ghc_wrapper produce these arguments on its own for the sake of modularity (so the caller isn't concerned with how the callee performs the work).

Do you have any opinion on what the interface should be?

aherrmann: > [..] I think it it is preferable to have the ghc_wrapper produce these arguments on its own for the sake of modularity (so the caller isn't concerned with how the callee performs the work).

Generally, I'd agree with this point of view.

> it is easier to accomplish the task if we keep this `show_iface_file`

However, I can see how this may be the case, given that `ghc_wrapper` is a bash script, and the information is coming in in a relatively unstructured format.

My impression is that `ghc_wrapper.sh` is growing beyond the scale where a bash script is still reasonable. A better approach may be the one also used by the Cabal actions: A `py_binary` that takes a JSON file as an input, such that the invoking rule can provide the needed metadata in a structured format. However, that's a wider refactoring. So, that would belong into its own PR.

facundominguez: @GuillaumeGen, please, leave it as is for now.

facundominguez (commented):
This is starting to look correct :tada:

I contributed some edits to the ABI hash note.

haskell/private/ghc_wrapper.sh
```diff
@@ -6,21 +6,33 @@ export PATH=${PATH:-} # otherwise GCC fails on Windows
...
-    "${compile_flags[@]}" "${extra_args[@]}" 2>&1 \
-      | while IFS= read -r line; do [[ $line =~ ^(Loaded|Warning: the following files would be used as linker inputs, but linking is not being done:) ]] || echo "$line"; done >&2
+# Drop messages that GHC produces on features that we rely upon.
+#
+# "Loaded" is emitted when using GHC environment files, which we
+# use as poor man's response files for GHC.
+#
+# "Warning: the following files ..." is produced when we tell GHC
+# to load object files in the interpreter in the build action
+# of haskell_module which doesn't do any linking.
```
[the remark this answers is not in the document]
facundominguez: This comment should go inside `drop_loaded_and_warning ()`, I think.

haskell/experimental/private/module.bzl
```diff
@@ -532,20 +564,75 @@ def _merge_narrowed_deps_dicts(rule_label, narrowed_deps):
...
         transitive_dyn_objects = transitive_dyn_objects,
     )
 
 def interfaces_as_list(with_shared, o):
     if with_shared:
-        return [o.hi, o.dyn_hi]
+        return [o.hi, o.dyn_hi, o.abi]
     else:
-        return [o.hi]
+        return [o.hi, o.abi]
```
facundominguez, line 582: Is this change necessary? I would think we shouldn't mix interface and abi files here.

haskell/experimental/private/module.bzl
```diff
@@ -532,20 +564,75 @@ def _merge_narrowed_deps_dicts(rule_label, narrowed_deps):
                 str(rule_label),
                 str(dep.label),
             ))
+        abis.update(lib_info.per_module_abi)
         _merge_depset_dicts(transitive_interfaces, lib_info.per_module_transitive_interfaces)
         _merge_depset_dicts(transitive_objects, lib_info.per_module_transitive_objects)
         _merge_depset_dicts(transitive_dyn_objects, lib_info.per_module_transitive_dyn_objects)
     return struct(
+        abis = abis,
```
facundominguez, line 572: This change requires updating the documentation of the function.

haskell/experimental/private/module.bzl
```diff
@@ -714,6 +822,7 @@ def build_haskell_modules(
         per_module_transitive_interfaces = per_module_transitive_interfaces0,
         per_module_transitive_objects = per_module_transitive_objects0,
         per_module_transitive_dyn_objects = per_module_transitive_dyn_objects0,
+        per_module_abi = per_module_abi0,
```
facundominguez, line 825: This change also requires updating the documentation of the function.

haskell/experimental/private/module.bzl
```diff
@@ -347,10 +351,14 @@ def _build_haskell_module(
         progress_message = "HaskellBuildObject {} {}".format(hs.label, module.label),
         env = hs.env,
         arguments = args,
-        extra_name = module.label.package.replace("/", "_") + "_" + module.label.name,
```
facundominguez, line 350: I'd leave this line as it was, unless you want to use the factored out `module_name` elsewhere.

facundominguez:
I created https://github.com/tweag/rules_haskell/issues/1783 to discuss testing of incremental builds.

haskell/toolchain.bzl
aherrmann: Given that it's now part of the GHC wrapper, this can be done in a separate PR.

haskell/experimental/private/module.bzl
```diff
@@ -742,3 +852,5 @@ def haskell_module_impl(ctx):
             transitive_module_dep_labels = transitive_module_dep_labels,
         ),
     ]
+
+    return result
```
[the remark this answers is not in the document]
GuillaumeGen: Bad reasons. I added a print here when I explored this function and when I suppressed the print command, I did not revert this.

facundominguez (approved):
LGTM!

haskell/toolchain.bzl
```diff
@@ -56,6 +70,41 @@ def _run_ghc(hs, cc, inputs, outputs, mnemonic, arguments, env, params_file = No
...
+        interface_with_abis_args = hs.actions.args()
+        interface_with_abis_args.set_param_file_format("multiline")
+        interface_with_abis_args.add_all(interface_inputs)
+        hs.actions.write(interface_with_abis_list, interface_with_abis_args)
+
+        # The interface_with_abis_list should be used as unused_inputs_list.
+        # However, since it is more traditional to generate the unused_inputs_list within the action itself,
+        # and this proposition is already quite hackish, we wanted to stick to the usage,
+        # expecting it to offer us a stronger reliability, even with the future evolution of Bazel.
+        unused_inputs_list = hs.actions.declare_file("unused_%s_%s" % (hs.name, extra_name))
```
facundominguez, line 96: How about making the new declared files that are introduced in this PR use paths that depend on all the parameters that the earlier files did? For instance, all of the following depend on `hs.name` and `extra_name`
```
show_iface_file = hs.actions.declare_file("show_iface_%s_%s" % (hs.name, extra_name))
interface_with_abis_list = hs.actions.declare_file("interfaces_%s_%s" % (hs.name, extra_name))
unused_inputs_list = hs.actions.declare_file("unused_%s_%s" % (hs.name, extra_name))
```
while the preexisting files depend on `hs.name`, `extra_name`, and `mnemonic`.
```
compile_flags_file = hs.actions.declare_file("compile_flags_%s_%s_%s" % (hs.name, extra_name, mnemonic))
extra_args_file = hs.actions.declare_file("extra_args_%s_%s_%s" % (hs.name, extra_name, mnemonic))
```

facundominguez:
This is almost ready to merge. The only thing we are waiting before proceeding is to check that we understand the effects of this PR on the code bases we work with. Firstly, we want to be sure that recompilation is avoided for all the files that we expect, and secondly, we want to know that any overhead introduced by reading interface files to generate ABI files is tolerable.

haskell/toolchain.bzl
facundominguez: Turns out this is not necessary as we are only creating this files with a single mnemonic so far.

haskell/toolchain.bzl
googleson78: Tests fail for me (NixOS) because of a missing sed. Created an issue - #1800

PR #1772, opened 2022-07-08
Test all supported bazel versions in integration tests

k1nkreet:
This PR is an invitation to discuss which Bazel versions we are considering as supported, since we can explicitly specify all Bazel versions which would be tested on CI.
As a start I've added all major versions from the interval specified in [haskell/private/versions.bzl](https://github.com/tweag/rules_haskell/blob/cfed08d1915bffab8c59edffbc9f1a07eda5b959/haskell/private/versions.bzl#L114)

The second question is which nixpkgs Bazel versions we are going to test. [Current implementation](https://github.com/tweag/rules_haskell/blob/cfed08d1915bffab8c59edffbc9f1a07eda5b959/bazel_versions.bzl#L5) offers to specify a list of Nixpkgs packages to test, but it will use nixpkgs version pinned in `rules_haskell/nixpkgs/default.nix` (there is no bazel_5 package for ex.) which is a bit unfair to `rules_haskell's` users. As a suggestion maybe we should use pair `(nixpkgs revision, bazel package)` to identify bazel packages to test. Or just keeping `rules_haskell/nixpkgs/default.nix` up-to-date will be good enough?


aherrmann:
> As a start I've added all major versions from the interval specified in [haskell/private/versions.bzl](https://github.com/tweag/rules_haskell/blob/cfed08d1915bffab8c59edffbc9f1a07eda5b959/haskell/private/versions.bzl#L114)

IIUC, this corresponds to all *minor* versions in the range.
According to Bazel's [versioning policy](https://docs.bazel.build/versions/main/versioning.html#understanding-versioning-on-bazel)
> * A major release contains features that are not backward compatible with the previous release.
> * A minor release contains new backward-compatible features.

Perhaps a lower and upper bound for each major version would be enough? E.g. if we start relying on a new minor release feature, we may no longer be compatible with a previous minor of the same major release family.

k1nkreet:
> Perhaps a lower and upper bound for each major version would be enough? E.g. if we start relying on a new minor release feature, we may no longer be compatible with a previous minor of the same major release family.

Agree

aherrmann:
Looks like CI is unhappy about something.

k1nkreet:
Should we test Bazel 5.x also? @aherrmann what do you think?

aherrmann:
> Should we test Bazel 5.x also? @aherrmann what do you think?

Yes, though, feel free to do that in a separate PR if it needs fixing some compatibility issues.

aherrmann (approved):
Nice, thank you!

PR #1773, opened 2022-07-11
Enable test error output on Bazel CI

avdv:
Having error output is tremendously useful after the fact when tests failed on CI where you cannot simply inspect the log files.


aherrmann (approved):
Thank you!

PR #1774, opened 2022-07-11
Keep Github actions up-to-date

avdv:
I noticed that a few Github actions we are using are already a bit old.

Let's keep them up-to-date automatically.

See https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot


aherrmann (approved):
Nice, thank you!

PR #1781, opened 2022-07-12
Add Bazel 5 to the integration testing on CI

k1nkreet:

PR #1793, opened 2022-07-22
Correct checksum instructions and update hash

avdv:
After the release, I noticed that the checksum for the tag was apparently not correct.

Also, the max Bazel version was not yet adapted in the `start` script after #1781 .


ylecornec (approved):
It looks good to me :+1:

PR #1796, opened 2022-07-28
Validate hidden modules

### LINKED ISSUE #1687: Clarify the meaning of hidden_modules and fix the code
The documentation of the `hidden_modules` attribute of the `haskell_library` rules currently says:
```
Modules that should be unavailable for import by dependencies.
```

The text itself is wrong, since dependencies of the `haskell_library` rule cannot import any modules of it. But there are also a couple of situations that need to be explained.

### Non-existent modules

The following instantiation will produce an error in the analysis phase
```
haskell_library(
    name = "lib",
    srcs = [
        "src/A.hs",
        "src/B.hs",
    ],
    hidden_modules = ["C"],
)
```
The error is:
```
        File ".../76262e73dad9f87ef6526dd61590ea62/external/rules_haskell/haskell/private/haskell_impl.bzl", line 505, column 27, in haskell_library_impl
                set.mutable_difference(exposed_modules, set.from_list(other_modules))              
        File ".../76262e73dad9f87ef6526dd61590ea62/external/rules_haskell/haskell/private/set.bzl", line 103, column 26, in _mutable_difference
                s0._set_items.pop(item)                                                                                                                                                               
Error in pop: KeyError: "C"
```

This is because `mutable_difference` is calling `pop(item)` instead of `pop(item, default = None)`. However, I would argue that it should be considered an error to list hidden modules that do not exist. If we agree on this, the documentation should say so, and the code should check it and produce an intelligible message. 

### Rexported modules

It turns out that `hidden_modules` does not only hide modules of this library but also modules in `reexported_modules`. And I'd feel inclined again to argue that this should rather be considered an error, and that documentation should say so, and that an appropriate error message should be produced in this situation.

If not considered an error, then at least it should be explained in the documentation that `hidden_modules` affects the reexported modules as well.

**Environment**
 - Version of the rules: master

martyall:
fixes #1687 

I tested this by just manually tweaking test files and monitoring for errors, because I was unsure of how to write a test expecting a failure. If one is required, can you please point me to an example?

### DIFF
--- a/haskell/private/haskell_impl.bzl
+++ b/haskell/private/haskell_impl.bzl
@@ -463,12 +463,10 @@ def haskell_library_impl(ctx):
     srcs_files, import_dir_map = _prepare_srcs(ctx.attr.srcs)
     module_map = determine_module_names(srcs_files)
 
-
     # Validate that hidden modules appear as modules in src list:
     hidden_minus_all_modules = set.difference(set.from_list(ctx.attr.hidden_modules), set.from_list(module_map.keys()))
     if not hidden_minus_all_modules == set.empty():
         fail("""Hidden modules must be a subset of all modules, found additional hidden modules {}""".format(set.to_list(hidden_minus_all_modules)))
-
 
     package_name = getattr(ctx.attr, "package_name", None)
     version = getattr(ctx.attr, "version", None)
--- a/haskell/private/haskell_impl.bzl
+++ b/haskell/private/haskell_impl.bzl
@@ -463,6 +463,13 @@ def haskell_library_impl(ctx):
     srcs_files, import_dir_map = _prepare_srcs(ctx.attr.srcs)
     module_map = determine_module_names(srcs_files)
 
+
+    # Validate that hidden modules appear as modules in src list:
+    hidden_minus_all_modules = set.difference(set.from_list(ctx.attr.hidden_modules), set.from_list(module_map.keys()))
+    if not hidden_minus_all_modules == set.empty():
+        fail("""Hidden modules must be a subset of all modules, found additional hidden modules {}""".format(set.to_list(hidden_minus_all_modules)))
+
+
     package_name = getattr(ctx.attr, "package_name", None)
     version = getattr(ctx.attr, "version", None)
     my_pkg_id = pkg_id.new(ctx.label, package_name, version)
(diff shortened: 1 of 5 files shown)


haskell/private/haskell_impl.bzl
```diff
@@ -526,7 +526,15 @@ def haskell_library_impl(ctx):
...
     other_modules = ctx.attr.hidden_modules
     exposed_modules_reexports = _exposed_modules_reexports(ctx.attr.reexported_modules)
+
     haskell_module_names = [haskell_module_from_target(m) for m in modules]
+
+    # Validate that hidden modules appear as modules in src list or modules list, depending which appears:
+    declared_modules = haskell_module_names if modules else module_map.keys()
+    hidden_minus_declared_modules = set.difference(set.from_list(ctx.attr.hidden_modules), set.from_list(declared_modules))
+    if not hidden_minus_declared_modules == set.empty():
+        fail("""Hidden modules must be a subset of all modules, found additional hidden modules {}""".format(set.to_list(hidden_minus_declared_modules)))
```
aherrmann, line 536: Orthogonal to this PR, but we should really get rid of `private/set.bzl` and replace it by [Skylib](https://github.com/bazelbuild/bazel-skylib/blob/main/lib/sets.bzl). I've opened https://github.com/tweag/rules_haskell/issues/1799 to track this.

aherrmann (commented):
Thanks for looking into this!

You can find tests for `haskell_module` [here](https://github.com/tweag/rules_haskell/tree/d8afc5739639f01e734b833426d70d10155be9bd/tests/haskell_module) and tests for hidden modules [here](https://github.com/tweag/rules_haskell/tree/master/tests/hidden-modules).

One possible way to test this would be to use [failure testing](https://bazel.build/rules/testing#failure-testing).

martyall:
Here is the result of running `bazel test :test-missing-module` in the `tests/hidden-modules` directory. 

Screenshot from 2022-08-24 12-23-56

How do I make sure this gets run with the rest of the tests?

aherrmann:
> How do I make sure this gets run with the rest of the tests?

CI invokes `bazel test //...`, so, the tests will be run without further steps required.

aherrmann (approved):
Thank you, that looks good!

Side note, you can push your branch to the tweag-org's rules_haskell repo to get CI to run on it.

martyall:
are these tests failures that don't seem to have anything to do with this work... Do you see anything related to this PR in the output?

aherrmann:
```
       //:buildifier_test                                                       FAILED in 0.2s
```
It's the formatting test that fails. You can reformat your code by invoking buildifier on it (`bazel run //:buildifier-fix`).

martyall:
> Side note, you can push your branch to the tweag-org's rules_haskell repo to get CI to run on it.

Yeah it seems like it doesn't always run automatically for external PRs, but when I pushed the branch of the same name to this repo, it triggered... weird

aherrmann:
> Yeah it seems like it doesn't always run automatically for external PRs, but when I pushed the branch of the same name to this repo, it triggered... weird

Yes, that's unfortunately how it is at the moment. It's the same issue as described here: https://github.com/tweag/rules_nixpkgs/issues/243.

martyall:
CI is failing pretty spectacularly on windows

aherrmann:
@martyall that looks like an issue with the remote cache. I've raised it [here](https://buildbuddy.slack.com/archives/CUY16GNK1/p1661840437979189). I've triggered a rerun to see if it was transient.

PR #1801, opened 2022-08-11
Use PATH from posix toolchain in normal compilation

### LINKED ISSUE #1800: Tests don't succeed due to sed dependency
On NixOS (I assume this is the cause), running
```
> bazel test //tests/haskell_module/repl:haskell_module_repl_cross_library_deps_test_nixpkgs_bazel_4
```
fails, complaining about a missing sed. I assume this is (directly) related to this discussion - https://github.com/tweag/rules_haskell/pull/1771#discussion_r918747509, and the solution should also be contained there.

googleson78:
This fixes the issue with sed(used in ghc_wrapper.sh) being gotten from the environment instead of the posix toolchain.

Fixes #1800

### DIFF
--- a/haskell/toolchain.bzl
+++ b/haskell/toolchain.bzl
@@ -271,7 +271,14 @@ def _haskell_toolchain_impl(ctx):
             tools_for_ghc_pkg = ctx.files.tools,
         )
     else:
-        tools_config = default_tools_config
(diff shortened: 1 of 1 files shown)


aherrmann (approved):
Thank you, LGTM!

PR #1802, opened 2022-08-11
Fix linker error when using haskell_module with th and local library

kozak:


WORKSPACE
```diff
@@ -103,6 +103,27 @@ stack_snapshot(
     },
 )
 
+new_local_repository(
+    name = "system-libs",
+    # pkg-config --variable=libdir x11
```
googleson78, line 108: Seems like a leftover?

kozak: Oh I am sorry, don't think this is ready for review yet. It fixes the problem, but the test case requires a local library which isn't portable.

googleson78: I converted the PR into a draft, feel free to undo that once it's ready :)

haskell/experimental/private/module.bzl
```diff
@@ -338,6 +350,7 @@ def _build_haskell_module(
                     narrowed_deps_info.deps_hs_libraries,
                     narrowed_deps_info.empty_hs_libraries,
                     object_inputs,
+                    depset(static_libs + dynamic_libs),
```
kozak, line 353: We've added this because of the error in PR description. However this problems doesn't happen for `zlib` so we are not sure if we are correctly including the `libphonenumber` library.  Does adding  dependencies to `run_ghc` here make sense from you perspective @googleson78  ?

WORKSPACE
```diff
@@ -103,6 +103,27 @@ stack_snapshot(
     },
 )
 
+new_local_repository(
```
kozak, line 106: @googleson78 do you have any idea of a better / more portable way to provide the library to the test? Unfortunately it has a bunch of dependencies https://github.com/google/libphonenumber/tree/master/cpp. I don't know much about nix, but maybe there is a nix package for it? (https://github.com/google/libphonenumber/tree/master/cpp)

googleson78:
I rebased your branch in master and modified your test to be fetching the dependency from nix in https://github.com/tweag/rules_haskell/tree/fix-haskell-module-th-linking. In particular, https://github.com/tweag/rules_haskell/commit/1340c40e218dcb77ae383c10be98eb8b6fb01c0f is the commit which makes the dependency change (and the only new commit there). 

Feel free to cherry-pick that or reset your branch to that one (I had to rebase because tests fail on NixOS from where you branched). 

Note however that there is still some work to be done regarding the test, as it must be disabled(?) when not running `rules_nixpkgs`, which I need to dig a bit more to find out how to do.

googleson78:
Getting this to work with the bindist route will require some effort, while disabling the test for bindists feels unsatisfying as well. Can you specifically on reproduce the issue with `libphonenumber`, or is it possible to also do so with any C library? (e.g. reproducing with zlib would be optimal here)

kozak:
Thanks for your help! 
Unfortunately this issue doesn't manifest itself with zlib, this was the first thing that I tried :). Hard to say about different C libs.

Issue #1742, opened 2022-05-16
Support Asterius on Darwin arm64 / Apple M1


avdv:
**Is your feature request related to a problem? Please describe.**

When testing `rules_haskell` I noticed that the nodejs toolchain could not be found. The reason for this is that the default nodejs version (currently 14.17.5 https://github.com/bazelbuild/rules_nodejs/blob/4.4.6/nodejs/repositories.bzl#L11) does not support darwin_arm64 yet, and creation of the corresponding `nodejs_darwin_arm64` repository is simply skipped by rules_nodejs.

The first nodejs version that supports running on that platform is the current LTS version 16.x.

**Describe the solution you'd like**

Upgrade rules_nodejs to version 5.x where 16.x is now the default: 

https://github.com/bazelbuild/rules_nodejs/pull/3107

Additionally, the `--experimental-wasm-bigint` flag needs to be removed when calling `node` since it is no longer understood by nodejs >= 16.x.

To that end, we also need to use a more recent Asterius bundle since the `--experimental-wasm-bigint` flag has been in use by Asterius itself, but was removed as of commit tweag/asterius@9c75ad777d1dc23cc33cc35ed38a63cca2eddb28.

**Describe alternatives you've considered**

Instead of upgrading rules_nodejs we could also specify the `node_version` parameter when calling `node_repositories`:
```
load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories")

node_repositories(
    node_version = "16.13.0",
)
```

**Additional context**

See https://github.com/bazelbuild/rules_nodejs/wiki/Migrating-to-5.0

Depends on PR #1739

avdv:
Fixed by #1810

PR #1810, opened 2022-08-22
Update nodejs

ylecornec:
This PR makes rules_haskell use node 16 as mentioned in issue #1742.

- For bindists, `rules_nodejs` is updated so that node 16 is now the default.
- For nix, `nodejs-16_x` is now used instead of the default `nodejs`.

#### rules_nodejs update.

 The way to use rules_nodejs 5.0.0 with nix changed quite a bit:

- The node from nix is still installed via the `_nixpkgs_nodejs` macro but was modified a bit because `node` is now expected at a specific path.
- In addition, the name of the external repository containing node is expected to end with the os name. Because of this we declare the same repository multiple times under all the possible names [here](https://github.com/tweag/rules_haskell/pull/1810/files?diff=unified&w=0#diff-9d21ccabe299eeb98999e18b704133d701be776e90d771eb972751729b9878f2R110), but only the one corresponding to the current os will be used. (At first I defined a new repository rule to get the os_name and recover it with a `load` command in the `WORKSPACE` file, but declaring all the possible names seems simpler).
- `yarn_install` now has a `yarn` attribute that can point to a nix installed yarn, which we install with `_nixpkgs_yarn`.
- We must now declare the node toolchain ourselves (in the `_declare_nix_node_toolchain_impl` rule) and register it.
- The name of the nodejs toolchain type changed to `@rules_nodejs//nodejs:toolchain_type`.

#### shellcheck

I encountered some shellcheck errors, mostly because it was trying to analyze template files which are not complete yet, so I added directives to these errors.

#### The `--experimental-wasm-bigint`option
This option is not needed anymore with node 16, so the asterius bundle was updated to a new version to fix this (as mentioned in #1742).

### DIFF
--- a/.netlify/build.sh
+++ b/.netlify/build.sh
@@ -2,7 +2,7 @@
 
 set -eux
 
-export PATH=$HOME/bin:$PATH
+export PATH="$HOME/bin:$PATH"
 
 # XXX We don't want to be using the Nixpkgs CC toolchain, because
 # Nixpkgs is not available. But currently we can only override the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 [#1790]: https://github.com/tweag/rules_haskell/pull/1790
 
+### Changed
+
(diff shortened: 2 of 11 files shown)


avdv (commented):
LGTM :100: 

Could you just add an entry to the Changelog.md file, please?

avdv (approved):
Thank you!

PR #1828, opened 2022-11-04
Fix empty glob results

### LINKED ISSUE #1827: rules_haskell does not build with incompatible_disallow_empty_glob
See: https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1320#018435e1-7742-44cb-911b-c21b89fb6e95

avdv:
Fixes #1827 

I did enable the `--incompatible_disallow_empty_glob` flag and found a few problems with `glob` calls by it.

Unfortunately, we cannot yet flip the switch to disallow empty glob results, since the embedded bazel_tools repository of the Bazel version we are using is not yet ready:

https://github.com/bazelbuild/bazel/blob/f64b7553607e1d3572611cc5011c498e3cd4505c/tools/jdk/jdk.BUILD#L165-L169

That is the reason the bindist checks fail with an error:
```
ERROR: Traceback (most recent call last):
	File "/private/var/tmp/_bazel_runner/d1e3bdaa0c37ce8d00ecff62b05849c3/external/local_jdk/BUILD.bazel", line 168, column 16, in <toplevel>
		srcs = glob(["conf/**"]),
Error in glob: glob pattern 'conf/**' didn't match anything, but allow_empty is set to False (the default value of allow_empty can be set with --incompatible_disallow_empty_glob).
ERROR: /private/var/tmp/_bazel_runner/d1e3bdaa0c37ce8d00ecff62b05849c3/external/bazel_tools/tools/jdk/BUILD:336:6: Target '@local_jdk//:jdk' contains an error and its package is in error and referenced by '@bazel_tools//tools/jdk:jdk'
```

After review, I'll remove the flag again so CI will pass.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -341,14 +341,14 @@ nixpkgs_package(
 
 nixpkgs_package(
     name = "nixpkgs_lz4",
-    attribute_path = "lz4",
+    attribute_path = "lz4.out",
     build_file_content = """
 package(default_visibility = ["//visibility:public"])
 load("@rules_cc//cc:defs.bzl", "cc_library")
 
 cc_library(
   name = "nixpkgs_lz4",
-  srcs = glob(["lib/liblz4.dylib", "lib/liblz4.so*"]),
+  srcs = glob(["lib/liblz4.dylib", "lib/liblz4.so*"], allow_empty = True),
   includes = ["include"],
 )
     """,
--- a/.bazelrc
+++ b/.bazelrc
@@ -12,6 +12,7 @@
 startup --host_jvm_args=-Djdk.tls.client.protocols=TLSv1.2
 
 common --incompatible_require_linker_input_cc_api
+common --incompatible_disallow_empty_glob
 
 # test environment does not propagate locales by default some tests reads files
 # written in UTF8, we need to propagate the correct environment variables, such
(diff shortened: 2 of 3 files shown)


WORKSPACE
```diff
@@ -341,14 +341,14 @@ nixpkgs_package(
...
-    attribute_path = "lz4",
+    attribute_path = "lz4.out",
     build_file_content = """
 package(default_visibility = ["//visibility:public"])
 load("@rules_cc//cc:defs.bzl", "cc_library")
 
 cc_library(
   name = "nixpkgs_lz4",
-  srcs = glob(["lib/liblz4.dylib", "lib/liblz4.so*"]),
+  srcs = glob(["lib/liblz4.dylib", "lib/liblz4.so*"], allow_empty = True),
```
k1nkreet, line 351: It's not clear for me why `allow_empty` is needed here, could you clarify?

avdv: It is because not only the result of the glob needs to be non-empty, but *all* patterns have to match something.

Depending on the OS, there will be `.so` files or `.dylib` but never both.

Naïvely, I tried to use `select()` on the `platform:os`, but since the dict given to select is evaluated eagerly, I ran into an empty glob error again.

k1nkreet (approved):
LGTM

PR #1829, opened 2022-11-04
Fix another config_setting visibility setting.

gregestren:
This seems enough to fully pass rules_haskll on Blaze's incompatible flag CI pipeline:
https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1320#018435e1-7742-44cb-911b-c21b89fb6e95 (I ran that test locally with this patch and the build succeeded).

For https://github.com/bazelbuild/bazel/issues/12933


ylecornec (approved):
Thank you!

PR #1830, opened 2022-11-07
Add support for GHC 9.4.3

GuillaumeGen:


GuillaumeGen:
Not as straightforward as for previous versions.

avdv:
Seems like they changed the layout of the Linux distribution and made it similar to the Windows and Darwin ones...

Maybe you could have a look at #1825 where I began modeling the GHC bindist layout in code, instead of configuring the prefixes for each version. That should make maintenance easier once the dust has settled and GHC uses a single canonical layout for all platforms...

GuillaumeGen:
> GHC uses a single canonical layout for all platforms

Is it something that the GHC development team targets?

I agree that not manually enriching this list of prefix would be better, but I thought that it was the right version to support GHC-9.4, since Stackage started to investigate moving their nightly build to 9.4 one month ago, so 9.4.3 will probably be the first 9.4 version widely used.

GuillaumeGen:
Outside of the small layout modifications, which were quite annoying to investigate, but not hard to fix, there seem to be a new error on Windows, related to `lld` linker not accepting option `-d`.
I don't know what is making this call to `lld` and why it changed with GHC-9.4.3.

avdv:
We added support for GHC 9.4.5 in https://github.com/tweag/rules_haskell/pull/1890, but we also see a linker error on Windows  (see https://github.com/tweag/rules_haskell/issues/1892).

PR #1833, opened 2022-11-23
Refine Continuous Integration workflow

avdv:
- skip writes to Buildbuddy without API key
- run workflow on push requests and on manual dispatch


aherrmann (approved):
> by manually triggering it from Github's UI.

Nice, thank you!

PR #1834, opened 2022-11-25
Replace (most of) custom set library by Skylib's sets.bzl

### LINKED ISSUE #1799: Replace custom "set" library by Skylib
rules_haskell contains a home-grown set library in [`haskell/private/set.bzl`](https://github.com/tweag/rules_haskell/blob/d8afc5739639f01e734b833426d70d10155be9bd/haskell/private/set.bzl). However, it's implementation is very similar to the [one provided by Skylib](https://github.com/bazelbuild/bazel-skylib/blob/main/lib/sets.bzl). We should switch to the Skylib provided one and remove the homegrown version.

evertedsphere:
See #1799.

https://github.com/bazelbuild/bazel-skylib/pull/415 tracks the addition of `mutable_{union,difference}` to Skylib, but Skylib's merge process isn't the fastest. For the moment, I've left our implementations in, in the interest of getting this merged quickly. I'll remove those in a future PR, which will allow us to close the issue linked above.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 *.swp
 *.swo
 __pycache__
+/.direnv
+/.envrc
--- a/haskell/private/path_utils.bzl
+++ b/haskell/private/path_utils.bzl
@@ -1,7 +1,7 @@
 """Utilities for module and path manipulations."""
 
 load("@bazel_skylib//lib:paths.bzl", "paths")
-load(":private/set.bzl", "set")
+load("@bazel_skylib//lib:sets.bzl", "sets")
 
 def is_haskell_extension(extension):
     """Whether the given extension defines a Haskell source file."""
--- a/haskell/repl.bzl
+++ b/haskell/repl.bzl
@@ -34,6 +34,7 @@ load(
 )
 load(":private/java.bzl", "java_interop_info")
 load(":private/set.bzl", "set")
+load("@bazel_skylib//lib:sets.bzl", "sets")
 
 HaskellReplLoadInfo = provider(
     doc = """Haskell REPL target information.
--- a/haskell/c2hs.bzl
+++ b/haskell/c2hs.bzl
@@ -18,6 +18,7 @@ load(
     "target_unique_name",
 )
 load(":private/set.bzl", "set")
+load("@bazel_skylib//lib:sets.bzl", "sets")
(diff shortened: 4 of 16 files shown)


.gitignore
```diff
@@ -3,3 +3,5 @@
 *.swp
 *.swo
 __pycache__
+/.direnv
+/.envrc
```
googleson78, line 7: Would you mind splitting this off to a different commit?

haskell/private/path_utils.bzl
```diff
@@ -1,7 +1,9 @@
 """Utilities for module and path manipulations."""
 
 load("@bazel_skylib//lib:paths.bzl", "paths")
-load(":private/set.bzl", "set")
+
+# load(":private/set.bzl", "set")
```
googleson78, line 5: I think you forgot to delete this line?

haskell/repl.bzl
```diff
@@ -203,7 +204,7 @@ def _create_HaskellReplCollectInfo(target, ctx):
         load_infos[target.label] = HaskellReplLoadInfo(
             source_files = hs_info.source_files,
             boot_files = hs_info.boot_files,
-            import_dirs = set.to_depset(hs_info.import_dirs),
+            import_dirs = depset(sets.to_list(hs_info.import_dirs)),
```
googleson78, line 207: Could this perhaps be another target for upstreaming?

evertedsphere: Good point. I've made a note of it and will add it to my Skylib PR.

googleson78 (changes requested):
I've left some very minor comments that could be addressed.

.gitignore
evertedsphere: Done.

haskell/private/path_utils.bzl
evertedsphere: Good catch.

haskell/repl.bzl
googleson78: If we're planning on upstreaming that, I think the `to_depset` function should then still remain in our internal sets library, so that it's more conveniently replaceable everywhere it's used in the future.

evertedsphere: Done.

googleson78 (approved):

Issue #1832, opened 2022-11-21
splitFileName and minusFileName Variable not in scope causes builds to fail


istathar:
**Describe the bug**
Autogenerated _Paths\__ modules are failing to compile

```
../../../bazel-out/k8-fastbuild/bin/external/stackage/regex-base-0.94.0.2/regex-base-0.94.0.2/build/autogen/Paths_regex_base.hs:48:17: error:
    Variable not in scope: splitFileName :: FilePath -> (a, b0)
   |
48 |   let (dir,_) = splitFileName exePath
   |                 ^^^^^^^^^^^^^

../../../bazel-out/k8-fastbuild/bin/external/stackage/regex-base-0.94.0.2/regex-base-0.94.0.2/build/autogen/Paths_regex_base.hs:49:16: error:
    Variable not in scope: minusFileName :: t0 -> String -> String
   |
49 |   return ((dir `minusFileName` "bin") `joinFileName` dirRel)

```

**To Reproduce**

This started happening when we tried to upgrade from a `nightly` snapshot from mid July to `lts-20.1`. We know we had to downgrade to GHC 9.2.4, that doesn't appear to be a problem. We've upgraded NixPkgs, Bazel, **rules_haskell**, anything we can think of, but the problem keeps reoccuring. It's not just **regex-base** as above; it's also happening with **transformers-compat**, **unliftio-core**, ...

A few months ago this problem cropped up in relation to "relocatable", https://github.com/tweag/rules_haskell/pull/1749#issuecomment-1143713623. The fix there (to not enable relocatable) was reverted. I've since tried that patch again, but still no luck; disabling relocatable makes the bild fail much earlier.

I tried latest version of **rules_haskell** from Git (no difference), tried upgrading **Cabal** (couldn't get that to build), 

**Expected behavior**

Be delighted if we could keep using Bazel to build the Haskell code, but unsure what to try next. Any suggestions would be appreciated!

**Environment**
 - Nix on Linux
 - Bazel version: 5.2.0
 - Version of the rules: both 0.15 and master

avdv:
Hi @istathar, thank you for the bug report!

Unfortunately, this is a known issue and we cannot really do anything about it. Specifically, it is a regression in Cabal introduced in 3.6.0 which was only recently fixed in https://github.com/haskell/cabal/pull/8220

Once there is a new Cabal release, we should look into how we can support using that with existing GHC bindists.

GuillaumeGen:
[...]

PR #1842, opened 2022-12-06
Building the most straightforward tests with GHC Bindist compiled with Hadrian

GuillaumeGen:
This is using a locally built bindist using Hadrian.
This is currently passing 153 of the 206 tests when running `bazel test //... --keep_going`.

The main identified issues for failing tests are:
 - The issue with Cabal dependency calling undeclared `splitFileName` and `minusFileName` as described in #1832 
 - The Hadrian bindist does not distribute a `ghci` binary. As far as I understood, this should be moved to `ghc --interactive` to pass tests related to REPL.
 - Some header files not found like "ghcplatform.h".
 - Some type interfaces changed between GHC 8.10.7 and 9.2.5

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -64,7 +64,7 @@ load(
     "haskell_register_ghc_bindists",
 )
 
-haskell_register_ghc_bindists(version = GHC_VERSION)
+#haskell_register_ghc_bindists(version = GHC_VERSION)
 
 load(
     "@rules_haskell//haskell/asterius:repositories.bzl",


WORKSPACE
```diff
@@ -308,14 +308,18 @@ rules_haskell_asterius_toolchains(
...
+    "@rules_haskell//haskell:ghc_bindist_hadrian.bzl",
+    "haskell_register_ghc_bindists_hadrian",
 )
 
-haskell_register_ghc_bindists(
+haskell_register_ghc_bindists_hadrian(
+    url = "file:///home/guillaume/ExternalPrograms/ghc/_build/bindist/ghc-9.2.5-x86_64-unknown-linux.tar.xz",
+    sha256 = "e27724de38998dd6c3fb46ac5df4cf6818d779c1b749ea1afcd0b64e55b8217e",
+    strip_prefix = "ghc-9.2.5-x86_64-unknown-linux",
+    version = "9.2.5",
```
aherrmann, line 319: Some of the issues are probably unrelated to the hadrian toolchain and instead just due to the version bump. It might make sense to tackle those in a separate PR and [bump the GHC version under test](https://github.com/tweag/rules_haskell/blob/master/constants.bzl#L1).

haskell/ghc_bindist_hadrian.bzl
```diff
@@ -0,0 +1,439 @@
...
+        locale = locale,
+    )
+    local_sh_posix_repo_name = "rules_haskell_sh_posix_local"
+    if local_sh_posix_repo_name not in native.existing_rules():
+        sh_posix_configure(name = local_sh_posix_repo_name)
+    local_python_repo_name = "rules_haskell_python_local"
+    if local_python_repo_name not in native.existing_rules():
+        _configure_python3_toolchain(name = local_python_repo_name)
+
+def _configure_python3_toolchain_impl(repository_ctx):
```
aherrmann, line 319: Fine for development, but we'll want to clean this up and deduplicate with `haskell/ghc_bindist.bzl`.

haskell/repl.bzl
```diff
@@ -483,8 +483,8 @@ def _create_repl(hs, cc, posix, ctx, repl_info, output):
         is_executable = True,
         substitutions = {
             "%{ENV}": render_env(env),
-            "%{TOOL}": hs.tools.ghci.path,
-            "%{ARGS}": "(" + " ".join(
+            "%{TOOL}": hs.tools.ghc.path,
+            "%{ARGS}": "--interactive (" + " ".join(
```
aherrmann, line 487: Does this work for other GHC versions as well? If so, great! If not, we'll have to gate this to not break the other GHC versions.

haskell/ghc_bindist_hadrian.bzl
```diff
@@ -0,0 +1,439 @@
...
+        "copy_filegroups_to_this_package",
+        name = "generated_docdir_filegroup",
+        srcs = [":{}".format(docdir)],
+    )
+
+    generated_include_filegroup = define_rule(
+        "copy_filegroups_to_this_package",
+        name = "generated_include_filegroup",
+        srcs = [":{}".format(docdir)],
+    )
```
aherrmann, line 117: You can move these into [`haskell/toolchain.bzl` next to `_hadrian_bindist_settings`](https://github.com/tweag/rules_haskell/pull/1842/files#diff-f83278238101cead941a012a809724f4fcac6997d0106d70678372f5c533ca84R590) as
```
copy_filegroups_to_this_package(
    name = "generated_bin_filegroup",
    srcs = [":bin"],
)
...
```
and the definition of `copy_filegroups_to_this_package` as well.

Reasons why that may be preferable:
1. The `ghc.BUILD.tpl` is also used by the other bindist toolchains and needs to remain valid for those use cases until we can supersede them with the Hadrian one. But, the `%{generated_*_filegroup}` placeholders are only used for the Hadrian case. Invoking the copy-files rules within `haskell_toolchain` macro means that these placeholders are no longer needed.
2. It's preferable to not mix repository rule and regular rule definitions in the same file, so that changes to the rule definition don't also invalidate the repository rule and cause unnecessary refetching. If the copy-file rule is moved into `haskell_toolchain` then the repository rules and regular rules are separated and the import of `ghc_bindist_hadrian.bzl` in `ghc.BUILD.tpl` is also no longer needed.
3. It puts both concerns into the same place: generating the settings file and placing the bindist files next to it.

WORKSPACE
GuillaumeGen: You are completely right. Separating issues is always beneficial and it will help both review and understanding the cause of the failing tests.

haskell/ghc_bindist_hadrian.bzl
GuillaumeGen: Right.

haskell/repl.bzl
GuillaumeGen: Good question. To investigate.

PR #1848, opened 2022-12-22
CI: Install libtinfo5 which is required by GHC bindists

avdv:
```
external/rules_haskell_ghc_linux_amd64/lib/bin/ghc-pkg: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory
```


aherrmann (approved):
Thank you!

avdv:
Trying to unblock this PR by setting `bes_upload_mode` to `wait_for_upload_complete` and adding a timeout.

PR #1858, opened 2023-01-06
Avoid :reload in the GHCi script

### LINKED ISSUE #1851: REPL modules are loaded twice after switching to GHC 9.4.3
**Describe the bug**
After updating our stack from GHC 9.2.5 to GHC 9.4.3, ghcid reloads are noticeably longer.
We are using a target using `haskell_repl` to load into ghcid.
I noticed that when running our REPL target like `bazel run :repl`, modules are loaded twice.
We have our modules loading, then `Ok, 145 modules loaded.` and then they all load again a second time.
Using `:r` in ghci then doesn't seem to have this issue even when updating a source file.

**To Reproduce**
Use GHC 9.4.3 on a project.

**Expected behavior**
Entering the REPL should only load modules once.

**Environment**
 - OS name + version: **NixOS 22.11**
 - Bazel version: **4.2.2**
 - Version of the rules: **0.16**

aherrmann:
Closes #1851

Since GHC version 9.4.1 a `:reload` in the GHCi script triggers a secondary recompilation of the modules, effectively doubling GHCi start-up times.

The `:reload` is required to clear the bindings of the start-up code that runs to load the modules into scope using the `:module` command.

This logic was needed when rules_haskell didn't know the module names for modules included in Haskell targets. However, rules_haskell has changed such that it does now know the module names, at least for haskell_library, haskell_binary, and haskell_test targets, which are the only ones relevant for loading sources into the REPL.

This commit changes the GHCi startup script to directly load the modules based on the module names tracked by rules_haskell. The custom logic in the startup script is no longer needed, and the `:reload` is no longer needed since no more bindings need to be cleaned up.


mrkkrp (approved):
This is much cleaner. I remember I did not follow the path of inferring module names because of the complications with the main function/module, which can be pretty hairy. I see that you handle that in `determine_module_names` now.

Issue #1855, opened 2023-01-05
Upgrade supported Bazel version to 5.x and 6.x


avdv:
**Is your feature request related to a problem? Please describe.**

There have been some problems using Bazel 4 lately (e.g. https://github.com/tweag/rules_haskell/pull/1848#issuecomment-1362855246) which have been fixed in Bazel 5.

Also, [Bazel 6](https://github.com/bazelbuild/bazel/releases/tag/6.0.0) has been released a couple of weeks ago and we want to support that.

**Describe the solution you'd like**

- [x] update Bazel version to 5.x on CI (#1849)
- [x] adapt minimum Bazel version to 5.0 in `haskell/private/versions.bzl` and in the `start` script
- [x] add proper support for Bazel 6 (related: #1853)
- [x] update changelog 
- [x] go through the release notes and see if we can get rid of any `--incompatible_*` or `--experimental_*` flags that have been flipped (e.g. `--incompatible_disallow_empty_glob` could be set to `true` now (it's the default in Bazel 6).

**Additional context**

avdv:
Fixed by #1860

Issue #1853, opened 2023-01-04
Building documentation fails with Bazel 6


ylecornec:
Building documentation targets from the `docs` directory with Bazel 6 fails with the following error:

```
./external/bazel_tools/tools/cpp/cc_configure.bzl:194:26: name 'module_extension' is not defined
com.google.devtools.build.skydoc.SkydocMain$StarlarkEvaluationException: name 'module_extension' is not defined
	at com.google.devtools.build.skydoc.SkydocMain.recursiveEval(SkydocMain.java:393)
	at com.google.devtools.build.skydoc.SkydocMain.recursiveEval(SkydocMain.java:402)
	at com.google.devtools.build.skydoc.SkydocMain.recursiveEval(SkydocMain.java:402)
	at com.google.devtools.build.skydoc.SkydocMain.eval(SkydocMain.java:260)
	at com.google.devtools.build.skydoc.SkydocMain.main(SkydocMain.java:163)
Stardoc documentation generation failed: name 'module_extension' is not defined
```

It seems that stardoc fails to parse files from `@bazel_tools` which use the recent `module_extension` feature.

See this similar issue: https://github.com/bazelbuild/stardoc/issues/123

PR #1860, opened 2023-01-11
Bazel 6 support

ylecornec:
This PR addresses the remaining items of issue #1855
- Stardoc is updated because of issue #1853
- Commit https://github.com/tweag/rules_haskell/pull/1860/commits/3be8027a8b7cc517b0ee725347c0d9b7ba84420e  adds a missing import to the `cabal_wrapper` python rule. It seems that things were working fine before due to [this issue](https://github.com/bazelbuild/rules_python/issues/382) which is fixed when using bazel6 + python 3.11
 
I did not try to replace the `crosstool_top` option yet as it seems to works fine with bazel 6, so replacing it can be done in another PR.


avdv (approved):
Thank you!

PR #1863, opened 2023-01-27
Declare cc toolchain for windows

ylecornec:
This PR declares the `cc_toolchain` used for windows, so that it can be found when using the `--incompatible_enable_cc_toolchain_resolution` option.
 As a result using the `crosstool_top` option is not required anymore.


aherrmann (approved):
Thank you! Good work!

PR #1868, opened 2023-02-15
Prepare ghc update

ylecornec:
This PR contains changes that prepare for the ghc update but should also work with the current version.

- It recovers include directories for `c2hs` via `ghc-pkg field rts include-dirs` instead of relying on `ghc --print-libdir`.
This should be more stable across ghc versions: https://gitlab.haskell.org/ghc/ghc/-/issues/21609#note_435800

Because of the path and command length limits on windows:

- Change the CI `output_user_root` on windows to a shorter path (this requires bumping the `cache-silo-key` value to invalidate the remote cache).

- When compiling a cabal package on windows, create a temporary package database in which we declare all the dependencies before invoking cabal. This generates shorter command lines than passing the databases of dependencies as arguments, both when calling cabal, and then when cabal calls ghc.


aherrmann (approved):
Thank you! Looks good!

PR #1873, opened 2023-03-03
Skip writing to the remote cache with no API key

avdv:
Related to https://github.com/tweag/rules_haskell/pull/1833 -- there was one more use of the empty API key.

For PRs from forks (and also dependabot) access to the repository secrets is not permitted, so the API key for Buildbuddy ends up as empty.

This would lead to an error:
```
WARNING: BES was not properly closed
ERROR: The Build Event Protocol upload failed: Not retrying publishBuildEvents, no more attempts left: status='Status{code=UNAUTHENTICATED, description=failed to parse API key
```


ylecornec (approved):
It looks good to me :+1:

PR #1884, opened 2023-03-24
build_bazel_rules_nodejs -> rules_js

ylecornec:
This PR replaces `build_bazel_rules_nodejs` which is does not support bzlmod and is deprecated in favor of `rules_js`.
Most notable changes include:

### Workspace rules
  - rules_js uses `pnpm` instead of `yarn`
      - `yarn.lock` is replaced by the `pnpm-lock.yaml` file
      - The `asterius_dependencies_*` repository rules which can setup the `nodejs` environment were adapted.
      - Since we do not use the `yarn_install` repo rule anymore, we do not need to declare multiple identical repos for the nix nodejs toolchain (see [here](https://github.com/tweag/rules_haskell/pull/1884/files#diff-9d21ccabe299eeb98999e18b704133d701be776e90d771eb972751729b9878f2L96-L113)).
  - WebAssembly version was updated to `5.76.0` (which superseeds https://github.com/tweag/rules_haskell/pull/1875)

### asterius_binary and asterius_test
The underlying `nodejs_{binary,test}` rules where switched to `js_{binary,test}`.

The interface changed a bit because these rules (with the `copy_data_to_bin` option) copy their inputs to the output tree.

So the entrypoint file is now a predeclared output of the `ahc_dist` rule. This way it is recognised as already in the output tree when passed as input to `js_{binary,test}`.

### asterius_webpack 
While `asterius_webpack` used to be a custom rule, we now rely on the `bin.webpack_cli` rule provided by rules_nodejs, with a wrapper around it in order to transition to the configuration where WebAssembly targets are built for the browser.

### The `asterius_binary_from_genrule` test
While running a `js_binary` with `bazel run` works, it is not trivial to use it as a build action in a `genrule` when it has runtime dependencies. The runfiles tree which for now exists in the `exec` configuration would likely need to be moved to the `BINDIR` where the action is executed.
Since the handling of WebAssembly may need significant refactoring when switching from asterius to ghc's wasm backend in the future, this test is disabled and left for future work.


avdv (approved):
Nice work, thank you!

PR #1886, opened 2023-04-12
Refactor WORKSPACE file

ylecornec:
Depends on #1884

This PR refactors the WORKSPACE file by declaring most external repositories definitions to `non_modules_deps_*.bzl` files,
so they can be used later with `bzlmod` through module extensions (See [here](https://github.com/tweag/rules_nixpkgs/blob/abfcb76ba2b865e41b994c37ab8444f6fb58dc2a/testing/rust/MODULE.bazel#L39-L42) for an example in `rules_nixpkgs`).

- A `setup_stack` option was added to `stack_snapshot` to optionally disable the installation of stack. Without bzlmod we rely on `native.existing_rule` to detect whether the stack repo is already set up, but this method is not compatible with bzlmod (See [here](https://github.com/bazelbuild/bazel/issues/15377)). This option should only be necessary until a `stack_snapshot` module extension is implemented for bzlmod.

- Some more repository rules are updated with an extra `register` parameter to optionally disable the registration of toolchains, because trying to do so fails in bzlmod.


avdv (approved):
Nice work, thank you!

PR #1887, opened 2023-04-24
stack_snapshot bzlmod support

ylecornec:
This PR fixes some labels in files generated by the `stack_snapshot` rule under bzlmod.

- With bzlmod, `repository_ctx.name` contains the canonical name of the repository and not the `name` that was passed to the repository rule. So a new `unmangled_repo_name` string attribute was added as a workaround to remember this name.

- Starting from bazel 6, `str` serializes a label to its canonical name, which is needed for bzlmod support (and can even replace the  `_label_to_string` function in non bzlmod). However on bazel 5, we continue to use the old `_label_to_string` function and do not try to support bzlmod (since it is experimental there and still evolving).

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -42,6 +42,7 @@ load(
     "get_library_files",
     "haskell_cc_libraries_aspect",
 )
+load(":private/versions.bzl", "check_bazel_version")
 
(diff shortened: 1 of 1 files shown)


avdv (approved):
Thanks!

PR #1888, opened 2023-04-28
bzlmod python fix

ylecornec:
This PR:
- Fixes one more python `import` statement, which refers to a repository name which changes with bzlmod.
- Makes use of raw strings for regular expressions in some python scripts, to fix `invalid escape sequence` warnings.


avdv (approved):
Thank you!

PR #1890, opened 2023-05-03
Add GHC bindist version 9.4.5 and 9.6.1

### LINKED ISSUE #1889: Binary distribution of GHC 9.6.1 not available
The latest version of GHC is 9.6.1 and is the version I want to use. But it appears that rules_haskell cannot download the binary.

avdv:
This adds GHC bindists for the latest stable GHC releases for 9.4 and 9.6. Fixes #1889 

Notable changes are that starting with GHC 9.6.1 the bindist packages are packaged with Hadrian and thus are relocatable without patching the scripts in `bin`.

Note that the test-patches workflow has been run for these successfully, except for GHC 9.4.5 on Windows. See https://github.com/tweag/rules_haskell/issues/1892  For the time being, until this gets resolved, I have excluded this version from the `test-ghc-patches` workflow.

### DIFF
--- a/haskell/ghc_bindist.bzl
+++ b/haskell/ghc_bindist.bzl
@@ -407,6 +407,8 @@ def ghc_bindist(
             "9.2.3": ["@rules_haskell//haskell:assets/ghc_9_2_3_win.patch"],
             "9.2.4": ["@rules_haskell//haskell:assets/ghc_9_2_4_win.patch"],
             "9.2.5": ["@rules_haskell//haskell:assets/ghc_9_2_5_win.patch"],
+            "9.4.5": ["@rules_haskell//haskell:assets/ghc_9_4_5_win.patch"],
+            "9.6.1": ["@rules_haskell//haskell:assets/ghc_9_6_1_win.patch"],
         }.get(version)
 
     if target == "darwin_amd64":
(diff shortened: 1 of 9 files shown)


haskell/ghc_bindist.bzl
```diff
@@ -388,7 +411,9 @@ def ghc_bindist(
             "9.2.3": ["@rules_haskell//haskell:assets/ghc_9_2_3_win.patch"],
             "9.2.4": ["@rules_haskell//haskell:assets/ghc_9_2_4_win.patch"],
             "9.2.5": ["@rules_haskell//haskell:assets/ghc_9_2_5_win.patch"],
-        }.get(version)
+            "9.4.5": ["@rules_haskell//haskell:assets/ghc_9_4_5_win.patch"],
+            "9.6.1": ["@rules_haskell//haskell:assets/ghc_9_6_1_win.patch"],
+        }.get(version, "".format())
```
ylecornec, line 416: Is the default value `"".format()` here on purpose ?

ylecornec (approved):
Thanks! I have a small comment but otherwise it looks good to me :+1:

haskell/ghc_bindist.bzl
avdv: Good catch! I started to refactor this, but then decided against it. I'll fix it.

PR #1903, opened 2023-05-26
Move tests to `rules_haskell_test` module

ylecornec:
To prepare for bzlmod support, this PR moves most tests from rules_haskell to a dedicated module called `rules_haskell_tests`.

Testing `rules_haskell` from another module is more realistic (which is particularly important with bzlmod since different modules have different visibility and repository mappings). In addition, once bzlmod is activated, the `rules_haskell_test` module will be able to depend on the `rules_haskell_nix` module which will be used to register the the nix toolchains (to avoid a dependency between `rules_haskell` and `rules_nixpkgs`).

#### Some comments
- Some targets are still present in rules_haskell's test folder for tests that do not test the public api but are internal to the module such as the `shellcheck` ones or `//tests/package_configuration:package_configuration_test`.

- Dependencies from the WORKSPACE and the `non_module_deps_*` files are split between the two modules. And the dependencies from `non_module_dev_dep_*` files will be development dependencies with bzlmod.

#### About the CI

- The CI was updated so that tests for `rules_haskell` and `rules_haskell_tests` happen in parallel. So there are more jobs but the tests for the `rules_haskell` module should be quick since most tests are now in `rules_haskell_tests`.

- The config was also updated so that `rules_haskell` and `rules_haskell_test` use separate caches on windows. This works around `undeclared inclusions` issues such as [this one](https://github.com/tweag/rules_haskell/actions/runs/5092117961/jobs/9153045522#step:8:148).
This can happen because the `*.d` files (that keep track of C dependencies for hermeticity purposes) such as the one file from this [cache hit](https://app.buildbuddy.io/invocation/0642ff72-9b53-4551-b5cb-d4f750f03357?actionDigest=6c28ad4feed4e90e7779186799948c703d068c909061a90c1b0f922e02ec1752#action) contain absolute paths that are not the same for both module and should not be shared.

- The mergify configuration was [updated](https://github.com/tweag/rules_haskell/pull/1903/files#diff-6a30e47dd51449847422017b16878889b848f6602ac56d5a2e67ce8a19afade2) but I think this is not effective until the PR is merged, so it will have to be merged manually.


avdv (approved):
Thank you, LGTM!

Could you please adjust the README to include a few words on how to run tests now?

BTW, for rules_sh we added the test module (as a local repository) to the root WORKSPACE too, so you could run all tests with `bazel test //... @rules_sh_tests//...`. Maybe this could be done here too, making it a bit easier to run the tests? What do you think?

aherrmann:
> BTW, for rules_sh we added the test module (as a local repository) to the root WORKSPACE too, so you could run all tests with `bazel test //... @rules_sh_tests//...`. Maybe this could be done here too, making it a bit easier to run the tests? What do you think?

I would not spend much effort in that direction. WORKSPACE are not loaded transitively, so testing `bazel test @rules_sh_test//...` is quite different from `cd test && bazel test ...`. It's also not applicable to the bzlmod world since library modules are not allowed to use local overrides.

PR #1905, opened 2023-06-08
pre bzlmod refactoring

ylecornec:
Depends on #1903

This PR contains some more refactoring in preparation for bzlmod.
- 84be61104a59502998bcc722375bfad0066d6c65 adds a `rules_haskell_dependencies_bzlmod` function that only contains dependencies that are not available as modules. So their definition is shared between bzlmod and workspace modes.
- f5d8e9fdece4b1fe1fc1de7d9254187256edf2f8 moves dependencies that are not used by bzlmod after all, from `non_module_deps` files directly to the WORKSPACE file.
- Because buildifier is a development dependency of rules_haskell, buildifier targets should not be in the same package as public ones. So they are moved to dedicated `buildifier` packages.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,5 +1,3 @@
-load("@com_github_bazelbuild_buildtools//buildifier:def.bzl", "buildifier", "buildifier_test")
-
 exports_files(["start"])
 
 exports_files(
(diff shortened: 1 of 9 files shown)


avdv (approved):
This is progressing nicely, well done!

Issue #1899, opened 2023-05-10
Add a module extension for Stackage packages


ylecornec:
- [x] https://github.com/tweag/rules_haskell/pull/1917
- [x] See comment: https://github.com/tweag/rules_haskell/pull/1917#discussion_r1247942790

PR #1906, opened 2023-06-09
Make stack_snapshot usable from another's module extension

ylecornec:
Depends on #1905

This PR makes the `stack_snapshot` repository rule usable from a `non_module_dep` extension of another module than `rules_haskell` (such as `rules_haskell_tests`).

These changes are needed because with bzlmod, when `stack_snapshot` is called by an extension of another module, rules_haskell does not have visibility of generated repositories (such as `rules_haskell_stack`) and cannot build labels to its targets using the `Label` constructor.

- To work around this, an optional `label_builder` parameter can be passed to `stack_snapshot`, which can be used to build labels from the context of the caller modules extension and obtain their canonical names. Once we provide our own module extension to install stackage packages (#1899), this `label_builder` parameter should not be necessary anymore.

- For the same reason, the `stack_update` attribute to `_stack_snapshot` is now always present (so we do not have to hardcode its label), and is made into a string so that we only fetch the `stack_update` repository when explicitly reading from it.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -1536,8 +1536,7 @@ def _download_packages(repository_ctx, snapshot, pinned):
 
     if stack_unpack:
         # Enforce dependency on stack_update.
-        # Hard-coded label since `stack_update` is `None` in this case.
-        repository_ctx.read(Label("@rules_haskell_stack_update//:stack_update"))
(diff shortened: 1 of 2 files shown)


avdv (approved):
Thank you.

PR #1911, opened 2023-06-19
Use bazel 6 in the CI

### LINKED ISSUE #1846: Flag --incompatible_disable_starlark_host_transitions will break rules_haskell in Bazel 7.0
Incompatible flag `--incompatible_disable_starlark_host_transitions` will be enabled by default in the next major release (Bazel 7.0), thus breaking rules_haskell. Please migrate to fix this and unblock the flip of this flag.

The flag is documented here: [bazelbuild/bazel#17032](https://github.com/bazelbuild/bazel/issues/17032).

Please check the following CI builds for build and test results:

- [Ubuntu 18.04 OpenJDK 11](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1365#0185154b-0509-4312-8b28-ee73057e3232)

Never heard of incompatible flags before? We have [documentation](https://docs.bazel.build/versions/master/backward-compatibility.html) that explains everything.
If you have any questions, please file an issue in https://github.com/bazelbuild/continuous-integration.

ylecornec:
Depends on #1909.

This PR updates the CI to bazel version `6.2.1`
This is necessary to:
  - update `rules_go` to a recent version and activate the `--incompatible_disable_starlark_host_transitions` to hopefully resolve #1846.
  - run the bzlmod tests in the future.

Some notes:
- `nixpkg` is updated to a version that contains bazel `6.2.1`.
- The default go version for this version of nixpkgs is `1.20`, which requires updating `rules_go` and rules_nixpkgs (to a version containing [this fix](https://github.com/tweag/rules_nixpkgs/pull/385)).
- With the nix setup, we now use the go toolchain from [rules_go_configure](https://github.com/tweag/rules_haskell/blob/master/non_module_dev_deps.bzl#L120C5-L125) instead of the `host` one. This is done in preparation for bzlmod  because it is not possible to use `is_nix_shell` in the `MODULE.bazel` file.


avdv (approved):
Thanks!

rules_haskell_tests/WORKSPACE
```diff
@@ -48,13 +48,12 @@ rules_proto_dependencies()
 rules_proto_toolchains()
 
 # For buildifier
-# starting from 0.29, rules_go requires bazel >= 4.2.0
 http_archive(
```
avdv, line 51: @ylecornec do you know which Bazel version rules_go 0.39.1 requires? AFAIK, 0.35 requires at least Bazel 5.1, so we might want to note somewhere that we require at least version x.y for developing on this repo? Or just use a `.bazelversion` file?

ylecornec: Ah indeed, it requires at least Bazel `5.3`. 

Release [0.38](https://github.com/bazelbuild/rules_go/releases/tag/v0.38.0) is the one with the fix for `incompatible_disable_starlark_host_transitions` and requiring this version.

PR #1913, opened 2023-06-27
Make the flags passed to Stack match Cabal

iteratee:
Cabal uses '+' to enable a flag, and '-' to disable a flag. Stack builds a dictionary that has either 'flag: true', or 'flag: false'. The behavior before was passing inconsistent flags to the different tools. Have cabal.bzl handle all the flags that we could pass to cabal. This helps to aid people new to rules_haskell. It also allows negating a flag, which didn't work before.


avdv:
@mergifyio rebase

PR #1914, opened 2023-06-27
haskell_toolchains extension

ylecornec:
Depends on #1912

This PR adds the `haskell_toolchains` module extension which is used to configure and register the haskell bindists, via the `bindist` and `bindists` tags.

- The `bindists` tag corresponds to `haskell_register_ghc_bindists` and registers toolchains for all platforms.
- The `bindist` tag corresponds to `ghc_bindist` and is used to register toolchains one at a time. Inside the same module, the toolchains declared with `bindist` take precedence over the `bindists` ones.

The `rules_haskell` module calls the `bindists` tag itself to register default toolchains that can be used without any configuration. Alternatively, we can register customized toolchains, as done by the `rules_haskell_test` [module](https://github.com/tweag/rules_haskell/pull/1914/files?diff=unified&w=0#diff-c385cb2f1e4a4c4652191639dc4ccb1c8e048d9ebc33d61b12dae27c57b45fccR338-R363) (the priority of toolchains respects bazel's iteration order over modules where the root module comes first).

#### We cannot use load in the `MODULE.bazel` files at the moment
The configuration variables such as `test_repl_ghci_args` are duplicated inside the [MODULE.bazel](https://github.com/tweag/rules_haskell/pull/1914/files?diff=unified&w=0#diff-c385cb2f1e4a4c4652191639dc4ccb1c8e048d9ebc33d61b12dae27c57b45fccR302-R336) file because it is not possible to `load` them at the moment (see https://github.com/bazelbuild/bazel/issues/17880). For the same reason it is not possible to rely on `is_windows` to configure the `cabalops` variable so the windows toolchains are declared separately in this file.
 
#### Documentation
The documentation is not generated because stardoc does not support it yet (https://github.com/bazelbuild/stardoc/issues/123).

#### Toolchain declaration
It is not possible to register aliases to toolchain with bzlmod yet (https://github.com/bazelbuild/bazel/issues/16298), so `ghc_bindist.bzl` was refactored a bit to make `ghc_bindist_toolchain_declaration` public.  This way the `haskell_toolchains` module extension can call it to build its own toolchain declarations.


rules_haskell_tests/MODULE.bazel
```diff
@@ -308,3 +293,71 @@ use_repo(
...
+]
+
+cabalopts_windows = test_cabalopts + [
+    # To avoid ghcide linking errors with heapsize on Windows of the form
+    #
+    #   unknown symbol `heap_view_closurePtrs'
+    #
+    # See https://github.com/haskell/ghcide/pull/954
+    "--disable-library-for-ghci",
+]
```
aherrmann, line 336: > The configuration variables such as `test_repl_ghci_args` are duplicated inside the [MODULE.bazel](https://github.com/tweag/rules_haskell/pull/1914/files?diff=unified&w=0#diff-c385cb2f1e4a4c4652191639dc4ccb1c8e048d9ebc33d61b12dae27c57b45fccR302-R336) file because it is not possible to `load` them at the moment (see [bazelbuild/bazel#17880](https://github.com/bazelbuild/bazel/issues/17880)). For the same reason it is not possible to rely on `is_windows` to configure the `cabalops` variable so the windows toolchains are declared separately in this file.

Perhaps another approach could be for the module extension/tags to accept a label to a JSON file holding this configuration? (Either way, out of scope for this PR)

MODULE.bazel
```diff
@@ -83,6 +80,33 @@ use_repo(
...
+register_toolchains(
+    "@rules_haskell_ghc_nixpkgs_toolchain//:toolchain",
+    dev_dependency = True,
+)
+
+register_toolchains(
+    "@all_bindist_toolchains//:all",
+    "@rules_haskell_python_local//:toolchain",
+    "@rules_haskell_sh_posix_local//:all",
+)
```
aherrmann, line 108: Happy to stick to this for this PR.
In the long run, similarly to https://github.com/tweag/rules_haskell/issues/1915, I think the default POSIX toolchain should be something that rules_sh already registers in its own MODULE.bazel file.
In other words, under bzlmod rules_haskell should only register Haskell toolchains itself (ignoring dev dependencies). Other toolchains should be registered by other modules. (rules_haskell_test can of course still register whatever is needed for testing).

extensions/haskell_toolchains.bzl
```diff
@@ -0,0 +1,165 @@
...
+
+_all_toolchains = repository_rule(
+    implementation = _all_toolchains_impl,
+    attrs = {
+        "toolchains": attr.string_list(
+            doc = "All the `toolchain(...)` declarations for the bindist toolchains as string",
+        ),
+    },
+    doc = """ Used to generate the `all_bindist_toolchains` external repository.
+    We can then invoque `register_toolchains("@all_bindist_toolchains//:all")` in the MODULE.bazel file.
```
aherrmann, line 82: typo `invoque --> invoke`.

extensions/haskell_toolchains.bzl
```diff
@@ -0,0 +1,165 @@
...
+    doc = """See [rules_haskell_toolchains](toolchain.html#rules_haskell_toolchains).
+    Only the first `bindists` tag is taken into account (according to the iteration order over modules).
+    """,
+)
+
+_bindist_tag = tag_class(
+    attrs = {
+        "name": attr.string(
+            mandatory = True,
+            doc = "A unique name for the repository",
```
aherrmann, line 41: I think this needs a bit more explanation. As I understand it, it's only a unique name for the tag within the calling module. I.e. it doesn't have to be a globally unique name, correct?

extensions/haskell_toolchains.bzl
```diff
@@ -0,0 +1,165 @@
...
+    # The code could be simplified a bit by registering aliases to the
+    # `toolchain` rules, once the following issue is resolved
+    # https://github.com/bazelbuild/bazel/issues/16298
+    toolchain_declarations = []
+
+    toolchain_names = []
+    found_bindists = False
+    for module in mctx.modules:
+        for bindist_tag in module.tags.bindist:
+            name = "bindist_{}_{}_{}".format(module.name, module.version, bindist_tag.name)
```
aherrmann, line 99: Is `bindist_tag.name` actually needed, or could it be replaced by an autogenerated unique suffix (e.g. a counter, or perhaps [a `hash`](https://bazel.build/rules/lib/globals/all#hash) plus collision counter to reduce invalidation)? Asking in the spirit of [this discussion](https://github.com/bazel-contrib/rules-template/pull/46#discussion_r1097502233). Since the name is scoped to within the calling MODULE and not global it doesn't seem too onerous, but, maybe it can be avoided altogether.

extensions/haskell_toolchains.bzl
```diff
@@ -0,0 +1,165 @@
...
+            fail(
+                """Module "{module}~{version}" used the "bindists" tag more than once.""".format(
+                    module = module.name,
+                    version = module.version,
+                ),
+            )
+        for bindists_tag in module.tags.bindists:
+            # We only consider the first `bindists` tag, because subsequent
+            # ones would have the same constraints and lower priority.
+            if not found_bindists:
```
aherrmann, line 139: Could this be simplified to something along the lines of the following?
```
if module.tags.bindists:
    bindists_tag = module.tags.bindists[0]
    ...
```
I.e. drop the `found_bindists` boolean.

aherrmann (approved):
Good work, that looks great!

Only a couple small comments.

Btw, I really appreciate how easy this PR and #1912 are to review commit by commit. Thank you!

extensions/haskell_toolchains.bzl
ylecornec: Ah indeed I think we can do without the name attribute (and without auto-generated suffixes).

I updated the code to rely on the `target` attribute instead.

Before, if two `bindist` tags shared the same `target` attribute, the toolchains would have the same constraints so the second one would be effectively ignored. So now, a module is not allowed to use two `bindist` tags with the same `target`, and `(module.name, module.version, bindist_tag.target)` can be used as an identifier.

MODULE.bazel
ylecornec: I removed the registration of `rules_haskell_sh_posix_local` toolchain and the one from `rules_sh` is used instead :+1: 

The last version of `rules_python` also registers a default toolchain but is does not work when bazel is run as root (see [this issue](https://github.com/bazelbuild/rules_python/issues/1169)), so I kept the `rules_haskell_python_local` for now.

PR #1916, opened 2023-06-29
Bazel ci fixes

ylecornec:
Depends on #1912

This PR contains two changes to bazel's CI:

- It updates `.bazelrc.common` to only use BuildBuddy for the build event service in our CI, but not in bazel's one.
  This fixes [this error](https://buildkite.com/bazel/rules-haskell-haskell/builds/1548#0188d62b-0887-4d42-8ff2-184d0b246e41/298-436) from the `Try update last green commit` job, which also contains the following warning at the top:

    ```
     '--config=ci-common' (source command line options) was expanded and now overrides the explicit option --bes_backend=buildeventservice.googleapis.com with --bes_backend=grpcs://remote.buildbuddy.io
    ```
- It adds the tests from the `rules_haskell_tests` module to bazel's CI.
  I needed to specify the platform explicitly when using a matrix because of [this error](https://buildkite.com/bazel/rules-haskell-haskell/builds/1573#0189062b-de32-4c1b-8fcb-32e6d0c59814/104-116) which looks a bit like an issue in bazel's ci. It looks like the default platform gets expanded to `ubuntu1804_config_01` [here](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/bazelci.py#L1056), which does not exist in the `PLATFORM` dictionary.

The CI was run manually against this commit [here](https://buildkite.com/bazel/rules-haskell-haskell/builds/1580#01890647-935a-4bc5-a78c-07fb0b35509c).

Note: updating the latest green commit should make bazel's CI run against master in order to be sure that #1846 is fixed.


.bazelrc.common
```diff
@@ -58,8 +60,6 @@ build:ci-common --repository_cache=~/repo-cache/
 build:ci-common --experimental_repository_cache_hardlinks
 
 # Use a remote cache during CI
-build:ci-common --bes_results_url=https://app.buildbuddy.io/invocation/
-build:ci-common --bes_backend=grpcs://remote.buildbuddy.io
```
avdv, line 62: Maybe it would make sense to still leverage the cache in general, but disable remote uploads with `--noremote_upload_local_results`, and re-enable remote uploads for the "ci" config with `--remote_upload_local_results`. Not sure this is possible.

avdv (approved):
Thanks, LGTM, but I left a comment.

PR #1917, opened 2023-06-30
stack_snapshot module extension

ylecornec:
This PR adds the `stack_snapshot.bzl` extension, which creates a `stackage` repository that can be shared between the different modules. It should also be possible to declare the extension as [isolated](https://github.com/bazelbuild/bazel/pull/18529) in the next release of Bazel, when we do not want to share the repository.

#### The package tag
- Packages are declare and configured with the same `package` tag (for instance [here](https://github.com/tweag/rules_haskell/compare/ylecornec/stack_snapshot?expand=1#diff-c385cb2f1e4a4c4652191639dc4ccb1c8e048d9ebc33d61b12dae27c57b45fccR454-R468)). So this tag has a `hidden` attribute if we only want to configure it but not make it visible.
- Only the `rules_haskell` and `root` modules can configure a package (with the root module taking precedence) to not make the configuration too complex, so for other modules, we only remember that they use a package (and set the `components` and `component_dependencies` fields if present). 

#### The configuration tags
- `snapshot` which is used to configure the snapshot using a `name` or a `local_snapshot` is set to a default value by the `rules_haskell` module, but can be overridden by the root module.
- `stack_snapshot_json` can only be set by the root module.
    - `stack_snapshot_json` has a new `os` argument to add OS contraints, because the lockfiles are different on windows and we cannot use `is_windows` in the `MODULE.bazel` file as we do in the `WORSPACE` file. See [here](https://github.com/tweag/rules_haskell/compare/ylecornec/stack_snapshot?expand=1#diff-c385cb2f1e4a4c4652191639dc4ccb1c8e048d9ebc33d61b12dae27c57b45fccR483-R489) for example.
    - If they are multiple `stack_snapshot_json` the first one matching the os constraints will be selected.
    - If the `os` attribute is empty, it means the json file is compatible with all os (so it can be used last as a default case).

- `verbose`, `haddock`,`netrc`, `tools`, `stack`, `haddock`can only be set by the root module if needed, to change the default values.

### Tests
The `stackage` repository is now defined in the `MODULE.bazel` files and shared between `rules_haskell` and `rules_haskell_tests` (however it was only a dev dependency of rules_haskell so no real gain yet).
The `rules_haskell` module only declares the packages it needs, and `rules_haskell_tests` adds the rest.


extensions/stack_snapshot.bzl
```diff
@@ -0,0 +1,295 @@
...
+        # Check that a package is configure at most one time per module
+        if sets.contains(packages_in_module, package_name):
+            fail("""Module "{module_name}~{module_version}" configured package "{package_name}" multiple times.""".format(
+                module_version = module.version,
+                module_name = module.name,
+                package_name = package_name,
+            ))
+        sets.insert(packages_in_module, package_name)
+
+        # If the package is already configured, we can only only make it visible.
```
aherrmann, line 160: typo: `only only --> only`.

extensions/stack_snapshot.bzl
```diff
@@ -0,0 +1,295 @@
...
+        if root_or_rules_haskell:
+            # Only the root module or rules_haskell can modify the
+            # following configuration settings. To make the configuration
+            # easier to understand.
+            if package_tag.setup_deps:
+                conf.setup_deps[package_name] = package_tag.setup_deps
+            if package_tag.flags:
+                conf.flags[package_name] = package_tag.flags
+            if package_tag.vendored:
+                conf.vendored_packages[package_name] = package_tag.vendored
```
aherrmann, line 186: I'm happy to keep it this way in this PR. For a follow-up it's worth considering whether these should raise an error on non-root/non-rules_haskell modules. Silently ignoring tag properties could be quite confusing to unsuspecting users.

extensions/stack_snapshot.bzl
```diff
@@ -0,0 +1,295 @@
...
+            # rules_haskell can also set the snapshot to be used as a default.
+            if module.tags.snapshot:
+                _assert_unique_tag(module.tags.snapshot, "snapshot", module)
+                snapshot_tag = module.tags.snapshot[0]
+                if "snapshot" not in kwargs and "local_snapshot" not in kwargs:
+                    if snapshot_tag.local_snapshot:
+                        kwargs["local_snapshot"] = snapshot_tag.local_snapshot
+                    if snapshot_tag.name:
+                        kwargs["snapshot_tag"] = snapshot_tag.name
+        if module == root_module:
```
aherrmann, line 209: Similarly, worth considering whether we should raise an error if a non-root module invokes these tags.

aherrmann (approved):
Thank you, that looks good!

PR #1918, opened 2023-06-30
Nix-based haskell toolchains

ylecornec:
This PR adds the `nix_haskell_toolchains` extension which can be used in bzlmod to register nix-based toolchains.

As a first step, this extension provides a `new` tag with the same interface as the existing `haskell_register_ghc_nixpkgs`.
However, we may want to change some parts of this interface in the future, in particular the attributes related to the posix toolchain could possibly be replaced by an extension from `rules_haskell_posix` once one is available.

### The `declare_toolchain` extension 
The `nix_haskell_toolchains` extension makes calls to `nixpkgs_package` according to the tags used by the different modules. Because of this, we cannot call `use_repo` in the `MODULE file` to make all of these visible to `rules_haskell_nix`, in order to resolve the paths to the `ghc` binaries which are needed later by the `_ghc_nixpkgs_haskell_toolchain` repository rule.

These repositories are however visible from the `nix_haskell_toolchains_configurations` repository which is also generated by the `nix_haskell_toolchains` extension. So we use this repository to resolve the labels to the `ghc` binaries, and it is loaded by the `declare_toolchains` extension, which sets up and generates the `all_nix_toolchains` repository.

### The `nixpkgs.bzl` file was refactored
  -  to separate the `nixpkgs_package` calls from the generation of the other toolchain repositories.
  - to make the `declare_toolchain` extension able to generate all the toolchain declarations in the same `BUILD` file (similarly to the bindist case).

### Toolchains
- The `rules_haskell_nix` module registers a default toolchain in it's `MODULE.bazel` file.
- The `rules_hakell_tests` module also registers a toolchain used for the tests (it was actually already merged by accident [here](https://github.com/tweag/rules_haskell/blob/master/rules_haskell_tests/MODULE.bazel#L361-L376)), but not used until now.

### Toolchain names
It seems that in the same `BUILD` file, toolchains are registered in alphabetical order, not in the order of the file.
(See [here](https://github.com/bazelbuild/rules_go/blob/e116651f92a15f4e5dd63ea09dd9ae2b2dd8e2f3/go/private/extensions.bzl#L135-L139)). So similarly to rules_go, a padded index is now used in the toolchain names, to ensure that the alphabetical order is the same as the order of the list. The name of the toolchain is not meaningful anymore, but it should be fine because the `toolchain` field (which is used for `--toolchain_resolution_debug`) is.


avdv (approved):
Nice, thank you!

PR #1921, opened 2023-07-14
Make nixpkgs GHC 9.6.x work

### LINKED ISSUE #1920: X is not a toolchain library
Hi,

I have upgraded a project to use GHC 9.6.2 (from 9.2.5).
I'm using nix to provide haskell dependencies using `ghcWithPackages` like so:

```nix
ghc = final.haskellPackages.ghcWithPackages (p:
  with p; [
     aeson
     aeson-pretty
    ...etc
```

I have a `BUILD` file with many targets and I have all the nix provided libraries declared as `haskell_toolchain_library`:

```bzl
haskell_toolchain_library(name="aeson")
haskell_toolchain_library(name="aeson-pretty")
haskell_toolchain_library(name="amazonka")
haskell_toolchain_library(name="amazonka-cognito-idp")
haskell_toolchain_library(name="amazonka-core")
haskell_toolchain_library(name="amazonka-dynamodb")
...etc
```

This used to work previously but now I can't load REPL or build library because of the following error:

```sh
ERROR: /home/romain/Code/matryo-product/backend/BUILD.bazel:58:26: in haskell_toolchain_library rule //backend:aeson-pretty:
        File "/home/romain/.cache/bazel/_bazel_romain/7404dc8029ea6151e966016273940d3d/external/rules_haskell/haskell/private/haskell_impl.bzl", line 801, column 13, in haskell_toolchain_library_impl
                fail(
Error in fail:
aeson-pretty is not a toolchain library.
Check that it ships with your version of GHC.
The following toolchain libraries are available:
{}
```

The library specified in the error (here `aeson-pretty`) changes with each try so i think it's a more general issue.
The libraries are available with the nix provided GHC according to `ghc-pkg list`

```sh
ghc-pkg list | grep aeson-pretty
    aeson-pretty-0.8.9
```

I'm using latest commit for `rules_haskell` and `rules_nixpkgs`.

Has something changed in the way we should use rules_haskell with Nix recently ?

### LINKED ISSUE #2031: add full support for GHC 9.6 nixpkgs

avdv:
Fixes #1920 and fixes #2031

Note that I had to skip the profiling test `//tests/haskell_module/dep-narrowing-th:Test` on macos for GHC 9.6 as it crashes calling `ghc-iserv`, see https://github.com/tweag/rules_haskell/issues/2073.

### DIFF
--- a/haskell/private/actions/compile.bzl
+++ b/haskell/private/actions/compile.bzl
@@ -124,6 +124,15 @@ def _compilation_defaults(
     compile_flags += hs.toolchain.ghcopts
     compile_flags += user_compile_flags
 
+    if hs.toolchain.is_darwin:
+        # assume `otool` and `install_name_tool` are available at the same location as `ar`
+        ar_bindir = paths.dirname(cc.tools.ar)
+
+        compile_flags += [
+            "-pgmotool=" + paths.join(ar_bindir, "otool"),
+            "-pgminstall_name_tool=" + paths.join(ar_bindir, "install_name_tool"),
+        ]
+
     package_ids = []
     all_plugins = plugins + non_default_plugins
     for plugin in all_plugins:
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -55,12 +55,15 @@ jobs:
         ghc:
           - 9.2.8
           - 9.4.6
+          - 9.6.2
         exclude:
           - module: rules_haskell_nix
             bzlmod: false
           # TODO: in a MODULE.bazel file we declare version specific dependencies, would need to use stack snapshot json
           #       and stack config per GHC version
           - ghc: 9.4.6
+            bzlmod: true
+          - ghc: 9.6.2
             bzlmod: true
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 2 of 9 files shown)


haskell/private/actions/compile.bzl
```diff
@@ -124,6 +124,15 @@ def _compilation_defaults(
...
     compile_flags += user_compile_flags
 
+    if hs.toolchain.is_darwin:
+        # assume `otool` and `install_name_tool` are available at the same location as `ar`
+        ar_bindir = paths.dirname(cc.tools.ar)
+
+        compile_flags += [
+            "-pgmotool=" + paths.join(ar_bindir, "otool"),
+            "-pgminstall_name_tool=" + paths.join(ar_bindir, "install_name_tool"),
+        ]
```
aherrmann, line 134: I suspect this may fail for users that try to use a hermetic bindist cc toolchain like grail's llvm toolchain. I'm fine with this for now as it is an improvement over the status quo. But, from what I understand so far, this may not cover all use-cases, yet.

aherrmann (approved):
Thank you!

haskell/private/actions/compile.bzl
avdv: Yes, we probably will need a better solution to this in the future.

PR #1922, opened 2023-07-30
fix include directories passed to hsc2hs

mraszyk:
New versions of ghc, e.g., ghc-9.6.2, do not contain an include directory under `$({ghc} --print-libdir)/include`. Hence, this PR uses `{ghc_pkg}` to find the include directories analogously to this [commit](https://github.com/tweag/rules_haskell/commit/7727edcac8606a19775d1f2a7463e503b6d976af).


avdv:
Thank you for your contribution!

Ideally, I would like to add support for GHC 9.4 (which seems to have the same problem, evidenced [here](https://github.com/tweag/gazelle_cabal/actions/runs/5519379463/job/14943156322?pr=49)) in https://github.com/tweag/rules_haskell/pull/1923 first

Also, I noticed that the hsc2hs which is included in GHC is configured with:
```
includedir="<<PREFIX>>/include"
...
Iflag="-I$includedir/"
```
(with `<<PREFIX>>` being the path given to the `configure --prefix=path` command)

The problem is that this directory does not exist in the installation, although it is present when unpacking the tarball. So this might actually be a matter of missing an installation step. I have created an [upstream bug](https://gitlab.haskell.org/ghc/ghc/-/issues/23769).

mraszyk:
Thank you for your further investigation and filing an upstream bug! Looking forward to a resolution.

avdv:
This [commit](https://github.com/tweag/rules_haskell/commit/7727edcac8606a19775d1f2a7463e503b6d976af) also seems very much related, since it fixes the same problem for c2hs.

mraszyk:
> This [commit](https://github.com/tweag/rules_haskell/commit/7727edcac8606a19775d1f2a7463e503b6d976af) also seems very much related, since it fixes the same problem for c2hs.

@avdv Do you suggest I use the same logic for obtaining the include dirs in this PR, too?

avdv:
> Do you suggest I use the same logic for obtaining the include dirs in this PR, too?

Yes, it seems a bit cleaner than fetching all include dirs unconditionally. :+1:

mraszyk:
> > Do you suggest I use the same logic for obtaining the include dirs in this PR, too?
> 
> Yes, it seems a bit cleaner than fetching all include dirs unconditionally. 👍

Sure, done.

mraszyk:
CI on Windows failed with
```
ERROR: The Build Event Protocol upload failed: Not retrying publishBuildEvents, no more attempts left: status='Status{code=UNAVAILABLE, description=io exception, cause=java.io.IOException: An existing connection was forcibly closed by the remote host
```
which doesn't seem related to this MR.

PR #1926, opened 2023-08-02
Do not skip worker on Windows

avdv:
It had been skipped because of an old GHC version on Windows, but this should be recent enough now.

This reverts commit 45340b1.

Also add some of the suggested Bazel flags by BuildBuddy. Apparently, this helps getting successful Windows CI builds.

depends on #1928


avdv:
@dependabot rebase

avdv:
@mergifyio rebase

Issue #1878, opened 2023-03-20
GHC version support


avdv:
* define a clear GHC version support range (keep this economical w.r.t. test coverage).

* evaluate test coverage and expand where needed

   * e.g. [GHC bindist patches undergo start-script tests](https://github.com/tweag/rules_haskell/blob/487535b8b10d496d8aa11aa4e9f1c91da476cf61/.github/workflows/patch-test.yaml#L78).
    * but changes to rules are not checked on the full GHC version range

* keep up-to-date
    * GHC upgrade(s)
    * what do we do on the blockers for ARM cross-compilation support?
    * can we use tooling (dependabot/renovate) to stay up-to-date going forward?

avdv:
Ideally, we should strive to support (and test) the recent stable GHC [releases](https://www.haskell.org/ghc/download.html):

> Current Stable Releases
>
>    * [9.6.2](https://www.haskell.org/ghc/download_ghc_9_6_2.html)
>    * [9.4.5](https://www.haskell.org/ghc/download_ghc_9_4_5.html)
>    * [9.2.8](https://www.haskell.org/ghc/download_ghc_9_2_8.html)

Also, we should take into account the supported GHC versions from [stackage](https://www.stackage.org/):

* 9.2.8 -> LTS 20.26
* 9.4.5 -> LTS 21.5
* 9.6.2 -> nightly

aherrmann:
@avdv is this one done now or is there still work remaining here?

avdv:
> is this one done now or is there still work remaining here?

We want to also support GHC 9.8 and eventually remove GHC 9.2 from testing since it is not a stable release anymore (but that can wait until we decide it's the right time)

aherrmann:
@avdv Thanks for clarifying, sounds good!

PR #1944, opened 2023-08-15
Add workflow to update GHC bindists

avdv:
The workflow is scheduled once a week, and can be triggered manually.

Currently, this does not (yet) try to update the GHC version that is used for tests.

And the created PR do not trigger CI; we would need to use a different token for that.

cf. #1878

### DIFF
--- a/.github/update-ghc.py
+++ b/.github/update-ghc.py
@@ -102,6 +102,10 @@ def main():
             with open('haskell/private/ghc_bindist_generated.bzl', 'r+') as generated:
                 run(['buildifier'], check=True, input=bzl, stdout=generated)
                 generated.truncate()
+
+            if 'GITHUB_OUTPUT' in os.environ:
+                with open(os.environ["GITHUB_OUTPUT"], 'a') as output:
+                    print(f"latest={ latest_release }", file=output)
     else:
         print("no tags found for prefix", prefix, file=sys.stderr)
 
--- a/.github/update-ghc.py
+++ b/.github/update-ghc.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python
+
+import os, sys, re
+import importlib
+import importlib.util
+import importlib.machinery
+import http.client
+import json
+from pathlib import PurePosixPath, Path
+from pprint import pprint
+from urllib.parse import quote
+from subprocess import Popen, PIPE, STDOUT, DEVNULL, check_output, run
+
+def import_path(path):
+    module_name = os.path.basename(path).replace('-', '_')
+    spec = importlib.util.spec_from_loader(
+        module_name,
+        importlib.machinery.SourceFileLoader(module_name, path)
+    )
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    sys.modules[module_name] = module
+    return module
+
+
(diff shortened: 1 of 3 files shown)


.github/workflows/update-ghc.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Create Pull Request
+        if: steps.ghc_update.outputs.latest != ''
+        uses: peter-evans/create-pull-request@v5
+        with:
+          commit-message: "[automation] Add GHC version "
+          title: "[update] GHC ${{ steps.ghc_update.latest }}"
+          branch: "automation/update-ghc-${{ matrix.ghc }}"
+          base: master
+          # TODO: needs a token so that workflows for the PR get triggered
+          #token: "${{ secrets.PR_TOKEN }}"
```
aherrmann, line 36: I think you have the required permissions to set such a token, or do you need help on that?

.github/workflows/update-ghc.yaml
```diff
@@ -0,0 +1,37 @@
...
+  update_ghc:
+    name: GHC ${{ matrix.ghc }} Update
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ghc:
+          - '9.2'
+          - '9.4'
+          - '9.6'
```
aherrmann, line 18: Just to confirm my understanding. Why do we loop over major-minor versions via a matrix instead of doing that within the update script with a single action run? Is it so that if multiple updates fall onto the same run we still get separate PRs per major-minor version?

.github/update-ghc.py
```diff
@@ -0,0 +1,114 @@
...
+def parse_version(s):
+    return tuple(map(int, s.split(".")))
+
+def main():
+    SCRIPT_PATH = Path(__file__)
+
+    if len(sys.argv) != 2:
+        sys.exit(f"usage: {SCRIPT_PATH.name} GHC_MAJOR_MINOR")
+
+    bindist = import_path(str(SCRIPT_PATH.parent.parent.joinpath('haskell/private/ghc_bindist_generated.bzl')))
```
aherrmann, line 58: An easier route could be to replace `haskell/private/ghc_bindist_generated.bzl` by a JSON file. The data is only loaded into the repository rule in `haskell/ghc_bindist.bzl`, so a combination of `repository_ctx.read` and `json.decode` could achieve the same thing. I think the data is only stored in a `.bzl` file for historic reasons (older versions of Bazel didn't expose JSON parsing in the Starlark API).

aherrmann (approved):
Thank you. A few questions, and I think this could be done easier via JSON. That said, AFAIU this re-uses the previous code from the manual update script. So, it's not much new complexity. So, I'm happy to merge this and switch to the JSON route in a follow-up.

.github/workflows/update-ghc.yaml
avdv: Yes, the reason was that I wanted to have separate PRs for each stable GHC major.minor version.

.github/workflows/update-ghc.yaml
avdv: I just wanted to check how other projects have set this up, maybe I could re-use a token -- do we have a bot user maybe?

Otherwise, I'll just create a new PAT and add it as a secret.

PR #1945, opened 2023-08-18
Use a JSON file to keep GHC bindist information

avdv:
This makes maintenance a bit easier. The downside is that we are no longer able specify valid values for `version`attribute in `attr.string(values = ...)`. So users might see the error a bit later, when we try to determine the hash and download URL.

And we also generate toolchain and bindist repositories (when using the `haskell_register_ghc_bindists` macro) for *all* supported targets regardless of the version now. Before, we would only generate these for targets which are available for the given version.


aherrmann (approved):
Thank you.

> The downside is that we are no longer able specify valid values for version attribute [...]

I think in principle one could have a repo rule that loads from the JSON file and exposes a Starlark variable that can be used to specify the valid values. But, that is probably not necessary.

> And we also generate toolchain and bindist repositories (when using the haskell_register_ghc_bindists macro) [...]

IIUC the module extensions still only registers for those platforms for which a release is available, correct?

avdv:
> I think in principle one could have a repo rule that loads from the JSON file and exposes a Starlark variable that can be used to specify the valid values. But, that is probably not necessary.

Yeah, I thought this would be too much effort as the difference for the user is neglectable.

> IIUC the module extensions still only registers for those platforms for which a release is available, correct?

No, when using the `bindists` tag on the `haskell_toolchains` extension, the `haskell_register_ghc_bindists` macro is called which will result in bindists for all platforms to be created.

aherrmann:
> > IIUC the module extensions still only registers for those platforms for which a release is available, correct?
> 
> No, when using the `bindists` tag on the `haskell_toolchains` extension, the `haskell_register_ghc_bindists` macro is called which will result in bindists for all platforms to be created.

I see, I think in this case we could avoid generating toolchain targets for unsupported platforms reasonably well, because the module extension can ([and already does](https://github.com/tweag/rules_haskell/blob/5b45a8bbaf96c7ee5d65d1921290ba9b54c5ed91/haskell/ghc_bindist.bzl#L383)) read the JSON file.

avdv:
> I see, I think in this case we could avoid generating toolchain targets for unsupported platforms reasonably well, because the module extension can ([and already does](https://github.com/tweag/rules_haskell/blob/5b45a8bbaf96c7ee5d65d1921290ba9b54c5ed91/haskell/ghc_bindist.bzl#L383)) read the JSON file.

I have added another parameter to the `haskell_register_ghc_bindists` macro which can be used to restrict the generated bindists. The module uses the GHC information from the JSON file to pass in only the supported target platforms. PTAL

aherrmann (approved):
Looks good, thank you!

PR #1948, opened 2023-09-01
Simplify mergify config

avdv:
Mergify automatically checks for requirements of branch protection rules, so we don't need to keep them in sync manually.

Currently, PR #1923 is not queued by mergify because the conditions are not met since the job names changed.


aherrmann (approved):
Thank you!

PR #1962, opened 2023-09-15
Allow using multiple stack snapshots in a MODULE.bazel

### LINKED ISSUE #1960: Support to define multiple stackage snapshots in module extenson

avdv:
Fixes #1960


aherrmann:
Why not simply use isolated extensions to that end as [implemented here](https://github.com/bazelbuild/bazel/pull/18529)? The feature should be available as of Bazel 6.3.0.

The main concern with naming snapshots is that it can easily lead to collisions across modules. E.g. what if two separate Bazel modules declare two separate snapshots called 'vector'?

avdv:
> Why not simply use isolated extensions to that end as [implemented here](https://github.com/bazelbuild/bazel/pull/18529)? The feature should be available as of Bazel 6.3.0.

Oh yes, we are on Bazel 6.2.0 currently. Will need to update.

aherrmann:
> With hoping that's the way forward, closing this PR.

It wasn't closed, yet. I'm assuming that that was the intention.

PR #1968, opened 2023-09-26
Import stackage as dev-dependency

aherrmann:
Mark the stack_snapshot imports within rules_haskell itself as dev_dependencies.
Otherwise downstream users of rules_haskell will have the same packages imported into their Stackage snapshot as used by rules_haskell for testing purposes. This can cause problems if downstream users want to use a different Stackage snapshot that may not include one of the dependencies imported by rules_haskell.

Note, due to a reference to `@stackage` from [within the `stack_snapshot` implementation](https://github.com/tweag/rules_haskell/blob/e31617dcd76264f00a833625f70e1e46e91fbc57/haskell/cabal.bzl#L2709) we still need to import the repository `@stackage` in all cases. This is done by using separate references to the extension, one marked as dev-dependency and the other not. However, they both refer to the same underlying instance of the module extension.


avdv (approved):
Thank you!

PR #1969, opened 2023-09-28
Only use GHC_VERSION if rules_haskell{,_test} is the root module

### LINKED ISSUE #1967: Internal GHC version selection exposed to users
**Describe the bug**
The internal GHC version selection mechanism introduced in https://github.com/tweag/rules_haskell/pull/1923 does affect down-stream use-sites.

**To Reproduce**
The bzlmod version of the Gazelle Cabal example test suite introduced in https://github.com/tweag/gazelle_cabal/pull/60, shows the following message in its [build log](https://github.com/tweag/gazelle_cabal/actions/runs/6312745862/job/17139411637?pr=60#step:7:661):
```
DEBUG: /home/runner/.cache/bazel/_bazel_runner/d4965fa66ed01d1aeec095f0355d1939/external/rules_haskell~override/haskell/private/ghc_ci.bzl:11:14: Using GHC version 9.4.5 from env variable `GHC_VERSION`
```
In this case the GHC toolchain is configured via Nix.

**Expected behavior**
I did not expect the GHC version selection logic used for rules_haskell's own CI to trigger in a use-site like the Gazelle Cabal example.

**Environment**
 - OS name + version: Ubuntu 22.04
 - Bazel version: 6.2.0
 - Version of the rules: 155ba21c7d315cc421ee3f17a1c09b758cc7279d

**Additional context**
none

avdv:
module_ctx.modules only contains the modules that use the extension, if the root module does not use it, the first module in that list is not the root module.

Fixes #1967

### DIFF
--- a/extensions/ghc_version.bzl
+++ b/extensions/ghc_version.bzl
@@ -14,9 +14,9 @@ load(
 )
 
 def _ghc_default_version_impl(mctx):
-    root_module = mctx.modules[0].name
+    root_module = mctx.modules[0]
 
-    if root_module in ["rules_haskell", "rules_haskell_tests"]:
+    if root_module.is_root and root_module.name in ["rules_haskell", "rules_haskell_tests"]:
         _ghc_version(name = "rules_haskell_ghc_version")
     else:
         _ghc_default_version(name = "rules_haskell_ghc_version")


aherrmann (approved):
Good catch!

Issue #1933, opened 2023-08-10
Add full support for GHC 9.4 bindist


avdv:
Currently we support GHC 9.4 bindist on a best effort basis, checking if the start example can be build using it.

There have been some reports that this is not good enough:

- https://github.com/tweag/rules_haskell/issues/1920
- https://github.com/tweag/rules_haskell/issues/1892
- #1900 

We should add GHC 9.4 to our CI matrix and test using that version.

PR #1978, opened 2023-10-18
Use GHC 9.4 for tests in rules_haskell_test module

avdv:
https://github.com/tweag/rules_haskell/issues/1933 added the `ghc` variable to the CI matrix but this was only used for the rules_haskell module.

This PR also makes use of the variable for the rules_haskell_test module in order to exercise the majority of tests for GHC 9.4 too.


aherrmann (approved):
LGTM, thank you!

PR #1980, opened 2023-10-18
Explicitly disable bzlmod by default

### LINKED ISSUE #1977: rules_haskell broken on Bazel's CI @HEAD - module not found in registries
**Describe the bug**
See https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3391#018b3bcb-68aa-4987-9304-53f1d05f0529

```
bazel --nosystem_rc --nohome_rc version

Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar

Build time: Thu Jan 01 00:00:00 1970 (0)

Build timestamp: Thu Jan 01 00:00:00 1970 (0)

Build timestamp as int: 0



bazel --nosystem_rc --nohome_rc info

[31m[1mERROR: [0min module dependency chain <root> -> rules_nixpkgs_core@0.9.0: module not found in registries: rules_nixpkgs_core@0.9.0

[0mTraceback (most recent call last):

  File "bazelci.py", line 3998, in <module>

    sys.exit(main())

  File "bazelci.py", line 3978, in main

    bazel_version=task_config.get("bazel") or configs.get("bazel"),

  File "bazelci.py", line 1354, in execute_commands

    bazel_version = print_bazel_version_info(bazel_binary, platform)

  File "bazelci.py", line 1647, in print_bazel_version_info

    [bazel_binary] + common_startup_flags() + ["--nosystem_rc", "--nohome_rc", "info"]

  File "bazelci.py", line 2655, in execute_command

    else None,  # capture_stderr=True when we want exceptions to contain stderr

  File "/usr/lib/python3.6/subprocess.py", line 438, in run

    output=stdout, stderr=stderr)

subprocess.CalledProcessError: Command '['bazel', '--nosystem_rc', '--nohome_rc', 'info']' returned non-zero exit status 2.
```

**To Reproduce**

Run `bazel --nosystem_rc --nohome_rc version` with latest green Bazel / RC version.

Note: seems like `--enable_bzlmod` is now flipped to true and that's why it fails with the above error.

avdv:
Upstream Bazel has flipped the enable_bzlmod flag to true by default, but we are not ready for it yet.

Fixes #1977

Note, since we require Bazel 6 now, we no longer need to use the `experimental` prefix for the `--enable_bzlmod` flag.


.bazelci/presubmit.yml
```diff
@@ -12,12 +12,16 @@ tasks:
     shell_commands:
       - "sudo apt -y update && sudo apt -y install libgmp-dev"
     build_flags:
+      # disable bzlmod explicitly, Bazel flipped the default to `true`
+      - "--noenable_bzlmod"
```
avdv, line 16: Unfortunately, this fix will not work since CI is running `bazel --nosystem_rc --nohome_rc info` before the build even starts and thus the build_flags are not taken into account yet. We would have to add this to our own bazelrc instead.

ylecornec (approved):
It looks good to me :+1:

PR #1994, opened 2023-10-20
Upgrade default GHC to 9.2.8

avdv:
depends on #1988


WORKSPACE
```diff
@@ -159,6 +159,7 @@ stack_snapshot(
         "_" + str(GHC_VERSION) if GHC_VERSION else "",
     ),
     packages = [
+        "Cabal",
```
avdv, line 162: Note, this is used here
https://github.com/tweag/rules_haskell/blob/8271ec29c47b3ed545234635a67b239d43854689/rules_haskell_tests/tests/repl-targets/hs_lib_repl_test/WORKSPACE#L132-L136

What this means is that it uses the Cabal from the git commit given in the stackage yaml config.

So instead of adding a `http_archive` for Cabal like this
https://github.com/tweag/rules_haskell/blob/8271ec29c47b3ed545234635a67b239d43854689/extensions/rules_haskell_dependencies.bzl#L28-L43
and then using `@Cabal//:Cabal` in `setup_deps` we might just use the cabal from the stackage snapshot directly as above. This would probably mean that every stack_snapshot would build it's own Cabal package instead of re-using the one from the Cabal repo. WDYT?

avdv: As discussed, I'll try to remove the redundant @Cabal repositories and use the Cabal from stackage directly.

haskell/cabal.bzl
```diff
@@ -2297,9 +2298,12 @@ load("@{workspace}//:packages.bzl", "packages")
...
 """.format(
-            workspace = workspace,
-            package = package,
-        ))
+                workspace = workspace,
+                package = package,
+            ))
+    else:
+        # a repository rule has to create a file
+        repository_ctx.file("no_executables", executable = False)
```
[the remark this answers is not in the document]
avdv: :+1: yes, changed.

PR #1999, opened 2023-10-26
Add GHC 9.8 to update-ghc workflow

avdv:
GHC 9.8.1 has been released a few weeks ago. This PR enables updates to the generated bindist file for this stable release too.

* Simplify and refactor the `gen_ghc_bindist.py` script to keep the GHC versions we generate entries for in the same place.
  These are now just stored in `ghc_bindist_generated.json` file.
* Add `9.8` to the `ghc` matrix variable to keep it updated

### DIFF
--- a/.github/update-ghc.py
+++ b/.github/update-ghc.py
@@ -93,8 +93,7 @@ def main():
             gen_script = gen.read()
 
             added_version = replace.sub(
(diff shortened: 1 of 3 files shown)


ylecornec (approved):
It looks good to me :+1:

PR #2005, opened 2023-11-01
Actually run in bzlmod mode on CI

avdv:
I have noticed that we actually never ran rules_haskell_test in bzlmod mode on CI.

This was due to a missing space character in the inline bash script of the workflow.yml file which would always yield `false`.

Additionally, we need to ensure that we can actually override Bazel settings using `.bazelrc.local` file which we are using on CI.

Note, this PR disables CI jobs for GHC 9.4 with bzlmod for now. The reason is that we would need to find a way to read a different stack_snapshot and json file depending on the GHC version. This is tracked here: https://github.com/tweag/rules_haskell/issues/2006

For Windows, we need to work around a path problem passing when `file:%workspace%/registry` which is expanded to an invalid file: URI without a path component. See https://github.com/bazelbuild/bazel/issues/20015

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,2 +1,6 @@
 import %workspace%/.bazelrc.common
 import %workspace%/.bazelrc.bzlmod
+
+# User Configuration
+# ------------------
+try-import %workspace%/.bazelrc.local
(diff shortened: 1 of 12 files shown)


aherrmann (approved):
Thank you!

PR #2008, opened 2023-11-06
Upgrade rules_nixpkgs modules to 0.10.0

avdv:
* remove rules_nixpkgs_core from our local registry, it is available in the BCR now
* upgrade modules in the registry and at their usage sites


aherrmann (commented):
remove rules_nixpkgs_core from our local registry, it is available in the BCR now

Not sure if that's the right thing. We want to develop all our rules_nixpkgs modules at HEAD. Meaning, if we make a change to rules_nixpkgs_core, then CI should test rules_nixpkgs_cc against that changed version of rules_nixpkgs_core. If we don't have the custom registry, then we probably need module overrides in bazelrc.

avdv:
> Not sure if that's the right thing. We want to develop all our rules_nixpkgs modules at HEAD.

yes, but this is rules_haskell. :-) we are just consuming rules_nixpkgs_* and only update on demand.

aherrmann (approved):
> yes, but this is rules_haskell. :-)

Oh, sorry! I thought this was in rules_nixpkgs :sweat_smile: 

Thank you, yes, this is great!

PR #2024, opened 2023-11-07
Add tar.gz archive to each release

avdv:
* generate a `tar.gz` archive instead of relying on Github's tag archive being immutable
* add a `bazel_dep` line to the release notes


avdv:
BTW, I tested this workflow and here's the result: https://github.com/tweag/rules_haskell/releases/tag/untagged-f8eac7679e3b92a44076

aherrmann (approved):
Thank you!

PR #2046, opened 2023-11-16
Add --with-bzlmod option to start script

### LINKED ISSUE #2044: Adapt start script to support bzlmod

avdv:
Fixes #2044 

Note, this only makes `--with-bzlmod=true|false` available for `--use-bindists` since the rules_haskell_nix module is not yet publicly available.


ylecornec (approved):
It looks good to me :+1:

PR #2050, opened 2023-11-20
Cleanup: remove deprecated options and targets

avdv:
1. remove `compiler_flags` option (users will get an error message now telling them to use `ghcopts` instead, to ease the migration -- this error handling should be removed after the next version was released)
2. remove REPL alias introduced in #460
3. remove deprecated nixpkgs platform aliases
4. move nixpkgs bzl_library to //docs (in order to get rid of any references to rules_nixpkgs_* in the core rules_haskell module)
5. enable lint rules and remove unused variables and load statements
6. remove unneeded indirect repositories from rules_haskell_tests module (as reported by gazelle's go_deps extension)

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -10,8 +10,41 @@ env:
   cache-version: 0
 
 jobs:
+  lint:
+    name: Format & Lint
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}-lint
+      cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Configure
+        env:
+          BUILDBUDDY_API_KEY: ${{ secrets.BUILDBUDDY_API_KEY }}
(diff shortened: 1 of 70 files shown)


avdv:
ping @aherrmann

.github/workflows/workflow.yaml
```diff
@@ -90,13 +123,13 @@ jobs:
...
             '
-      - name: Build & test - rules_haskell_nix
+      - name: Build - rules_haskell_nix
         if: matrix.module == 'rules_haskell_nix'
         run: |
           nix-shell --arg docTools false --argstr ghcVersion ${{ matrix.ghc }} --pure --run '
             set -euo pipefail
             cd rules_haskell_nix
-            bazel test //...
+            bazel build //...
```
aherrmann, line 132: I guess this is done because `rules_haskell_nix` does not contain any tests (apart from the manual buildifier test)?
We need not to forget to switch this back if we every add tests again.
I wonder if it may be safer to add a dummy test and keep this as `bazel test //...`.

aherrmann (approved):
Nice cleanup, thank you!

PR #2065, opened 2023-11-30
chore: refactor Bazel remote cache config

cgrindel:
Related to tweag/scalable-builds-group#107.

### DIFF
--- a/.bazelrc.common
+++ b/.bazelrc.common
@@ -56,6 +56,9 @@ build:remote-cache --remote_timeout=3600
 build:remote-cache --keep_backend_build_event_connections_alive=false
 # All clients except CI should be configured as read-only
 build:remote-cache --noremote_upload_local_results
+build:remote-cache --remote_timeout=3600
+build:remote-cache --experimental_remote_cache_compression
+build:remote-cache --experimental_remote_build_event_upload=minimal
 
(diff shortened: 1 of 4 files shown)


.bazelrc.common
```diff
@@ -71,7 +82,6 @@ build:ci-windows-bindist --bes_upload_mode=wait_for_upload_complete --bes_timeou
 # (see https://github.com/buildbuddy-io/buildbuddy/issues/4467)
 # increase the retries as a workaround.
 build:ci-windows-bindist --experimental_build_event_upload_max_retries=256
-build:remote-cache --remote_cache=grpcs://remote.buildbuddy.io
 build:ci-common --remote_timeout=3600
 # Avoid failures of the form `deadline exceeded after 14999958197ns DEADLINE_EXCEEDED`.
 # See https://github.com/tweag/rules_haskell/issues/1498.
```
avdv, line 87: This, and the line below should be redundant now, right?

avdv (approved):
Thank you.

.bazelrc.common
cgrindel: Again, good catch. Apparently, I was asleep when I put this PR together. I did another pass and grouped a few more remote cache flags

PR #2075, opened 2023-12-09
chore: use `tweag/run-nix-shell`

cgrindel:
Related to https://github.com/tweag/scalable-builds-group/issues/138.


.github/workflows/workflow.yaml
```diff
@@ -96,29 +98,35 @@ jobs:
...
       - name: Build & test - rules_haskell_nix
         if: matrix.module == 'rules_haskell_nix'
-        run: |
-          nix-shell --arg docTools false --argstr ghcVersion ${{ matrix.ghc }} --pure --run '
+        uses: tweag/run-nix-shell@v0
+        with:
+          options: ${{ env.NIX_SHELL_ARGS }}
+          run: |
             set -euo pipefail
             cd rules_haskell_nix
```
avdv, line 122: Could we use `working-directory:` instead and drop the `cd command` and `set -euo pipefail`?

.github/workflows/workflow.yaml
```diff
@@ -96,29 +98,35 @@ jobs:
...
       - name: Build & test - rules_haskell_tests
         if: matrix.module == 'rules_haskell_tests'
-        run: |
-          nix-shell --arg docTools false --argstr ghcVersion ${{ matrix.ghc }} --pure --run '
+        uses: tweag/run-nix-shell@v0
+        with:
+          options: ${{ env.NIX_SHELL_ARGS }}
+          run: |
             set -euo pipefail
             cd rules_haskell_tests
```
avdv, line 131: Likewise, but keep the `set -euo pipefail` since this runs multiple commands.

avdv (changes requested):
Thank you, just some minor remarks.

avdv (approved):
Thank you!

.github/workflows/workflow.yaml
cgrindel: It looks like I need to add an attribute to `run-nix-shell` to specify the location of `shell.nix`. I just added [an issue](https://github.com/tweag/run-nix-shell/issues/4) to track. Are you OK with me leaving the `cd ...` for now?

PR #2076, opened 2023-12-09
chore: use magic-nix-cache-action

cgrindel:


cgrindel:
The magic cache does not appear to shorten the build time for this repo.

PR #2078, opened 2023-12-11
chore: use `extractions/netrc` to configure GitHub API authentication

cgrindel:


avdv (approved):
LGTM, and since we only use github tokens as password, we should not be effected by https://github.com/extractions/netrc/issues/4

cgrindel:
> since we only use github tokens as password, we should not be effected by https://github.com/extractions/netrc/issues/4

@avdv Good catch.

PR #2080, opened 2023-12-13
chore: support arm64 on MacOS

cgrindel:


haskell/ghc_bindist.bzl
```diff
@@ -301,8 +289,21 @@ _ghc_bindist = repository_rule(
...
+            default = [],
+            doc =
+                "A list of files that are to be applied as patches afer " +
+                "extracting the archive.",
+        ),
+        "repl_ghci_args": attr.string_list(),
+        "target": attr.string(),
+        "version": attr.string(
+            default = _GHC_DEFAULT_VERSION,
+            doc = "The desired GHC version",
```
avdv, line 306: Are these changes done by buildifier? Are we missing an option here that enforces sorting the dict keys?

avdv (approved):
LGTM, thanks!

haskell/ghc_bindist.bzl
cgrindel: I reverted the unintended formatting changes for now.

PR #2088, opened 2023-12-19
chore: remove redundant Bazel configuration names

cgrindel:
- Remove `ci-linux-*` and `ci-macos-*` as they merely applied their respective `linux-*` and `macos-*` configuration names.

Related to #2087.


avdv:
Could you also remove those from the `.bazelci/presubmit.yml` config?

cgrindel:
> Could you also remove those from the `.bazelci/presubmit.yml` config?

It appears that they are not referenced there. I just did a search for `ci-linux-` and `ci-macos-`. There were no results.

avdv (approved):
> > Could you also remove those from the `.bazelci/presubmit.yml` config?
> 
> It appears that they are not referenced there. I just did a search for `ci-linux-` and `ci-macos-`. There were no results.

You're right. I don't know where dreamed that up...

LGTM, thank you!

Issue #2089, opened 2023-12-19
Fix failing CI in `rules_haskell`


cgrindel:

avdv:
This seems to be a bit flaky, as this does not happen very often. Seems like the bazel processes get killed:
```
Exit Code: ExitFailure (-9)
```
Maybe this is some sort of timeout?

cgrindel:
It could be. I have been trying to set up my personal MacOS laptop with Nix to try things out. (My Modus laptop is arm64 and  I have not completed the work getting rules_haskell to work over there.) Unfortunately, the Nix install is failing. I must not have cleaned up from a previous Nix install attempt. 😣

cgrindel:
[This issue](https://github.com/commercialhaskell/stack/issues/3017) seems to suggest that it may be running out of memory.

cgrindel:
For code locations where `run-tests` starts servers, see @avdv [comment](https://github.com/tweag/rules_haskell/pull/2090#discussion_r1432807881).

PR #2090, opened 2023-12-19
fix: shutdown Bazel before executing `run-tests`

cgrindel:
The `run-tests` can fail with out-of-memory errors (e.g. `Exit Code: ExitFailure (-9)`). We shutdown Bazel from the previous step to alleviate some of the memory pressure.

Related to #2089.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -139,6 +139,8 @@ jobs:
               ./tests/run-start-script.sh --use-nix --with-bzlmod=${{ matrix.bzlmod }}
             fi
             bazel build //tests:run-tests
+            # Shutdown Bazel to free up memory
+            bazel shutdown
             ./bazel-ci-bin/tests/run-tests
             bazel coverage //...
 
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -130,6 +130,9 @@ jobs:
           options: ${{ env.NIX_SHELL_ARGS }}
           run: |
             set -euo pipefail
+            # DEBUG BEGIN
+            set -x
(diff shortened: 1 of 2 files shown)


.github/workflows/workflow.yaml
```diff
@@ -139,6 +139,8 @@ jobs:
               ./tests/run-start-script.sh --use-nix --with-bzlmod=${{ matrix.bzlmod }}
             fi
             bazel build //tests:run-tests
+            # Shutdown Bazel to free up memory
+            bazel shutdown
```
avdv, line 143: That might already help, but `RunTest.hs` is invoking bazel again and starts a few server instances. Maybe we should also add shutdown calls there...

cgrindel: @avdv Do you mean after `run-tests` has completed but before it returns to this script?

avdv: I mean here:
https://github.com/tweag/rules_haskell/blob/c2c6d3f7a5f51b49b65d98ede37f20914e591475/rules_haskell_tests/tests/RunTests.hs#L153-L156
and here:
https://github.com/tweag/rules_haskell/blob/c2c6d3f7a5f51b49b65d98ede37f20914e591475/rules_haskell_tests/tests/RunTests.hs#L149-L151

The spec items are run sequentially and all the items before these lines are run inside the current workspace, but these items run in a different workspace.

cgrindel: I added a task to #2089.

avdv (approved):

cgrindel:
Replaced by #2092. Avoid triggering the `Test GHC patches`.

PR #2096, opened 2023-12-21
chore: shutdown Bazel servers launched by RunTests.hs

cgrindel:
Per [this conversation](https://github.com/tweag/rules_haskell/pull/2090#discussion_r1432807881), to prevent out-of-memory (OOM) errors, we will shutdown Bazel servers started by `RunTests.hs`.

Related to #2089.


rules_haskell_tests/tests/RunTests.hs
```diff
@@ -149,10 +149,12 @@ main = hspec $ do
   it "bazel test examples" $ do
     assertSuccess $ (bazel ["build", "//..."]) { Process.cwd = Just "../examples" }
     assertSuccess $ (bazel ["test", "//..."]) { Process.cwd = Just "../examples" }
+    assertSuccess $ (bazel ["shutdown"]) { Process.cwd = Just "../examples" }
```
avdv, line 152: Maybe we also need to shutdown the bazel server from the previous spec items?

cgrindel: Agreed. I am not sure where I should put the shutdown. Not being familiar with Haskell or the test framework, I am not sure if the tests and test cases are executed in order from top to bottom. That is on my list of things to do today. If you have a suggestion on where to put the shutdown, please let me know.

avdv: Probably the easiest option is to shutdown the server before starting the `examples` and `tutorial` run? E.g.
```
it "bazel test examples" $ do
    assertSuccess $ bazel ["shutdown"]
    assertSuccess $ (bazel ["build", "//..."]) { Process.cwd = Just "../examples" }
    assertSuccess $ (bazel ["test", "//..."]) { Process.cwd = Just "../examples" }
    assertSuccess $ (bazel ["shutdown"]) { Process.cwd = Just "../examples" }
```
likewise for the tutorial. This way it does not matter in which order the items are run and you can also run individual items if you want.

cgrindel: Good point. I will add that with comments explaining why we added these.

rules_haskell_tests/tests/RunTests.hs
```diff
@@ -165,6 +164,58 @@ bazel args = Process.proc "bazel" args
...
+
+topPath :: String
+topPath = "/usr/bin/top"
+
+-- | Print information about the current memory state to debug intermittent failures
+-- Related to https://github.com/tweag/rules_haskell/issues/2089
+printMemory :: String -> IO ()
+printMemory msg = do
+  -- Do not attempt to run top, if it does not exist.
+  (exitCode, _, _) <- Process.readProcessWithExitCode "test" [topPath] ""
```
avdv, line 205: Does this really check if the file exists? Don't you need to call it like `test -f path-to-file`?

You could also use `System.Directory.doesFileExist` instead.

avdv (approved):
Otherwise, LGTM.

rules_haskell_tests/tests/RunTests.hs
cgrindel: It doesn't. 🙂 I forgot the `-e`.  I will try using the suggested API.

cgrindel:
@avdv For now, I recommend that we disable the test that exhibits the error. I will continue my investigation. However, these errors are hindering other work items. Hence, I am asking for another code review.

PR #2098, opened 2024-01-03
Fix toolchain libraries creation when using `ghcWithPackages`

### LINKED ISSUE #2097: rts dependency in cc_library no longer works (can't find HsFFI.h)
**Describe the bug**
I have a `cc_library` with a file that includes `HsFFI.h`. It depends on the `rts` `haskell_toolchain_library`. As of some commit around the range of the two commits below, that include fails:

```
c-toxcore-hs/src/hsbracket.c:1:10: fatal error: 'HsFFI.h' file not found
```

**To Reproduce**
Commit d14584e7f227fee55d857694d6ba7dc7e74d8f3c is fine.
Commit 895a4a2973f03006e2c461b3551fce22038df1f7 is broken.

**Environment**
 - OS name + version: NixOS 23.11
 - Bazel version: 6.4.0
 - Version of the rules: See above.

**Additional context**
I tried roughly bisecting the git history but didn't investigate a lot. The git history is very non-linear, so the linear github view is a bit confusing. I didn't look at the actual git repo.

avdv:
Fixes #2097

When using `ghcWithPackages` package db config files are symlinked into the package database
directory referencing the real ghc store path.

We assume the real ghc file is also accessible from the toplevel ghc-with-packages store path,
which is symlinked into the toolchain repository.


iphydf:
I can confirm this fixes the bug. toktok-stack builds completely and all tests pass with this PR.

avdv:
Thank you @ylecornec for adding the test and pushing this through!

Issue #2099, opened 2024-01-04
Refactor Bazel configuration to reduce/eliminate platform-specific flags


cgrindel:
**Is your feature request related to a problem? Please describe.**

This repository defines named configurations based upon platform (e.g. `linux-nixpkgs`, `macos-nixpkgs`, `linux-bindist`, `macos-bindist`). In theory, some/all of the configuration could be migrated to platform discovery features using `exec_compatible_with` or `select()`.

**Describe the solution you'd like**

Review the OS-specific configuration settings and propose alternative implementations.

cgrindel:
[How to detect if in a Nix shell: nix-shell vs nix shell.](https://discourse.nixos.org/t/in-nix-shell-env-variable-in-nix-shell-versus-nix-shell/15933)

cgrindel:
More on how to detect if in Nix shell
- https://github.com/NixOS/nix/issues/6677
- [Related discussion on Discourse](https://discourse.nixos.org/t/custom-prompts-or-shell-depth-indicator-for-nix-shell-nix-develop/29942)

cgrindel:
Quick update:
- It turns out in Bazel 6.4.0, toolchain resolution occurs before incompatible targets. This causes failures when trying to specify `target_compatible_with`. #2106 
- I started trying to upgrade `rules_hakell` to Bazel 7. #2107 
  - There are a number of errors.
  - One of the errors that has been difficult to track down is that [`cc_library` targets](https://github.com/tweag/rules_haskell/blob/56c24284a8baf9981612705179d96e3957d40998/rules_haskell_tests/tests/data/BUILD.bazel#L5-L10) build differently under Bazel 7 vs Bazel 6.x on MacOS.

PR #2102, opened 2024-01-05
chore: configure OS-specific variables in `.bazelrc`

cgrindel:
- Add `--enable_platform_specific_config` to `.bazelrc.common`.
- Move MacOS- and Windows-specific environment variables to `.bazelrc.common` under their OS config names.
- Update code that generates bazelrc files for integration tests to include the new entries.

Related to https://github.com/tweag/rules_haskell/issues/2099.


avdv (approved):
Nice, thank you.

Should we also remove the `BAZEL_USE_CPP_ONLY_TOOLCHAIN` variable from the `shell.nix` file?

cgrindel:
@mergifyio squash title+body

PR #2104, opened 2024-01-08
Remove `haskell_toolchain_libraries` rule

avdv:
This internal rule was only needed as a workaround, but toolchain transitions have been implemented and it is not needed anymore.


aherrmann (approved):
Nice, thank you!

PR #2115, opened 2024-02-05
Prepare release 0.19

ylecornec:


avdv:
Thank you, LGTM.

But we also need to fix the publish-to-BCR config, according to this comment: https://github.com/bazel-contrib/publish-to-bcr/issues/98#issuecomment-1865916836

Could you also take care of this here?

PR #2120, opened 2024-02-08
Update presubmit.yml for the BCR

ylecornec:
This PR updates the presubmit.yml for the BRC after it was modified a bit in the following [PR](
https://github.com/bazelbuild/bazel-central-registry/pull/1447).

In the end calling `xcodebuild -runFirstLaunch` installed the required library on the macos arm64 runner (IIUC `code-select --install` requires a desktop environment).


avdv (approved):
Thank you!

Issue #2006, opened 2023-11-03
Test GHC 9.4 with bzlmod on CI


avdv:
**Is your feature request related to a problem? Please describe.**
We disabled GHC 9.4.x jobs for bzlmod on CI since we need to provide a version specific stack snapshot and json file.

**Describe the solution you'd like**
Find a way to pass corresponding stack yaml and json files according to the currently active GHC version. 

Ideally, this should only have an impact on our CI run, not on user's of rules_haskell.

**Describe alternatives you've considered**

**Additional context**
See https://github.com/tweag/rules_haskell/pull/2005

PR #2121, opened 2024-02-13
Set default ghc verion 9.4.6

ylecornec:
This PR sets ghc 9.4.6 as the default version used by rules_haskell (at the moment the default version is the only one tested with bzlmod (See https://github.com/tweag/rules_haskell/issues/2006)).

The current version (9.2.8) is not listed anymore in the list of [current releases](https://www.haskell.org/ghc/download.html) and we will stop testing against it when adding support for 9.8.


aherrmann (approved):
LGTM

PR #2122, opened 2024-02-14
fix `use_stack` documentation

prednaz:
closes https://github.com/tweag/rules_haskell/issues/1759.

Here is a screenshot.
Screenshot 2024-02-14 at 22-43-13 Cabal packages

PR #2123, opened 2024-02-20
improve testing and linting experience for newcomers

prednaz:
added a `.bazelrc.local` symbolic link to `rules_haskell_tests` following the example of https://github.com/tweag/rules_haskell/pull/750 and in reaction to https://github.com/tweag/rules_haskell/pull/2122#issuecomment-1944652674.

`(bazel test //... && cd rules_haskell_tests && bazel test //...)` and `(bazel run //buildifier && cd rules_haskell_tests && bazel run //buildifier)` run successfully now in a freshly cloned repository after `echo "build --config=linux-nixpkgs" >> .bazelrc.local`.

using round brackets around the bash commands will restore the current working directory afterwards.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -98,7 +98,6 @@ jobs:
           common --enable_bzlmod=${{ matrix.bzlmod }}
           EOF
           cp .bazelrc.local rules_haskell_nix
-          cp .bazelrc.local rules_haskell_tests
       - name: Build & test - rules_haskell
         if: matrix.module == 'rules_haskell'
         uses: tweag/run-nix-shell@v0
@@ -208,7 +207,6 @@ jobs:
           $bzlmod_cache_silo_key
           common --enable_bzlmod=${{ matrix.bzlmod }}
           EOF
-          cp .bazelrc.local rules_haskell_tests
       - name: Build & test - rules_haskell
         if: matrix.module == 'rules_haskell'
         shell: bash
--- a/README.md
+++ b/README.md
@@ -352,20 +352,20 @@ on every development system (`python`, `ghc`, `go`, …).
 To build and run tests locally, execute:
 
 ```
-$ bazel test //... && cd rules_haskell_tests && bazel test //...
+$ (bazel test //... && cd rules_haskell_tests && bazel test //...)
 ```
 
 Starlark code in this project is formatted according to the output of
 [buildifier]. You can check that the formatting is correct using:
 
 ```
-$ bazel run //buildifier && cd rules_haskell_tests && bazel run //buildifier
+$ (bazel run //buildifier && cd rules_haskell_tests && bazel run //buildifier)
 ```
 
 If tests fail then run the following to fix the formatting:
 
 ```
-$ git rebase --exec "bazel run //buildifier:buildifier-fix && cd rules_haskell_tests && bazel run //buildifier:buildifier-fix" <first commit>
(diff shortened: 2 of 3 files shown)


prednaz:
i cannot understand why ci is failing. `nix-shell --pure --command 'cd rules_haskell_tests && bazel build //tests:run-tests'` works on my machine.

ylecornec:
It looks like the CI is failing when `bzlmod` is activated, you should be able to try it locally with the `--enable_bzlmod` flag
(or setting `build --enable_bzlmod` in the `.bazelrc.local`).

I think the problem happens because the local module registry has a relative path for `rules_haskell` defined [here](
https://github.com/tweag/rules_haskell/blob/master/registry/modules/rules_haskell/0.19/source.json#L3) (because we want to use the version of rules_haskell that is currently checked out ).

So by making `registry` a symlink in `rules_haskell_test` it looks like this path is now resolved starting from `rules_haskell_test` instead of `rules_haskell` and some files are not found.

prednaz:
@ylecornec thanks, that is a good explanation. in this case, i will give up on my idea of unifying those `.bazelrc*` files and restrict this pull request to the changes i described above instead.

ylecornec (commented):
This forces the `.bazelrc.local` files to be the same in `rules_haskell` and `rules_haskell_test` but I think it should be ok.

In general we must be careful with such symlinks (see https://github.com/tweag/rules_nixpkgs/issues/373) but it should be fine here since rules_haskell_test is only used for testing.

It also seems a bit risky to have a dangling symlink by default if there is no root `.bazelrc.local` but `try-import` seems to handle it fine.

However, I think if we commit the `rules_haskell_tests/.bazelrc.local` file, there should be an exception in `.gitignore` : `!/rules_haskell_tests/.bazelrc.local` (because at the moment all `.bazelrc.local` file are ignored).

aherrmann:
I'd prefer not to check in `.bazelrc.local` files. These are meant for *local* modifications that are not shared across contributors.
If there is an intent to make sure that both `rules_haskell` and `rules_haskell_tests` take the top-level `.bazelrc.local` into account, then this could be achieved with `try-import` statements.

prednaz:
> If there is an intent to make sure that both `rules_haskell` and `rules_haskell_tests` take the top-level `.bazelrc.local` into account, then this could be achieved with `try-import` statements.

yes, that was the intent in order to avoid newcomers facing https://github.com/tweag/rules_haskell/pull/2122#issuecomment-1944652674. the alternative you suggest sounds better to me too except if combined with https://github.com/tweag/rules_haskell/pull/2124 because then there would be a `try-import %workspace%/../.bazelrc.local` in the top level workspace, which seems like a bad idea.

cgrindel:
> If there is an intent to make sure that both rules_haskell and rules_haskell_tests take the top-level .bazelrc.local into account, then this could be achieved with try-import statements.

Agreed. We can try-import in the parent and the `rules_haskell_tests` directory.

aherrmann:
> except if combined with #2124 because then there would be a `try-import %workspace%/../.bazelrc.local` in the top level workspace, which seems like a bad idea.

I agree that the `../` import in the top-level `.bazelrc` is a bad idea. But, I don't think it's necessary to fulfill #2124. "Unifying" the `.bazelrc` files does not have to mean to physically have a single file and symlink to it from all other instances. Instead it can mean to have the shared configuration in a `.bazelrc.common` that is imported by the other `.bazelrc` files that need it. [rules_nixpkgs](https://github.com/tweag/rules_nixpkgs/) does this. Symlinks can be problematic. For example, if we want to publish a sub-tree as a standalone extension, e.g. rules_haskell_nixpkgs, then upwards symlinks break the Zip archives, see https://github.com/tweag/rules_nixpkgs/issues/373.

PR #2124, opened 2024-02-20
unify `rules_haskell_tests/.bazelrc*` files

prednaz:
unify `.bazelrc*` files and `rules_haskell_tests/.bazelrc* files` via symbolic links following the example of https://github.com/tweag/rules_haskell/pull/750.

### DIFF
--- a/rules_haskell_tests/.bazelrc
+++ b/rules_haskell_tests/.bazelrc
@@ -1,6 +1,1 @@
-import %workspace%/../.bazelrc.common
-import %workspace%/.bazelrc.bzlmod
-
-# User Configuration
-# ------------------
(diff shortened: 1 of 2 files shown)


cgrindel:
While this change removes some code duplication, I am worried that this change does not help readability. One needs to reason that the file that actually lives at the root of the repository needs to be interpreted as being evaluated from the `rules_haskell_tests` directory. For example, where should one put the `.bazelrc.local` for `rules_haskell_tests`.

prednaz:
closing this because for the reason stated in https://github.com/tweag/rules_haskell/pull/2123#issuecomment-1959078641.

Issue #1145, opened 2019-11-14
Transitive re-exporting does not work


eltix:
**Describe the bug**

When there are at least two levels of re-exporting, the innermost modules do not get exported.

**To Reproduce**
Consider the following example:

```bazel
haskell_library(
    name = "lib",
    exports = [
        ":a",
        ":b",
        ":c",
    ],
    deps = [
        ":a",
        ":b",
        ":c",
    ],
)

haskell_library(
    name="pipeline",
    exports = [":lib"],
    deps=[
      "@hackage//:aeson",
      ":lib",
     ]

haskell_binary(
    name="exe",
    srcs=["src/Main.hs"],
    deps=[
        "@hackage//:base",
        ":pipeline",
    ],
)
```

The above will fail upon building with this error:
```
src/Main.hs:21:1: error:
    Could not load module ‘A’
    It is a member of the hidden package ‘a’.
    You can run ‘:set -package a’ to expose it.
    (Note: this unloads all the modules in the current scope.)
    Use -v to see a list of the files searched for.
   |
21 | import qualified A  as A
```

where `A.hs` is a Haskell source file of the library `":a"`

Everything works fine when `":lib"` is added to the deps of the haskell binary.

**Expected behavior**
We shouldn't have to add `":lib"` to the deps of the haskell binary since it's already re-exported by `":pipeline"`

**Environment**
 - OS name + version: Ubuntu 19.10
 - Bazel version: "Build label: 0.28.0- (@non-git)"
 - Version of the rules: "23c6b64cae7db676485a1e401698520f4cdd8254"

aherrmann:
At a cursory look, fixing this will probably require tracking exports transitively [here](https://github.com/tweag/rules_haskell/blob/7204b88fdcbaf2098d381790bf5d146e62b4a05c/haskell/private/haskell_impl.bzl#L617-L621) instead of just the direct exports. This should probably be done by using a `depset` instead of a list to track these.

aherrmann:
@prednaz was this resolved by #2125? If so, please close. :slightly_smiling_face:

prednaz:
fixed by https://github.com/tweag/rules_haskell/pull/2125

PR #2125, opened 2024-02-20
use `depset` for reexports in `HaskellLibraryInfo`

prednaz:
- [x] fix https://github.com/tweag/rules_haskell/issues/1145
- [x] add tests
- [ ] investigate whether the call sites of `all_package_ids` and `all_dependencies_package_ids` expect the old or new behavior to be correct


haskell/providers.bzl
```diff
@@ -31,12 +31,12 @@ HaskellLibraryInfo = provider(
     fields = {
         "package_id": "Workspace unique package identifier.",
         "version": "Package version.",
-        "exports": "List of other `HaskellLibraryInfo` that this package reexports",
+        "exports": "List of `package_id`s that this package exports, that is reexports and this package's own `package_id`",
```
avdv, line 34: small typo:
```suggestion
        "exports": "List of `package_id`s that this package exports, that it reexports and this package's own `package_id`",
```

avdv (approved):
Nice, thank you!

haskell/providers.bzl
prednaz: hm, not a typo but clumsy english. i meant to go for [this idiom](https://www.dictionary.com/browse/that%20is). here are some ideas to make it more clear.

```suggestion
        "exports": "List of `package_id`s that this package exports, that is, reexports and this package's own `package_id`",
```

```suggestion
        "exports": "List of `package_id`s that this package exports, that is to say, reexports and this package's own `package_id`",
```

```suggestion
        "exports": "List of `package_id`s that this package exports, i.e. reexports and this package's own `package_id`",
```

Issue #2131, opened 2024-03-01
proto-len protoc plugin failures on windows with ghc 9.6.2 and 9.8.1


ylecornec:
- This [commit](https://github.com/tweag/rules_haskell/commit/baf6f8e4118eef89f79ca64eecd6abbe4b603d21) contains the error message for ghc 9.6.2
- Similar errors happen with ghc 9.8.1
    - https://github.com/tweag/rules_haskell/actions/runs/7992627631/job/21826580388#step:11:958
    - https://github.com/tweag/rules_haskell/actions/runs/7992627631/job/21826579493#step:10:194

It is not clear where the problem comes from, but in both cases we temporarily rely on forks of `proto-lens` and `ghc-source-gen` packages in which we updated package bounds and fixed some compilation issues (which seems to work on other platforms according to the tests).

- For ghc 9.6,  we could maybe try to use upstream packages now, as well as update to 9.6.4.

- I do not know if this is related but the error message is similar in these ghc issues:
  - https://gitlab.haskell.org/ghc/ghc/-/issues/24016
  - https://gitlab.haskell.org/ghc/ghc/-/issues/24443

Issue #2127, opened 2024-02-21
Haddock 2.30 faillure in profiling mode


ylecornec:
It seems that haddock 2.30 which comes with ghc 9.8.1 now requires `.hi` interface files as input.

See https://github.com/haskell/haddock/blob/ghc-9.8/doc/invoking.rst#avoiding-recompilation

This is a problem when profiling is enabled since only `p_hi`  files are present in the sandbox in this case.

**To Reproduce**

The following command fails on the `ylecornec/ghc_9_8_1` branch (at least with a nix provided ghc).

```
$ cd rules_haskell_test && GHC_VERSION=9.8.1 bazel build -c dbg //tests:toto
Failed to load interface for ‘LZ4’
```

- If we copy a non profiling `lz4.hi` file to the sandbox next to `lz4.p_hi` the build is successful
- If we change the extension of profiled interface files from `p_hi` to `hi` the file is found by haddock but there is a different error:
    ```
    Exception when reading interface file  bazel-out/k8-dbg/bin/tests/testsZSlz4/_iface/LZ4.hi
      mismatched interface file profile tag (wanted "", got "p")
    ```

Issue #2073, opened 2023-12-07
ghc-iserv BUS error / SIGSEGV on Darwin with GHC 9.6 / 9.8


avdv:
**Describe the bug**

Testing `//tests/haskell_module/dep-narrowing-th:Test` on amd64 Darwin using GHC 9.6 crashes ghc-iserv with a BUS error when using a GHC from nixpkgs, and with a SIGSEGV error using a GHC bindist.

### nixpkgs GHC 9.6.2:
```
ERROR: /Users/runner/work/rules_haskell/rules_haskell/rules_haskell_tests/tests/haskell_module/dep-narrowing-th/BUILD.bazel:88:16: HaskellBuildObject @//tests/haskell_module/dep-narrowing-th:lib @//tests/haskell_module/dep-narrowing-th:TestModule2 failed: (Exit 1): ghc_wrapper failed: error executing command (from target //tests/haskell_module/dep-narrowing-th:lib) 
         (cd /private/var/tmp/_bazel_runner/6f93c0710d32c05eb86fd5414239ece6/sandbox/processwrapper-sandbox/3795/execroot/rules_haskell_tests && \
         exec env - \
           LANG=C.UTF-8 \
           MUST_EXTRACT_ABI=true \
           PATH=/nix/store/f0hpls21prmic0kg5b4yz5wgg1jdradr-posix-toolchain/bin:/nix/store/f0hpls21prmic0kg5b4yz5wgg1jdradr-posix-toolchain/bin \
           RULES_HASKELL_DOCDIR_PATH=/nix/store/slndd05yb65viiv9g7b05fb9y38rj0kq-ghc-9.6.2/lib/ghc-9.6.2/lib/../../../../qpzm6rba12ifz005ksjx302vyrzri7rq-ghc-9.6.2-doc/share/doc/ghc/html/libraries/base-4.18.0.0 \
           RULES_HASKELL_GHC_PATH=external/rules_haskell_ghc_nixpkgs/bin/ghc \
           RULES_HASKELL_GHC_PKG_PATH=external/rules_haskell_ghc_nixpkgs/bin/ghc-pkg \
           RULES_HASKELL_LIBDIR_PATH=/nix/store/slndd05yb65viiv9g7b05fb9y38rj0kq-ghc-9.6.2/lib/ghc-9.6.2/lib \
[...]

PR #2132, opened 2024-03-01
Support for ghc 9.8.1

ylecornec:
This PR adds support for ghc 9.8.1

In particular:
- ghc 9.2.8 is replaced by 9.8.1 in the CI.
- ghc now installs libraries in directories named `name-version-suffix`, where suffix which can be either a hash, or `-inplace`:   See https://gitlab.haskell.org/ghc/ghc/-/issues/23788. This only seems to affect the `test-haddock` test which was refactored a bit as these hashes vary between platforms.
- As for version 9.6.2, windows testing is deactivated in the CI. See issue https://github.com/tweag/rules_haskell/issues/2131.
  It may be possible to test targets that do not depend on these failing one but I did not try it yet. The targets failing are at least:
    - `@stackage//:language-c` which is a dependency of the `c2hs` toolchain
    - `//rule_info:rule_info_proto`
- The `//tests:test-haddock` is disabled because of #2127
- #2073 still seems to happen with 9.8.1


avdv (approved):
Nice work! :100:

Issue #2119, opened 2024-02-08
Upgrade to new rules_nixpkgs release


avdv:
**Is your feature request related to a problem? Please describe.**
There have been made several fixes in rules_nixpkgs that we need.

* https://github.com/tweag/rules_nixpkgs/issues/461
* https://github.com/tweag/rules_nixpkgs/pull/479

**Describe the solution you'd like**
Wait for a new rules_nixpkgs release, and upgrade to that when it is available.

cc: @benradf  @prednaz

PR #2136, opened 2024-03-19
Do not depend on system tools on Darwin

avdv:
Our cc wrapper was always using the system tools from a fixed location on Darwin, which is not hermetic
(and can lead to problems when using nixpkgs) and requires the Xcode command line tools to be installed.

Apply the same heuristic as for the otool and install_name tool for the haskell toolchain and assume
the tools are available at the same location as the `ar` binary.

Depends on https://github.com/tweag/rules_nixpkgs/pull/479 and depends on #2119

### DIFF
--- a/haskell/private/cc_wrapper.py.tpl
+++ b/haskell/private/cc_wrapper.py.tpl
@@ -931,12 +931,16 @@ def darwin_rewrite_load_commands(rewrites, output):
     if args:
         subprocess.check_call([INSTALL_NAME_TOOL] + args + [output])
         # Resign the binary after patching it.
+        # Fall back to /usr/bin/codesign if the `CODESIGN` executable is not available
+        # (this might happen when using a default cc toolchain from a nix shell on Darwin instead
+        #  of using a nixpkgs_cc_configure'd toolchain).
+        codesign = CODESIGN if os.access(CODESIGN, os.X_OK) else "/usr/bin/codesign"
         # This is necessary on MacOS Monterey on M1.
         # The moving back and forth is necessary because the OS caches the signature.
         # See this note from nixpkgs for reference:
         # https://github.com/NixOS/nixpkgs/blob/5855ff74f511423e3e2646248598b3ffff229223/pkgs/os-specific/darwin/signing-utils/utils.sh#L1-L6
         os.rename(output, f"{output}.resign")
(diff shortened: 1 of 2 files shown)


avdv:
@mergifyio rebase

haskell/private/cc_wrapper.py.tpl
```diff
@@ -930,12 +931,16 @@ def darwin_rewrite_load_commands(rewrites, output):
     if args:
         subprocess.check_call([INSTALL_NAME_TOOL] + args + [output])
         # Resign the binary after patching it.
+        # Fall back to /usr/bin/codesign if the `CODESIGN` executable is not available
+        # (this might happen when using a default cc toolchain from a nix shell on Darwin instead
+        #  of using a nixpkgs_cc_configure'd toolchain).
+        codesign = CODESIGN if os.access(CODESIGN, os.X_OK) else "/usr/bin/codesign"
```
ylecornec, line 937: Should we do similar checks for `INSTALL_NAME_TOOL` and `OTOOL` ?

haskell/private/cc_wrapper.py.tpl
```diff
@@ -930,12 +931,16 @@ def darwin_rewrite_load_commands(rewrites, output):
...
         # This is necessary on MacOS Monterey on M1.
         # The moving back and forth is necessary because the OS caches the signature.
         # See this note from nixpkgs for reference:
         # https://github.com/NixOS/nixpkgs/blob/5855ff74f511423e3e2646248598b3ffff229223/pkgs/os-specific/darwin/signing-utils/utils.sh#L1-L6
         os.rename(output, f"{output}.resign")
-        subprocess.check_call([CODESIGN] + ["-f", "-s", "-"] + [f"{output}.resign"])
+        subprocess.check_call([codesign] + ["-f", "-s", "-"] + [f"{output}.resign"], env = {'CODESIGN_ALLOCATE': CODESIGN_ALLOCATE})
```
ylecornec, line 943: Maybe if `{:bindir:}/codesign` is not available then we should not set the `CODESIGN_ALLOCATE` ? Since it also points to `{:bindir}`.

ylecornec (approved):
I left a few comments, but otherwise it looks good to me :+1:

PR #2137, opened 2024-03-20
Update example in haskell-use-cases.rst to rules_oci

iteratee:
rules_docker is not just deprecated, the repository is archived and read only. Change the example in the use cases doc to be a rules_oci example.

Missing from the example is build-timestamp stamping. That isn't sufficient reason to not update the example with the current state of rules_docker.


avdv (approved):
Looks good, thank you @iteratee !

PR #2158, opened 2024-04-11
Do not create empty libraries

### LINKED ISSUE #2155: Avoid creating empty libraries
In [`haskell/private/haskell_impl.bzl`](https://github.com/tweag/rules_haskell/blob/e92d0d75424de9d7ac83bb52248daab224efc1e4/haskell/private/haskell_impl.bzl#L398-L431) we create an empty library, from an empty .c file.

The reasoning behind this is best described by this comment at the top of that file:
```
# GHC 8.10.x wants to load the shared libraries corresponding to packages needed
# for running TemplateHaskell splices. It wants to do this even when all the
# necessary object files are passed in the command line.
#
# In order to satisfy GHC, and yet avoid passing the linked library as input, we
# create a ficticious package which points to an empty shared library. The
# ficticious and the real package share the same interface files.
#
# Avoiding to pass the real shared library as input is necessary when building
# individual modules with haskell_module, otherwise building the module would
# need to wait until all of the modules of library dependencies have been built.
```
However, ghc only loads the library since it is given in the `extra-libraries` stanza of the package conf file.

Do not create the `extra-libraries` field in the package conf file for the empty package db's and remove the empty_libs handling.

avdv:
Fixes #2155


haskell/private/actions/package.bzl
```diff
@@ -64,7 +64,7 @@ def package(
       exposed_modules: List of exposed modules.
       other_modules: List of hidden modules.
       my_pkg_id: Package id object for this package.
-      has_hs_library: Whether hs-libraries should be non-null.
+      has_hs_library: Whether hs-libraries should be created.
```
avdv, line 67: I thought about refactoring this function a bit further, but AFAICS it is part of the public interface of the haskell toolchain in `actions.package` I did not went further.

aherrmann (approved):
Thank you!

PR #2117, opened 2024-02-07
chore: upgrade Bazel to 6.5.0

cgrindel:

PR #2107, opened 2024-01-09
chore: upgrade to bazel 7

cgrindel:
Related to #2099.

Depends on https://github.com/tweag/rules_haskell/issues/2119

PR #2180, opened 2024-04-30
Support Bazel 7

avdv:
After #2117, since Bazel and the project dependencies use separate nixpkgs revisions, this adds support for Bazel 7.

Supersedes #2107


avdv:
@mergifyio rebase

avdv:
Currently, rules_haskell_tests are failing on MacOS using nixpkgs:

1. https://app.buildbuddy.io/invocation/300047f3-2d31-44d6-9d75-5d9ca4783b4f
2. https://app.buildbuddy.io/invocation/5765ca58-0c3d-441b-8104-c9c09585cd19
3. https://app.buildbuddy.io/invocation/74d80558-c7a7-4838-aac4-3d8c63907695
4. https://app.buildbuddy.io/invocation/7af7ebe8-9e34-41ed-a7ec-ceb23d0baa34

The tests that fail are:

1. `//tests/recompilation:recompilation_test_nixpkgs_bazel_7`
2. `//tests/repl-targets:hs_lib_repl_test_nixpkgs_bazel_7`

The latter fails with:
```
       INFO: Running command line: bazel-bin/hs-lib@repl@repl -ignore-dot-ghci -e 'show (foo 10) ++ bar ++ baz ++ gen'
       <command line>: User-specified static library could not be loaded (/private/var/tmp/_bazel_runner/6f93c0710d32c05eb86fd5414239ece6/f463254a63c950347e0840a543f8ccec/execroot/_main//bazel-out/darwin_x86_64-fastbuild/bin/libourclibrary.a)
       Loading static libraries is not supported in this configuration.
       Try using a dynamic library instead.
```
This seems to be due to an internal change of the cc_rules. With Bazel 6, a dynamic library was built, but with Bazel 7 a static one is produced.

Trying to switch from `cc_library` to `cc_shared_library` should make this work, but we do not yet handle this type of library in rules_haskell since we expect C libraries to have a CcInfo provider which the cc_shared_library does not provide.

malt3:
The root cause has been found and will be fixed by https://github.com/bazelbuild/bazel/pull/23438 at some point. Now working on a temporary workaround in rules_nixpkgs.

malt3:
It looks like this feature was removed on purpose. I think we should add a `dynamic_deps` attribute to `haskell_library` that expects a label with `CcSharedLibraryInfo` provider.
I can take a stab at this.

malt3:
The support for `cc_shared_library` is now in draft state in #2232.

malt3:
Superseded by #2232

Issue #2181, opened 2024-05-02
build on Bazel CI broken


avdv:
Even though the badge on our README says buildkite is green, it is broken and thus the last green commit hasn't been updated in a while...

[![Build status](https://badge.buildkite.com/1de3270f1df070a99978adb6efa180d0c32eac58e0fd1938d1.svg?branch=master)](https://buildkite.com/bazel/rules-haskell-haskell)

_Originally posted by @avdv in https://github.com/tweag/rules_haskell/issues/2170#issuecomment-2089777803_

```
(00:23:25) ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/b29c7eb0e49d1926dfb75bf93b64c07e/external/ghcide/BUILD.bazel:684:22: Error while validating output TreeArtifact File:[[<execution_root>]bazel-out/k8-fastbuild/bin]external/ghcide/data-array-byte-0.1.0.1/_install/data-array-byte-0.1.0.1_iface : java.lang.InterruptedException
(00:23:25) [6,242 / 7,700] 52 actions running
    HaskellCabalLibrary @Cabal//:Cabal; 36s remote-cache, linux-sandbox
    HaskellBuildBinary @//tests/binary-with-lib:binary-with-lib; 36s remote-cache, linux-sandbox
    HaskellCabalLibrary @Cabal//:Cabal [for tool]; 36s remote-cache, linux-sandbox
    HaskellCabalLibrary @ghcide//:data-array-byte; 36s remote-cache, linux-sandbox
    HaskellBuildBinary @//tests/haskell_toolchain_library:binary; 36s remote-cache, linux-sandbox
    HaskellCabalLibrary @Cabal//:Cabal [for tool]; 36s remote-cache, linux-sandbox
    HaskellCabalLibrary @Cabal//:Cabal [for tool]; 36s remote-cache, linux-sandbox
    //tests/library-empty:empty_lib; 36s remote-cache, linux-sandbox ...
    Fetching bazel-out/k8-fastbuild/bin/tests/haddock/haddock-transitive/ghc-boot-th-9.4.6/doc-index.json; 31s
    Fetching bazel-out/k8-fastbuild/bin/tests/haddock/haddock-transitive/ghc-boot-th-9.4.6/src/GHC.Lexeme.html; 31s
    Fetching bazel-out/k8-fastbuild/bin/tests/haddock/haddock-transitive/ghc-boot-th-9.4.6/index.html; 31s
[...]

PR #2182, opened 2024-05-03
Set asterius tests to manual

ylecornec:
Asterius has been [deprecated](https://github.com/tweag/asterius) for a while so we do not need to maintain these tests anymore.

At first sight, it looks like they might be responsible for https://github.com/tweag/rules_haskell/issues/2181,
so making them manual may solve this issue until we remove them.


avdv (approved):
Yes, that makes sense. Thank you @ylecornec

PR #2202, opened 2024-05-30
Fix using the single threaded RTS lib

### LINKED ISSUE #2188: [Bazel CI] library tests are failing with Bazel@HEAD
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3841#018f5b83-19fe-48b2-b301-77acb6e1c285

 Platform: Ubuntu

 Logs: 
```
FAILED: //tests/library-empty:library-empty (Summary)

FAILED: //tests/library-with-static-cc-dep:library-with-static-cc-dep-dynamic (Summary)
``` 
 Culprit: https://github.com/bazelbuild/bazel/commit/2482322fedb463e0bbecd29c5e8e6d0f087ed884

 CC Greenteam @mai93 @Wyverald

avdv:
Since GHC 9.4.1, the rts package has `hs-libraries: HSrts-1.0.2 Cffi` in its package config.

We have special handling for HSrts (and Cffi) but that fails due to the name change.

This caused the build on Bazel CI to fail since the single threaded rts library
was symlinked into the solib directory instead of the multi threaded variant:

```
libHSrts-1.0.2_thr-ghc9.4.6.so: cannot open shared object file: No such file or directory
```

Fixes #2188

----

Note: Bazel CI is green on this branch: https://buildkite.com/bazel/rules-haskell-haskell/builds/2032

### DIFF
--- a/haskell/private/pkgdb_to_bzl.py
+++ b/haskell/private/pkgdb_to_bzl.py
@@ -84,10 +84,11 @@ def path_to_label(path, pkgroot, output=None):
         else:
             print("WARN: could not handle", path, file=sys.stderr)
 
-def hs_library_pattern(name, mode = "static", profiling = False):
+def hs_library_pattern(package_name, name, mode = "static", profiling = False):
     """Convert hs-libraries entry to glob patterns.
 
     Args:
+        package_name: The name of the package.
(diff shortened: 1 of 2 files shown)


aherrmann (approved):
Nice work, thank you!

PR #2218, opened 2024-07-10
Add support for unit-based hie-bios script for ghc 9.4+

iteratee:
Starting with GHC 9.4, GHC, and therefore HLS indirectly, will segregate the command line flags for different compilation units if the flags for that unit are placed in a file and passed as -unit @unitfile. GHC will accept multiple unitfiles on the command line and keep the arguments separate.

As an example of something this fixes, currently if you have to haskell_library targets, A and B, and A depends on B and B imports external library C, the current hie-bios script makes modules from C visible for import into A, which will fail to compile unless A also explicitly depends on C.

  A -> B -> C, but not A -> C (directly)

Add a multi flag to a repl. This currently doesn't affect the base repl target, it only affects the hie-bios target that also gets generated. This change would be a good basis for starting a repl with multiple unit files as arguments.

For a multi-repl, we carefully segregate out what should be transitive, and what should not be transitive. For example, package databases should be transitive, but package ids should not be.

In order to correctly build the unit files for each unit, we collect the haskell dependency graph as a part of _create_HaskellReplCollectInfo. Then we can use this dependency graph to build up HaskellReplDepInfo and HaskellReplLoadInfo for haskell targets from the HaskellReplLoadInfo and HaskellReplDepInfo from the corresponding structs of their direct dependencies. Because we are going to use some of these intermediate products to produce the corresponding unit files, it is important that we have more control as we build these up.

To build the unit files, we walk the haskell dependency graph in post order. We build the load_info and dep_info as above, and if we determine that this is a source-loaded dependency, then we pair the load_info and dep_info into a HaskellReplInfo. Each HaskellReplInfo corresponds to a unit file. The actual script just outputs "-hide-all-packages" and "-unit @unitfile" for each of the unit files generated in this way.

The filenames and include paths are output relative to the workspace root. There was not an obvious way to produce the absolute paths for the unit files. Producing the absolute paths in the non-multi hie-bios script relies on shell features that aren't available if we're just dumping a text file.


avdv (approved):
Thank, you @iteratee ! That looks really good. Would you be able to add a test for GHC >= 9.4 so we don't regress in the future?

iteratee:
Sorry that it took me a while. I wrote a test and made sure that it was passing.

iteratee:
It looks like the 4 CI failures on my most recent push were incidental.

Can I get a re-run?

iteratee:
> It looks like the 4 CI failures on my most recent push were incidental.
> 
> Can I get a re-run?

I realized that I needed to re-word a commit message anyway. So I did that in order to self-service restart CI.

avdv:
Thank you @iteratee !

PR #2228, opened 2024-08-09
Include package dbs of source loaded dependencies in unit files

iteratee:
Package dbs are one of the few things that need to be transitively included in a unit file. Fix the following case: There are two units A and B, and A depends on B. B depends on an external dependency C. In this case the unit file for A needs to include the package db for C. HLS and GHC will not pick up the package db indirectly through B's unit file. Fix this by passing a set of package-db-only HaskellReplDepInfo when merging.


avdv (approved):
Thank you!

PR #2232, opened 2024-08-28
support Bazel 7 cc_shared_library

malt3:
This change is based on #2180 and adds support for `cc_shared_library`. This is needed for macOS and Bazel 7+, since `cc_library` no longer emits dynamic libraries on macOS ([feature removal](https://github.com/bazelbuild/bazel/commit/ec5553352f2f661d39ac4cf665dd9b3c779e614c), [linker flag removal](https://github.com/bazelbuild/bazel/commit/4853dfd02ac7440a04caada830b7b61b6081bdfe)).

This implementation differs from the one in `cc_binary` in some ways: `cc_binary` uses a separate attribute `dynamic_deps` for `CcSharedLibraryInfo`, while `haskell_library` / `haskell_binary` use `deps` for both `CcInfo` and `CcSharedLibraryInfo`.
We also do not implement the same deduplication logic as `cc_binary` where we filter `CcInfo` deps which are already provided by a `CcSharedLibraryInfo`. This may lead to duplicate symbols. A more correct implementation would need to use aspects and might require the use of a separate attribute (`dynamic_deps`).

Open issues:

- [x] Fix `//tests/recompilation:recompilation_test_nixpkgs_bazel_7`, which depends on the macOS [`Foundation` framework](https://developer.apple.com/documentation/foundation). <strike>Probably broken by the removal of `--undefined dynamic_lookup`</strike> was broken by [`-fobjc-link-runtime`](https://cs.opensource.google/bazel/bazel/+/refs/tags/7.3.1:tools/cpp/unix_cc_toolchain_config.bzl;l=1412-1427) but only worked accidentally (needed nixpkgs cc_toolchain). 
- [x] Make tests work on Bazel 6 and 7 (use cc_shared_library only conditionally and make `//tests/repl-targets:hs_lib_repl_test_nixpkgs_bazel_6` work again)

Supersedes #2107 and #2180


aherrmann (commented):
> We can probably refactor the testing code so that cc_shared_library is used iff we are on Bazel 7 or greater. 

Perhaps [bazel_features](https://github.com/bazel-contrib/bazel_features/blob/main/features.bzl) can help. It doesn't seem to list a feature for cc_shared_library, yet. But, we could add one.

> A more correct implementation would need to use aspects and might require the use of a separate attribute (dynamic_deps).

I think that's true. Nonetheless, I think this is a good step forward.

malt3:
Adding back the linker flags `--undefined dynamic_lookup` did not fix the missing Foundation framework at link time. I'll try to investigate why we depend on the framework in the first place and what other things might have changed.

avdv:
> Adding back the linker flags `--undefined dynamic_lookup` did not fix the missing Foundation framework at link time.

I think that issue occurs when using an automatically configured local cc toolchain which is picking up a C compiler from inside a nix shell. The nix C compiler requires some environment variables to be set in order to locate the frameworks and these are not available inside of the Bazel actions (that is why we write these flags to files in the `nix-support` directory for the cc toolchain from rules_nixpkgs: https://github.com/tweag/rules_nixpkgs/blob/a11818b2173100ab8122712ae3694b711f956cbd/toolchains/cc/cc.nix#L30C1-L50C7).

The failing [test](https://github.com/tweag/rules_haskell/blob/96e231409b772859926914e76296e894982db600/rules_haskell_tests/tests/recompilation/recompilation_workspace/WORKSPACE) uses `haskell_register_ghc_nixpkgs` but does not configure a nixpkgs cc toolchain. That is probably not something that should be supported, but it worked with Bazel 6 / older rules_cc versions.

malt3:
> The failing [test](https://github.com/tweag/rules_haskell/blob/96e231409b772859926914e76296e894982db600/rules_haskell_tests/tests/recompilation/recompilation_workspace/WORKSPACE) uses `haskell_register_ghc_nixpkgs` but does not configure a nixpkgs cc toolchain. That is probably not something that should be supported, but it worked with Bazel 6 / older rules_cc versions.

So a better fix would be to add the nixpkgs cc toolchain to that WORKSPACE, correct?

avdv:
> So a better fix would be to add the nixpkgs cc toolchain to that WORKSPACE, correct?

Yes. :crossed_fingers:

PR #2239, opened 2024-09-04
Explicitly enable workspace when Bzlmod is disabled

Wyverald:
Fixes https://github.com/tweag/rules_haskell/issues/2237


Wyverald:
Oof. This flag (--enable_workspace) is only available in Bazel 7.1.0, and it looks like the CI setup is very unhappy about that. I'll probably just leave this to more capable hands.

PR #2256, opened 2024-11-07
Increase GHC bindist install timeout

### LINKED ISSUE #2254: Timeout when installing GHC
**Describe the bug**

I'm seeing timeouts when installing the GHC bindist with rules_haskell:

```
ERROR: An error occurred during the fetch of repository 'rules_haskell_ghc_linux_amd64':
	File "/var/tmp/bazel-output/external/rules_haskell/haskell/ghc_bindist.bzl", line 192, column 31, in _ghc_bindist_impl
		execute_or_fail_loudly(
	File "/var/tmp/bazel-output/external/rules_haskell/haskell/private/workspace_utils.bzl", line 25, column 13, in execute_or_fail_loudly
		fail("\n".join(["Command failed: " + " ".join(arguments), exec_result.stderr]))
Error in fail: Command failed: make install -f relocatable.mk
Timed out
```

For now I'm patching the rules to bump the `execute` timeout to 3 times the default (`3 * 600` seconds). The command is installing (copying) several Gigabytes of data and it seems to take longer than ten minutes on machines with slow IO.

```diff
diff --git a/haskell/private/workspace_utils.bzl b/haskell/private/workspace_utils.bzl
index df851329..980b3e16 100644
--- a/haskell/private/workspace_utils.bzl
+++ b/haskell/private/workspace_utils.bzl
@@ -19,6 +19,8 @@ def execute_or_fail_loudly(
         environment = environment,
         quiet = True,
         working_directory = working_directory,
+        # triple the default timeout because e.g. copying GHC is slow (1.5G)
+        timeout = 3 * 600,
     )
     if exec_result.return_code != 0:
         arguments = [_as_string(x) for x in arguments]
```

Would it make sense to add `timeout` as an argument to `execute_or_fail_loudly` and specify a greater timeout in the `make install` call? Or is there a reason why it would get completely stuck and throw a timeout error? Unfortunately I have not managed to reproduce this reliably and can't confirm that increasing the timeout fixes the issue for good (i.e. can't rule out there's another cause beyond GHC being absurdly large)

**Environment**
 - OS name + version: Linux, Ubuntu Focal 20.04.6 LTS
 - Bazel version: 7.3.1
 - Version of the rules: 1.0

nmattia:
Fixes #2254 

This bumps the default timeout for installing the GHC bindist to 30 minutes. Because the GHC install is so big (1+ GB) the command can take a while to complete on some systems.

### DIFF
--- a/haskell/ghc_bindist.bzl
+++ b/haskell/ghc_bindist.bzl
@@ -201,6 +201,8 @@ include Makefile""")
             # https://source.chromium.org/chromium/chromium/src/+/62848c8d298690e086e49a9832278ff56b6976b5.
             environment = {"ZERO_AR_DATE": "1"},
             working_directory = unpack_dir,
+            # use a big timeout because copying GHC is slow (1.5G)
+            timeout = 30 * 60,
         )
 
         if not is_hadrian_dist:
--- a/haskell/private/workspace_utils.bzl
+++ b/haskell/private/workspace_utils.bzl
@@ -2,13 +2,15 @@ def execute_or_fail_loudly(
         repository_ctx,
         arguments,
         environment = {},
-        working_directory = ""):
+        working_directory = "",
+        timeout = 600):
     """Execute the given command
 
     Fails if the command does not exit with exit-code 0.
 
     Args:
       arguments: List, the command line to execute.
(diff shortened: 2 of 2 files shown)


avdv (approved):
LGTM, thank you!

PR #2258, opened 2024-11-15
Fix Update GHC workflow

avdv:
The workflow failed for a few weeks because of an error:
```
  File "/home/runner/work/rules_haskell/rules_haskell/haskell/gen_ghc_bindist.py", line 11, in <module>
    from distutils.version import StrictVersion
ModuleNotFoundError: No module named 'distutils'
```

The `distutils` module is deprecated since Python 3.10 and has been removed in Python 3.12 now.

Apparently, this workflow is using Python 3.12 since the `nixpkgs` registry entry by default refers to
"github:NixOS/nixpkgs/nixpkgs-unstable".

This change uses the `packaging.version` module instead, switching to `nix-shell` to make use of the
nixpkgs revision that is configured in `nixpkgs/default.nix` and to provide the `packaging` pip package.

### DIFF
--- a/.github/workflows/update-ghc.yaml
+++ b/.github/workflows/update-ghc.yaml
@@ -2,9 +2,8 @@ name: Update GHC
 on:
   schedule:
     # run weekly on Thursday
-    - cron:  '0 15 * * THU'
+    - cron: '0 15 * * THU'
   workflow_dispatch: # allows manual triggering
-
 jobs:
   update_ghc:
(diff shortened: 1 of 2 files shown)


avdv:
Note, running the "Update GHC" workflow now is successful again: https://github.com/tweag/rules_haskell/actions/runs/11852416689

It found an update and created a PR here: #2259

ylecornec (approved):

Issue #2274, opened 2024-12-10
Cannot build happy 2.x from hackage


avdv:
**Describe the bug**
After updating the stackage snapshot to 2024-10-10, we get happy version 2.0.2 from hackage.

This version is no longer a simple executable but the package has been split into the command line tool and a library with multiple sublibraries.

First off, this means we have to declare dependencies between components of the library manually:

```
stack_snapshot(
     name = "stackage",
     components = { 
         "happy-lib": [                                                                                                                                                      
            "lib",                                                                                                                                                          
            "lib:grammar",                                                                                                                                                  
            "lib:backend-glr",                                                                                                                                              
            "lib:backend-lalr",                                                                                                                                             
            "lib:frontend",                                                                                                                                                 
            "lib:tabular",                                                                                                                                                  
        ],                                                                                                                                                                  
     },
     components_dependencies = {
        "happy-lib": repr({                                                                                                                                                 
            "lib:backend-glr": ["lib:grammar", "lib:tabular"],                                                                                                              
            "lib:backend-lalr": ["lib:grammar", "lib:tabular"],                                                                                                             
            "lib:frontend": ["lib:grammar"],                                                                                                                                
            "lib:happy-lib": ["lib:grammar", "lib:backend-glr", "lib:backend-lalr", "lib:frontend", "lib:tabular"],                                                         
            "lib:tabular": ["lib:grammar"],                                                                                                                                 
        }),                                                                                                                                                                 
     },
   ...
)
```

Next, the problem is that the happy-lib component is basically empty (ie. it does not produce a static nor shared library and no haddock) since it just re-exports modules from sub-libraries.

Furthermore, the backend sublibraries depend on the `Paths_happy_lib` module (see [here](https://github.com/haskell/happy/blob/e6ad1ed736182aa039b106ae004f22132ddd1af9/lib/happy-lib.cabal#L127)) to locate its data-dir which fails at runtime since the location of the data-dir in the runfiles tree is completely different to what the Cabal generated (relocatable) Paths modules expects:
```
happy: /home/claudio/.cache/bazel/_bazel_claudio/f3d395587a4c32cd78e0bbc3318c2a29/execroot/rules_haskell_tests/bazel-out/k8-opt-exec-C7777A24/bin/external/stackage/happy-2.0.2/_install/happy-lib-2.0.2-backend-lalr_data/HappyTemplate.hs: openFile: does not exist (No such file or directory)
```

[...]

Issue #2200, opened 2024-05-28
Cannot build zlib 0.7.1.0 on Windows


avdv:
**Describe the bug**

When using a recent stack snapshot that includes zlib 0.7.1.0, this version depends on the internal zlib-clib library either if explicitly requested using the `bundled-c-zlib` flag, or when disabling the `pkg-config` flag and the platform is Windows.

See https://github.com/haskell/zlib/pull/70

**To Reproduce**
```
stack_snapshot(
    name = "stackage",
    extra_deps = {"zlib": ["@zlib.dev//:zlib"]},
    packages = ["zlib"],

    # disable calling pkg-config
    flags = {"zlib": ["-pkg-config"]},

    snapshot = "nightly-2024-05-24",
)
```

Trying to build `@stackage//:zlib` results in:
```
ERROR: C:/users/runneradmin/_bazel_runneradmin/z5des2pa/external/stackage/BUILD.bazel:29:22: output 'external/stackage/zlib-clib-1.3.1/_install/zlib-clib-1.3.1_haddock/zlib-clib.haddock' was not created
ERROR: C:/users/runneradmin/_bazel_runneradmin/z5des2pa/external/stackage/BUILD.bazel:29:22: HaskellCabalLibrary @stackage//:zlib-clib failed: not all outputs were created or valid
```

The reason for this error is that the zlib-clib library is a C only library which does not produce any haddock.

**Expected behavior**

**Environment**
 - OS name + version: Windows
 - Bazel version: 6.5.0
 - Version of the rules: https://github.com/tweag/rules_haskell/commit/1cd404c28b5d95fd2e8258eccdc2779ce2053b17

**Additional context**

The only possible workaround is to disable haddock generation altogether:
```
stack_snapshot(
   name = "stackage",
   ...
   haddock = False,
)
```

jonathanlking:
I believe I'm experiencing the same issue on [lts-22.26](https://www.stackage.org/lts-22.26) with [libyaml](https://www.stackage.org/lts-22.26/package/libyaml-0.1.4) which now has a [libyaml-clib](https://www.stackage.org/lts-22.26/package/libyaml-clib-0.2.5).

[...]

Issue #1302, opened 2020-04-19
"output ... was not created" when building base-noprelude


joprice:
I am attempting to use an alternative prelude, but am getting the following error when building base-noprelude, against a stack_snapshot of lts-14.0:

```ERROR: /private/var/tmp/_bazel_josephprice/b90ee53fffa6993c9bf551d9de451812/external/stackage/BUILD.bazel:93:1: output 'external/stackage/base-noprelude-4.12.0.0/_install/lib/libHSbase-noprelude-4.12.0.0.a' was not created
ERROR: /private/var/tmp/_bazel_josephprice/b90ee53fffa6993c9bf551d9de451812/external/stackage/BUILD.bazel:93:1: output 'external/stackage/base-noprelude-4.12.0.0/_install/base-noprelude-4.12.0.0_haddock/base-noprelude.haddock' was not created
ERROR: /private/var/tmp/_bazel_josephprice/b90ee53fffa6993c9bf551d9de451812/external/stackage/BUILD.bazel:93:1: output 'external/stackage/base-noprelude-4.12.0.0/_install/lib/libHSbase-noprelude-4.12.0.0-ghc8.6.5.dylib' was not created```

**Environment**
 - OS name + version: OSX 10.15.3 
 - Bazel version: 2.0.0
 - Version of the rules: cf6beb4e3748dce6dc401a95ab8cacb803c5d944

aherrmann:
Thanks for raising this issue.

With Bazel all inputs and outputs have to be predeclared. In case of `haskell_cabal_library`, which `stack_snapshot` uses under the hood, we predeclare the library and optionally haddock outputs. If they are not created, then Bazel will fail with the error that you observe.

Right now `stack_snapshot` works around known empty packages, e.g. `bytestring-builder`, by [defining an empty `haskell_library` target instead](https://github.com/tweag/rules_haskell/blob/cf6beb4e3748dce6dc401a95ab8cacb803c5d944/haskell/cabal.bzl#L998-L1011). However, this will probably not work for `base-noprelude`. While it is an empty package in the sense that it doesn't create a library, it does create a non-empty package configuration that forwards `base` modules. At this point this is not supported by `haskell_cabal_library`.

However, `haskell_library` supports module reexports. So, I think the following approach should allow you to achieve the effect of `base-noprelude` with `rules_haskell` as of now.

1) Define a `haskell_library` that mimics `base-noprelude`. say `//third_party/base-noprelude:base-no-prelude`, like so:
[...]

PR #2287, opened 2025-02-13
cabal: Add support for empty libraries

iteratee:
Previously, an empty library that only exported modules from another library failed to build. Bazel would error out when the .a file wasn't created. These libraries are becoming more common in the Haskell ecosystem, so we need to support them.

There is a straightforward way to handle these libraries. We make the package database the default output. We avoid creating an output file for any libraries if the library is marked as empty. This is all straightforward bazel dependency/output management.

Adjust the cabal wrapper to fix up the package database even in the case that there is no library. Otherwise we end up with an empty package database, which isn't what we're looking for.

### DIFF
--- a/haskell/cabal.bzl
+++ b/haskell/cabal.bzl
@@ -483,7 +483,7 @@ def _haskell_cabal_library_impl(ctx):
         "_install/{}_data".format(package_id),
         sibling = cabal,
     )
-    with_haddock = ctx.attr.haddock and hs.tools_config.supports_haddock
+    with_haddock = ctx.attr.haddock and hs.tools_config.supports_haddock and not ctx.attr.empty_library
     if with_haddock:
         haddock_file = hs.actions.declare_file(
             "_install/{}_haddock/{}.haddock".format(package_id, package_name),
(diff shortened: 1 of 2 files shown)


avdv:
Thank you, that looks good so far.

It would be nice if this would be usable from `stack_snapshot` and also from the module extension.

With this in place we should be able to remove the workaround for empty packages [here](https://github.com/tweag/rules_haskell/blob/5d091280cd649e39868f3a99b912158f720b9f77/haskell/cabal.bzl#L134-L141), right?

Could you also add a test case (I was using the `nats` package to test this after removing that from the empty packages blacklist) please?

This is related to https://github.com/tweag/rules_haskell/pull/1458, https://github.com/tweag/rules_haskell/issues/1302 and (partially) also https://github.com/tweag/rules_haskell/issues/2274 and https://github.com/tweag/rules_haskell/issues/2200.

For the latter, we would need a similar mechanism to pass through the `haddock` attribute to the `haskell_cabal_libary` call from `stack_snapshot`  / the module extension for specific packages, just like the `empty_library` attribute.

This gets a bit tedious over time, especially since Bazel's attr is a bit limited and does not support arbitrary values in dicts, just strings. We would have to add `empty_libraries` and a `disable_haddock` list attributes for these... and maybe more in the future.

I was thinking about adding a rule which mimics the cabal attributes and returns a Provider that can be used to pass additional information to the haskell_cabal_library rule. Something along the lines of:

```
cabal_args = provider(...)

haskell_cabal_args = rule(
  ...
  attrs = {
    "is_empty": "if package is empty",
    ...
  },
  provides = [cabal_args],
)

haskell_cabal_library = rule(
   ...
   attrs = {
     ...

    "args": attr.label(providers = [cabal_args]),
  },
)

_stack_snapshot = repository_rule(
   ....
   attrs = {
     ...

    "package_args": attr.string_keyed_label_dict(providers = [cabal_args]),
  },
)
```
One could then pass any additional args to the `stack_snapshot` macro:
```
stack_snapshot(
   ...
   package_args = {
     "nats": { "is_empty": True },
     "zlib-clib": { "haddock": False },
  }
   ...
)
```
... which would generate `haskell_cabal_args` targets for each entry of the given dict and pass the appropriate label to the `args` attribute of the internal _stack_snapshot repository rule.

WDYT?

iteratee:
Sorry for the delay. I'll pick this back up again this week.

iteratee:
I have a first pass at using a provider to handle this. I only handled empty libraries with a provider.

Let me know if this along the lines that you were thinking.

With the provider, it was easier to see how to handle this for a stack snapshot. I added to the component parsing with `empty_lib` and `empty_lib:<sublib>`. This makes it easy to add a package with an empty (sub)library without needing to completely vendor the package.

This allowed me to remove the longstanding empty package blacklist and just add those packages to the list of known defaults for stack snapshots.

iteratee:
Was hoping to get some progress on this.

This fixes one of the biggest weaknesses that `rules_haskell` has dealing with cabal packages. There's even a hardcoded list in order to make it work at all.

iteratee:
> Also, it would be nice to wire this up in the module, so it can be used from a `MODULE.bazel` file.

I added a commit that I think does the right thing. I'll try and test it with modules. I hadn't yet converted over because ghc from nixpkgs wasn't working with modules for a while. I'll double check and try and switch over.

iteratee:
Can I get access to the buildbuddy secret so that my CI runs will be cached?

haskell/cabal.bzl
```diff
@@ -2237,6 +2314,7 @@ _stack_snapshot = repository_rule(
         "verbose": attr.bool(default = False),
         "custom_toolchain_libraries": attr.string_list(default = []),
         "enable_custom_toolchain_libraries": attr.bool(default = False),
+        "components_args": attr.string_dict(),
```
avdv: Likewise, could we use `attr.label_keyed_string_dict(providers = [HaskellCabalLibraryArgs])` here instead, with a TODO to switch to `string_keyed_label_dict` some time later.

This makes it explicit what to expect here.

Just call `_invert` before passing its value to the `_stack_snapshot` rule.

PR #2288, opened 2025-02-13
cabal: Honor LOCALE_ARCHIVE settings when building cabal targets

iteratee:
LOCALE_ARCHIVE is a nixpkgs/nixos setting to help glibc find the correct `locale-archive` file. A recent change has made setting this mandatory for newer nixpkgs snapshots. This setting wasn't being honored for `haskell_cabal_library` and `haskell_cabal_binary`. Pass it through for these targets. Also fix `cabal_wrapper.py` to make the path absolute before changing directories.


avdv (approved):
Thank you!

PR #2316, opened 2025-06-25
Fix small issues with CI examples jobs

avdv:
- **CI: Use windows-2022 explicitly, instead of latest**
- **CI: Run tests in test step for examples**


avdv:
@mergifyio requeue

PR #2344, opened 2025-10-01
stack.package: docs: flag enable/disable syntax

illia-bobyr:
While the "List of cabal flags" implies that `+flag` and `-flag` syntax can be used, it is somewhat subtle and requires knowledge of cabal flags syntax.  Documenting it directly here would not hurt.


avdv (approved):
Thank you!

avdv:
I just merged the changes manually, thanks again @ilya-bobyr !

PR #2422, opened 2026-05-15
docs: update old nixos wiki link

znmz:
I'm working on replacing all references to `nixos.wiki` with `wiki.nixos.org` across all nix-related repos. This increases visibility of the official wiki so that someday there will no longer be any confusion.


avdv (approved):
Thank you!

PR #2427, opened 2026-05-20
Guide: improve code snippet syntax highlighting

axman6:
When reading the guide, I noticed there were a lot of code blocks that lacked useful highlighting. I've added the appropriate annotations according to https://pygments.org/docs/lexers/#lexers-for-various-shells. It hasn't helped as much as I'd hoped, but does improve things somewhat (and VS Code picks up on the annotations and highlights things inline, which is nice for editing).

Starlark blocks have been marked `python` as there isn't a Starlark lexer.

### DIFF
--- a/docs/haskell-use-cases.rst
+++ b/docs/haskell-use-cases.rst
@@ -6,7 +6,9 @@ Common Haskell Build Use Cases
 Starting a new project
 ----------------------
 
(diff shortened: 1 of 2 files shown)


avdv (approved):
Nice, thank you!

PR #2434, opened 2026-06-26
coverage: return InstrumentedFilesInfo from haskell_library/binary/test

briantopping:
**Problem**

rules_haskell ships HPC coverage, but its rule implementations never return Bazel's `InstrumentedFilesInfo`. Without it, `bazel coverage` can't deterministically instrument the transitive Haskell libraries in a test's closure: a library is realized in both instrumented and non-instrumented configurations within one `coverage //...` run, and a test linking the non-instrumented variant produces no `<lib>_.hpc` — so its covered modules silently score 0%.

**Fix**

Return `coverage_common.instrumented_files_info` at the provider sites:
- `_haskell_binary_common_impl` (haskell_binary + haskell_test): `source_attributes = ["srcs"]`, `dependency_attributes = ["deps", "narrowed_deps"]`.
- `haskell_library_impl`: `dependency_attributes` additionally includes `"exports"`.

`haskell_library_impl` is reused by `haskell_proto_aspect` via a patched-ctx struct (the existing `real_ctx` indirection), which isn't a genuine rule ctx and which `coverage_common.instrumented_files_info` rejects — so the provider is attached only for a real rule ctx (no `real_ctx` attribute), preserving the aspect's provider contract.

**Why upstream**

Generic coverage-correctness fix, nothing project-specific — any user running `bazel coverage` over a multi-library Haskell target gets non-deterministic transitive-library coverage today.

**Verification**

buildifier clean; cherry-picks cleanly onto current master. Carried in a downstream fork where it makes HPC→lcov scoring deterministic. Not yet exercised with a fresh `bazel coverage` on master here — happy to confirm if a maintainer prefers.

### DIFF
--- a/haskell/private/haskell_impl.bzl
+++ b/haskell/private/haskell_impl.bzl
@@ -390,6 +390,18 @@ def _haskell_binary_common_impl(ctx, is_test):
                 collect_data = True,
             ),
         ),
+        # Advertise sources and transitive deps for HPC coverage. Without an
+        # InstrumentedFilesInfo provider, Bazel does not deterministically
+        # instrument transitive Haskell libraries, so their .mix (HPC metadata)
+        # is not surfaced through the provider graph and hpc-codecov scores
+        # executed library modules at 0%. haskell_binary/haskell_test expose
+        # `deps` and `narrowed_deps` (but not `exports`).
+        coverage_common.instrumented_files_info(
+            ctx,
+            source_attributes = ["srcs"],
+            dependency_attributes = ["deps", "narrowed_deps"],
+            extensions = ["hs", "hsc", "lhs", "chs"],
+        ),
         OutputGroupInfo(**compile_info_output_groups(
             name = ctx.label.name,
             workspace_name = ctx.workspace_name,
@@ -722,7 +734,7 @@ def haskell_library_impl(ctx):
         cc_shared_library_infos = [dep[CcSharedLibraryInfo] for dep in deps if CcSharedLibraryInfo in dep],
     )
 
-    return [
+    providers = [
         hs_info,
         out_cc_info,
         out_cc_shared_library_info,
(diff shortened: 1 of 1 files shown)


avdv:
Thank you! 

Does this have any effect on the tests in rules_haskell_tests, e.g. the two-libs one ?

briantopping:
Good question — no, it doesn't change the two-libs result, and the CI backs that up: the diff only touches `haskell_impl.bzl`, the two-libs test is unmodified, and the full `rules_haskell_tests` matrix passes on this PR — including `two-libs` with its strict `expected_covered_expressions_percentage = 55` / `expected_uncovered_expression_count = 4` (ubuntu + macos, GHC 9.4.8 / 9.6.5 / 9.8.2, coverage on and off).

The reason they don't interact: `two-libs` measures coverage through rules_haskell's HPC path — `experimental_coverage_source_patterns = [":two"]`, `strict_coverage_analysis`, the `.tix`/`.mix` → `hpc report` in the coverage wrapper. `InstrumentedFilesInfo` feeds Bazel's *native* coverage machinery (`--collect_code_coverage` → lcov), a separate path — so it doesn't change what HPC measures. The PR just lets `haskell_library`/`haskell_binary` participate in native/lcov coverage deterministically (without the provider Bazel doesn't know the instrumented set); it's additive to the HPC-based tests, not a change to them.

avdv (approved):
Also LGTM. Thank you!

briantopping:
Thanks for all you do, Claudio!  I'm grateful to be getting in enough usage to add a little polish. I do use Claude as you can see, but there's a human (me!) behind everything. I have everything running in Bazel, so I suspect that might create some valuable use cases. Everything has a build requirement on Linux/amd64, but my desktop is Darwin/aarch. If I can find small slices that eventually bring that to life, I will be sure to submit them. Cheers!

PR #2435, opened 2026-06-26
coverage: persist .tix and .mix into TEST_UNDECLARED_OUTPUTS_DIR

briantopping:
**Problem**

The coverage wrapper runs `hpc report` but leaves the generated `.tix` in the ephemeral test sandbox. Tooling that post-processes coverage (e.g. HPC→lcov for a coverage service) then has to scrape Bazel's sandbox-stash pool, which evicts `.tix` files once a suite grows large — silently under-counting coverage.

**Fix**

When `TEST_UNDECLARED_OUTPUTS_DIR` is set, copy each test's `.tix` and its `.mix` dirs into it, so Bazel preserves them as a stable per-target artifact under `bazel-testlogs/<target>/test.outputs/`. No behavior change when the variable is unset (non-coverage runs).

**Why upstream**

Generic + low-risk: a guarded copy gated on `TEST_UNDECLARED_OUTPUTS_DIR`, touching only coverage runs. Anyone post-processing HPC output across a large suite hits the eviction.

**Verification**

The insert sits between the test run and `hpc report`; the report path is unchanged. Carried in a downstream fork where it makes large-suite HPC→lcov scoring complete and stable.

### DIFF
--- a/haskell/private/coverage_wrapper.sh.tpl
+++ b/haskell/private/coverage_wrapper.sh.tpl
@@ -74,6 +74,23 @@ done
 
 # run the test binary, and then generate the report
 $binary_path "$@" > /dev/null 2>&1
+
+# Persist the coverage inputs (.tix + the .mix dirs) into the test's
+# undeclared outputs. Bazel keeps these as a stable per-target artifact
+# under bazel-testlogs/<target>/test.outputs/, so the lcov post-processor
+# (scripts/coverage.sh) reads complete, deterministic per-target data.
+# Without this the .tix lives ONLY in the ephemeral sandbox, which
+# Bazel's sandbox-stash pool evicts once there are many tests — silently
+# under-counting coverage (a gate that can falter is worse than none).
+if [ -n "${TEST_UNDECLARED_OUTPUTS_DIR:-}" ] && [ -f "$tix_file_path" ]; then
+  cp "$tix_file_path" "$TEST_UNDECLARED_OUTPUTS_DIR/" 2>/dev/null || true
+  mkdir -p "$TEST_UNDECLARED_OUTPUTS_DIR/mix"
+  for _hd in "${hpc_dir_args[@]}"; do
+    _d="${_hd#--hpcdir=}"
+    [ -d "$_d" ] && cp -r "$_d" "$TEST_UNDECLARED_OUTPUTS_DIR/mix/" 2>/dev/null || true
+  done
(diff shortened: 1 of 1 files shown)


avdv (approved):
Thank you, LGTM!

haskell/private/coverage_wrapper.sh.tpl
```diff
@@ -74,6 +74,23 @@ done
...
+# (scripts/coverage.sh) reads complete, deterministic per-target data.
+# Without this the .tix lives ONLY in the ephemeral sandbox, which
+# Bazel's sandbox-stash pool evicts once there are many tests — silently
+# under-counting coverage (a gate that can falter is worse than none).
+if [ -n "${TEST_UNDECLARED_OUTPUTS_DIR:-}" ] && [ -f "$tix_file_path" ]; then
+  cp "$tix_file_path" "$TEST_UNDECLARED_OUTPUTS_DIR/" 2>/dev/null || true
+  mkdir -p "$TEST_UNDECLARED_OUTPUTS_DIR/mix"
+  for _hd in "${hpc_dir_args[@]}"; do
+    _d="${_hd#--hpcdir=}"
+    [ -d "$_d" ] && cp -r "$_d" "$TEST_UNDECLARED_OUTPUTS_DIR/mix/" 2>/dev/null || true
```
avdv: This triggered a shellcheck error: https://github.com/tweag/rules_haskell/actions/runs/28252469379/job/84067437775#step:10:1693

Is there a specific reason why we need to ignore errors for the `cp` command ?

briantopping: Good catch — and no, there's no reason to swallow them silently. The block runs after the test under `set -euo pipefail`, so I'd made it non-fatal on purpose (a coverage-persist hiccup shouldn't fail an otherwise-passing test) — but `2>/dev/null || true` did that by *hiding* the error, which under-counts coverage, the exact failure this change exists to prevent.

Fixed in 7c8a178: it now warns to stderr instead of hiding the error, and the `[ -d ] && cp || true` is an explicit `if` (which also clears the SC2015). If you'd rather a persist failure hard-fail the test instead of warn, happy to drop the `|| echo` and let it propagate under `set -e` — your call.

avdv: Thank you, that's fine for me! :+1:

PR #2452, opened 2026-09-16
Mark Nix toolchain extensions as os_dependent and arch_dependent

### LINKED ISSUE #2449: Mark Nix toolchain extensions as `os_dependent`  and `arch_dependent`
Both `declare_nix_toolchains` and `nix_haskell_toolchains` are platform-dependent module extensions, but neither declares itself as such. Bazel's lockfile therefore stores a single `"general"` entry for each, which is wrong when the extension is evaluated on more than one platform.

* `declare_nix_toolchains` (`private/declare_toolchains.bzl`) calls `get_cpu_value(mctx)` and reads `mctx.os.name` directly:

  ```python
  ghc_nixpkgs_toolchain_declaration(
      ...
      cpu_value = get_cpu_value(mctx),
      os_name = mctx.os.name,
      ...
  )
  ```

* `nix_haskell_toolchains` (`extensions/nix_haskell_toolchains.bzl`) reads the platform indirectly through `default_constraints(mctx)` (from `rules_nixpkgs_core`), which generates platform-appropriate exec/target constraints. It also calls `nixpkgs_package()` to fetch GHC from Nix, where the derivation and store path are inherently platform-specific.

### Proposed fix

Set `os_dependent = True, arch_dependent = True` on both extensions. This tells Bazel to store separate lockfile entries per `(os, arch)` pair, as documented in the [Bazel Q3 2023 community update](https://blog.bazel.build/2023/10/10/bazel-q3-2023-community-update.html#lockfile-improvements).

jonathanlking:
Fixes https://github.com/tweag/rules_haskell/issues/2449.


avdv (approved):
Thank you!

PR #2455, opened 2026-09-17
Make C-library symlink directory names deterministic

### LINKED ISSUE #2448: Make C-library symlink directory names in `pkgdb_to_bzl.py` deterministic
`haskell/private/pkgdb_to_bzl.py` generates symlink directory names for external C-library paths (e.g. libffi, gmp, numa) using an insertion counter over a global `symlinks` dict:

```python
lnk = "lnk_{}".format(len(symlinks))
```

i.e. the first symlink is named `lnk_0`, the second is `lnk_1`, and so on.

The name depends entirely on how many _other_ paths have already been assigned a symlink name.
The names are a function of enumeration order and that order is nondeterministic.

The main loop iterates over `.conf` files via `glob.glob()`:

```python
for conf in glob.glob(os.path.join(package_conf_dir, '*.conf')):
```

`glob.glob()` does not sort its results — the [Python documentation](https://docs.python.org/3/library/glob.html#glob.glob) notes that ordering depends on the filesystem.

The `lnk_N` names appear in the generated `haskell_import()` rules, affecting packages such as `base`, `ghc-bignum`, `ghci`, `rts`, and `terminfo`. Since virtually every Haskell package transitively depends on `base`, a change in these symlink names propagates through most of the action graph. Bazel hashes action inputs to compute remote-cache keys, so different `lnk_N` names for the same underlying path means the cache key diverges and cross-environment cache reuse breaks.

### Proposed fix

Instead of using a loop counter to generate the symlink name, use a deterministic function of the path itself. We've been using the first 16 hex digits of the SHA-256 digest of the path string, but any hash function would work:

```patch
--- a/haskell/private/pkgdb_to_bzl.py
+++ b/haskell/private/pkgdb_to_bzl.py
@@ -17,6 +17,7 @@
 import json
 from pathlib import Path

+import hashlib
 import package_configuration

 def match_glob(root_dir, pattern):
@@ -72,7 +73,7 @@
             global symlinks
             lnk = symlinks.get(path)
             if not lnk:
-                lnk = "lnk_{}".format(len(symlinks))
+                lnk = "lnk_{}".format(hashlib.sha256(path.encode("utf-8")).hexdigest()[:16])
                 symlinks[path] = lnk
                 output.append("#SYMLINK: {} {}".format(path.replace('\\', '/'), lnk))
             return lnk
```

jonathanlking:
Fixes https://github.com/tweag/rules_haskell/issues/2448.


avdv (approved):
Thank you!

